PR #1419, opened 2012-12-28
Enable XSD caching with Xerces

easirois:
Merged content from Radu and removed disabling XSD grammar cache from  XMLGrammarPoolImplUtils as set from #1401.  That said, I'm hoping to find some middle ground where the grammar cache is only disable or enable in DITA OT based on info that we can get back back from Xerces or Xerces override when looking for a particular XSD document.


jelovirt:
@robander I think this should be closed without merging. It requires patching a third party library and we don't want to go there. If someone has need for this, they can use this patch in their fork or try to get it into Xerces itself.

PR #1445, opened 2013-01-25
Proposal for better Glossary processing

reinierk:
Changed stylesheets to get Glossary items similar to this:
>**x86-64 (x86-64)**
  A 64-bit extension of IA-32, the 32-bit generation
  of the x86 instruction set. It supports vastly
  larger amounts of virtual memory and physical
  ...
  compatible with 16-bit and 32-bit x86 code. The
  x86-64 specification is not compatible with the
  Intel Itanium (IA-64) architecture.
  **Synonyms:** Intel 64
  **Acronyms:** x86-64, x64, EM64T
  **Related Links**
  Intel Itanium architecture on page 234


jelovirt:
I think this would be good enhancement, but the patch has redundant whitespace changes that need to be fixed and merge conflicts. @reinierk are you able to rebase this on top of develop?

jelovirt:
Closing as abandoned

Issue #1545, opened 2013-08-12
Inconsistencies between plugins when flagging the root element of a topic


nigelwhitaker:
Flagging the root element of a topic does work in for example the html plugin.  But it doesn't work in pdf2.  

There's a comment in the code (plugins/org.dita.pdf2/xsl/fo/flagging-preprocess.xsl) which suggests that flagging shouldn't work:

   
   <xsl:template match="*[contains(@class, ' topic/topic ')]" 

But I suspect this comment predates the 1.7 flagging updates.

I would like to make flagging work on the topic root element for pdf2 and have worked out how to fix the code.

Also discussed here:  http://tech.groups.yahoo.com/group/dita-users/message/31152

Nigel

raducoravu:
:+1: One of our users is also trying to flag an entire topic/task in the PDF and it does not work. DITA OT 2.3.2.

raducoravu:
This is somehow also related to:
https://github.com/dita-ot/dita-ot/issues/1913
which would consider that setting the flagging condition directly on the topicref could also work in the PDF.

PR #1606, opened 2013-12-02
Topic level styling for pdf2 plugin

nigelwhitaker:
For issue #1545.

The branch also includes some other changes I needed to make in order to get the 'Build' and 'Usage' stages to work  when I tested out the plugin.


jelovirt:
Since 1.x will not patched anymore, closing as obsolete. Also the patch is stale and would require work to rebase on develop.

PR #1707, opened 2014-07-07
outer.control for files *not* in or below...

bbg3:


infotexture:
Fixed in PR #2066, so this can be closed.

PR #1720, opened 2014-08-05
XSLT optimizations for PDF stylesheets

eerohele:
Backported from version 2.0 (#1672). This commit only includes the XSLT1-compatible changes in that PR.

The table/fig/fn numbering optimizations in #1672 use node comparison operators, which are only available in XPath 2, so we can't use them directly in 1.8 (which is XSLT1-only). It might be possible to backport the changes to XSLT1 somehow, but I'm not sure how much of an effect that'd have, and I can't get exact profiling data because the XSLT stylesheets in 1.8 use EXSLT extensions, which means I can't easily use Saxon HE to profile the stylesheets. Saxon B doesn't have the profiling feature as far as I can tell.

These changes were tested using [these fixtures](https://github.com/eerohele/dita-ot-issues/tree/master/fixtures/1672) and the DITA 1.2 Specification, just as #1672. DITA-OT produces the same `topic.fo` file in both cases both before and after applying these changes, so this shouldn't break anything.

No profiling data since I can't profile the stylesheets, but a quick test suggests that these changes shave about 1.5 minutes off the entire processing time for the DITA 1.2 Specification.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -47,6 +47,13 @@ See the accompanying license.txt file for applicable licenses.
 
     <xsl:key name="id" match="*[@id]" use="@id"/>
     <xsl:key name="map-id" match="opentopic:map//*[@id]" use="@id"/>
+    <xsl:key name="topic-id"
+                 match="*[@id][contains(@class, ' topic/topic ')] |
+                        ot-placeholder:*[@id]"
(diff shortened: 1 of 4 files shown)


jelovirt:
@eerohele could you resubmit this PR into `1.8.6`, if you still want this to go in?

eerohele:
Ah, right, completely forgot about this... yeah, I'll look into it.

eerohele:
Submitted #2063 against `hotfix/1.8.6`; closing this one.

Issue #1677, opened 2014-05-09
Handle by default author and publisher content in <topicmeta> from a regular DITA Map


raducoravu:
If I have a DITA Map with content like:

```
         <map title="Eclipse content aggregated by a map">
        <topicmeta>
        <author>RADU C</author>
          <publisher>SYNCRO</publisher>
       </topicmeta>
```

and I publish it to PDF, by default I should get the author and publisher names on the front page of the PDF publication, the same as when I publish a bookmap.

In the "plugins\org.dita.pdf2\xsl\fo\front-matter.xsl" right now the "createFrontMatter_1.0" template delegates only to bookmap topicmeta:

```
     <xsl:apply-templates select="$map//*[contains(@class,' bookmap/bookmeta ')]"/>
```

and maybe it should do something like:

```
    <xsl:apply-templates select="($map//*[contains(@class,' map/topicmeta ')])[1]"/>
```

Also in the XSLT "DITA-OT\plugins\org.dita.pdf2\xsl\fo\map-elements.xsl" there is a template which matches:

```
    <xsl:template match="*[contains(@class, ' map/topicmeta ')]">
```

and has higher priority than the one in "front-matter.xsl", I do not exactly know why this happens.

PR #1722, opened 2014-08-08
Add author info onto PDF cover for normal maps #1677

eerohele:
[Fixtures](https://github.com/eerohele/dita-ot-issues/tree/master/fixtures/1677).

I'm not sure the `<xsl:template match="*[contains(@class, ' map/topicmeta ')][not(parent::opentopic:map)]">` template is actually 100% necessary, but I suppose it's a bit easier to add processing for `<topicmeta>` elements that aren't children of `<map>` if it's there since you can theoretically just do something like `<xsl:apply-templates select="$whatever/*[contains(@class, ' map/topicmeta ')]"/>` and have some kind of default processing.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/front-matter.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/front-matter.xsl
@@ -37,12 +37,6 @@ See the accompanying license.txt file for applicable licenses.
     exclude-result-prefixes="opentopic"
     version="2.0">
 
-    <xsl:template match="*[contains(@class, ' map/topicmeta ')]">
-        <fo:block-container xsl:use-attribute-sets="__frontmatter__owner__container">
-            <xsl:apply-templates/>
-        </fo:block-container>
-    </xsl:template>
-
     <xsl:template match="*[contains(@class, ' topic/author ')]">
         <fo:block xsl:use-attribute-sets="author" >
             <xsl:apply-templates/>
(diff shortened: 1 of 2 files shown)


robander:
Tested this change in 2.2.4 - it can add a lot more than the author onto covers. I created a couple of maps that use an excessive amount of metadata, including elements that have no general purpose at the map level, such as `<linktext>`. In a normal map, this change adds a lot of this material on the cover; in a bookmap, it adds less, but still more than I want. In both cases my cover gets navigation titles from child topics (this may be due to other changes in processing since the pull request was created).

I have created maps like this for testing in the past but couldn't find them; this time, I've posted them here for later reuse: https://github.com/robander/metadita/tree/master/sampledocs/metadata

robander:
One thought ... currently, the `map/topicmeta` match does nothing. This PR runs `apply-templates` on the content so that author falls through, but so does a lot more.

We could run `apply-templates` in the same location but with something like `mode="topicmeta-front-cover"`, and the default rule for that mode does nothing. This would make it easier to customize to pick and choose what shows up (though I suppose it would make the order difficult). For 3.0, we can evaluate which of the many metadata options should show up on the default cover. I do like the idea that author shows up, for example.

eerohele:
Yes, this PR seems to need more attention for 2.2.4. I'll look into it when I get the chance. Thanks for testing it!

jelovirt:
Closing this as nothing has happened with this PR in a year. Because the only generic solution for the cover page is to output the title, outputting author info on the cover page should be a customization to the output.

Issue #1414, opened 2012-12-07
The ANT build should fail by default if the XSL-FO processor cannot create the PDF file


raducoravu:
By default the ANT process finishes with success even if the creation of the PDF by the XSL-FO process failed. This is bad, the user might have the impression that the PDF was updated and in fact it was not.

There are three cases:

1) Apache FOP is used. Unfortunately the Apache FOP ANT task never fails, it just reports errors in the console, see:

https://issues.apache.org/bugzilla/show_bug.cgi?id=54219

so this should either be patched in the DITA OT or wait for a fix on the Apache FOP side.

2) RenderX XEP is used. The build file has a parameter "xep.failOnError" but by default it is set to "false". In my opinion the default should be changed to "true".

3) Antenna House is called by using <exec>. In my opinion it should also set a failonerror="true" on the task or at least have a parameter for this like the XEP task.

raducoravu:
Found a problem when generating PDF using XEP:

DITA-OT3.x\plugins\org.dita.pdf2.xep\build_xep.xml:183: You must nest a condition into <condition>

so this line:

      <condition property="xep.failOnError" value="true">

should probably be replaced with:

    <condition property="xep.failOnError" value="true">
      <not><isset property="xep.failOnError"/></not>
    </condition>

jelovirt:
We can just use

PR #1831, opened 2014-12-23
Fail Ant build if PDF rendering fails #1414

eerohele:
The way this is implemented for FOP is a hack, but I don't think we should keep waiting for this to be fixed in FOP, either.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/resource/messages.xml
+++ b/src/main/plugins/org.dita.pdf2/resource/messages.xml
@@ -93,4 +93,8 @@ See the accompanying LICENSE file for applicable license.
     <reason>Found a table row with more entries than allowed.</reason>
     <response/>
   </message>
+  <message id="PDFX013F" type="FATAL">
+    <reason>The PDF file '%1' could not be generated.</reason>
+    <response/>
+  </message>
 </messages>  
--- a/src/main/plugins/org.dita.pdf2.axf/build_axf.xml
+++ b/src/main/plugins/org.dita.pdf2.axf/build_axf.xml
@@ -110,6 +110,8 @@ See the accompanying LICENSE file for applicable license.
       </and>
     </condition>
 
+    <property name="axf.failOnError" value="true"/>
+
     <!-- run.sh on Linux, AHFCmd.exe or XSLCmd.exe on Windows -->
     <condition property="fo.ah.program.name" value="run.sh">
       <and>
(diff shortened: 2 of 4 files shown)


jelovirt:
Merged this PR by hand and noticed a problem. If AXF executable is made to fail the build on error, the log will not be processed and the user will not get any error message about what went wrong. The log processor should be made to support `@failOnError` attribute that would act the same way as the same attribute on `<exec>` except the log would be processed.

jelovirt:
This PR is hideously old, but the fixes should still be merged at some point.

eerohele:
I had completely forgotten about this PR, to be honest.

If you're willing to keep it open for a while longer, I can try take a look at it this week.

eerohele:
I think it'd be simpler to do something like this:

```xml
    <exec executable="${axf.cmd}" resultproperty="errCode" output="${dita.temp.dir}/pdf2.ah.log">
      ...
    </exec>
    <log-processor file="${dita.temp.dir}/pdf2.ah.log" taskname="ahf"/>
    <fail>
      <condition>
        <not>
          <equals arg1="${errCode}" arg="0"/>
        </not>
      </condition>
    </fail>
```

eerohele:
`<dita-ot-fail>` instead of `<fail>`, of course…

Pushed some changes. Ran some quick tests and everything seems to work fine for me.

Suggestions regarding the error message (`PDFX013F`) and a possible `<response>` appreciated, though (/cc @infotexture).

xephon2:
If you're touching the Ant code, you might consider implementing #2463 along the way.

raducoravu:
@xephon2 maybe it's not a good idea to remove the output PDF file all the times before the publishing is over. Maybe if the publishing fails for some other reason, end user wants to remain with previous PDF.

xephon2:
@raducoravu that's offtopic, so shuld be discussed in #2463

src/main/plugins/org.dita.pdf2/resource/messages.xml
```diff
@@ -93,4 +93,8 @@ See the accompanying LICENSE file for applicable license.
     <reason>Found a table row with more entries than allowed.</reason>
     <response/>
   </message>
+  <message id="PDFX013F" type="FATAL">
+    <reason>Generating PDF file '%1' was unsuccessful.</reason>
```
infotexture: I suggest we simplify this to something like:

> The PDF file '%1' could not be generated.

Not sure if we can be more specific w/r/t the reasons for failure, so a rather generic message like this may be best. 

If we _do_ know more about the cause, we might want to add details to the `<response>`.

eerohele: >The PDF file ’%1’ could not be generated.

:+1: sounds much better. I’ll update the PR soon.

PR #1844, opened 2015-01-28
Implemented improved table geometry handling for HTML output

drmacro:


drmacro:
This is an enhancement to the current HTML table generation processing, which simply pushes absolute measurements out the HTML as-is, which doesn't work in many, if not most, browsers for units other than px.

This code calculates the column geometry once during tgroup processing then makes that geometry available to the doentry template, where it's easier to calculate the appropriate width value.

Also adds a new run-time parameter to control whether or not absolute measurements should be converted to percentages (the default) or maintained as absolute measurements (converted to pixel widths).

This adds three new utility functions for working with measurement values. (These functions could be used from the image handling code but I didn't want to touch that code as part of this change).

jelovirt:
Table processing has been updated with #1871 and HTML stylesheets have been updated with #1873. These need to be merged to this PR.

raducoravu:
+1 from one of our users who used inches on the @colwidth.

jelovirt:
Calculating the column widths in XSLT and generating hard coded width definitions in HTML completely negates the idea of responsive layout. Also, the branch is stale and still contains debug code.

Closing without merging. This functionality should be implemented in a plugin for uses who need hard coded widths in HTML.

PR #1894, opened 2015-04-15
Fix unitless length-to-pixels

lagivan:
Do not remove decimal point when unitless length has two decimal digits (e.g. "88.21" was replaced with "8821").

I have the following example:
```xml
<image href="images/sot669_sv.svg" width="432.27" height="88.21" placement="break" class="- topic/image "/>
```
There `width` and `height` values are replaced by `43227` and `8821`. This happens because `$numeric-value` already does not include the decimal point after the `number` conversion:

https://github.com/dita-ot/dita-ot/blob/develop/src/main/xsl/common/dita-utilities.xsl#L178


jelovirt:
Fixed this in ba56c0e by using the same rounding rules as for all other units and cleaned to code.

Issue #1886, opened 2015-04-03
Error running DITA-OT 2.0.1 with XEP


gsejourne:
Faced this when trying to upgrade to DITA-OT 2.0.1:
May be more a doc bug, you'll be judge.

If you follow the install instructions that are available at http://www.dita-ot.org/2.0/readme/installing-minimal-package-on-linux.html you end up stuck for the PDF renderer section:

It says:
Optional: For PDF output, set up the XSL-FO processor:

Processor	Action
RenderX	Add the RenderX installation directory to the local.properties file as the xep.dir property, for example:
xep.dir=/usr/share/java/xep

Unfortunately, this does not seem to be enough to have xep run properly on a DITA-OT 2.0 install.

It's eventually missing the "pdf.formatter=xep" instruction at least, but also fails when run:

[xep] Error: Could not find or load main class com.idiominc.ws.opentopic.fo.xep.Runner
[xep] Java Result: 1

If it requires additional steps, they should be described in the install guide.

I'd be happy to contribute on the doc update... if I get the list of required steps to follow to have xep run properly.

gsejourne:
Found the reason in the end:

- the fo.jar shipped with the client package does not contain the XEP classes.
So, this is what I did:
1/ Downloaded the 2.0.1 sources for data-ot from github
2/ Compiled the various classes
3/ the fo.jar file still did not contain the xep classes.
4/ Found out that In the ./src/main/plugins/org.dita.pdf2/buildPackage.xml, looks like the if="xep-available" condition for building the xep class should be removed altogether...
I don't think it makes sense to force copy the XEP directory in the build environment...

So, in the end, I copied my XEP install dir over to the build library path (compile.path), and compiled the classes again: it worked.

I can suggest 3 possible fixes here:

1/ Clearly document the fact that XEP support is not available from the Client distribution package,
2/ Clearly document the fact that if you want XEP support, you have to compile the java classes yourself, after copying your XEP install dir in the build ${compile.path} lib directory.
3/ Disable the if="xep-available" test in the buildPackage altogether, as I don't see how it would harm to have it in the fo.jar and not use it, and would make things much simpler for people using XEP.

Let me know what you think the best solution would be, and I'll send a merge-request for the changes (be them doc or code).

jelovirt:
@robander I think you build the 2.0.1 distribution packages. Does your build clone contain XEP?

jelovirt:
[...]

PR #1895, opened 2015-04-16
Enable XEP compilation

leesmith2:
Configure xep.dir in build.properties.
Thanks to comments in: https://github.com/dita-ot/dita-ot/issues/1886

### DIFF
--- a/src/main/plugins/org.dita.pdf2/buildPackage.xml
+++ b/src/main/plugins/org.dita.pdf2/buildPackage.xml
@@ -2,6 +2,8 @@
 <project name="build_FO_plugin" default="package-plugin" basedir=".">
   <description>Build jar for FO output</description>
 
+  <property file="${basedir}/../../../../build.properties"/><!-- let it find xep.dir -->
+  
   <property name="src.dir" location="${basedir}/src/"/>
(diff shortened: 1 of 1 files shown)


jelovirt:
Gradle based build has been modified to use dummy XEP classes and XEP itself is no longer needed to build the `fo.jar`.

jrhooker:
I've read this as slowly and carefully as I can and I still can't tell if this is supposed to have fixed the issue with needing to rebuild the OT from source in order to use XEP or ... what? I'm running 2.2.2 and getting the same error as the original poster.

jelovirt:
If you're using 2.2, you should not need to recompile XEP. How are you running OT?

jrhooker:
From the command line in Windows. I'm quite sure I've set up the classpath and xep.dir correctly.

jrhooker:
When I drop in the fo.jar file from 1.8.5, I can generate PDFs with XEP but with a raft of resource resolution issues. When I return to the fo.jar file that ships with 2.2.2, I'm right back to the java class error. Given this, I don't think that this problem could have its roots in my local settings.

jelovirt:
When you run it on the command line, are you using `dita` command or running Ant directly?

jrhooker:
dita command.

PR #1909, opened 2015-05-04
Change title of garage sample maps.

nakohdo:
Give the garage sample a more meaningful title.


robander:
:+1: admittedly minor, but good to replace the silly title with one related the content.

Issue #1924, opened 2015-05-28
"exclude" in ditaval results in ant BUILD FAILED


quipsy-karg:
I just migrated from DITA-OT 1.x to 2.0.1 and noticed that the same, unchanged project fails the ant build with BUILD FAILED result when there is an "exclude" in ditaval:
```
<val>
	<prop att="audience" val="developer" action="exclude"/>
</val>
```
As soon as I replace 'exclude' by either 'include' or 'flag', the result is BUILD SUCCESSFUL.

I assume that this is not my fault as the project runs well on DITA 1.x, so this looks like a bug.

jelovirt:
Please link to a Gist with the verbose log output.

quipsy-karg:
https://gist.github.com/quipsy-karg/b70ca43f22dd95bfc5ee

jelovirt:
Seems to be a issue in PDF output. When particular elements are removed, the resulting FO file is invalid.

quipsy-karg:
I hereby confirm that the solution provided by eerohele works like a charm even in my really complex filtering real-world environment. Thank You for providing this solution! :-)

@jelovirt Please merge to master! :-)

robander:
Fixed with commit 0d73b27

PR #1950, opened 2015-06-23
Fix invalid FO caused by filtered simpletable #1924

eerohele:
If all `<strow>` elements are filtered out of a `<simpletable>` element, the FO file has an empty `<fo:table-body/>` element, which is invalid.

Adding the requisite descendant elements of `<fo:table-body>` fixes the issue.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
@@ -838,9 +838,20 @@
             </xsl:apply-templates>
 
             <fo:table-body xsl:use-attribute-sets="simpletable__body">
-                <xsl:apply-templates select="*[contains(@class, ' topic/strow ')]">
-                    <xsl:with-param name="number-cells" select="$number-cells"/>
-                </xsl:apply-templates>
+                <xsl:choose>
+                  <xsl:when test="empty(*[contains(@class, ' topic/strow ')])">
+                    <fo:table-row>
+                        <fo:table-cell>
+                            <fo:block/>
+                        </fo:table-cell>
+                    </fo:table-row>
+                  </xsl:when>
+                  <xsl:otherwise>
+                    <xsl:apply-templates select="*[contains(@class, ' topic/strow ')]">
+                        <xsl:with-param name="number-cells" select="$number-cells"/>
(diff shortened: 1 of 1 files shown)


jelovirt:
If filtering removes all rows from a table, I'm not sure if it's suitable to create empty rows/cells. I would prefer to not generate table body at all, but XSL `fo:table` requires one or more `fo:table-body` children.

@robander, what's your preference? Remove the whole table if there is no content, or create empty cells? I would prefer to not output an empty table, since if filtering removes all body content, keeping just the header row feels wrong.

robander:
@jelovirt I agree, preference would be to remove the table if there is no authored content. If it has a `<sthead>` header row in the source, but body rows are missing or filtered out, I'd still want the header row to appear since we have no idea what is there.

And yes, the awkward wording is explicitly to address the case of generated headings, where I would not want the generated headings to appear above empty content.

eerohele:
Yeah, I was actually first going to submit a PR that removes the whole table if all `<strow>` elements are filtered out, but I figured it might be slightly confusing and counter-intuitive to omit the entire table than leave just the (admittedly rather silly-looking) `<sthead>` row. After all, we would then be filtering out authored content that isn't marked up to be filtered out.

I'm not sure how to address the case of generated headings because I'm not sure when a heading would be automatically generated for a `<simpletable>` unless an user has created a customization for it.

robander:
Right, we can't know if a table will have generated headings (though some do with core DITA-OT processing). Basically I meant - 
* If the table has only `<strow>` elements, and all are filtered out, remove the table. This could remove generated headings, and that's find because there is no real content.
* If the table has non-filtered `<sthead>` in the source, even when all `<strow>` elements are filtered out, it should appear because that header row has actual source content. This is probably the case for generating an empty body cell in an `<strow>`, if that's still needed to make the table valid in FO (or maybe one cell for each cell in the header).

jelovirt:
@eerohele right now you generate an empty cell if the table is empty. `fo:table-cell` [contents][1] is

> `(%block;)+`

Thus we should generate an empty normal area there.

Also, could you rebase on develop to remove the merge commit.

[1]: https://www.w3.org/TR/xsl/#fo_table-cell

eerohele:
Done.

Issue #1861, opened 2015-02-26
copy-to Not Applied to Topics Referenced by Key In Navigation tree


drmacro:
Given this map:

```
<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "map.dtd">
<map>
 <title>Test of @copy-to For Keyreffed Topics</title>
 <topicgroup>
   <keydef keys="topic-01" href="topic01.dita"/>
 </topicgroup>
  <topicref href="root-topic.dita">
    <topichead>
      <topicmeta>
        <navtitle>Use 1 of topic01 (no copy-to)</navtitle>
      </topicmeta>
      <topicref keyref="topic-01"/>
    </topichead>
    <topichead>
      <topicmeta>
        <navtitle>Use 2 of topic01 (copy-to use-02.dita)</navtitle>
      </topicmeta>
      <topicref keyref="topic-01" copy-to="use-02.dita"/>
    </topichead>
    <topichead>
      <topicmeta>
        <navtitle>Use 3 of topic01 (copy-to use-03.dita, keys="use-03")</navtitle>
      </topicmeta>
      <topicref keyref="topic-01" keys="use-03" 
        copy-to="use-03.dita"/>
    </topichead>
  </topicref>
</map>
```

The preprocessing does not generate the copied-to topic "use-03.dita". A key-based xref to key "use-03" thus fails (because the topic does not exist).

I haven't yet looked at the code to see what is causing this behavior but it's definitely a problem.

drmacro:
Here is gist with full test data set:

https://gist.github.com/c5e4bf92110a3695d150.git

eerohele:
This happens because copy-to files are generated in the `debug-filter` stage, at which point keyrefs aren't resolved yet.

I wonder if it would be an option to extract copy-to-related code into a separate pipeline stage that would be executed after the `keyref` stage (and possibly `conref` and `conrefpush` stages)? I haven't examined what the repercussions of that would be, but this can't work with the current preprocessing order, at least.

Insofar as this issue is related to preprocessing order, it might have dependencies to #1482 and #1605 at least.

drmacro:
This particular markup configuration still fails in 2.5.4 and 3.0.4

With 2.5.4 it fails with an NPE in the copy-to phase

With 3.0.4 HTML the copy-to is simply not reflected in the temp files, so the referenced topics aren't there.

[...]

Issue #1605, opened 2013-11-24
Two levels of map ref causes good key ref to fail


drmacro:
This may be the same as issue 1600 or maybe not. 

Given a map that includes a map that then includes another map with a keydef, an xref to that keydef that is proven correct by Oxygen fails to resolve in the OT.

Test data set and OT log are available here:

http://github.com/drmacro/dita-13-spec/tree/master/temp

eerohele:
Fixed by #1962.

jelovirt:
@eerohele,could you also add 2.2 milestone as thus fix kinda goes to it.

Issue #1482, opened 2013-03-25
Incorrect processing of subject scheme maps


keberlein:
## General description

DITA-OT 1.7 does not handle subject scheme maps properly in regard to validating the attribute values against the specified subject schemes. While the simplest test yields the expected warning messages when the content contains attribute values not specified in the subject scheme, it requires that the subjectScheme map be referenced with type="subjectScheme" specified.  (The DITA 1.2 spec  does not suggest this behavior; if it is a DITA-OT requirement, we need to document it.) 

More realistic test cases using schemeref elements simply fail, however, regardless of whether type="subjectScheme" is specified.

The DITA maps and topic are located at https://gist.github.com/keberlein/5237276 .
## Test content

The following subject definitions and enumeration:

```
<subjectdef keys="operating-systems">
    <subjectdef keys="linux">
        <subjectdef keys="redhat"/>
        <subjectdef keys="suse"/>
    </subjectdef>
    <subjectdef keys="unix"/>
    <subjectdef keys="windows">
        <subjectdef keys="windows7"/>
        <subjectdef keys="windows8"/>
    </subjectdef>
</subjectdef>
```

One DITA topic with following values for the platform attribute: ubuntu, mandriva, solaris
## Test 1

**Test-1.ditamap** contains the following content:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "map.dtd">
<map>
  <title>Example of a map referencing a subject scheme</title>
  <mapref href="baseScheme-with-enumeration.ditamap" type="subjectScheme"/>
  <topicref href="t-creating-controlled-values.dita"/>
```

</map>

**The expected results are generated --  warning messages**

```
[filter] [DOTJ049W][WARN] The attribute value platform="mandriva" on element "step" does not comply with the specified subject scheme. According to the subject scheme map, the following values are valid for the platform attribute: redhat,windows8,windows7,windows,suse,linux,unix,operating-systems
 …
 [filter] [DOTJ049W][WARN] The attribute value platform="solaris" on element "step" does not comply with the specified subject scheme. According to the subject scheme map, the following values are valid for the platform attribute: redhat,windows8,windows7,windows,suse,linux,unix,operating-systems
```

[...]

PR #1961, opened 2015-07-05
Move mapref processing to beginning of preprocess

jelovirt:
This will simply keyref processing and also enable easier implementation of DITA 1.3 branch filtering because all processing can be done on a single map.


eerohele:
We should check whether this PR and/or #1962 fixes these issues (at least, there may be others):

- #1482
- #1605
- #1861

Issue #1637, opened 2014-02-09
Support DITA 1.3 branch profiling


jelovirt:
DITA 1.3 proposal 13059a.

Add support for branch filtering.

jelovirt:
Initial implementation via #1969

PR #1969, opened 2015-07-12
Add initial implementation DITA 1.3 branch filtering

jelovirt:
Add DITA 1.3 branch filtering support for #1637

The implementation is a separate module that is ran before keyref processing. The process will

-   Initially split braches so that each branch will only contain a single `ditavalref`
-   Generate `copy-to` attributes for each branch generated `topicref`
-   Filter the map based on branch filters
-   Rewrite duplicate generated `copy-to´ targets with `-#` suffix
-   Copy and filter generated `copy-to` targets
-   Filter topics that were not branch generated

### DIFF
--- a/src/main/java/org/dita/dost/module/BranchFilterModule.java
+++ b/src/main/java/org/dita/dost/module/BranchFilterModule.java
@@ -1,0 +1,565 @@
+/*
+ * This file is part of the DITA Open Toolkit project.
+ * See the accompanying license.txt file for applicable licenses.
+ */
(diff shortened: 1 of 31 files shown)


jelovirt:
Seems the branch filtering will aggressively fix duplicate `href` attributes. This should be disabled as we should only fix duplicates when filters are used.

robander:
Basic function seems to be working well - :+1: for merging. Still some details to work out but this is a good start.

PR #1977, opened 2015-07-20
Feature/1.8.6/1972

drmacro:


jelovirt:
The `.gitattributes` change is a workaround to the files being incorrectly checked in the first place, or not normalized correctly. It's the files that should be fixed, not the configuration.

drmacro:
Sorry, didn't intend for the .gitattributes change to be part of the pull request--that was just a local workaround so my repo would be in a clean state.

jelovirt:
Just revert 87b45db

jelovirt:
My vote is :-1: for this PR as it adds new functionality to a bug fix release. We've followed that rule and I want to keep doing that. For new features, there should be a new release.

drmacro:
By new release do you mean 1.8.7 or 1.9?

I would consider this to be a bug fix in that the plugin should have always been constructed this way. It doesn't add any new processing or behavior, just enables extension where it should have always been enabled.

jelovirt:
New release would be 1.9.

This doesn't break backwards compatibility but it's still a new feature as an extension point is a feature.

robander:
Agree with @jelovirt - we've held to the rule of not adding new function to bug fix releases, and I think it's a good rule to keep in place. I consider adding new extension points to be a feature, which should be documented, even if they should have been there in the past.

drmacro:
What would be involved in making a 1.9? Since I have some funding for this effort I might be able to do the work if it's not too involved.

DeltaXML needs this functionality for customers who can't necessarily move to 2.x immediately.

jelovirt:
Closing as a new feature to bug fix release branch.

PR #1983, opened 2015-07-27
Factored out main business logic to separate import to enable overrides.

drmacro:
I realized that with all the template in the top-level XSLT module they couldn't be overridden by imported modules.

I've refactored everything except the imports and params in flagging-preprocess_template.xsl into a separate *Impl.xsl

I've tested with 1.x and 2.x using my flagging plugin.


robander:
I think this is now obsolete - the modified XSL is no longer part of the PDF process, as it has moved to use the common preprocess module for flagging. [Related: we should probably mark these XSL files deprecated so that we can remove them in a later release.]

Issue #1991, opened 2015-08-06
Problem with chunking (DITA OT 2.x).


raducoravu:
I uploaded a zip here:

http://www.oxygenxml.com/forum/files/ug.zip

If the usermanual DITA Map gets published to plain XHTML the "eppo-installation-linux" HTML file should contain inside it the content from the two subtopics but it no longer does that.

The used chunk method is:

      <topicref href="topics/eppo-installation-linux.dita" chunk="select-branch to-content" navtitle="Linux Installation" locktitle="yes">
        <topicref href="concepts/installation-installer-linux.dita" toc="no"/>
        <topicref href="concepts/installation-requirements-linux.dita" toc="no"/>
      </topicref>

jelovirt:
@raducoravu are you able to verify whether this is a regression bug or something that's never worked?

jelovirt:
`select-branch` is redundant since you're targetting the root topic in `eppo-installation-linux.dita`. Have you tried removing it, in case the problem is there?

raducoravu:
Still does not work if I remove "select-branch". It works with DITA OT 1.8. But it seems that it does not work even with the stable DITA OT 2.1 release.

jelovirt:
Ok, so it's a regression bug.

raducoravu:
I tested the changes and I confirm Chris' commit fixes this issue. Thanks Chris :)

PR #2009, opened 2015-08-21
Ensuring @chunk inside topicgroups functions as expected.

ctnitchie:
Currently if @chunk is set on a topicref inside a topicgroup, it is not being processed. Since when chunking occurs, submaps are represented as topicgroups, this means that chunk instructions in sub-maps are also ignored. This attempts to resolve the issue.


jelovirt:
:+1: for merge. The `develop` branch integration tests are broken since b094e44, those should be fixed before this is merged, even if there's no overlapping code changes.

robander:
Agreed, :+1: but good policy to wait until integration tests are running.

ctnitchie:
Incidentally I think this also fixes #1991.

PR #2031, opened 2015-09-05
Feature/parse no coderefs

queshaw:
In GenMapAndTopicListModule.categorizeReferenceFile it checks for the referenced file in coderefTargetSet, but that set is not updated by GenListModuleReader. The result appears to be that the referenced code is parsed as XML, which fails when it's not XML.

This change adds a test of coderefTargetSet in the reader.

The point of all this is that I am looking into a problem reported on dita-users list about coderefs when using generate.copy.outer=3. Prior to solutions I can think of, first the coderef target shouldn't be parsed as XML when it is not XML, I think.

The method TestGenMapAndTopicListModule.testFileContentBelow is different than the others by testing the generated files. I point that out in case there is something wrong with that.

That same method also invokes generate within the method to test the non-existent case where it would fail to generate due to parsing the coderef target. That is probably confusing but I wasn't sure how else to test that.

Another possibly weird thing is TestUtil.getSourceDitaDir which guesses where src/main is based on the class path.


jelovirt:
Started to look into this and the way coderefs are currently processed is quite right. Not sure this is the correct way to handle this issue, preferably a longer term solution should be used.

jelovirt:
Commit 16bc438 is a fix to the same issue. It simplified coderef processing a bit.

The unit test improvements should be resubmitted in a separate PR.

PR #2032, opened 2015-09-06
Test coderef handling change

queshaw:
This adds a test for https://github.com/dita-ot/dita-ot/commit/16bc43859c38862d921b33521b3ec288157f5d5c

If you undo that commit, the tests in this pull request will fail because referenced json is parsed as XML.

I am just adding this because it was code already existing in another pull request.

Note that it tests the generated output instead of the copied resources, e.g. tempDirBelow instead of e. If that is the right approach, maybe e should be changed to the generated output in the other methods and I can submit another pull request for that, or someone else can do it.

The method testFileContentBelow runs generate as a way to test that the json won't be parsed.

Also, TestUtils.getSourceDitaDir guesses that the root of build/resources/tests, on the classpath, is src/main, which could be wrong.


queshaw:
I've changed the test to use Job file info map to check for subtarget and reverted changes to TestUtils.java.

jelovirt:
@queshaw could you squash the branch and force push to update the PR to a single commit?

jelovirt:
Sorry, what I meant to say is that could you squash the branch to be on top the the latest `develop` code? (So that it doesn't have e.g. `README.markdown` or `.gitignore` changes.

queshaw:
I suspect that might not be what you want either. Sorry.

I think my branch now only differs from where it diverged by my changes. But, the other changes in develop now differ.

queshaw:
Okay, I squashed changes into develop on my fork.

queshaw:
I think I have fixed things and there are only my changes comitted.

I also noticed that I had named a directory "aux" which is a device name on windows. So, I changed that.

jelovirt:
The change broadens the scope of a unit tests. Coderef should be tested in it's own unit test, not in one for gen-list.

Issue #2020, opened 2015-08-28
CHM encoding of HTML content inside topics should always be UTF-8


raducoravu:
One of our users had the following problem:

Let's say I have a DITA topic, I have no language specified to it's considered to be en-us.
At a certain point in the topic I use a certain non-ASCII character `<ph>&#10003;</ph>`.

After publishing to CHM, the character is replaced with an invalid character in the HTML output.
This happens because the "org.dita.dost.util.ConvertLang" Java task has looked into the `/org/dita/dost/util/codepages.xml` and considered that for en-us it should use "iso-8859-1" which cannot encode that specific character. Looking in the same codepages.xml file the encoding is UTF-8 for certain languages.

In my opinion the encoding used for the HTML content obtained from the DITA topics (not for the HHP and HHC files which are created separately) should always pe UTF-8 and the codepages.xml should be eliminated.

What I could find on the web about this:

> If the HTML is encoded as Unicode (UTF-8 or UTF-16 aka UNICODE) non-English chars won't be handed correctly in the HH navigation (TOC, Index, Search). The embedded browser (content area on right of the help viewer) will however display the topic text fine since this is a UNICODE enabled control.

Or at least for en-us we could use utf-8 because it is a perfect superset of ASCII which can encode all English chars.

jelovirt:
I'm unable to test this as Windows is needed. Can someone confirm that CHM will work on "all" cases if the actual content pages are output as UTF-8 and only TOC, index, and search are converted to language specific encoding?

robander:
We've been using UTF-8 for the content for years, while converting the HHP, HHC, and HHK files to a specific codepage for the locale. It's worked for all of the languages we use (all those currently in the toolkit).

robander:
Looking at this again - it looks like there were a couple of closed pull requests but the function hasn't been added yet in 2.2. I think this would be a good change. It might also take care of #1383, at least for text in the content (at least for those not in the TOC / index / search).

raducoravu:
I think I closed this one myself because there were failed tests after my changes. Maybe you could try to make changes on your side and run the tests, see if they should be rectified in some way.

raducoravu:
+1 

Had one more end user using reserved chars in the DITA content like "&#xe020;" and expecting them to be carried as such to the HTML content in order to be displayed with a custom font they have.

robander:
[...]

PR #2036, opened 2015-09-09
UTF-8 encoding in CHM for en-us languages

raducoravu:
Possible fix for:
https://github.com/dita-ot/dita-ot/issues/2020

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -1244,7 +1244,7 @@ See the accompanying license.txt file for applicable licenses.
         </xsl:variable>
         <xsl:choose>
             <!--  Disable chapter summary processing when mini TOC is created -->
-            <xsl:when test="$topicType = ('topicChapter', 'topicAppendix')"/>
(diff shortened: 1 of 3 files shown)


raducoravu:
It seems I managed to somehow commit 3 fixes for three different issues on the same pull request. Probably there should have been 3 individual push requests.
Also the first commit makes a test fail because it was expecting an ISO encoding for a certain output CHM HTML file.
So I'm stuck, I don't know what to do now :)

jelovirt:
You've used `develop` branch to create the PR commits and that causes issues when a pull request is not accepted. The approach we'd like people to use is to [always create a branch for PR](https://github.com/dita-ot/dita-ot/blob/develop/CONTRIBUTING.md#always-create-a-branch-for-your-changes). I can cherry-pick these changes now, but in the future it's way easier to use a separate branch for each feature/fix. (If you're coming from SVN and are not used to using branches as much, it's just a shift in how you approach the workflow.)

raducoravu:
Great, you can accept whatever change seems appropriate for you. You are right, the treatment of topic groups as transparent could have also been done (and done more elegantly) with a mode, I tried to make the smallest change possible. But you can reject that.

PR #2041, opened 2015-09-10
Add axf.opt Ant parameter

eerohele:
It would be useful to be able to pass the location of the Antenna House Formatter option file without having to set the environment variable.

This is not an elegant solution, but probably serves as a stopgap measure. It's hard to come up with an elegant solution without refactoring the entire buildfile.


jelovirt:
Would it not be more simple to replace `has.axf.option` property and just use `axf.opt` to handle that?

eerohele:
The first thing I tried was something like this:

```xml
<property name="axf.opt" location="${env.AXF_OPT}"/>

...


...

<target name="transform.fo2pdf.ah.hasoption" if="axf.opt">
``

PR #2042, opened 2015-09-10
Include maps when looking for props specializations (1.8.5 patch)

queshaw:
This branches from the 1.8.5 tag.

GenListModuleReader updates the list of props specializations when reading a topic, but not when reading a map.

This appears as a problem when reading a map like:

```
<map id="map">
  <title>Map</title>
  <keydef keys="some" blah="value">
    ...
  </keydef>
</map>
```

where blah is a specialization of props.

I have a test (junit) I could add, but several other unrelated tests fail so I thought I should check if this change is wanted before adding a test.


jelovirt:
This PR seems to try to merge 1.8.5 code into current `develop` branch and that's not correct. You should send the PR to `hotfix/1.8.6` branch.

queshaw:
Sorry. I've created another pull request:

https://github.com/dita-ot/dita-ot/pull/2045

Issue #2047, opened 2015-09-11
PDF flagging should use common preprocess code


robander:
A couple years ago we introduced a preprocess step to add flagging information from DITAVAL documents into DITA content. This makes it easy for any output format to support flagging without having to re-implement the logic for flagging. XHTML code was converted to use the preprocess step at that time.

The PDF2 plugin runs the common preprocess step but ignores the result. It runs an extra processing step to set up flagging, so the logic is duplicated. We should remove the extra step and make use of the common preprocess logic.

I've got the basics working now but need additional updates for elements where fall-through processing of flags would result in out of context images (such as flagged `<tbody>` or `<ol>`).

robander:
Commit includes initial implementation plus support for lists, where flagging inside of the main list causes context errors. Remaining tasks to fix context issues are mostly table elements and images (flagging currently disabled for those, needs to be re-enabled with context issues resolved).

robander:
This feature makes the recent addition of the following extension point obsolete, as we no longer have any reason to run the step:
```xml
  <extension-point id="dita.xsl.xslfo.flagging-preprocess" name="PDF flagging preprocess import"/>
```

jelovirt:
@robander is this ready?

robander:
Yes, ready [for some reason I thought I'd added that comment a couple weeks ago].

jelovirt:
The extension-point `dita.xsl.xslfo.flagging-preprocess` was never deprecated in docs and it hasn't been supported since 2.2. I removed it in 7868113 and streamlined the Ant configuration by removing the old flagging target completely.

PR #2049, opened 2015-09-14
Refactor PDF flagging to use common preprocessing #2047

robander:
Replaces the current PDF2 flagging step (currently converts `stage1.xml` to `stage1a.xml` in the PDF process) with code that takes advantage of the flagging information added during preprocess.


jelovirt:
What's the status of this PR?

robander:
It is complete and ready to merge unless anybody has concerns, but I see it now has conflicts that were not there when submitted.

PR #2051, opened 2015-09-16
Catch null FileInfo object being referenced in move-meta.

queshaw:
If .job.xml references a file that is not in the temp directory, a null pointer exception was being thrown in two places.


jelovirt:
How did this occur, that `FileInfo`was not available?

Side note, `FileInfo` objects are not just for temporary files, they can refer to an external resource that's not copied to the temporary directory.

queshaw:
I inserted an element named "invalid" into a topic in the taskbook
sample, in order to cause a parse error.
Could the fileinfo object refer to an external resource in this method?
So, the error message I added is wrong, since it prepends the temp
directory path?

jelovirt:
The code that follows is

```java
final URI targetFileName = job.tempDir.toURI().resolve(fi.uri);
```
Thus it should be in temporary directory. Note that I've been changing all messages (and basically _all_ code in general) to use URIs, not Files. This is because the goal is to make OT be able to read from URIs and use e.g. non-file system based temporary storage etc. So that error message should really use URI output.

Issue #2034, opened 2015-09-09
when transfer .ditamap file to odt format output,the title of the odt file is not correct


JoeWeiQiao:
when transfer .ditamap file to odt format output,the title of the odt file is not correct.
I try the pdf format,the title of pdf file is correct.
so maybe there is something wrong in the 'org.dita.odt' directory.
Here is the sample the dita install pkg give:
![screenshot from 2015-09-09 14 20 42](https://cloud.githubusercontent.com/assets/14193577/9754644/026f3ce8-56fe-11e5-8098-b7de7a276d69.png)
the correct title should be 'Working in the garage' which is in the sequence.ditamap file in the sample directory.
since we need to use dita to manage our user manual ,so pls help to give some feedback as soon,thanks

jelovirt:
Fixed by #2059

PR #2059, opened 2015-09-22
Fix ODT title generation #2034

eerohele:


jelovirt:
FYI, I started a _very_ heavy handed ODT refactoring earlier this year, but I haven't had time to finish it. We can merge this fix since my refactoring will probably not land in a while.

Issue #2062, opened 2015-09-23
Support "artlbl" option in PDF


robander:
We have the argument `args.artlbl` in HTML based transforms - useful for review, it shows the name of an image alongside that image. Any objections if I add support for that same parameter to PDF2? I know that @keberlein has clients that would like it, and I've got some users who expect it as well.

jelovirt:
:+1:

The name of the argument is ugly, but we should probably keep it the same in HTML and PDF.

robander:
Agreed. Implementing with same argument name `args.artlbl`, for XSL parameter I'll pass it in with the slightly more meaningful `artLabel`. For inline images, the label will appear immediately after the image; for standalone images, it will appear on a line following the image. I picked bold italic as the default style for the label so that it stands out a bit, but could easily change that if something else is preferred.

PR #2072, opened 2015-09-29
Support for args.artlbl #2062

robander:


robander:
Sure, sounds good

robander:
Ah, right, I didn't do that because for inline, I added a space on either side of the `@href` value, so it didn't bump up against the image or following text. Not sure if that's needed though.

jelovirt:
Since for inline content will be the only child of the wrapping `<fo:block>`, you can use the same inline content.

robander:
Good point.

robander:
Updated, and tested with standalone / inline images.

Issue #2071, opened 2015-09-29
PDF index should not drop terms based on first character


robander:
Index sorting in PDF2 is based on groups with specified characters. If a primary index begins with a character from a group, it goes in that group. If the starting character is not part of a group, the term is dropped with an error message. For example, all of the following are dropped from English indexes:
```xml
<indexterm>© copyright sample</indexterm>
<indexterm>пример</indexterm>
<indexterm>Å and other accents</indexterm>
<indexterm>Übersample</indexterm>
```

Rather than dropping terms, we should move unexpected terms to the "Special characters" section of the index. We should keep a message but maybe lower to Info level (sometimes useful to know, as in the second two samples, but often the correct behavior, as in the first).

I've got this working when the index grouping file consistently uses the key "Specials" for the group that includes special characters. This is inconsistent today; some use it, some have no key for the group, some translate the key. The key itself is just an ID, and never appears in rendered output; as part of this, I plan to add that specific key to the special character group in all languages. 

If the sort can't find that key, I think the only option is to continue the current behavior (drop the term, with an error). We could create another group with the key, but that would give bad results if the group already exists using a different key (presumably, part of a plugin or customization dir that already modifies the sort).

robander:
Looks like the key was missing for DE, IT, FR, was in Chinese for zh-TW, and was "/" for JA. I'll update all of those to use the key Specials.

robander:
@infotexture this feature adds a message to the message guide, not sure if additional updates are necessary? Message is just that a term has been moved to the "Special characters" section of the index.

infotexture:
No further updates necessary. 

The `user-guide/DITA-messages.xml` topic is generated automatically during the docs build and the new message appears there as expected once that runs.

PR #2073, opened 2015-09-29
Avoid dropping unexpected index terms #2071

robander:


robander:
Yep, I see. I'd copied the code from another part of the file that loops through terms. Looks like there are 2 other spots from the original code that use this redundant variable.

robander:
Other instance I'd copied from: https://github.com/robander/dita-ot/blob/cc60b9f65c559cf61456682769a195b8bc05db54/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexGroupProcessor.java#L112

And it appears again just below the new code: https://github.com/robander/dita-ot/blob/cc60b9f65c559cf61456682769a195b8bc05db54/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexGroupProcessor.java#L167

jelovirt:
The existing code should be refactored at some point to remove the redundant `key2` and just use `key`.

robander:
I can add that in to the pull request, if that sounds good?

PR #2076, opened 2015-09-30
Remove default finger image for notes in PDF

robander:


robander:
I think this is ready to go, unless there are concerns?

FWIW I considered doing this as an option, but this is one place where (from what I've heard anyway) nobody likes the current behavior. So it seems like adding an option "Turn on the finger image", which is off by default, might be overkill. We could go that way though if anybody prefers?

jelovirt:
Since an image to signal or augment notes is by no means absolutely correct or only way, I see no reason to add a toggle. If someone wants to add the old finger image, it's pretty easy to do it.

The finger images themselves could be removed in 3.0.

PR #2084, opened 2015-10-04
Ignore new build artifacts from split_pdf feature

infotexture:


jelovirt:
Implemented as part of 867e45f.

PR #2087, opened 2015-10-07
Improve programming domain semantics for HTML output

eerohele:


eerohele:
>Shouldn't we keep the class="codeph" on the `<code>` element?

Calling the `commonattributes` template already takes care of that — the literal `@class` attribute was unnecessary.

I think one point of debate is whether to implement the `<pre><code>` change for both XHTML & HTML5 or just HTML5. For XHTML, `<code>` seems to be classified as an inline element, but having `<pre><code>` is valid XHTML and seems to work just fine. I would keep it, but I don't object to having it just for HTML5.

Looks like the integration tests would need to be updated if this PR merged, in any case.

robander:
I think it's good to use it in XHTML - we've got some XHTML output that has used this for a long time now with no issues, and any modern tool is going to better recognize it as code that way.

eerohele:
>Framing should be handled with CSS, not `<hr>`. ... Same change should be made to `<pre>`

Agreed. I did it this way to retain backwards compatibility, but it's definitely a job better left for CSS.

I'll update the PR tomorrow.

eerohele:
Force-pushed to remove the `<hr>` elements.

eerohele:
`<filepath>` belongs to the software domain, but it should also be `<code>` instead of `<span>`. I can update this PR accordingly when I get the chance.

infotexture:
> `<filepath>` belongs to the software domain, but it should also be `<code>` instead of `<span>`

The same logic could be applied to `<markupname>` and the new elements in the DITA 1.3 XML mention domain. They are rendered in monospace by the `pdf2` plugin, and should probably also be wrapped in `<code>` for HTML output.

PR #2088, opened 2015-10-07
Synchronize task string use from PDF2 with common

robander:
Currently, PDF2 variable files contain their own copy of all task headings (About this task, Procedure, etc). These should be reused from the common string files (which also makes adding new languages just a bit easier). Unfortunately they use a different ID for each string, and they also have a unique ID for unordered steps (as opposed to regular steps), though in all cases the string currently matches.

In addition, the X/HTML code supports flags for both Optional and Required steps; PDF2 currently only flags Optional steps. PDF2 should flag Required steps just as X/HTML does. This is harder to reuse because PDF2 uses one variable (including the colon) while X/HTML uses two; to avoid ugly little backwards incompatible changes, I'm not going to try to reuse the strings here.

This pull request includes the following changes:
* In common string files, add an entry for `task_procedure_unordered` alongside `task_procedure`, allowing any transform type to customize these as PDF2 can today
* In XHTML and ODT, add support for this new string when using `<steps-unordered>`
* In PDF2:
  * Remove current translations for "Task NNN".
  * Add translation for "Required step"
  * For languages that listed it as "TODO", add translation for "Optional step"
  * In rendering code, check first for the old PDF2 string; anybody customizing the string gets their value. If no value is found, use the common string.

### DIFF
--- a/src/main/plugins/org.dita.odt/xsl/xslodt/dita2odt-task.xsl
+++ b/src/main/plugins/org.dita.odt/xsl/xslodt/dita2odt-task.xsl
@@ -136,7 +136,12 @@
     <xsl:apply-templates select="." mode="generate-task-label">
       <xsl:with-param name="use-label">
         <xsl:call-template name="getVariable">
-          <xsl:with-param name="id" select="'task_procedure'"/>
+          <xsl:with-param name="id">
+            <xsl:choose>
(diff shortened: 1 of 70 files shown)


robander:
Ready to merge, unless anybody has feedback...

PR #2090, opened 2015-10-08
Add axf.jar into log-processor taskdef classpath

eerohele:
The `dita` command works without this change because `axf.jar` is imported into the classpath in `org.dita.pdf2.axf/plugin.xml`, but running DITA-OT via Ant will fail because it can't find the `org.dita.dost.pdf2.AhfLogProcessorTask` class.

### DIFF
--- a/src/main/plugins/org.dita.pdf2.axf/build_axf.xml
+++ b/src/main/plugins/org.dita.pdf2.axf/build_axf.xml
@@ -5,6 +5,11 @@
 -->
 <project name="org.dita.pdf2.axf">
 
+  <path id="axf.class.path">
+    <path refid="project.class.path"/>
+    <pathelement location="${dita.plugin.org.dita.pdf2.axf.dir}/lib/axf.jar"/>
+  </path>
+
   <target name="transform.fo2pdf.ah.test-use">
     <condition property="use.ah.pdf.formatter">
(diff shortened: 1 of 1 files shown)


jelovirt:
Does this work in 2.1 without fo.jar in classpath? Since AXF code used to be there.

eerohele:
If I understand you correctly, if I check out the 2.1 tag and run:

```bash
$ ant -logfile /tmp/log.txt -debug -lib lib:resources -f build.xml \
      -Dargs.input=$PWD/samples/hierarchy.ditamap -Dtranstype=pdf \
      -Dpdf.formatter=ah -Daxf.path=/usr/local/AHFormatterV62
```

It does work, and the debug log says:

```
transform.fo2pdf.ah.init:
<snip/>
Finding class org.dita.dost.pdf2.AhfLogProcessorTask
Loaded from /Users/eerohele/Git/dita-ot/src/main/plugins/org.dita.pdf2/lib/fo.jar org/dita/dost/pdf2/AhfLogProcessorTask.class
```

So it does work without having to explicitly add `fo.jar` into the classpath when running via Ant.

jelovirt:
I would not like to manage any classpath in Ant to keep clear separation of concens, but because it has worked before, this should be merged.

PR #2097, opened 2015-10-11
Improve HTML5 and XHTML table border processing

jelovirt:
Improve HTML5 and XHTML table border processing to match expected output, allow easier integration to CSS frameworks, and output valid documents.

### DIFF
--- a/src/main/plugins/org.dita.xhtml/resource/commonltr.scss
+++ b/src/main/plugins/org.dita.xhtml/resource/commonltr.scss
@@ -1,0 +1,402 @@
+/*
+ | This file is part of the DITA Open Toolkit project.
+ | See the accompanying license.txt file for applicable licenses.
+*/
+/*
+ | (c) Copyright IBM Corp. 2004, 2005 All Rights Reserved.
+ */
+.unresolved {
+  background-color: skyblue;
+}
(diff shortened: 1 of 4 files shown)


jelovirt:
@infotexture @eerohele This also adds SASS stylesheets that would replace the CSS stylesheets as the normative stylesheets. Gradle build can then be extended to compile them to CSS during build with https://plugins.gradle.org/plugin/com.github.robfletcher.compass

eerohele:
Sounds good. Have you tested what the performance is like with that plugin? Just wondering because there's two varieties of SASS compilers, libsass and Ruby Sass, and libsass is apparently around 4000% faster than Ruby Sass. It sounds like that plugin uses Ruby Sass, but I didn't test.

Another option would be to use something like [cssnext](http://cssnext.io/), which is nice in the sense that as browsers improve their CSS support, it will become unnecessary, but it has the downside of being dependent on Node.js.

infotexture:
+1 for Sass/Compass. 

Mixins support will be useful for CSS modularization idea dicussed last month _(1 .scss file per vocabulary domain, compiled to CSS at build time)_

jelovirt:
Implemented in 509fdf4 and 7b6933c.

Issue #2086, opened 2015-10-07
Optimize PDF index generation


eerohele:
When publishing the DITA 1.2 spec with FOP, for instance, index generation is the slowest part of the PDF2 stylesheets. `index_fop.xsl` especially could be rewritten using XSL keys for better performance. `index.xsl`, too, has some expensive XPath expressions.

jelovirt:
Implemented in PR #2098

PR #2098, opened 2015-10-11
Optimize FOP PDF index generation #2086

eerohele:
This refactoring reduces index processing time for the FOP-specific stylesheet from about 4 seconds to about 200 milliseconds for the DITA 1.2 specification.

The XSL-FO this code produces is correct insofar as the original code is correct: the resulting XSL-FO is the same for [this set of test files](https://github.com/eerohele/dita-ot-issues/tree/master/fixtures/2086) and the DITA 1.2 specification both before and after this change.

I tried refactoring the code in `index.xsl`, too, but it didn't yield any gains, so I think we can leave it as it is for now.


eerohele:
Hmm, not sure why the Travis build failed. Tests pass for me locally.

PR #2103, opened 2015-10-18
Add sidebar TOC option to HTML5

jelovirt:
Add option to generate a sidebar TOC to HTML5 topic pages.

### DIFF
--- a/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
+++ b/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
@@ -62,5 +62,4 @@
     <copy file="${args.css.real}" todir="${user.csspath.real}"/>
   </target>
   
-  
 </project>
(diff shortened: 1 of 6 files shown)


jelovirt:
SASS to handle the TOC is still missing.

jelovirt:
Right now my preference would be to implement this in HTML5, but leave it unstyled. This may sound like a weird default, but if you create styled output, you'll probably take e.g. Bootstrap, Foundation, or some other framework into use and we don't want to mandate the use of a given framework. By not providing page layout CSS in `html5`, we make it easier to provide them in a transtype based on HTML5, something like `webhelp` that would be HTML5 + Bootstrap + JS.

xephon2:
I'm working on a DITA-OT plugin that adds some Bootstrap specifc elements and classes (based on [org.dita-ot.html](http://github.com/dita-ot/org.dita-ot.html)) to a Markdown/HTML mixed output that directly deploys to a Jekyll instance that is based on [Tom's Jekyll Documentation Theme](http://idratherbewriting.com/documentation-theme-jekyll/). I'll publish it here: [github.com/doctales](https://github.com/doctales) in a few weeks.

PR #2107, opened 2015-10-22
Element name generalisation optimisation

jelovirt:
XLST optimisation based on http://www.elovirta.com/2014/06/10/optimizing-dita-ot-xslt-stylesheets.html

## XPath rewriting

A simple regexp on XSLT attributes with the following search/replace expression:

```regexp
\*\[contains\(@class, *' (map|topic)/(\w+?) '\)\]
$2
```
This only works for base elements. For specializations, the rewrite rule needs to know which base type to use as the element name; the predicate is kept as is.


jelovirt:
Important thing to note here is that this optimisation _may_ not be any or that much faster than the normal case. Only by measuring the final results we'll know if this will pay off.

raducoravu:
At some point Michael Kay was interested to work on the Saxon side on this:

http://www.biglist.com/lists/lists.mulberrytech.com/xsl-list/archives/201501/msg00060.html

jelovirt:
A [paper] on the topic was presented in XML London 2015.

[paper]: http://www.saxonica.com/papers/xmllondon-2015jl.pdf

eerohele:
Ran some initial benchmarks.

Figures are for five runs with the DITA-OT Gradle plugin to eliminate JVM warmup time, starting with a cold JVM in both cases, `html5` transtype.

#### DITA 1.2 Specification

```
Before:
718.90 real        14.55 user         1.72 sys

After
417.81 real        13.69 user         1.61 sys
```

#### &lt;oXygen/&gt; XML User Manual

```
Before:
721.22 real        11.01 user         1.76 sys

After:
499.88 real         9.44 user         1.18 sys
```

In other words, the processing time is reduced by something like 30%–40%. A single run on the &lt;oXygen/&gt; XML User Manual is almost an entire minute faster post-optimization. I haven't tested, but I would expect the results for the `pdf` transtype to be quite similar or even better.

I just ran a simple `diff -r` on the output directories and there are no differences, so no content seems to be left out, either.

The gains are probably quite heavily dependent on the input document: I think I tried this optimization with the DITA-OT user guide earlier, which is significantly smaller than either of these documents, and the performance improvement was quite unimpressive, but for larger documents the impact seems to be much, well, larger.

In any case, looking quite good at this point. :+1:

Issue #1816, opened 2014-12-08
abbreviated-form and term keyref links are not resolved when chunk="to-content" [DOT 1.8 and 2.0]]


raducoravu:
Let's say I have a DITA Map like:

        <!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "http://docs.oasis-open.org/dita/v1.1/OS/dtd/map.dtd">
        <map title="Growing Flowers">
          <topicref href="topics/introduction.dita"/>
          <topicref href="glossary/glossary_overview.dita" chunk="to-content">
            <topicref href="glossary/ot.dita" keys="opentoolkit"/>
          </topicref>
        </map>

with "introduction.dita" having the content:

            <!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "http://docs.oasis-open.org/dita/v1.1/OS/dtd/topic.dtd">
            <topic id="introduction">
                <title>Introduction</title>
                <body>
                    <p>Look out for <term keyref="opentoolkit"/> and for <abbreviated-form keyref="opentoolkit"/> but not for <xref keyref="opentoolkit"/>.</p>
                </body>
            </topic>

and "glossary_overview.dita" having the content:

      <!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
      <topic id="topic_dfl_sn4_3q">
        <title>TEST</title>
        <body>
          <p></p>
        </body>
      </topic>

and "ot.dita" having the content:

      <!DOCTYPE glossentry PUBLIC "-//OASIS//DTD DITA Glossary//EN" "glossary.dtd">
      <glossentry id="wmd">
        <glossterm>Weapons of Mass
            Destruction</glossterm>
        <glossBody>
          <glossSurfaceForm>Weapons of Mass
            Destruction (WMD)</glossSurfaceForm>
          <glossAlt>
            <glossAcronym>WMD</glossAcronym>
          </glossAlt>
        </glossBody>
      </glossentry>

Publishing the DITA Map to XHTML, both the abbreviated-form and term keyrefs are not properly solved. The xref with the same keyref is properly solved.

The console output contains:

[...]

PR #2111, opened 2015-10-26
Fix href rewriting for chunked topic/term keyrefs #1816

eerohele:


eerohele:
34ba175 should correct both issues.

At least one issue remains, though, which is DITA 1.3-style references like this:

```xml
<xref href="#./p5"/>
```

Should we rewrite `dita-ot:get-topic-id` to return the actual topic ID instead of `.` in cases like that?

jelovirt:
Same topic fragments should be dealt with the SAX filter for it. It's ran before chunking, so it should work. @eerohele, how is it broken?

eerohele:
>Same topic fragments should be dealt with the SAX filter for it. It's ran before chunking, so it should work. @eerohele, how is it broken?

Ah, all right, I didn't know that. In that case, we should be fine in that respect.

Looks like I made the tests fail, though — will look into it.

eerohele:
Knew there was a reason I had used `tokenize()` instead of `substring-before()` before... should be fixed now.

PR #2118, opened 2015-10-27
PDF: Prevent table rows from breaking across pages

infotexture:


jelovirt:
While this isn't correct for all tables, for most cases this is pretty good default.

PR #2123, opened 2015-11-01
Update `html5.nav-toc` description

infotexture:
Clarify description & add a trailing period to separate it from the stem sentence with the list of supported values generated by docs `resources/params.xsl`.


infotexture:
The description could of course be revised along with the implementation if that changes, but I removed the `<nav>` element reference.

jelovirt:
Squash merged as cd02389761d81c614a0abb9111625da2c3180e0c

Issue #2126, opened 2015-11-04
Index page should contain link to page also for parent indexterm


raducoravu:
Let's say I have a DITA topic with a prolog like:
```xml
<prolog>
  <metadata>
    <keywords>
      <indexterm>Info<indexterm>Locale</indexterm></indexterm>
    </keywords>
  </metadata>
</prolog>
```
and I publish to PDF using Apache FOP. In the index page the "Info" word has no page link next to it, only the sub index "Locale" term does.
When publishing with XEP, both index terms have the page link next to them.

raducoravu:
At some point in the XSLT:

plugins/org.dita.pdf2/xsl/fo/index.xsl

there is an xsl:if:

     <xsl:if test="contains($isNormalChilds,'true ')">

which applies templates in a certain mode:

     <xsl:apply-templates select="." mode="make-index-ref">

We could add to that apply-templates an extra parameter:

        <xsl:with-param name="idxs" select="opentopic-index:refID"/>

because it is used to generate the link in the index_fop.xsl stylesheet.
This does not seem to influence XEP because it overrides this base template completely in the index_xep.xsl.

raducoravu:
I could issue a pull request if you agree with my proposed solution.

jelovirt:
:+1: for PR.

raducoravu:
Does the fix look right or should I close the pull request?

robander:
I'm not sure what to think about this one. The behavior you describe for FOP above is the one I've always assumed correct (it is the design I've been given for various index programs in IBM). But, I'd also like the outputs to be consistent.

raducoravu:
No problem, you can choose to close and not merge this one, we'll see if we have future feedback from others regarding this behavior.

robander:
There is a related issue #1903 that suggests merging the parent/child into one line for this case, rather than adding the page number to the parent. I don't have strong feelings for or against that request - if it came with a pull request, I'd probably be fine with it - but I generally think that would be a better approach than duplicating the page number.

jelovirt:
I agree with @robander here, there should not be two page numbers, especially with the original example. If the example was

```xml
<prolog>
  <metadata>
    <keywords>
      <indexterm>Info</indexterm>
      <indexterm>Info<indexterm>Locale</indexterm></indexterm>
    </keywords>
  </metadata>
</prolog>
```
that would be a different case.

drmacro:
[...]

PR #2127, opened 2015-11-04
Generate link to page also for parent index terms

raducoravu:
Possible fix for:
https://github.com/dita-ot/dita-ot/issues/2126


jelovirt:
Closing as rejected, since e.g. CMOS doesn't promote this approach.

PR #2132, opened 2015-11-07
Fix `base-html` description for generated docs

infotexture:
After [removing](https://github.com/dita-ot/docs/commit/1dbb3147aac2c9909efc0dfab968b5673c519ab8#diff-f7ba90227a330f8f7cac9e6fab83097bL59) the _“Ant parameters: ”_ prefix in dita-ot/docs@1dbb314, the description of the `base-html` transtype should be changed to correct the title of the generated docs topic:

This commit changes the resulting topic title
* _“HTML-based output formats parameters”_ to 
* _“HTML-based output parameters”_


jelovirt:
Cherry-picked in ca2205118fd83e98e071852f773b5216c428ba8d

PR #2149, opened 2015-11-24
Support PDF bookmarks for index letter headings

robander:
Many of the indexes I work with can grow quite large, spanning many pages. Currently, when using an index, the PDF bookmarks only have a single entry for the index. This pull request adds support for letter headings under the "Index" bookmark, with one entry for each letter heading that appears in the index.

Details:

There is a new parameter in the XSL, `INDEX-BOOKMARKS`. 
* When set to 0, no letter bookmarks will appear.
* When set to any other number, bookmarks will appear if the total number of entries equals or exceeds the value of `INDEX-BOOKMARKS`. Default is currently set to 100, so that if the index only spans one or two pages, no extra bookmarks will appear.
* Setting to 1 will force bookmarks to appear for any document that uses an index.

While doing this, I refactored `bookmarks.xsl` just slightly, to move index processing out of the larger `createBookmarks` named template. This makes it easier to override or customize, should anyone need to do so.

The only change for `index.xsl` is to add an ID to each letter group, which is needed to enable the bookmarks.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/ExtensibleAntInvoker.java
+++ b/src/main/java/org/dita/dost/invoker/ExtensibleAntInvoker.java
@@ -8,18 +8,16 @@
  */
 package org.dita.dost.invoker;
 
+import static java.util.Arrays.*;
+import static java.util.Collections.*;
 import static org.dita.dost.util.Constants.*;
 
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
+
 import org.apache.tools.ant.BuildException;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.Task;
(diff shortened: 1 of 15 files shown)


robander:
I think the feature here is pretty straightforward, I did wonder whether to make the parameter something that could be set in a build, but I ended up thinking it's too detailed for that. If you want to change the threshold at which entries appear, you can do so in an override / plugin.

jelovirt:
I would prefer `INDEX-BOOKMARKS` to be in lowercase, or is there some reason to have it in uppercase? Also, the name could be more describing, like `bookmarks.index-group-size`.

robander:
No reason to have it in upper case, just that I wasn't paying much attention to the variable name. I've updated to use the suggested value.

jelovirt:
@robander could you rebase on `develop`.

robander:
Yes. Going to take a little work - lots of conflicting changes and I don't remember much of what I did to get this working based on 2.1.

robander:
Pretty sure the only actual changes I made, or intended to make, are from the last two XSL files in the commit -- `bookmarks.xsl` and `index.xsl`. Tested 2.2.5 with just those 2 updates and the bookmarks are working. (I was pretty sure I didn't touch Java for this, and confused why they show up with the pull request).

I tried rebasing off of `develop` and got lots of conflicts because of the other files that show up in that commit. Is there an easy way to force update everything else and just keep those two changes? Or should I just do a new branch / new PR with the simple updates?

robander:
Tried rebasing again (after doing that successfully for another PR), and it claims I'm up to date, but pushing that doesn't do anything to this PR. At this point I think I'll make a new branch that just updates the 2 files and submit a PR with that.

robander:
Found it easier to open a new PR, #2350, closing this one.

PR #2150, opened 2015-11-24
Add basic DITA-OT support for Bosnian, Montenegrin strings

robander:
Adding in support for generated text for Bosnian and Montenegrin.

For Bosnian, the expected `xml:lang` values are "bs" or "bs-BA".

For Montenegrin, the expected `xml:lang` value is "sr-Latn-ME". (Default "sr" is Serbian-Cyrillic, this is the Montenegrin variant with Latin character set.)

As usual, values in `xml:lang` should not be case sensitive.


robander:
@jelovirt this one should be ready to merge. It's not related to the other larger i18n PR which extends existing languages for PDF; this one is adding brand new languages to the common set.

PR #2158, opened 2015-12-02
Add an updated generalisation stylesheet that uses Saxon 6.5.

georgebina:


jelovirt:
:-1: since we don't want to add XSLT 1.0 specific code. We have a hard dependency to XSLT 2.0 and as a result of that, to Saxon versions that support XSLT 2.0.

Is the `generalize.xsl` actually used anywhere? Instead, we should add generalization support to https://github.com/dita-ot/org.dita.normalize

georgebina:
It will be easy to update that to XSLT 2 - I assumed XSLT 1.0 is required as this is what the existing `generalize.xsl` used. The existing stylesheet is used only once, in: `dita-ot/src/test/testsuite/testcase/unknownforeign/unknowforeign_generalize.xml`. See also my comments on Slack - `generalize.xsl` is broken.

georgebina:
See https://github.com/dita-ot/dita-ot/pull/2159 for the XSLT 2.0 version. I am not sure how this should be integrated into https://github.com/dita-ot/org.dita.normalize. If you think the place for these stylesheets is not here, we can just remove them, including `normalize.xsl`, `pretty.xsl` and maybe also `specialize.xsl`.

PR #2159, opened 2015-12-02
Implement DITA generalization in XSLT2.

georgebina:
This implements DITA generalization with XSLT 2.0. Probably we should also remove `generalize.xsl`.

### DIFF
--- a/src/main/xsl/generalizeXSLT2.xsl
+++ b/src/main/xsl/generalizeXSLT2.xsl
@@ -1,0 +1,86 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- This file is part of the DITA Open Toolkit project.
+     See the accompanying license.txt file for applicable licenses.-->
+<!-- (c) Copyright Syncro Soft SRL, 2015. All Rights Reserved. -->
+<!-- @author George Bina -->
+
+<!--  generalizeXSLT2.xsl 
+     | Convert specialied DITA topics into revertable, "generalized" form.
+     | Started with generalize.xsl then 
+     |   - fix errors
+     |   - remove the parameter that allowed to set a folder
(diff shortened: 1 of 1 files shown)


jelovirt:
Since generalize is not really needed and is a separate XSLT that is not part of any plugin, I think there's no reason to improve it.

The normalization plugin has issue dita-ot/org.dita.normalize#2 to add generalization and I think think that's the correct place to add this functionality.

georgebina:
I understand that generalization may not be part of DITA-OT core, however, as long as we have that there it should be correct. So I think that we should either remove the current generalization.xsl or replace it with an updated version.

jelovirt:
@robander what's your take? It's been broken for a while now and we haven't received any issue, nor have I seen anything on DITA Users. Could we just take it out in 2.3 and not wait for 3.0?

georgebina:
On the same note, I think there is a problem also with `specialize.xsl` because it removes the `@class` attributes but because there is no DOCTYPE declaration on the result file this information is lost, so the IMO stylesheet should keep class attributes. FWIW, an updated XSLT 2.0 styleshet that fixes this is available at https://github.com/georgebina/dita-ot/blob/fixes/specializationXSLT2.0/src/main/xsl/specializeXSLT2.xsl but if the decision will be to remove all these stylesheets then its place should be somewhere else...

easirois:
We could use the code the @jelovirt created for the normalize.xsl to add the doctype back in. That said, that one adds the original one back in, but we might be able to add logic for base DITA topic types. The only one that might cause an issue is the general task one. Generally the root element is the same value as part of the public identifier.

easirois:
https://gist.github.com/easirois/554c439161515494aa7b

jelovirt:
Rejected as normalization plugin handles the same thing with a full transtype integration.

PR #2168, opened 2015-12-10
Recursively ignore `.gradle/` & `build/` folders

infotexture:
The current patterns miss `src/main/plugins/org.dita.html5/build/` and the 1000+ files there, but it seems silly to add folders for individual plugins when we can ignore them all with simpler patterns.


jelovirt:
Cherry-picked as 1cd870bce51cb885101f8f10870a116584ed3e7d

PR #2174, opened 2015-12-20
Added support for small-caps

Hasimir:
* Added support for small-caps as: span class="sc"
* Will (should) cover transformations from other formats "passing
  through" DITA.
* Will also match the DITA for Publishers output when encountering its
  own transformations from the sc specialisation.

### DIFF
--- a/src/main/plugins/org.dita.xhtml/resource/commonltr.css
+++ b/src/main/plugins/org.dita.xhtml/resource/commonltr.css
@@ -281,6 +281,9 @@ ul.simple {
 .shortcut {
   text-decoration: underline;
 }
+.sc {
+  font-variant: small-caps;
+}
 /* Default of bold for definition list terms */
 .dlterm {
   font-weight: bold;
--- a/src/main/plugins/org.dita.xhtml/resource/commonrtl.css
+++ b/src/main/plugins/org.dita.xhtml/resource/commonrtl.css
@@ -284,6 +284,9 @@ ul.simple {
 .shortcut {
   text-decoration: underline;
 }
+.sc {
+  font-variant: small-caps;
+}
 /* Default of bold for definition list terms */
 .dlterm {
   font-weight: bold;


jelovirt:
:-1: since this is not general use, the XSLT doesn't support this, and the old `commonltr.css` should be considered legacy and kept as stable as possible. The same small-caps functionality can be easily added to custom stylesheets that style XHTML output.

@robander what's your take?

Hasimir:
That's cool, I'm surprised it got a look so quickly actually.  As for the D4P default, the obvious occurred to me a few minutes later and Eliot's got a much bigger pull request to sift through on d4p-html5 (which covers pretty much all its other variants and allowed me to address all the Roman numeral stuff - the really big numerals that hardly anyone even knows about).

On the other hand there's I think it's only 2 of the D4P sub-specialisations that are missing by default: small-capitals/sc is one and the other is ... underline I think ... yeah, underline, which is defined to u (which might clash with something in html5, but I guess that's what xsl is for ...

Anyway, all of this is a means to an end, currently that end is providing an alternative to FOP-based PDF generation that doesn't cost an arm and a leg (it will mean venturing out from Java, but not so far away that it stops being platform independent.  ;)

jelovirt:
If `sc` is a D4P element, then it definitely should not go into the default` commonltr.css`.

Hasimir:
well, it's also a html5 one, IIRC, so that might be another influence.  I'm pretty sure Eliot tried to match current web usage as much as possible so that people would effectively be interacting with something familiar.

Hasimir:
Er, no, hang on ... either I am remembering something else or just plain mistaken (re: h5), either way that last comment is wrong.  So all things being equal (and the other stuff pending for D4P, you're right, ditch this one.  Someone might want to add the same thing to that old XHTML1.x plugin for regular DTDs, but I'm not sure there.  It is a semi-standard font setting (which I think is what I was remembering incorrectly).

Regardless, it need not be in common*.css.

PR #2177, opened 2015-12-27
Mark AbstractPipelineModuleImpl members as protected... 

PStellmann:
...to allow them being accessed in a derived class being implemented in a plugin.


jelovirt:
:+1: but not sure if this is enough. The pipeline invoker also require changes to facilitate this.

PStellmann:
It already worked for me.
Though, it was not sufficient to register my plugin library in the dost classpath. Instead I had to add it to the library list when calling dita-ot from oxygenxml - but that's fine for me.

jelovirt:
Implemented in b3b8f86

Issue #2197, opened 2016-01-15
Specializing the link element does not work


xephon2:
The content is copied from [Stackoverflow - How can I specialize the link element (related-links)](http://stackoverflow.com/questions/34766407/how-can-i-specialize-the-link-element-related-links), because Eliot says, this is a bug.

- - - - - 

I'd like to create new elements, based on the link element. This would be more comfortable for the author than using the role and otherrole attributes to specify the link role.

Unfortunately the DITA-OT 2.2.1 fails with this message:

    Required item type of result of template related-links:link. 
    is element(Q{}link); supplied value has item type
    element(Q{}myelement)

Is this a valid approach, or should I stick to otherrole?

Update 1: The error occurs in the related-links.xsl:
```
<!-- Ungrouped links have the default-mode 
     template applied to them. (Can be overridden.) -->
<xsl:template match="*[contains(@class, ' topic/link ')]" 
              mode="related-links:link" 
              name="related-links:link."
              as="element(link)">
  <xsl:sequence select="."/>   <!--- error points to this line -->
</xsl:template>
```

This line causes the error [beetle.dita#L23](https://github.com/doctales/org.doctales.terminology/blob/0b7384483956473a7d019091baf7d60f331547ac/samples/beetle.dita#L23).
   	 
	
Both variants did not work:

```
<!ATTLIST hyponym %global-atts; class CDATA "- topic/link ">
```

```
<!ATTLIST hyponym %global-atts; class CDATA "- topic/link termentry/hypernym">
```

PR #2199, opened 2016-01-16
Allow specializations of the link element

xephon2:
This should solve [issue 2197 -  Specializing the link element does not work](https://github.com/dita-ot/dita-ot/issues/2197).

### DIFF
--- a/src/main/xsl/common/related-links.xsl
+++ b/src/main/xsl/common/related-links.xsl
@@ -81,7 +81,7 @@
 
     <!-- Ungrouped links have the default-mode template applied to them. (Can be overridden.) -->
     <xsl:template match="*[contains(@class, ' topic/link ')]" mode="related-links:link" name="related-links:link"
-                  as="element(link)">
+                  as="element()*">
       <xsl:sequence select="."/>
     </xsl:template>
 
--- a/src/main/xsl/common/related-links.xsl
+++ b/src/main/xsl/common/related-links.xsl
@@ -80,7 +80,7 @@
     </xsl:template>
 
     <!-- Ungrouped links have the default-mode template applied to them. (Can be overridden.) -->
-    <xsl:template match="*[contains(@class, ' topic/link ')]" mode="related-links:link" name="related-links:link."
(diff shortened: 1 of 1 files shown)


jelovirt:
Could you squash the commits into one.

jelovirt:
Squashed and merged in 882163d

raducoravu:
In the same XSLT there are multiple places where this is used:

    as="element(linklist)"

which would lead to future potential problems when specializing link lists (I think).

jelovirt:
@raducoravu :+1: those need to be fixed too. The other possibility is to normalize those link elements when that link file is generated. Do we need the semantics at that point?

raducoravu:
Do we need the semantics at that point?

Possibly if people would want to write custom XSLT template doing different stuff with their specialization.
So if we have extension points which would allow a custom XSLT template to act at that preprocessing level and override templates in the reltable XSLT processing, then yes.

raducoravu:
The XSLT:

     DITA-OT2.x\plugins\org.dita.xhtml\xsl\xslhtml\rel-links.xsl

also seems to use a:

     as="element(linklist)

in two places.

xephon2:
Radus suggestions are not implemented yet. `as="element(linklist)` appears in:

**PDF2**

* `~\dita-ot\src\main\plugins\org.dita.pdf2\xsl\fo\links.xsl`

**XHTML**

* `~\dita-ot\src\main\plugins\org.dita.xhtml\xsl\xslhtml\conceptdisplay.xsl`
* `~\dita-ot\src\main\plugins\org.dita.xhtml\xsl\xslhtml\refdisplay.xsl`
* `~\dita-ot\src\main\plugins\org.dita.xhtml\xsl\xslhtml\rel-links.xsl` (as Radu mentioned)
* `~\dita-ot\src\main\plugins\org.dita.xhtml\xsl\xslhtml\taskdisplay.xsl`

I'm not sure whether they have to be fixed, too. If yes, should I send a PR, @jelovirt?

jelovirt:
I'll take this.

PR #2205, opened 2016-01-21
Remove src/main/samples and use samples submodule instead

shaneataylor:
The purpose for this change is to better enable using the samples files in the docs repo (for coderef and reference to example files) without removing these files from inclusion in the OT distros in the expected samples location.

It is hoped that this change will also facilitate efforts to keep the sample files current and relevant, and to include additional files, such as example .properties and .ditaval files.

A requirement for implementation of course is that https://github.com/shaneataylor/samples should be cloned to a new https://github.com/dita-ot/samples and the submodule reference in dita-ot be updated accordingly.


jelovirt:
:-1:

While I like the idea of removing `samples`, I don't want to introduce a new submodule.

shaneataylor:
Would it be better if the `samples` source was moved to the docs repo and maintained there? If so, should the docs build copy `samples` out as a sibling of the generated `doc` directory (i.e., to its current location) for distribution?

eerohele:
Maybe `samples` could live in its own repo in the `dita-ot` organization and the documentation could have a link to it, but just leave out the submodule? After all, I don't think it's really necessary to distribute the samples with DITA-OT.

Another option is just to delete the `samples` folder, which I wouldn't oppose, either.

shaneataylor:
@eerohele If we removed `samples` from the toolkit, then it could live with the docs with no need to exist as a separate repo.

@jelovirt @infotexture I think moving samples to the docs repo is the right thing to do, so I'll open pull requests to do that. I do think there is still value in distributing the sample files with the toolkit just like we include the docs (particularly if the samples are better documented), so I'll also update the docs build to copy the samples to output and open a separate PR for that change.

infotexture:
:+1: for moving `samples` to docs repo.

Since they are referenced in the the docs, I think it makes more sense to maintain them there rather than in OT core. This would also facilitate the autogeneration and `<coderef>` ideas discussed elsewhere.

PR #2206, opened 2016-01-21
Remove samples from OT (will add to docs)

shaneataylor:
See https://github.com/dita-ot/docs/pull/58 for the corresponding PR to include these files in the docs repo.


infotexture:
The corresponding docs pull request dita-ot/docs#58 has been integrated, so this can be merged to prevent duplication of the `samples` content in distribution packages.

PR #2217, opened 2016-02-02
Feature/process.note.other bug

drmacro:
Cast @othertype to string, otherwise the $title parameter is the attribute node, not the string value, so then doing <xsl:sequence select="$title"/> results in an attribute node, not a string (or node sequence).


jelovirt:
The `title` param for `process.note.common-processing` mode is untyped, thus the default value will actually be a `document-node()`. The `title` param value is used only by `xsl:value-of`, thus is should not make any difference whether we've passed in an attribute node or a string.

Only 24b23b1 from this PR is applicable, other commits should be ignored. Thus, don't merge this PR as is.

drmacro:
I used <xsl:sequence> to output $title on the assumption that it is or could be a node list, which it should be (because in theory it could have embedded markup of sort). Thus in my custom code that overrides the base note processing, I got an attribute node instead of a string.

If the intent of the code is that title is only a string the parameter should be typed as xs:string. If it can be a node sequence it should be typed as node()*. Leaving parameters like this untyped is very bad practice in XSLT 2+.

jelovirt:
Cherry picked in c9b10c53207b5594bbd82a207c4a0302787d552b

PR #2218, opened 2016-02-04
HTML5: Refactor Sass & lint partials

infotexture:
This pull request builds on the Sass refactoring changes from the January 14 Helsinki Hackathon, splitting additional rules from `_topic.scss` into Sass partials and adding an [scss-lint](https://github.com/brigade/scss-lint/) configuration file to ensure Sass code is formatted consistently.

Based on recommendations from The Sass Guidelines project & Hound CI:
* http://sass-guidelin.es/#scss-lint
* https://houndci.com/


infotexture:
The modified selector sort order yields significant diffs in the generated CSS, but the result should be functionally equivalent.

We may want to revisit the use of C-style comments to mark inline/block elements in the generated Sass partials for DITA domains, as C-style comments are passed through to the output even when empty rules that follow them are dropped by the compiler. Switching these to Sass inline comments would help to reduce noise in the output.

So far, the old CSS comments have been preserved in the Sass source and passed through to the CSS output on the assumption that they may be useful to users developing custom stylesheets. 

The alternative would be to toggle compiler options to generate compressed CSS _(without comments)_ for faster page load times, on the premise that stylesheet developers would turn to Sass source files for annotations.

Issue #2190, opened 2016-01-08
Referencing a file outside of the current director causes processing to fail


dagoss:
(tested with OT 2.2.1 pdf)

Create a map that references a file above the current directory:

	<map>
		<title>Title</title>

		<mapref href="../common/keydefs.ditamap"/>-->

		<topicref href="topics/topic1.dita"/>
		<topicref href="topics/topic2.dita"/>

	</map>

During processing, you'll get all sorts of errors about "path/to/temp/topics/topic1.dita does not exist or cannot be loaded"

If you looked in the temp/, all of the processed topics are inside a directory named after the build directory. For example, if the above map was built in "myproject/", you would find temp/myproject/topics/topic1.dita instead of temp/topics/topic1.dita that the error message says is missing.

This happens if generate.copy.outer is 1 or 3. It almost seems like generate.copy.out isn't working...

dagoss:
After some further investigation, this seems to have something to do with keydefs/keyrefs. For example, this is OK:

    <topicref href="../common/mytopic.dita"/>

But this fails:

    <keydef keys="mykey" href="../common/mytopic.dita"/>
    <topicref keyref="mykey"/>

This will even fail if the me has nothing to do with the external entity. For example, this fails:

    <keydef keys="mykey" href="topics/mytopic.dita"/>
    <topicref keyref="mykey"/>
    <topicref href="../common/mytopic.dita"/>

dagoss:
I'm attaching a minimal working example that illustrates this issue

[mwe.zip](https://github.com/dita-ot/dita-ot/files/83413/mwe.zip)

I also attempted to build this against OT 2.1.2, and it *does* build successfully with that version (OT2.2.1 was tested on both Windows and Linux; OT 2.1.2 was tested on Linux)

eerohele:
Reproduced and bisected; the bug was introduced in a4952323848288eaac31a13f943cef7bb4494d61.

eerohele:
>Reproduced and bisected; the bug was introduced in a495232.

Well, "bug was introduced" is wrong. The *error messages* appear because of the changes in that commit. The bug seems to be somewhere else, because prior to a495232, the keyrefs in the topicrefs aren't resolved at all, as the commit message for that commit indicates.

linok1:
Does this mean that this issue will be addressed in 2.3 version?

jelovirt:
[...]

PR #2223, opened 2016-02-06
Fix keyref resolving bug #2190

eerohele:
Previously, the target file URI was resolved against the root of the
temp directory. Now it's resolved against the source file URI, which
should always yield the correct result.

`KeyrefPaser.java` is a bit different in `develop` and `hotfix/2.2.3`, which is why I'm submitting this to `develop` for now, but I can backport this fix to `hotfix/2.2.3` and submit a separate PR if necessary.

### DIFF
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -384,17 +384,18 @@ public final class KeyrefPaser extends AbstractXMLFilter {
                 if (keyDef != null) {
                     if (currentElement != null) {
                         final NamedNodeMap attrs = elem.getAttributes();
-                        final URI target = keyDef != null ? keyDef.href : null;
-                        if (target != null && !target.toString().isEmpty()) {
+                        final URI href = keyDef.href;
+
+                        if (href != null && !href.toString().isEmpty()) {
                             if (TOPIC_IMAGE.matches(currentElement.type)) {
                                 valid = true;
-                                final URI targetOutput = normalizeHrefValue(URLUtils.getRelativePath(currentFile, job.tempDir.toURI().resolve(target)), elementId);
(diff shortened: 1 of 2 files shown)


src/main/java/org/dita/dost/writer/KeyrefPaser.java
```diff
@@ -384,17 +384,17 @@ private Attributes processElement(final Attributes atts) {
                 if (keyDef != null) {
                     if (currentElement != null) {
                         final NamedNodeMap attrs = elem.getAttributes();
-                        final URI target = keyDef != null ? keyDef.href : null;
+                        final URI target = keyDef.href;
```
eerohele: I removed the null check here because there's already a null check for `keyDef` on l. 384, so I think this is unnecessary, right?

eerohele:
Ah, the integration tests reveal I misunderstood something (although this particular Travis build failed for an unrelated reason).

Anyway, I'll have to revisit this.

eerohele:
All right, so I think the issue is that `keyDef.href` needs to be resolved against `keyDef.source` before this approach can work.

8ac5b22 passes the integration tests (locally) and fixes #2190, but I'm not sure about the approach I took to ensure `keyDef.source` is set. Need to look into that a bit more when I get the time.

eerohele:
3a72f89 passes the *unit* tests, too (stupidly only ran integration tests before).

What I'm wondering is whether it would be better to pass `currentFile` around as a function argument in `KeyrefReader.java` instead of setting it as an instance variable. I just went for a minimal changeset as a starting point.

jelovirt:
No point in passing `currentFile` to functions as arguments, because `KeyrefReader` is not multi-threaded. Just keep it as an instance variable.

jelovirt:
@eerohele are you able to rebase on `develop` to resolve the merge conflicts?

eerohele:
@jelovirt: done.

PR #2229, opened 2016-02-11
Use fo:wrapper instead of fo:inline for ID containers

eerohele:
With Antenna House, if you customize the `getTitle` template to generate an `<fo:block>` element (a pretty common scenario, I'd imagine) and add `space-before` to that `<fo:block>` element, the `<fo:inline>` elements whose sole purpose is to serve as reference targets affect the space before the title. Here are two screenshots that illustrate the issue:

#### `fo:inline`

inline

#### `fo:wrapper`

wrapper

The red rectangle that delineates the area generated by the title is larger in the `fo:inline` example.

I don't think the reference target elements should affect spacing in any way, so fixing this issue follows the path of least surprise, I think. I wasted quite many hours trying to figure out what was going on until I realized that it was the `fo:inline` elements that were the culprit.

This issue doesn't affect FOP — both `fo:inline` and `fo:wrapper` yield the same result with FOP. I don't have a XEP license to test with.

(This might be an Antenna House issue, but I doubt it. And even if it is, using `fo:wrapper` seems like a good workaround to me.)

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -720,14 +720,14 @@ See the accompanying license.txt file for applicable licenses.
                       <xsl:with-param name="marker-class-name" as="xs:string">current-h2</xsl:with-param>
                     </xsl:apply-templates>
                 </xsl:if>
-                <fo:inline id="{parent::node()/@id}"/>
-                <fo:inline>
+                <fo:wrapper id="{parent::node()/@id}"/>
+                <fo:wrapper>
                     <xsl:attribute name="id">
                         <xsl:call-template name="generate-toc-id">
                             <xsl:with-param name="element" select=".."/>
                         </xsl:call-template>
                     </xsl:attribute>
-                </fo:inline>
+                </fo:wrapper>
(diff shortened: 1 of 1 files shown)


drmacro:
For history, there was a time when some tools, probably FOP but maybe AH, threw away IDs on fo:wrapper. Clearly that is no longer the case. Definitely support this change.

PStellmann:
Using FOP I have additional space before a title and I could identify the empty fo:wrapper elements to cause this.

Mmy sample fo code looks like this (the spaces for indention were added by hand to make it easier to read and are not present in the actual code):
```
<fo:block border="solid blue">
	<fo:marker marker-class-name="current-h2">Title</fo:marker>
	<fo:wrapper id="unique_3"/>
	<fo:wrapper id="_OPENTOPIC_TOC_PROCESSING_d89e342"/>
	<fo:block border="solid red">Title</fo:block>
</fo:block>
```

producing this output:
dita-ot-title-wrapper
So the fo:wrapper elements generate a blank line.

When I change the fo:wrapper to fo:inline the result doesn't change.

When I change the fo:wrapper to fo:block the additional space disappears:
dita-ot-title-block

So I think at least for FOP fo:block should be used. I can't test it with XEP or AH, though.

PR #2233, opened 2016-02-17
Rewrite utility templates as functions and add XSpec tests

eerohele:
I didn't write XSpec tests for the `@href` resolving functions even though they could really use some because I wasn't quite sure how to best do it, frankly, since the functions return absolute paths and they can obviously be different on everyone's machine.

Also, I didn't write add one for the `dita-ot:get-variable` function because I just couldn't get the test to work even though the function works correctly. Rewriting the `findStrings` template it calls to facilitate testing would probably help.

Also, a heads up: I think at least some of the tests for `length-to-pixels` don't really work properly: I changed some to have incorrect expectations and they still passed. Don't know what's up with that.

### DIFF
--- a/src/main/xsl/common/functions.xsl
+++ b/src/main/xsl/common/functions.xsl
@@ -1,0 +1,137 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!-- This file is part of the DITA Open Toolkit project.
+     See the accompanying license.txt file for applicable licenses.-->
+<!-- (c) Copyright IBM Corp. 2004, 2005 All Rights Reserved. -->
+
+<xsl:stylesheet version="2.0"
+  xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:dita-ot="http://dita-ot.sourceforge.net/ns/201007/dita-ot"
+  exclude-result-prefixes="xs dita-ot">
+
+  <!-- href -->
+
+  <xsl:function name="dita-ot:resolve-href-path" as="xs:anyURI">
+    <xsl:param name="href" as="attribute(href)"/>
+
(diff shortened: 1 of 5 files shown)


eerohele:
Force-pushed missing scenarios.

src/main/xsl/common/functions.xsl
```diff
@@ -0,0 +1,137 @@
...
+      <xsl:with-param name="ctx" select="$ctx" tunnel="yes"/>
+      <xsl:with-param name="id" select="$id"/>
+      <xsl:with-param name="params" select="$params"/>
+      <xsl:with-param name="ancestorlang"
+        select="dita-ot:get-language-codes(dita-ot:get-current-language($ctx))"/>
+      <xsl:with-param name="defaultlang" select="dita-ot:get-language-codes($DEFAULTLANG)"/>
+    </xsl:call-template>
+  </xsl:function>
+
+  <xsl:function name="dita-ot:get-variable">
```
jelovirt: Could you add result datatype to all functions?

eerohele: Yeah, I didn't add one for these two because I wasn't 100% sure what it should optimally be. `xs:string*`, I guess? Or should it be `node()*`? `xs:string` would throw a fatal error if DITA-OT can't find the variable, so I don't think we want that.

eerohele:
Force-pushed missing data types (`xs:string*`).

eerohele:
Err... I'm not sure what's going on with the integration tests. They pass for me locally. Maybe this whole PR branch got borked somehow.

jelovirt:
The errors are e.g
>`Expected text value 'Parent topic:' but was 'Parent topic :' `

This is...weird, where did that whitespace come from.

jelovirt:
Reran the CI tests and now they pass. Random failures, it seems.

PR #2239, opened 2016-02-21
Refactor HTML5 table generation

eerohele:
If you think this PR is (or has the potential to be) merge-worthy, I'd like to write XSpec tests for the functions I added before it's merged. I just wanted to get some eyes on this thing before I get any further (although I probably got a bit carried away already).

Relying on data attributes for presentation might be a bit unorthodox, but I think in this case it's actually pretty clean and intuitive. Let me know what you think, though.

Tested on these files:

https://github.com/eerohele/dita-ot-issues/tree/master/fixtures/2239

Here's `topic1.html` before this PR:

https://nw.kapsi.fi/dita-ot/issues/2239/before/topic1.html

And after:

https://nw.kapsi.fi/dita-ot/issues/2239/after/topic1.html


jelovirt:
I'm not sure about the use of data attributes like that. What we're basically doing is adding the old HTML 4 era styling attributes back as data attributes. The HTML5 seems to be just declare all styling in CSS and generally speaking that's via CSS classes. The final result is the same, but still this feels a bit strange, since if I were to do this outside DITA and OT, from scratch, I would not use data attributes like this, I'd definitely use classes instead.

jelovirt:
I think the question is which is better,

```html
<td class="entry" data-align="left" data-colsep="1" data-rowsep="0" headers="d39e567 ">2:3</td>
```
or

```html
<td class="entry entry--align-left entry--colsep" headers="d39e567 ">2:3</td>
```

eerohele:
Fair enough — I can certainly refactor the code to use classes instead.

And I felt kinda weird using data attibutes for styling like this, too (and had the same thought re HTML4 styling attributes) but I just couldn't resist the temptation to make the XSLT code as minimal as possible and being able to use SASS/CSS almost as if directly styling the DITA elements.

Using BEM-style CSS class names definitely works, too — it just requires writing a bit more XSLT to generate the class names.

jelovirt:
Preferably the class name generation should be overrideable, to make OT content easily embeddable to existing class name structure. I don't feel strongly about data vs class, but I think classes are what webdevs are used to seeing and data attributes are usually used to store non-styling related data or metadata.

eerohele:
>Preferably the class name generation should be overrideable, to make OT content easily embeddable to existing class name structure.

Agreed. Overridability was one of the aspects I was going for with the data attribute -based approach, too. I'll see what I can come up with with the class-based version.

eerohele:
`baf2205` refactors the implementation to use classes instead of data attributes. Wanted to get the change up for comments early again.

Here's a new sample output file:

https://nw.kapsi.fi/dita-ot/issues/2239/after-class/topic1.html

To override the HTML `@class` attribute generation for a specific element or attribute, you can match it on the `css-class` mode:

```xsl
<xsl:template match="*[contains(@class, ' topic/table ')]" mode="css-class">
  <!-- Add my-custom-class to the @class attribute of every generated table -->
  <xsl:sequence select="'my-custom-class'"/>
  <!-- Generate a CSS class from @frame, @pgwide, and @scale attributes -->
  <xsl:apply-templates select="@frame, @pgwide, @scale" mode="#current"/>
</xsl:template>
```

The `dita-ot:css-class` function takes an attribute and generates a BEM-style CSS class from that attribute. It uses the name of the containing element, the name of the attribute, and the value of the attribute by default.

For example:

```
<table pgwide="1">

  =>

<table class="table--pgwide-1">
```

Reasonable or over-engineered?

infotexture:
Do these changes only apply to CALS tables, or do they also consider `<simpletable>`,  `<choicetable>`, etc. ?

I have clients that use a mix of `<table>` & `<simpletable>` markup within publications, and they are regularly surprised when processing & default styling differ. It would be good if the way OT handles different types of tables did not drift apart unnecessarily.

There are good reasons for using different variants in DITA markup, but the HTML output they generate should be as similar as possible to limit the number of CSS rules required to align them.

eerohele:
@infotexture only CALS tables at this stage, but my intention was to make this a first step in making processing for all table types as uniform as possible for the HTML5 transtype.

drmacro:
Whatever you do the original class values need to be maintained so that existing CSS doesn't break.

I definitely like the idea of having more complete class values for CSS styling--I'm dealing with that right now for a client's custom EPUB output where they want to use CSS to fully control the table rendering.

eerohele:
>Whatever you do the original class values need to be maintained so that existing CSS doesn't break.

It's difficult to do that *and* generate clean HTML5 and use intuitive and readable class names.

The current iteration of this PR removes the old `.cellrowborder` etc. classes and replaces them with more direct `.rowsep-1` and `.colsep-1` classes (could also be just `.colsep` and `.rowsep`, but this was more straightforward to implement).

It also removes the `.firstcol` class that was previously generated for `<table rowheader="firstcol">` (the first column uses the `<th scope="row">` instead so it can still be easily targeted with CSS).

As it stands, then, this PR definitely targets DITA-OT 3.0, where breaking background compatibility is "allowed".

It's no big effort to restore the old class names, of course, if necessary.

jelovirt:
Since this is for HTML5 plugin, not the old XHTML plugin, we have a lot more freedom in changing the class names. I'd be ok releasing this in 2.3, since 3.0 will take about a year to land.

drmacro:
As a data point, I just went through the exercise of overriding most of the table formatting in the 2.1 HTML transform, removing all generation of HTML-specific formatting controls and @style attributes, leaving only the @class values on table (adding value of the @frame attribute) and <td> and <th> (client doesn't currently need to apply styling at the row level).

But they are using the current @class values generated for entry elements to control the column and row separation rendering via CSS. It would be annoying to have to change that CSS to move to 2.3.

Since @class is an arbitrarily-long list of keywords I don't see the harm in retaining the current values in addition to new values other than general neatness.

At least make retaining them a runtime option in the 2.x timeframe.

eerohele:
Maybe retain them in 2.3 (provided that this change is included in that version) with a deprecation comment and remove in 3.0?

drmacro:
That would be reasonable. But I think it would be bad form to break working CSS from 2.2 to 2.3.

jelovirt:
@eerohele I'd like to push this to 2.3, since there's no quarantee that 3.0 will be the next version. We may end up releasing 2.4 next. (Though, _I_ will plan to start work on 3.0 next.)

eerohele:
@jelovirt OK. I'll restore the legacy CSS classes and add the XSpec tests.

jelovirt:
Should we mark the deprecated classes with [`@warn`](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#_6)?

eerohele:
Good idea — I'll do that.

eerohele:
I've now tried to restore the legacy CSS classes to the best of my ability. It's kinda difficult because the old code is pretty buggy in places, and I don't always understand what it tries to do, so if anyone has the time to test my changes, that'd be great.

New before and after:

https://nw.kapsi.fi/dita-ot/issues/2239/before/topic1.html
https://nw.kapsi.fi/dita-ot/issues/2239/after/topic1.html

I added XSpec tests for the functions that return atomic values. I couldn't figure out how to test a function that returns a node with XSpec, but I can add the tests for those later once/if I figure it out.

eerohele:
Refactored simpletable processing, too. Changed `table--frame-all` and `table--scale-50` etc. to just `frame-all` and `scale-50` so that they can be used with all elements that use [the display attributes group](https://www.oxygenxml.com/dita/1.3/specs/#langRef/attributes/displayAttributes.html).

Before and after:

https://nw.kapsi.fi/dita-ot/issues/2239/before/simpletable.html
https://nw.kapsi.fi/dita-ot/issues/2239/after/simpletable.html

I was unsure about how to handle simpletable cell borders, so I still have to implement that. In the previous code, simpletables always have a frame, regardless of the value of the `@frame` attribute, and cells always have a border. `@frame` now works as it should, but cells have no borders.

I'm leaning towards adding borders for simpletable cells by default. Opinions?

jelovirt:
:+1: for borders to simpletable

eerohele:
bfe29db adds borders for simpletable entries while retaining the possibility to use `@frame` to control whether the table has borders.

https://nw.kapsi.fi/dita-ot/issues/2239/after/simpletable.html

infotexture:
Should the `<simpletable>` rules from bfe29db be stored alongside the other table styles in [_tables.scss](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/sass/_tables.scss)?

I know the generated `_topic.scss` currently contains empty rules for these and other table-related classes, but I think it would be a bit more intuitive if table-related rules were stored in the same partial.

eerohele:
>Should the `<simpletable>` rules from bfe29db be stored alongside the other table styles in _tables.scss?

I'm fine with that — I just put them where the existing `.stentry` rules were. But it makes sense to split `_topic.scss` into smaller partials anyway, it'll grow to be to large otherwise.

Could also add `_simpletable.scss` and move all simpletable-related rules (empty or not) there?

infotexture:
I wouldn't create a separate partial for the `<simpletable>` rules. I think it makes more sense to put all table-related rules in one place, including those for `<choicetable>` and whatever else we'll need.

Regarding `<simpletable>` borders, I'd be inclined to handle them just like regular `<table>` borders, so the default presentation is the same for all table variants.

On the project website, we currently use the Sass `@extend` directive, so `<simpletable>` picks up the same styles used by standard tables to ensure consistent presentation:

```scss
.simpletable {
	@extend .table;
	@extend .table-condensed;
}
```

While the extended classes in this code are specific to Bootstrap, I think we could use a similar approach here to align table output regardless of which DITA element they are derived from.

eerohele:
I moved the simpletable rules into `_tables.scss`.

Simpletable entries now have borders by default, but regular table entries don't. Neither table type has frames by default.

Frames are a fairly straightforward thing to enable by default (basically just a matter of adding `table { border: 1px solid }`). Likewise, simpletable borders are easy because they have no `colsep` or `rowsep` or the burden of legacy CSS classes: just override `.stentry` via CSS.

Regular table borders, though, are a bit more complicated. the CSS becomes somewhat more awkward if we enable borders by default and try to make it easy to disable them by setting `colsep` or `rowsep` to `0` — especially with the legacy CSS classes mixing things up.

Personally, I kind of like the idea of regular tables being more of a blank slate sort of a thing and giving the user the ability to enable the frames and borders they want, instead of the other way around. Also, FWIW, HTML tables don't have any borders by default.

What do others think?

Other than that, there's one big thing standing in the way of this PR being finished, which is that `org.dita.html5/xsl/xslhtml/simpletable.xsl` overrides `<choicetable>` and `<properties>` processing, due to the way the XSL stylesheets are imported/included. For choicetables, it's not that big of a deal, but `<properties>` tables need a bit of work because of the optional cells they can have.

So:

- [ ] Refactor `<properties>` table processing
- [ ] Refactor `<choicetable>` processing

(I *may* have bitten off a bit more than I intended here...)

eerohele:
[XHTML](https://nw.kapsi.fi/dita-ot/issues/2239/xhtml/root.html) / [Before — HTML5](https://nw.kapsi.fi/dita-ot/issues/2239/before-all/root.html) / [After](https://nw.kapsi.fi/dita-ot/issues/2239/html5/root.html)

I think I'm done with this PR unless you have any further change requests. I'd like to avoid making further major changes in this PR so that it doesn't grow too big to handle.

If necessary, I think we can also tweak some aspects (like default frame behavior) later in other PRs, too.

jelovirt:
Anyone else want to review? Look ok to me.

infotexture:
We might want to consider removing the default column separators for `<simpletable>`, `<choicetable>`, and `<properties>`.

Since there's also no padding, text in subsequent columns appears directly next to the column separator, which impairs readability when compared to the previous “XHTML”  and “Before — HTML5” output.

If we leave it as is, customization would be required in most applications to either add padding or remove the column separators, but I think it would be better if the defaults were readable without customization.

eerohele:
>We might want to consider removing the default column separators for `<simpletable>`, `<choicetable>`, and `<properties>`.

I just added them because it was requested earlier. (See also my comment above regarding borders.)

I'm fine with either default, but since regular tables don't have row or column separators by default, it would be more consistent if simpletable (and specializations) didn't, either.

>Since there's also no padding, text in subsequent columns appears directly next to the column separator, which impairs readability when compared to the previous “XHTML” and “Before — HTML5” output.

The readability and the appearance of the tables is currently pretty atrocious, but I think improving that aspect could perhaps be done in a separate PR. I think it's just a bit of a slippery slope if we start working a lot with the appearance stuff in this PR — I'd like to keep this at a manageable size.

It is also a somewhat bigger question: is the point of DITA-OT to provide a starting point that is as easily customizable as possible, regardless of how it looks, or should we put some effort into making the default output reasonably readable and attractive?

From a technical point of view, the XHTML tables have padding because they use the legacy `@cellpadding` attribute, which is why the default padding is missing in HTML5.

robander:
> It is also a somewhat bigger question: is the point of DITA-OT to provide a starting point that is as easily customizable as possible, regardless of how it looks, or should we put some effort into making the default output reasonably readable and attractive?

Always a tough question but I think we have to put some effort into making the default output reasonably readable / attractive. It doesn't have to be flashy but it can't be bad. Too many people will never take the effort to customize, especially if their first impression is negative. There are a few people who come after me regularly for this, even those who know that default styles can/should be used as a basis for customization. I've been told it turns people off of DITA itself.

As to whether we do that in this PR or another, I have less of an opinion there. I agree it's nice to keep things manageable.

PR #2249, opened 2016-03-07
Add reference to PDF2 catalog into catalog-dita.xml

eerohele:
This is useful e.g. when profiling XSLT stylesheets. Without this,
running topic2fo_shell.xsl directly fails because the XSLT processor
can't resolve paths that use the cfg: scheme.


raducoravu:
Hopefully this does not interfere with the way in which a PDF customization folder can overwrite that catalog.

Issue #2243, opened 2016-02-26
Publishing fails when using term with keyref and chunking


raducoravu:
I tested using the latest DITA OT from the 2.2.3 branch on Windows, uploaded samples here:

http://www.oxygenxml.com/forum/files/chunkProblemWithTerm.zip

The failure is something like:

         [xslt] Processing C:\Users\radu_coravu\Desktop\zuzu\temp\xhtml\topics\register-license-key.dita to C:\Users\radu_coravu\Desktop\zuzu\out\xhtml\topics\register-license-key.html
         [xslt] C:\Users\radu_coravu\Documents\dot-develop\src\main\xsl\common\dita-utilities.xsl:394: Fatal Error! Base URI {file:/C:/Users/radu_coravu/Des...} is invalid: Invalid relative URI: Illegal character in path at index 2: ..\glossary\named-user.dita: file:/C:/Users/radu_coravu/Desktop/zuzu/temp/xhtml/topics/register-license-key.dita
         [xslt] Failed to process null
            
            BUILD FAILED
            C:\Users\radu_coravu\Documents\dot-develop\src\main\build.xml:41: The following error occurred while executing this line:
            C:\Users\radu_coravu\Documents\dot-develop\src\main\plugins\org.dita.xhtml\build_general.xml:162: Fatal error during transformation using C:\Users\radu_coravu\Documents\dot-develop\src\main\plugins\org.dita.xhtml\xsl\dita2xhtml.xsl: Base URI {file:/C:/Users/radu_coravu/Des...} is invalid: Invalid relative URI: Illegal character in path at index 2: ..\glossary\named-user.dita: file:/C:/Users/radu_coravu/Desktop/zuzu/temp/xhtml/topics/register-license-key.dita; SystemID: file:/C:/Users/radu_coravu/Documents/dot-develop/src/main/xsl/common/dita-utilities.xsl; Line#: 394; Column#: -1

raducoravu:
Looking at how the "register-license-key.dita" looks like in the temporary files folder, it has a link like this:

      <term keyref="glossentry_named-user" class="- topic/term " xtrf="file:/C:/Users/radu_coravu/Desktop/zuzu/tasks/unregister-license-key.dita" xtrc="term:1;42:97" href="..\glossary\named-user.dita" linking="none"/>

Because that "\" is used in the @href, the resolve-uri method breaks in the XSLT.

raducoravu:
[...]

PR #2274, opened 2016-03-23
Convert backslashes to slashes before passing through XSLT resolve-uri function

raducoravu:
Particular fix for #2243
Convert backslashes to slashes before passing through XSLT resolve-uri function


jelovirt:
This is a workaround in 2.2 branch and should be reverted from 2.3.

PR #2276, opened 2016-03-26
Use stable ID generation in HTML5

jelovirt:
The HTML5 output should use stable `@headers` and `@id` generation instead of the current one that is based on `generate-id()` which is not stable. Using a stable approach would allow e.g. DITA-OT site to have smaller change sets when `@headers` would not change between every conversion.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/xslhtml/functions.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/xslhtml/functions.xsl
@@ -38,10 +38,11 @@
     <xsl:param name="element" as="element()"/>
     
     <xsl:variable name="topic" select="$element/ancestor-or-self::*[contains(@class, ' topic/topic ')][1]" as="element()"/>
-    <xsl:variable name="topic-id" select="$topic/@id" as="xs:string"/>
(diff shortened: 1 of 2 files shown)


jelovirt:
@eerohele do you have time for a review?

jelovirt:
Added XSpec tests and copied ID related functions from XHTML to HTML5. This made tests simpler and in the future when we remove the dependency to XHTML this will be done in any case.

jelovirt:
The stable IDs could maybe be made even more stable by using `topic_id + closest_ancestor_element_id + counter_in_that_element`. What do you think? Then changes to topic content would be limited to closest ancestor element and the counter scope would no longer be the whole topic.

eerohele:
>@eerohele do you have time for a review?

I'm a bit late, but looks good to me, and a big :+1: for this change. I diff HTML outputs quite often when developing plugins and this change makes regression testing much more straightforward.

Pretty much the only (very minor) thing is that we *could* call [`dita-ot:get-closest-topic($element)`](https://github.com/dita-ot/dita-ot/blob/e5726607504ca72035b0a7a4570452a64b1b8f0d/src/main/xsl/common/dita-utilities.xsl#L384-L389) instead of doing `
$element/ancestor-or-self::*[contains(@class, ' topic/topic ')][1]`, but that'll require rejigging the XSpec tests a bit, because that function isn't currently available in the XSpec tests.

PR #2296, opened 2016-04-12
One more fallback font for Chinese

raducoravu:
Added SimSun as extra fallback font for Chinese and default language.


jelovirt:
CI fails for some reason, I will check why.

Issue #1876, opened 2015-03-25
Plugin Can't Override Templates in dita2html5Impl.xsl


shaneataylor:
The DITA-OT 2.0 HTML5 transform includes the topic short description as `<meta name="description">` but I need to change this to `<meta name="abstract">`.

My plugin already overrides dita2htmlImpl.xsl (using the dita.xsl.xhtml extension point) , but I can't override the templates in dita2html5Impl.xsl.

So far as I can determine, because of the structure of the toolkit's XSL for HTML5, the import precedence rules are preventing my override from being considered.

shaneataylor:
Running up against this problem again. With a little more knowledge, I see that the following HTML5 XSL extension points weren't fully implemented:

```
  <extension-point id="dita.xsl.html5" name="HTML/XHTML XSLT import"/>
  <extension-point id="dita.xsl.html5.toc" name="HTML/XHTML TOC XSLT import"/>
  <extension-point id="dita.xsl.html5.cover" name="HTML/XHTML Cover XSLT import"/>
```

Although they are declared in plugin.xml, they aren't supported by XSL template files in the HTML5 plugin, so the needed `<dita:extension>` elements aren't present in said templates to be resolved to imports for the overriding XSL, as in:

```
<dita:extension id="dita.xsl.html5" behavior="org.dita.dost.platform.ImportXSLAction" xmlns:dita="http://dita-ot.sourceforge.net"/>
```

shaneataylor:
Attached are the new/changed files (based on 2.2.4) that address this problem.

[org.dita.html5.zip](https://github.com/dita-ot/dita-ot/files/219962/org.dita.html5.zip)

jelovirt:
@shaneataylor please send a pull request instead

shaneataylor:
Just following up: does anyone have time to check & accept my pull requests for this and for the related issue #2312 ?

robander:
The pull request looks good to me - apologies for not taking a look more quickly.

robander:
Marking this as a bug (we declare extension points that are not usable), and for 2.2.5.

robander:
Merged the pull request #2313 last week and forgot to mark this closed - closing now.

PR #2305, opened 2016-04-15
Implement HTML5 XSL extension-points #1876

shaneataylor:


jelovirt:
When you introduce template generated files, you must remove the generated files from Git. Squash the changes into a single commit.

Issue #2307, opened 2016-04-18
Huge performance drop down when using huge database topics.


xephon2:
Hi, I noticed a huge performance drop down, if a huge database topic (a topic containing only blocks for reference) is used instead of multiple smaller database topics containing the same amount of blocks. The performance drop down is immense.

Test case:
[2307.zip](https://github.com/dita-ot/dita-ot/files/223172/2307.zip)

Build time for comparison using `DITA-OT 2.2.4`:
`map1.ditamap`: 1 database topic with 10000 dlentry elements: Total time: 428 minutes 56 seconds
`map2.ditamap`: 4 database topics with 2500 dlentry elements: Total time: 25 minutes 3 seconds
`map3.ditamap`: 10 database topics with 1000 dlentry elements: Total time: 3 minutes 56 seconds

Build time for comparison using `DITA-OT 1.8.5`:
`map1.ditamap`: 1 database topic with 10000 dlentry elements: Total time: 714 minutes 38 seconds

xephon2:
@eerohele, thanks a lot. :smiley:  :+1: 
Awesome!!! Your XSLT changes boost our publication time from > 60 min down to < 2 min.

infotexture:
@robander The fix was merged to `hotfix/2.2.5` with #2309. 

Can you add the milestone so it appears in the list of issues fixed in the release?

PR #2308, opened 2016-04-18
Fix conref performance issue #2307

eerohele:
`not(@conrefend = '')` returns `true` even for elements that don't have the `@conrefend` attribute. That would cause the expensive XPath expression at [conrefImpl.xsl#L435][1] to be evaluated every time, which
would cause the stylesheet execution to become prohibitively slow in cases where there are thousands of following siblings.

I also fixed other similar XPath expressions in the same stylesheet.

This change doesn't address cases where `@conrefend` is actually used, so the XPath expression at [conrefImpl.xsl#L435][1] should probably be refactored, too.

[1]:  https://github.com/dita-ot/dita-ot/blob/876053c933975fd4744a049dbb3c4c7433d3f417/src/main/xsl/preprocess/conrefImpl.xsl#L435

### DIFF
--- a/src/main/xsl/preprocess/conrefImpl.xsl
+++ b/src/main/xsl/preprocess/conrefImpl.xsl
@@ -73,9 +73,7 @@
   </xsl:template>
 
   <xsl:template match="@*" mode="get-source-attribute" as="xs:string?">
-    <xsl:if test="not(. = '')">
-      <xsl:value-of select="name()"/>
-    </xsl:if>
+    <xsl:value-of select="name()"/>
   </xsl:template>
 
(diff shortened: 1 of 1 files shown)


jelovirt:
@eerohele `hotfix/2.2.4` has been released. I will create a `hotfix/2.2.5`, could you cherry-pick 3cebad3 to that branch?

eerohele:
Ah, sorry. I'd missed the 2.2.4 release.

Created #2309 on top of `hotfix/2.2.5`.

infotexture:
We should remove the **2.2.5** milestone here, since this PR was closed without merging and replaced by #2309, which includes the same fix for #2307.

PR #2311, opened 2016-04-19
Avoid double dependency on copy-css task

raducoravu:
The copy-css task was called too early, before the user.csspath param was computed, leading to the CSSs being copied to a folder called "${user.csspath}" and not being used at all in the JavaHelp output.


robander:
:+1: for 2.2.5.

PR #2315, opened 2016-04-19
Use before and after operators rather than following-sibling 

drmacro:
Use the "<<" and ">>" operators to select the range rather than following-sibling.

Works in my (limited) testing.

I also added a check that the conref end target actually exists, which the current code does not have.

### DIFF
--- a/src/main/xsl/preprocess/conrefImpl.xsl
+++ b/src/main/xsl/preprocess/conrefImpl.xsl
@@ -818,53 +818,64 @@
 
     <xsl:choose>
       <xsl:when test="exists($conrefend)">
-        <xsl:for-each select="following-sibling::*[following-sibling::*[@id = $conrefend] or self::*[@id = $conrefend]]">
-          <xsl:choose>
-            <xsl:when test="@conref">
-              <xsl:apply-templates select=".">
-                <xsl:with-param name="source-attributes" select="$source-attributes"/>
-                <xsl:with-param name="conref-source-topicid" select="$conref-source-topicid"/>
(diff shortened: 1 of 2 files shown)


src/main/xsl/preprocess/conrefImpl.xsl
```diff
@@ -818,53 +818,64 @@
...
-            </xsl:otherwise>
-          </xsl:choose>
-        </xsl:for-each>
+        <xsl:variable name="current" as="element()" select="."/>
+        <xsl:variable name="conrefEndNode" as="element()?"
+          select="following-sibling::*[@id = ($conrefend)][1]"
+        />
+        <xsl:choose>
+          <xsl:when test="not($conrefEndNode)">
+            <xsl:message>Warning: Conref range: Cannot find range end element with ID "<xsl:value-of select="$conrefend"/>" </xsl:message>
```
robander: The message should use `output-message` and be pulled from `resources\messages.xml`, as with other generated messages - that way it also goes automatically into the docs (CC: @infotexture ). Most (all?) other messages in `conrefImpl.xsl` are created by processing the current node with a mode, so that overrides could tweak the message as needed -- see https://github.com/dita-ot/dita-ot/blob/develop/src/main/xsl/preprocess/conrefImpl.xsl#L819 for an example of that.

robander:
Looks good as far as I can tell, apart from one issue with the warning message that I noted with a comment in the commit.

drmacro:
OK, I can rework the message tomorrow. Out of time to work on this today.

robander:
Looks good to me.

Issue #2321, opened 2016-04-22
Topicset @type attribute warning in move-meta-entries


shaneataylor:
Version: DITA-OT 2.2.4
Transformation: HTML5 customization
Platform: OS X

Since upgrading my plugin to 2.2.4, I see warnings in my build log for topicsets:

> file:/Users/staylor/Dev/dita_source_files/m_a_common_topicsets.ditamap:12:69: [DOTX019W][WARN]: The type attribute on a topicref was set to 'topicset', but the topicref references a 'concept' topic. This may cause your links to sort incorrectly in the output. Note that the type attribute cascades in maps, so the value 'topicset' may come from an ancestor topicref.

My map doesn't specify `@type="topicset"` anywhere. I'm reasonably sure this is an error in xsl/preprocess/mappullImpl.xsl:

```
<xsl:template match="*" mode="mappull:verify-type-value">
    <xsl:param name="type" as="xs:string"/>          <!-- Specified type on the topicref -->
    <xsl:param name="actual-class" as="xs:string"/>  <!-- Class value on the target element -->
    <xsl:param name="actual-name" as="xs:string"/>   <!-- Name of the target element -->
    <xsl:param name="WORKDIR" as="xs:string">
      <xsl:apply-templates select="/processing-instruction('workdir-uri')[1]" mode="get-work-dir"/>
    </xsl:param>
    <xsl:choose>
      <!-- The type is correct; concept typed as concept, newtype defined as newtype -->
      <xsl:when test="$type=$actual-name"/>
      <!-- If the actual class contains the specified type; reference can be called topic,
           specializedReference can be called reference -->
      <xsl:when test="contains($actual-class,concat(' ',$type,'/',$type,' '))">
        <!-- commented out for bug:1771123 start -->
        <!--xsl:apply-templates select="." mode="ditamsg:type-mismatch-info">
          <xsl:with-param name="type" select="$type"/>
          <xsl:with-param name="actual-name" select="$actual-name"/>
        </xsl:apply-templates-->
        <!-- commented out for bug:1771123 end -->
      </xsl:when>
      <!-- Otherwise: incorrect type is specified -->
      <xsl:otherwise>
        <xsl:apply-templates select="." mode="ditamsg:type-mismatch-warning">
          <xsl:with-param name="type" select="$type"/>
          <xsl:with-param name="actual-name" select="$actual-name"/>
        </xsl:apply-templates>
      </xsl:otherwise>
    </xsl:choose>
[...]

PR #2322, opened 2016-04-22
Fix for issue #2321

shaneataylor:
This addresses #2321 by returning a `#none#` type for the targets of topicsetref elements (and their descendants). Current processing returns 'topicset' which causes build warnings because the referenced topics are concept, task, reference, etc.

### DIFF
--- a/src/main/xsl/preprocess/mappullImpl.xsl
+++ b/src/main/xsl/preprocess/mappullImpl.xsl
@@ -562,6 +562,20 @@ Other modes can be found within the code, and may or may not prove useful for ov
       </xsl:when>
     </xsl:choose>
   </xsl:template>
+  
+  <xsl:template match="*[ancestor-or-self::*[contains(@class,' mapgroup-d/topicsetref ')]]" 
+    mode="mappull:get-stuff_get-type" as="attribute()?">
+    <!-- <topicsetref> should not return a type -->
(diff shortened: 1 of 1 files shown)


robander:
Added a question in #2321 about this -- I think that either the current warning is valid, or else we need a more general fix. The more general one might actually be simpler than the one here; today, the code looks for a target that exactly matches `@type` (such as `type="topicset"` that refers to an actual `<topicset>`). If that fails, it looks for `@type/@type` in the `@class` value on the target - such as "task/task" to see if your reference is to a task or specialized task. We could make this more general and simply look for `/@type` -- this way `type="topicset"` would be valid for a reference to both `<topicset>` or to a specialized `<topicset>`.

drmacro:
I think that only looking for "/@type" is appropriate. The way the @type attribute is defined is somewhat ambiguous. 

The one wrinkle would be if the value of @type matches "{name}/{name}" then the check should be for "{name}/{name}", that is, an exact match for the module/name item in the @class value.

robander:
:+1: to all of what Eliot said.

robander:
Lots of discussion over in #2321 explaining why the fix should applied earlier in the processing stream (during `mapref` step rather than `mappull`). Marking this closed in expectation of a different fix.

PR #2325, opened 2016-04-27
Correct htmlflag include to import

drmacro:
This fixes the bug with including rather than importing the htmlflag.xsl module, which makes it impossible to override.

This shouldn't have any effect on existing code.


robander:
Chatted with Eliot about this one a little - should be easy to accept the actual change, once the commit history is sorted out: https://github.com/dita-ot/dita-ot/pull/2325/commits/d2d63285e14ed6d0b0e5754f3a686b027c9c2c3e

infotexture:
@drmacro @robander A bit of Git forensics was necessary to figure out what's going on here:

Looks like the `drmacro:feature/fix-htmlflag-import` branch was created using the `develop` branch as `<start_point>`, and then `hotfix/2.2.5` was merged into the feature branch with 2535763.

_(That's why all develop commits beginning with 5d9d3db appear in the list of changes above.)_

If merged as is, this would merge all of `develop` into the 2.2.5 hotfix, which is not what you intended.

I suggest we close this PR without merging, and instead:

1. Create a new feature branch from the tip of `hotfix/2.2.5` _(currently 5ce7a65)_ 
2. Cherry pick the changes from d2d6328 to the new feature branch
3. Create a new pull request from the new feature branch, using:
    * `dita-ot/dita-ot` as base fork, and `hotfix/2.2.5` as base branch
    * your fork as head fork, comparing to your new feature branch

When creating the new PR, only a single commit (with the changes from d2d6328) should appear.

robander:
Thanks for the deep dive, @infotexture 

I suspect we can skip the cherry-pick and just make the change -- @drmacro since the full extent of the change is just one line, it would be easier to create a new branch off of `hotfix/2.2.5`, make that change again, and do the pull request against `hotfix/2.2.5`.

Issue #1709, opened 2014-07-09
Re-check and rephrase messages.xml


raducoravu:
The messages XML document:

     DITA-OT/plugins/org.dita.pdf2/resource/messages.xml

contains messages shown by the DITA OT when running. Some of them seem to have been written by a non-English speaker. 
Example:

       There are multiple index entry found which is opening range for ID="%1" but there is only one which close it or ranges are overlapping.

      There are multiple index entry found which closing range for ID="%1".

Maybe they should be reviewed.

robander:
Saw this when scanning through older issues - I can take it now unless @infotexture wants to grab it, or has perhaps already done some work on it?

infotexture:
@robander — Please do. 

You probably know more about the context in which each of these messages is likely to appear. 

Let me know _(or send a docs PR)_ if there are any for which you'd like to add **Additional details** in docs via [DITA-messages-details.xml](https://github.com/dita-ot/docs/blob/develop/user-guide/DITA-messages-details.xml) .

robander:
I'll work on updating them. Some today have content in the `<response>` section that is really more appropriate for Additional detail, as in the following, where the `<response>` is not actually a response:

```xml
  <message id="DOTA067W" type="WARN">
    <reason>Ignoring index-see '%1' inside parent index entry '%2' because the parent indexterm contains indexterm children.</reason>
    <response>According to the DITA Specification, the index-see element should be ignored if the parent indexterm contains other indexterm children.</response>
  </message>
```

robander:
While working on this I noticed this in the PDF messages file:
```xml
  <!-- DOTA067W and DOTA068W actually come from Java code inside the PDF plug-in.
       DOTA prefix is innacurate. No reason to change at this point. -->
  <message id="DOTA067W" type="WARN">
    <reason>Ignoring index-see '%1' inside parent index entry '%2' because the parent indexterm contains indexterm children.</reason>
    <response>According to the DITA Specification, the index-see element should be ignored if the parent indexterm contains other indexterm children.</response>
  </message>
...
```

That makes me a bit nervous, to knowingly reuse the message prefix, even if this does jump 60 digits ahead of existing numbers. I won't worry about that now, but may open an issue for 2.3 or 3.0 to try and change message numbers so that they're unique to a plugin, leaving no chance for overlap.

robander:
It is no longer possible to reach this message, as preprocessing removes the empty href attribute:
```xml
[...]

PR #2327, opened 2016-04-28
Update message text and remove obsolete messages #1709

robander:
Update for #1709 about bad message text - this pull request fixes message text, and will go along with an expected update that moves extra information from messages into the guide. The update also removes two messages that no longer exist in the PDF code.


robander:
This commit has associated doc changes in a pull request at `docs/hotfix-2.2.5`: https://github.com/dita-ot/docs/pull/95

PR #2329, opened 2016-04-28
Made include of htmlflag.xsl into an import and moved before extensio…

drmacro:
…n point

Tested with my reworked org.dita-community.flagging.status plugin and verified that the override that updates `@domains` was applied.


robander:
Agree this should have been `xsl:import` instead of `xsl:include`, and I'm fine considering that a bug. I also can't think of any negative repercussion from moving it ahead of the import - as you say, this just means that what you previously couldn't override, you now can.

Issue #2317, opened 2016-04-21
PDF: Values set for @expanse attribute are not used


keberlein:
For the PDF2 transformation and the codeblock, fig, and pre elements, there is no difference when expanse="page" versus expanse="column". 

I would have expected expanse="page" to force the element to be rendered aligned with the page margin (where H1 begins) as opposed to being rendered aligned with the indentation that is specified for the content.

Per the DITA 1.3 spec at http://docs.oasis-open.org/dita/dita/v1.3/os/part1-base/langRef/attributes/displayAttributes.html#display-atts:

> **@expanse**
    Determines the horizontal placement of the element. Allowable values are:
> 
>   **page**
        Places the element on the left page margin for left-to-right presentation, or right page margin for right-to-left presentation. 
    **column**
        Aligns the element with the current column margin 
   **textline**
        Aligns the element with the left (for left to right presentation) or right (for right to left presentation) margin of the current text line and takes indention into account. 
    **spread**
        Indicates that, if possible, the object should be rendered across a multi-page spread. If the rendition target does not have anything corresponding to spreads then spread has the same meaning as "page". 
    **-dita-use-conref-target**
        See Using the -dita-use-conref-target value for more information. 
> 
> In DITA tables, in place of the @expanse attribute used by other DITA elements, the @pgwide attribute is used in order to conform to the OASIS Exchange Table Model. The @pgwide attribute has a similar semantic (1=page width; 0=resize to galley or column).
> 
> Some DITA processors or output formats might not be able to support all values.

----
Does the DITA-OT transformation NOT support the @expanse attribute?

I have a ZIP that contains a test set of files, but I cannot attach it here.

robander:
I don't see any support for `@expanse` on any element. It looks like it should be present on these (possible I've missed some):

- fig
- pre, and specializations (codeblock, msgblock, screen all have their own overrides in PDF)
- lines
- simpletable

We do currently support `@pgwide` on `<table>` -- when set to page, we add `start-indent="0"  end-indent="0" width="auto"` to get the full page spanning. [This doesn't seem to work in the initial chapter from a bookmap, as it just causes the table to expand to the full column, but I'm using it as the basis for `@expanse` support at the moment.]

robander:
[...]

PR #2332, opened 2016-04-29
Support page expanse from display attributes #2317

robander:
Fix support for `expanse="page"` on elements that use the attribute. Set up attribute sets for other values, though they are empty for now.


robander:
After thinking on this more, it should really be aimed at 2.3, rather than at a patch.

Issue #1190, opened 2012-05-02
Tables in RTF output are damaged, RTF output cannot be opened


sorin-ristache:
The RTF transformation (the wordrtf transtype) cannot create correct output for tables. If a topic contains a table the RTF output cannot be opened in RTF reader applications like Windows Wordpad.

For reproducing the problem just transform to RTF any DITA map that includes a topicref that includes a table like:

    <table frame="none">
      <title>Flowers</title>
      <tgroup cols="3">
        <colspec colname="c1" colnum="1" colwidth="150pt"/>
        <colspec colname="c2" colnum="2" colwidth="100pt"/>
        <colspec colname="newCol3" colnum="3" colwidth="200pt"/>
        <thead>
          <row>
            <entry>Flower</entry>
            <entry>Type</entry>
            <entry>Soil</entry>
          </row>
        </thead>
        <tbody>
          <row>
            <entry>Chrysanthemum</entry>
            <entry>perennial</entry>
            <entry>well drained</entry>
          </row>
          <row>
            <entry>Gardenia</entry>
            <entry>perennial</entry>
            <entry>acidic</entry>
          </row>
          <row>
            <entry>Gerbera</entry>
            <entry>annual</entry>
            <entry>sandy, well-drained</entry>
          </row>
          <row>
            <entry>Iris</entry>
            <entry>perennial</entry>
            <entry>slightly acidic</entry>
          </row>
          <row>
            <entry>Lilac</entry>
            <entry>perennial</entry>
            <entry>alkaline</entry>
          </row>
          <row>
            <entry>Salvia</entry>
            <entry>perennial</entry>
            <entry>average</entry>
          </row>
          <row>
            <entry>Snowdrop</entry>
            <entry>perennial</entry>
            <entry>humus-rich</entry>
          </row>
        </tbody>
      </tgroup>
    </table>

robander:
The pull request #1589 resolves this while proposing (effectively) an entire new plugin with a new design. The current suggestion in that pull request is to keep that plugin separate from the core project; assuming that path is taken, I'm not sure what to do with this feature. The existing RTF plugin has gotten very little attention and I do not expect a fix for this, but I don't like the fact that it crashes. My best idea at this point is to update the code so that it simply ignores non-proportional column widths.

robander:
Changing the tag from "feature" to "bug" -- all I'm attempting to do at this point is fix the build so that it does not crash with the sample table above.

[...]

PR #2333, opened 2016-04-29
Avoid crash based on colwidth as reported in #1190

robander:


robander:
Guessing there might be a more optimized fix, but the RTF code is not optimal in any way, so I consider this good enough. RTF is unlikely to get much more attention, so I'm planning to go with this...

PR #2341, opened 2016-05-05
Generate startcmd files during integration

jelovirt:
Generate `startcmd.sh` and `startcmd.bat` during integration. This allows us to get the correct JAR files from `lib` and plugins into the classpath.


robander:
:+1:

PR #2342, opened 2016-05-07
Ignore files generated from templates since #2313

infotexture:
* `src/main/plugins/org.dita.html5/xsl/dita2html5.xsl`
* `src/main/plugins/org.dita.html5/xsl/map2html5-cover.xsl`
* `src/main/plugins/org.dita.html5/xsl/map2html5toc.xsl`


jelovirt:
Merge by cherry-picking into bc4f94b15025365948223c9c28fcc9ab43ef5b8d.

PR #2343, opened 2016-05-09
Add missing variables to original string files

robander:
The Dutch variables file for PDF has always been missing a few strings, which will throw `DOTX001W` when these constructs are used:
- All strings for Glossary (translated heading, and header/footer variables)
- List of tables
- List of figures

In addition, German, Italian, French, and Spanish are missing the variable "Table of Contents Notices" that will cause `DOTX001W` when using a notices topic. Most other languages define this variable but leave it empty, so doing the same here.


jelovirt:
Rebase to clean history

robander:
OK - completed rebase - ready for review / merge.

jelovirt:
LGTM

PR #2344, opened 2016-05-09
Add Kazakh, Korean, Urdu, and Traditional Chinese support for PDF

robander:
These four languages are the last remaining that we support for XHTML, but not for PDF. 

Initial commit adds support for variables / generated text.

Next will be the `i18n` files and index. Hoping to get all of these in for 2.3, will make it as far as I can with them.


robander:
Most of our index collation files are relatively small (10 or 15K). The current zh-CN sort makes use of character ranges so it is also small.

I'm working on Korean right now, and I'm ending up with a giant collation file (a few hundred K), even when consolidating any ranges within a single group. Collation is based on the charts here: http://developer.mimer.com/charts/korean.htm

Don't know that there is any other option, but it looks like the file will be big. The zh-TW file, last one remaining after this, looks like it will be even larger: http://developer.mimer.com/charts/chinese_zhuyin.htm

There is a lot of white space (if we keep indenting as in other files), which means they will be much smaller in the distribution zip, but will still expand to a noticeable size. Not sure yet how well these will impact processing of Korean or Traditional Chinese docs, but again, I'm not sure of an alternative. @drmacro I know you've worked on indexing in the past - any thoughts?

jelovirt:
One option is to split CKJV support into a separate plugin.

robander:
Yes, could do that. It feels awkward to do it just because of the size of these two files, given that `ja` and `zh-CN` have been supported natively in the plugin from early on; I don't know that it would make sense to pull those out. FWIW, the indented collation file for Korean is coming in at nearly 700K. Without indent (but with each item still on a newline for readability), it goes down to 530K.

drmacro:
It's been too long since I actively worked on my old i18n library but I don't remember the collation and grouping configurations for any language to be that big, but it also leveraged ICU4J and may have just used whatever collation configuration it has for Korean.

When I get home I could probably dig up my old code from archives if it's useful.

robander:
Other index sorts I've worked on did not have this problem, but the design here differs considerably from my old programs. For those I let ICU sort everything, then inserted headings between boundaries. The PDF design forces you to set up every group with the heading, then ICU takes over within the group. Most of the Korean characters don't seem to be in sequence within a group (at least based on the Mimer reference, which has been good for languages I know more about); the result is that I have to list them all independently.

robander:
Looks like I was making the Korean sort more complicated than I should have - I don't think I was reading that chart properly. Checked in a simple collation that appears to work for a large test document. Now working on Traditional Chinese.

robander:
Our older SGML based index uses a preprocess routine to analyze and group terms to determine headings (using the big5 character set, not Unicode), rather than using the simple character based collation in our PDF2 process. I don't anticipate getting that in place for 2.3, so I've checked in a basic sort. It will group all Chinese characters up front, followed by the usual "Special characters" and "Numerics", and ending with the usual Latin character sort with group headings.

robander:
@infotexture - once this gets integrated, we will have parity between XHTML and PDF for the list of languages supported.

robander:
Any objections? If no, I'll merge this one tomorrow. It should be pretty straightforward and parallels other new PDF language support. The only extra documentation item worth noting is that the index support for `zh-TW` is not quite complete; terms should be sorted appropriately, but will not have headings or bookmarks for the Chinese terms in the index.

src/main/plugins/org.dita.pdf2/cfg/fo/i18n/ko.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- 
+Copyright © 2004-2006 by Idiom Technologies, Inc. All rights reserved.
```
jelovirt: This isn't Idiom code, thus remove Idiom copyright and just use the normal OT statement.

src/main/plugins/org.dita.pdf2/cfg/common/vars/zh_TW.xml
```diff
@@ -0,0 +1,344 @@
...
+    <variable id="other Note Image Path"></variable>
+
+    <!-- UNUSED: The title of the links to child topics -->
+    <variable id="Child topics"></variable>
+
+    <!-- UNUSED: The title of the links to Related references -->
+    <variable id="Related references"></variable>
+
+    <!--
+       UNUSED IN PDF2: Strings for online-help publishing.
```
jelovirt: Should we remove these unused strings in one comment later on?

robander: I was thinking I would do a single commit / pull request that marked these "deprecated" across all languages, then remove from every language in 3.0. I don't think there's a clear best practice about adding them in these 4 languages. I added them in all of the other languages for 2.2 / 2.3 just because they were part of English, and it seemed if they're in English they should be present in others.

src/main/plugins/org.dita.pdf2/cfg/fo/i18n/ko.xml
robander: Good point. How do you want me to handle the commits at that point. Earlier ones were mostly separated by what they were doing - add variables, add i18n, add individual index. I'm fine with squashing those all into one commit, along with this update, FWIW.

robander:
Merged the commits into one.

jelovirt:
LGTM

PR #2350, opened 2016-05-12
Support PDF bookmarks for index letter headings

robander:
Many of the indexes I work with can grow quite large, spanning many pages. Currently, when using an index, the PDF bookmarks only have a single entry for the index. This pull request adds support for letter headings under the "Index" bookmark, with one entry for each letter heading that appears in the index.

Details:

There is a new parameter in the XSL, `bookmarks.index-group-size`. 
* When set to 0, no letter bookmarks will appear.
* When set to any other number, bookmarks will appear if the total number of entries equals or exceeds the value of `bookmarks.index-group-size`. Default is currently set to 100, so that if the index only spans one or two pages, no extra bookmarks will appear.
* Setting to 1 will force bookmarks to appear for any document that uses an index.

While doing this, I refactored `bookmarks.xsl` just slightly, to move index processing out of the larger `createBookmarks` named template. This makes it easier to override or customize, should anyone need to do so.

The only change for `index.xsl` is to add an ID to each letter group, which is needed to enable the bookmarks.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/bookmarks.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/bookmarks.xsl
@@ -27,18 +27,25 @@ These terms and conditions supersede the terms and conditions in any
 licensing agreement to the extent that such terms and conditions conflict
 with those set forth herein.
 
-This file is part of the DITA Open Toolkit project hosted on Sourceforge.net. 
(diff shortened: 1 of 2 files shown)


robander:
Any objections? This should be pretty straightforward and responds to the code review comments in the original pull request. If no objections, I'll merge tomorrow.

jelovirt:
LGTM

Issue #2326, opened 2016-04-28
Fix PDF plugin message "PDFX063W"


robander:
The PDF code contains this section:

```xml
  <xsl:template name="brokenLinks">
    <xsl:param name="href"/>
    <!-- FIXME: There is no message PDFX063W -->
    <xsl:call-template name="output-message">
      <xsl:with-param name="msgnum">063</xsl:with-param>
      <xsl:with-param name="msgsev">W</xsl:with-param>
      <xsl:with-param name="msgparams">%1=<xsl:value-of select="$href"/></xsl:with-param>
    </xsl:call-template>
  </xsl:template>
```

Note the FIXME section -- we should set up message text for this if it is still useful, or remove it if it is not.

robander:
I can't manage to trigger the message.

The message is part of the named template `brokenLinks`. This is called from `buildBasicLinkDestination` which sets up `@external-destination` or `@internal-destination`.

The logic in `buildBasicLinkDestination` looks first for external links, then for peer, then for any link including #, and finally has a failsafe call to `brokenLinks`. 

As far as I can tell, every link or xref that is not external or peer will end up converted to something like `href="#unique_N"` in the merged file. Tested with valid DITA destinations, invalid DITA destinations, missing and valid local non-DITA files. In all cases, the merged file ends up with `#unique_N`. There is no case where the message can get called. And because the actual message does not exist, we do not have any message text to suggest why it was there to begin with. 

I suggest we remove the message. The only other alternative would be to add message text such as: "We do not know how you got this. Please share your source so we can figure it out. Thanks!"

robander:
Closed with #2351

PR #2351, opened 2016-05-12
Remove broken message that cannot be triggered #2326

robander:
Message `DOTX063W` is called but is undefined, along with a `FIXME` note. It appears that there is no way to arrive at this message; removing it as discussed in #2326.


jelovirt:
LGTM

PR #2352, opened 2016-05-12
Support page expanse from display attributes #2317

robander:
Add support for `expanse="page"` in PDF, and set up attribute sets for other values though they remain empty for now. Mirrors the way we currently support `@frame` on the same elements.

The `<msgblock>` element is part of the same group of elements that uses `display-atts`, but did not previously support any value. So, this PR also adds support for `@frame` and `@scale` on that element, the same way that they are supported on the related elements.


jelovirt:
LGTM

Issue #1116, opened 2012-04-25
args.eclipsehelp.toc/TOCROOT not used by map2eclipse.xsl


jelovirt:
Converted from [SourceForge issue 1568130](http://sourceforge.net/support/tracker.php?aid=1568130), submitted by dave_resch

The args.eclipsehelp.toc property is not used to set the name of the toc.xml file generated by  map2eclipse.xsl. However, it is passed as a param  (TOCROOT) to the map2plugin.xsl, so the resulting  plugin.xml file refers to the toc.xml file as  ${args.eclipsehelp.toc}.xml.

Thus, if you set args.eclipsehelp.toc to a non-default  value, you get a plug-in that points to a non-extant  toc.xml file.

robander:
Assigning to myself to either fix or close in 2.3 - I'm not sure what still applies from this issue.

robander:
Checking with 2.2.4 and confirmed this is still an issue. The TOC file now defaults to the name of the map, rather than to `toc.xml`. If you specify `args.eclipsehelp.toc`, the generated TOC file is unchanged, but the map references the specified value plus the `.xml` extension.

robander:
Given that this has been broken since 1.3, and the updated behavior (without the parameter) is better than the old behavior, my preference is to remove the broken parameter entirely in 2.3. I don't think it is necessary to deprecate the parameter, given that it has not worked for 7 or 8 years. Instead, we should remove it, and stop passing it to `map2plugin.xml`; @infotexture note that this also means removing it from the docs: http://www.dita-ot.org/dev/parameters/parameters-eclipsehelp.html

robander:
Submitted #2354 to remove the parameter - @infotexture I forgot that removing the parameter from `plugin.xml` would remove from the docs, so nevermind about that...

PR #2354, opened 2016-05-13
Remove obsolete parameter args.eclipsehelp.toc

robander:
As discussed in #1116, removing an obsolete parameter that has not been broken since 1.3.


jelovirt:
LGTM

Issue #2298, opened 2016-04-13
Link to CSS not properly computed for chunks


raducoravu:
Working with DITA OT 2.2.2.
I have uploaded a sample project here:

https://www.oxygenxml.com/forum/files/path2projectAndChunking.zip

When you produce XHTML from the main DITA Map, a unique file called "chunk....html" results in the output folder.
It does not properly point to the CSS, it points to it something like:

``` xml
<link rel="stylesheet" type="text/css" href="../commonltr.css" />
```

Looking in the temporary files folder, the chunk.dita contains the path 2 project URI something like:

``` xml
<?path2project-uri ../?>
```

but it should actually be:

``` xml
<?path2project-uri ./?>
```

So probably this issue comes from the Java stage.

raducoravu:
My fix breaks CSS links with chunking in other cases so I will keep digging.

raducoravu:
So the minimal example would be something like:

     <keydef href="Topics/target.dita" keys="target"/>
    <topicref keyref="target" chunk="to-content"/>

In the Java code the method "org.dita.dost.writer.ChunkTopicParser.processChunk(Element, File)" which processes the chunk checks at some point if the target.dita already exists on disk:

                    if (outputFileName.exists() && !MAP_MAP.matches(classValue)) {
                        final File t = outputFileName;
                        outputFileName = resolve(filePath, generateFilename());
                        conflictTable.put(outputFileName.getPath(), t.getPath());
                    }

and in our case the topic exists because it was referenced as a resource-only topic by the keydef element. So the name of the output file remains "Chunkxxx.html" + it remains in the root directory and the path to project PI inside it is invalid.

raducoravu:
:+1: Another end user encountered the same problem. Generated chunked topic had "Chunkxxx.html" name, path to CSS is invalid, path to referenced images is invalid.

raducoravu:
Can no longer reproduce problem using DITA OT 3.2.

PR #2355, opened 2016-05-13
Make Path to project relative to chunk file

raducoravu:
Possible fix for:
https://github.com/dita-ot/dita-ot/issues/2298

The path to project PI inside the chunk file needs to be computed relative to the chunk file and not to the original file location.


raducoravu:
Closing as invalid. Leads to side effects.

Issue #1476, opened 2013-03-14
TopicMerge sets unexpected @xml:lang attribute to topic.


ToshihikoMakita:
I'm facing a serious problem found in DITA-OT 1.7.3. The problem is that DITA-OT TopicMerge sets unexpected `@xml:lang` attribute to topic. Here is a example.

[bookmap: sample_ja.bookmap]
```xml
<bookmap xml:lang="ja" id="GUID-BC13DC41-C380-4E24-9047-B0E6018F4D86">
  <booktitle id="GUID-877192EE-6DCD-4695-AD85-E1F54ED545E5"> 
    <mainbooktitle>テスト</mainbooktitle> 
  </booktitle> 
  ...
  <chapter format="dita" href="sample.xml" scope="local" navtitle="サンプル"/>
</bookmap>
```

[sample.xml]
```xml
<concept id="concept_D941EF43C15A40D0BF155B7C5B8003F0">
  <title>コンセプトサンプル</title>
</concept>
```

[sample_ja_MERGED.xml]
```xml
<concept ditaarch:DITAArchVersion="1.2" domains="(topic concept)                            (topic hi-d)                             (topic ut-d)                             (topic indexing-d)                            (topic hazard-d)                            (topic abbrev-d)                            (topic pr-d)                             (topic sw-d)                            (topic ui-d)    " 
         class="- topic/topic concept/concept " xtrf="D:\DITA-OT1.7.3\samples\sample.xml" xtrc="concept:1;3:56"
         xml:lang="en" oid="concept_D941EF43C15A40D0BF155B7C5B8003F0" id="unique_1">
  <title class="- topic/title " xtrf="D:\DITA-OT1.7.3\samples\sample.xml" xtrc="title:1;4:8">コンセプトサンプル</title>
...
</concept>
```

As shown above TopicMerge sets topic/@xml:lang="en" as default value.

Until now we use DITA-OT 1.5.3 and our stylesheet honors `bookmap/@xml:lang`. The stylesheet inherit `bookmap/@xml:lang` to the topic and element level. So according to this situation, `bookmap/@xml:lang="ja"` should be honored.

Are there any TopicMerge option to avoid setting `@xml:lang="en"` as default value?

ToshihikoMakita:
Any progress on this issue?

This will not affect following user.

- CMS user (because CMS forces to specify topic/@xml:lnag explicitly.)
- English user
- PDF2 plug-in user (It only use map/@xml:lang only.)

[...]

PR #2356, opened 2016-05-13
Stop setting default lang on topics in merge file #1476

robander:
The `topic-merge` function adds `xml:lang="en"` to topics that have no language. Dropping this default allows PDF to use the map language as a default.


robander:
Need to update test case before this is ready.

jelovirt:
Could you squash?

jelovirt:
LGTM

robander:
Squashed, also removed additional one line comment that no longer applies.

PR #2357, opened 2016-05-16
Improve setting of default language

robander:
Our common language detection has a few problems today:

- We have a configuration property `default.language` but it is not used by any of our XSLT processing steps
- XSLT (in `dita-utilities.xsl`) hard codes a default of `en-us` (which conflicts slightly with the shipped value `en` in `configuration.properties`)
- Even when a topic or map sets `@xml:lang` on the root element, it is ignored for any variable that is pulled from the `/` context in XSLT. This is the case for several PDF variables, most noticeable are the strings for `Chapter` and `Index` within the PDF bookmarks. The reason is that processing looks for `@xml:lang` on ancestor or self, so no language is found from `/`.

This pull request makes the following updates:

- Pass in the configuration default to any XSLT step that is likely to do any language based processing. I've added it to `topicpull` and `flag` from the pre-process (both pull variables); it is also passed to all of the XSLT rendering steps that have any real NLS support today.
- Continue using the XSLT variable `$DEFAULTLANG` that has been used since DITA-OT 1.0, but with new logic:
 - If `@xml:lang` is present on the root element, that is the default. This fixes the problem of pulling variables from the `/` context.
 - Otherwise, use the configured DITA-OT default.

Worth noting: as a side effect, it is now much easier to test generated strings / variables in XHTML, by using a common set of topics with no `@xml:lang` and building with the property `default.language` to override the configured default.

### DIFF
--- a/src/main/plugins/org.dita.base/build_preprocess_template.xml
+++ b/src/main/plugins/org.dita.base/build_preprocess_template.xml
@@ -454,6 +454,7 @@
         <param name="TABLELINK" expression="${args.tablelink.style}" if="args.tablelink.style" />
         <param name="FIGURELINK" expression="${args.figurelink.style}" if="args.figurelink.style" />
         <param name="ONLYTOPICINMAP" expression="${onlytopic.in.map}" if="onlytopic.in.map"/>
+        <param name="defaultLanguage" expression="${default.language}"/>
         <dita:extension id="dita.preprocess.topicpull.param" behavior="org.dita.dost.platform.InsertAction"/>
         <xmlcatalog refid="dita.catalog"/>
       </xslt>
@@ -488,6 +489,7 @@
         <param name="BASEDIR" expression="${basedir}"/>
         <param name="OUTPUTDIR" expression="${output.dir}"/>
         <param name="DBG" expression="${args.debug}" if="args.debug"/>
+        <param name="defaultLanguage" expression="${default.language}"/>
         <dita:extension id="dita.preprocess.flag-module.param" behavior="org.dita.dost.platform.InsertAction"/>
         <xmlcatalog refid="dita.catalog"/>
       </xslt>
--- a/src/main/xsl/common/dita-utilities.xsl
+++ b/src/main/xsl/common/dita-utilities.xsl
@@ -12,7 +12,10 @@
 
   <xsl:include href="functions.xsl"/>
 
-  <xsl:param name="DEFAULTLANG">en-us</xsl:param>
+  <xsl:param name="defaultLanguage" select="'en'" as="xs:string"/>
+
+  <xsl:param name="DEFAULTLANG" select="if (/*/@xml:lang) then /*/@xml:lang else $defaultLanguage" as="xs:string"/>
+
   <xsl:param name="variableFiles.url" select="'plugin:org.dita.base:xsl/common/strings.xml'"/>
   
   <xsl:variable name="pixels-per-inch" select="number(96)"/>
--- a/src/main/plugins/org.dita.docbook/build_dita2docbook.xml
+++ b/src/main/plugins/org.dita.docbook/build_dita2docbook.xml
@@ -41,6 +41,7 @@
       classpathref="dost.class.path"
       style="${args.xsl}">
       <param name="outputdir" expression="${outputDir}" />
+      <param name="defaultLanguage" expression="${default.language}"/>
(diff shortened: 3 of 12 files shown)


src/main/xsl/common/dita-utilities.xsl
```diff
@@ -12,7 +12,10 @@
 
   <xsl:include href="functions.xsl"/>
 
-  <xsl:param name="DEFAULTLANG">en-us</xsl:param>
+  <xsl:param name="ditaot.default.language" select="'en'"/>
```
jelovirt: Remove `ditaot.` prefix and use camel-case.

robander:
I was working on updating test cases for this when I realized our existing function `dita-ot:get-first-topic-language` is broken. My change caused the unit test to fail for that function, as expected, but for the wrong reason. Test input used languages `en-us` and `en-US`, checked for a return value of `en-us` on each. This passed, but not because language was properly detected; instead, it was passing because `en-us` was the default. The function should be fixed (the test is good, but it was sort of a false pass).

src/main/plugins/org.dita.base/build_preprocess_template.xml
```diff
@@ -454,7 +454,7 @@
         <param name="TABLELINK" expression="${args.tablelink.style}" if="args.tablelink.style" />
         <param name="FIGURELINK" expression="${args.figurelink.style}" if="args.figurelink.style" />
         <param name="ONLYTOPICINMAP" expression="${onlytopic.in.map}" if="onlytopic.in.map"/>
-        <param name="ditaot.default.language" expression="${default.language}"/>
+        <param name="defaultLanguage" expression="${default.language}"/>
```
xephon2: Why using camel case? Other parameters are upper-case

robander: For that program they're all upper case, but there is no one convention across all of our XSLT. Most of the other programs use a mix, with camelCase quite common. Whatever name is used, it has to be the same in all cases where we use it.

src/main/xsl/common/dita-utilities.xsl
```diff
@@ -12,9 +12,9 @@
 
   <xsl:include href="functions.xsl"/>
 
-  <xsl:param name="ditaot.default.language" select="'en'"/>
+  <xsl:param name="defaultLanguage" select="'en'"/>
 
-  <xsl:param name="DEFAULTLANG" select="if (/*/@xml:lang) then /*/@xml:lang else $ditaot.default.language"/>
+  <xsl:param name="DEFAULTLANG" select="if (/*/@xml:lang) then /*/@xml:lang else $defaultLanguage"/>
```
xephon2: Would it be possible to merge both `<xsl:param>` to something like this?

```xml
<xsl:param name="defaultLanguage" select="if (/*/@xml:lang) then /*/@xml:lang else 'en'"/>
```

robander: I don't think so. The problem is that `defaultLanguage` should always be passed in from the build, from the value in `configuration.properties`. It represents the default setting for the toolkit. The only place it is / should be used in the code is as a fallback for `$DEFAULTLANG` when no value is found on the map. If we condense to one parameter, the value passed from the scripts will always override the value on the map.

robander:
This one hit a large number of integration test files that did not set a default language, and now use `en` (from the configuration file) instead of the old `en-us` (from `dita-utilities.xsl`). Hoping to have all that resolved tomorrow.

src/main/xsl/common/dita-utilities.xsl
```diff
@@ -12,7 +12,10 @@
 
   <xsl:include href="functions.xsl"/>
 
-  <xsl:param name="DEFAULTLANG">en-us</xsl:param>
+  <xsl:param name="defaultLanguage" select="'en'"/>
```
jelovirt: Add datatype with `@as`.

src/main/xsl/common/dita-utilities.xsl
```diff
@@ -12,7 +12,10 @@
...
+
+  <xsl:param name="DEFAULTLANG" select="if (/*/@xml:lang) then /*/@xml:lang else $defaultLanguage"/>
```
jelovirt: Add datatype with `@as`.

jelovirt:
LGTM

Issue #2359, opened 2016-05-17
hyperlinks to fn do not function in pdf


leighw:
Tested with OT 2.2.4, dita2pdf transform. Given this syntax in a topic:

```xml
<p>This is a paragraph with a footnote in it. <fn id="footnote-1">Here is a footnote.</fn>>/p>
<p>And here is a paragraph with a cross-reference to that footnote. <xref
 href="#topica/footnote-1" format="dita" type="fn"/></p>
```

The expected result is a functioning hyperlink to the footnote. No link is created. It's rendered in topic.fo as:
 
```xml
<fo:block space-after="0.6em" space-before="0.6em" text-indent="0em">Here is a
paragraph that cross-references a footnote.<fo:inline baseline-shift="super"
font-size="75%" keep-with-previous.within-line="always">1</fo:inline></fo:block>
```
In other words, plain superscripted text. The same syntax results in a functioning hyperlink in the dita2html transform.

robander:
I can reproduce this - with my simple topic, I get the correct footnote (superscript "1") but it is not a link, and presumably should be. Marked as a bug though this could arguably be called a feature, as I would guess it has never worked, and was apparently not part of the original PDF design.

```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE task PUBLIC "-//OASIS//DTD DITA Task//EN"
 "../../dtd/technicalContent/dtd/task.dtd">
<task id="changeoil" xml:lang="en-us">
<title>Changing the oil in your car</title>
<shortdesc>This isn't really about changing the oil anymore.</shortdesc>
<taskbody>
<context><p>Changing the oil regularly will help keep the engine in good condition. <fn id="testfn">this is a
test footnote</fn></p>
<p>Here is a link to the footnote: <xref href="#changeoil/testfn" format="dita" type="fn"/></p></context>
</taskbody>
</task>
```

jelovirt:
Implemented in #2364

PR #2361, opened 2016-05-17
Add linking for footnotes #2359

robander:
Suggested fix for issue #2359, would appreciate code review.

Adds `@id` within each `<fo:footnote-body>`. For simplicity, I used a generated ID in all cases.

Adds `<fo:basic-link>` around the footnote reference, whether generated or used by reference. The link goes to the ID within the footnote body.

Tested with FOP, with the sample file in #2329 (footnote with `@id`, footnote without).

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/common/index/ar.xml
+++ b/src/main/plugins/org.dita.pdf2/cfg/common/index/ar.xml
@@ -49,6 +49,7 @@
                 <group.key>Numbers</group.key>
                 <group.label>أرقام</group.label>
                 <group.members>
+                    <char.set>0</char.set>
                     <char.set>０</char.set>
                     <char.set>1</char.set>
                     <char.set>１</char.set>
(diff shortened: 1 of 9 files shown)


drmacro:
For this type of code where multiple things need to generate the same value for a reference or ID I like to put the ID-construction code in a local function.

robander:
Thanks. Any comments on this as a function? I've tested and got it working, so can redo the commit:
```
<xsl:function name="getFootnoteInternalID" as="xs:string">
      <xsl:param name="ctx" select="." as="element()"/>
      <xsl:text>fn</xsl:text>
      <xsl:sequence select="generate-id-($ctx)"/>
</xsl:function>
```

jelovirt:
That `getFootnoteInternalID` will generate two strings, it will not work. Use e.g. `concat` in `xsl:sequence`  to combine them.

robander:
Odd - it worked, gave me a linking PDF, but I'll update regardless as `concat` makes more sense and is what I originally had without the function.

robander:
Request moved to 2364 after this one was closed.

PR #2364, opened 2016-05-18
Add linking for footnotes #2359

robander:
Suggested fix for issue #2359 (originally submitted as pull request #2361, I mixed up a couple git commands and that one was automatically closed.)

Adds `@id` within each `<fo:footnote-body>`. For simplicity, I used a generated ID in all cases.

Adds `<fo:basic-link>` around the footnote reference, whether generated or used by reference. The link goes to the ID within the footnote body.

Tested with FOP, with the sample file in #2329 (footnote with `@id`, footnote without).

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -27,7 +27,7 @@ These terms and conditions supersede the terms and conditions in any
 licensing agreement to the extent that such terms and conditions conflict
 with those set forth herein.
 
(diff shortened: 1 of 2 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
```diff
@@ -2012,6 +2018,7 @@ See the accompanying license.txt file for applicable licenses.
                             <xsl:value-of select="count(key('enumerableByClass', 'topic/fn')[. &lt;&lt; current()]) + 1"/>
```
jelovirt: Could you extract this and duplicates of this into a function?

jelovirt:
LGTM

PR #2367, opened 2016-05-20
Reference checks not handling case where topic and non-topic element have same ID

drmacro:
Code was assuming a given ID would only occur once. User error can result in multiple instances of the same ID. Code should not fail utterly in that case.


drmacro:
This fix is bogus as there's a deeper problem in that the code assumes that topic IDs and non-topic IDs will be mutually exclusive, which they do not have to be.

Digging in deeper.

drmacro:
OK, refactored the code to centralize the topic reference lookup and correctly handle the case where topics and non-topic elements have the same ID. Also improved the error reporting and generally streamlined the code that was operating on the referenced topics.

It looks like the same problem may be in topicpull--checking that code now.

src/main/resources/messages_template.xml
```diff
@@ -689,6 +689,17 @@
     <reason>Conref range: Unable to find conref range end element with ID "%1".</reason>
     <response/>
   </message>
+
+  <message id="DOTX072W" type="WARN">
+    <reason>Failed to resolve topic ID "%1" to a topic in document "%2".</reason>
```
jelovirt: The location of error condition should not be part of the error message, it's populated from `@xtrf`.

drmacro:
Didn't realize that. Will correct.

src/main/resources/messages_template.xml
```diff
@@ -689,6 +689,17 @@
...
+    <response/>
+  </message>
+  
+  <message id="DOTX073W" type="WARN">
+    <reason>Got %1 topics for topic ID "%2" in document "%3". Using the first topic in the document.</reason>
```
jelovirt: This should be

```xml
<reason>Got %1 topics for topic ID "%2"".</reason>
 <response>Using the first topic in the document.<response>
```

drmacro:
In the case of message DOTX072W I want to include the URI of target document. Should I be getting it's URI from it's @xtrf value?

jelovirt:
I misunderstood, it's not the link source element, but rather target. Then it should just go to the message text as it already did.

drmacro:
OK, I updated the messaging so it reports the @xtrf value of the target document and also reports the @class value in the case where the target is not a topic.

drmacro:
With these latest changes the mappull processing appears to work correctly with my test doc (meaning that all the topicrefs are resolved, which they should be).

I'm still getting anomalous results from the topicpull process (reporting files as unavailable or unable to resolve link text when all the files should be present and available as the map passes Oxygen's validation).

drmacro:
This same general problem (and solution) applies to topicpullImpl.xsl as well. Will work on updating it as part of this pull request.

drmacro:
I know the commits need to be squashed but i'm not sure how or if I should do it since I've already pushed to this branch from my local dev environment.

drmacro:
Current code fails on three integration tests. Investigating.

jelovirt:
Closing as inactive.

Issue #2302, opened 2016-04-14
XHTML transtype problem (and solution): Content from RTL tables is aligned to the left 


dan-caprioara:
This is caused by DITA-OT 2.x generating a style attribute with "text-align:left" on all the table cells, no matter if they have or not the the align attribute set into the source,

**How to reproduce:**

Insert a table with arabic content into a topic. Put xml:lang="ar" and dir="rtl" on the root of the topic, then transform it to XHTML. You can use this fragment below, but the problem can be seen on english content in a RTL table also.
```

    <table id="table_chp_lxj_hl" rowsep="1" colsep="1" dir="rtl">
      <tgroup cols="2">
        <colspec colnum="1" colname="col1"/>
        <colspec colnum="2" colname="col2"/>
        <tbody>
          <row>
            <entry namest="col1" nameend="col2">النبات الحبي</entry>
          </row>
          <row>
            <entry namest="col1" nameend="col2"> النباتات الحبية
              المختلفة و منتجاتها</entry>
          </row>
        </tbody>
      </tgroup>
    </table>

```

The text from table cells are aligned to the left inside each of the cells.
This is wrong, it should be aligned to the right. For LTR tables it should be to the left.

The problem is in the file:

`DITA-OT2.x/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl`

in template:

`<xsl:template name="doentry">`

Here is computed the style attribute for each of the table entries. There are some default values for text-align and vertical-align set to "left" and "top". The default values should not be specified, so the browser can infer them from the parent elements/styles.

**The solution is:**
Remove the xsl:otherwise blocks.
```

 <xsl:variable name="align" as="xs:string?">
        <xsl:choose>
         <xsl:when test="@align">
           <xsl:value-of select="@align"/>
         </xsl:when>
          <xsl:when test="$group/@align">
            <xsl:value-of select="$group/@align"/>
          </xsl:when>
          <xsl:when test="$colspec/@align">
            <xsl:value-of select="$colspec/@align"/>
          </xsl:when>
          <!-- PATCH START   -->          
          <!--
          <xsl:otherwise>
            <xsl:value-of select="$table.align-default"/>
          </xsl:otherwise>
          -->
        </xsl:choose>
      </xsl:variable>
      <xsl:if test="exists($align)">
        <xsl:text>text-align:</xsl:text>
        <xsl:value-of select="$align"/>
        <xsl:text>;</xsl:text>
[...]

PR #2368, opened 2016-05-20
Remove default entry alignment to fix bidi #2302

robander:
See #2302 for details - if there is no alignment specified in the source files, do not place one in the X/HTML table entries, allowing the document language or direction to determine alignment.

### DIFF
--- a/src/main/plugins/org.dita.xhtml/xsl/xslhtml/tables.xsl
+++ b/src/main/plugins/org.dita.xhtml/xsl/xslhtml/tables.xsl
@@ -12,8 +12,6 @@
 
 <!-- =========== CALS (OASIS) TABLE =========== -->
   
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.xhtml/xsl/xslhtml/tables.xsl
```diff
@@ -459,7 +457,7 @@
             <xsl:value-of select="$colspec/@align"/>
           </xsl:when>
           <xsl:otherwise>
-            <xsl:value-of select="$table.align-default"/>
+            <!-- 2.3: Remove default (allow actual default to come from browser or CSS) -->
```
jelovirt: Remove this comment and the whole `xsl:otherwise`. We have version control for history.

jelovirt:
LGTM

PR #2369, opened 2016-05-21
HTML5: Use Sass `// comment` syntax per #2218

infotexture:
This PR replaces the C-style (multi-line) `/* comment */` syntax used to mark inline/block elements in the generated Sass partials for DITA domains, as these comments are passed through to the output even when empty rules that follow them are dropped by the compiler.

Switching these to Sass C++-style (single-line) `// comment` syntax reduces noise in the CSS output.

The old CSS comments have been preserved in the Sass source and are passed through to the CSS output on the assumption that they may be useful to users developing custom stylesheets.

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/domains/_abbreviated-form.scss
+++ b/src/main/plugins/org.dita.html5/sass/domains/_abbreviated-form.scss
@@ -1,3 +1,3 @@
-/* inline */
+// inline 
 .abbreviated-form {
 }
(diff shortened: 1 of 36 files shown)


infotexture:
This is a follow-up task to the Sass refactoring slated for inclusion in 2.3:

* Add Sass partial for XML mention domain #2195
* Add SASS shell files for DITA domains #2196
* HTML5: Refactor Sass & lint partials #2218

PR #2381, opened 2016-05-23
Ensures that having same ID for topic and non-topic element does not result in errors.

drmacro:
Refactored to use key tables for topic lookup by ID and centralize resolution of topic reference.

This is intended for inclusion in 3.0


jelovirt:
Closing as inactive

PR #2382, opened 2016-05-23
PDF: Remove background color from keycol

infotexture:
Currently, an `antiquewhite` background color is inconsistently applied to key column contents of `<simpletable>` and `<properties>`, but not in `<choicetable>`. 

Instead of filling the entire table cell, the background color is limited to the text content of the entry, leaving white artifacts in the cell padding as shown in the example below:

simpletable-keycol-bg-issue

This PR removes the background color from `<simpletable>` and `<properties>` key columns to sync presentation with `<choicetable>` and align PDF output with HTML-based formats.


infotexture:
We may also want to consider replacing the last vestige of `antiquewhite` in table heads with something a bit more neutral, such as the 6% gray shade `#f0f0f0`, which is already used for codeblock & screen backgrounds, but that is probably best treated as a separate issue.

robander:
I think having the text in bold is better than bold + shading that covers the text. I'd also be fine with shading that covers the full cell. No strong opinion on which is preferable. I have always found the color choice a bit odd.

jelovirt:
LGTM

PR #2383, opened 2016-05-23
Mark deprecated variables, tweak conref variable

robander:
This pull request covers related updates to the PDF variable files:

- [x] Update the "Content-reference" variable. I thought this was unused (and marked it as unused in new variable files). In fact, it is used when `@conref` did not resolve. I tweaked the comment to clarify the purpose, and changed the text from "Content-reference to" to the more accurate "Unresolved content reference to"
- [x] Mark the "Untitled section" variable as deprecated. This is not used in our code, is untranslated for almost every language, and I cannot see a case where we generate that in a default process.
- [x] Mark all of the online help variables as deprecated. I'm guessing (but cannot verify) that these were used by Suite-sol code unrelated to PDF, and that they stored them with PDF strings for convenience. As with the previous, these are untranslated for almost every language and never used by any DITA-OT code.

For the deprecated strings, my goal is to mark them deprecated in 2.3 so that we can easily remove them in 3.0.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/common/vars/ar.xml
+++ b/src/main/plugins/org.dita.pdf2/cfg/common/vars/ar.xml
@@ -154,14 +154,14 @@ See the accompanying license.txt file for applicable licenses.
          text in case it's not specified in source. -->
     <variable id="Cross-Reference">Cross-Reference to:</variable>
 
-    <!-- UNUSED: Default content-reference text prefix. Used to generate reference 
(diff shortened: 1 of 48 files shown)


jelovirt:
Merged locally and moved the deprecation comments to be on their own line. This makes line based diff easier to read and retains history for the actual variables.

robander:
This is marked as `3.0` but was merged for `2.3`, right? There should be a separate pull request (probably next week so I don't forget) that removes these for 3.0.

jelovirt:
@robander corrected milestone

PR #2384, opened 2016-05-24
PDF: Change table head background to neutral gray

infotexture:
This replaces the last vestige of "antiquewhite" with the 6% gray shade "#f0f0f0" that is used for codeblock & screen backgrounds.

See #2382.


jelovirt:
I'm not sure if this is the right solution either. Both `anitquewhite` and `#F0F0F0` are a matter of taste. I'd rather see the background colour removed completely, instead of changed into something that is not better or more right, just different.

raducoravu:
:+1: for removing it, I think the XHTML does not have it anyway so we would have consistency.

infotexture:
Also fine with me to remove it entirely. 

That would have the added benefit of aligning CALS table presentation with `<simpletable>` output, which currently has no shading in the table heads.

PR #2386, opened 2016-05-24
PDF: Remove background color from table heads

infotexture:
Sync default presentation with HTML-based output and `<simpletable>` per comments in #2384.

Closes #2384.


jelovirt:
LGTM

Issue #1777, opened 2014-11-04
1.3: Support @orient attribute on table


drmacro:
Proposal 13078

juliovaz:
Is there a target date for implementation?

robander:
This is not actively being worked on, so there is not a target date for implementation. I took a (very brief) look at it a few weeks ago when evaluating the backlog but did not have time to do anything for 2.3.

juliovaz:
Thanks for the update. Pulling together a presentation and want to make sure I speak the truth.

juliovaz:
Not sure why this is closed. The attribute doesn't do anything in PDF or HTML (not that you should need it for HTML).

jelovirt:
Do you mean the support in PDF is broken and it doesn't work?

juliovaz:
I tried on 2.2.4 and don't get land value to do anything

jelovirt:
Please look at this issue at Github site. The issue metadata on the right sidebar says "Milestone 2.3". That means this feature will be released in DITA-OT version 2.3 and is not available in earlier versions.

juliovaz:
Ah. I see. Thanks.

PR #2387, opened 2016-05-24
Add support for table @orient attribute #1777

eerohele:
[Test files](https://github.com/eerohele/dita-ot-issues/tree/master/fixtures/1777).

Sample output:

- [Antenna House](https://github.com/dita-ot/dita-ot/files/279806/root_ah.pdf)
- [FOP](https://github.com/dita-ot/dita-ot/files/279807/root_fop.pdf)

This is the smallest change that implements support for `@orient` per spec, so far as I can tell. Anything missing?

(Note to any potential users of this feature: Antenna House supports multi-page landscape tables — FOP doesn't.)

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/tables-attr.xsl
+++ b/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/tables-attr.xsl
@@ -72,6 +72,10 @@ See the accompanying license.txt file for applicable licenses.
   <xsl:attribute-set name="__tableframe__right" use-attribute-sets="common.border__right">
   </xsl:attribute-set>
 
+  <xsl:attribute-set name="table__container">
+    <xsl:attribute name="reference-orientation" select="if (@orient eq 'land') then 90 else 0"/>
+  </xsl:attribute-set>
+
   <xsl:attribute-set name="table" use-attribute-sets="base-font">
     <!--It is a table container -->
     <xsl:attribute name="space-after">10pt</xsl:attribute>
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
@@ -249,20 +249,22 @@
             <xsl:call-template name="getTableScale"/>
         </xsl:variable>
 
-        <fo:block xsl:use-attribute-sets="table">
-            <xsl:call-template name="commonattributes"/>
-            <xsl:if test="not(@id)">
-              <xsl:attribute name="id">
-                <xsl:call-template name="get-id"/>
-              </xsl:attribute>
-            </xsl:if>
-            <xsl:if test="exists($scale)">
-                <xsl:attribute name="font-size" select="concat($scale, '%')"/>
-            </xsl:if>
-            <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-startprop ')]" mode="outofline"/>
-            <xsl:apply-templates/>
-            <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-endprop ')]" mode="outofline"/>
-        </fo:block>
+        <fo:block-container xsl:use-attribute-sets="table__container">
+            <fo:block xsl:use-attribute-sets="table">
+                <xsl:call-template name="commonattributes"/>
+                <xsl:if test="not(@id)">
+                  <xsl:attribute name="id">
+                    <xsl:call-template name="get-id"/>
+                  </xsl:attribute>
+                </xsl:if>
+                <xsl:if test="exists($scale)">
+                    <xsl:attribute name="font-size" select="concat($scale, '%')"/>
+                </xsl:if>
+                <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-startprop ')]" mode="outofline"/>
+                <xsl:apply-templates/>
+                <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-endprop ')]" mode="outofline"/>
+            </fo:block>
+        </fo:block-container>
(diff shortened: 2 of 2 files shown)


robander:
:+1: LGTM. A review by somebody who knows XSL-FO a bit more intimately might be good, but the code looks right to me.

drmacro:
This change should work OK for one-page tables, but won't work generally for multi-page tables.

For multi-page tables you have to start a new page sequence with different orientation so that the table can break across pages.

My only concern with the current commit is that people will use it and then be distressed when it doesn't work for longer tables. But addressing this requirement also requires making the page sequence generation much more flexible, which is definitely beyond the current sprint.

robander:
And that's why I was hoping for a look by someone more familiar with FO issues... 

So without the page sequence rework that Eliot mentioned, the choices are: current behavior (never support `@orient`), or support it for tables that are small enough and (probably) give bad output with FOP for big tables. This _could_ be viewed as a regression for anybody using FOP, with a large table and a previously ignored `@orient="land"` setting, but I'm not sure that it should be.

eerohele:
>For multi-page tables you have to start a new page sequence with different orientation so that the table can break across pages.

Right, it won't work for multi-page tables for FOP users specifically. AH users should be fine. I don't have a XEP license, unfortunately, so I don't know how well it hand 

>My only concern with the current commit is that people will use it and then be distressed when it doesn't work for longer tables. But addressing this requirement also requires making the page sequence generation much more flexible, which is definitely beyond the current sprint.

I assume you mean that whenever there's an `@orient="land"` table, use a landscape orientation for the entire page instead?

If so, yeah, I thought about taking that approach but decided against it simply because I can't take up a task that big at the moment. I figured a basic support would be better than nothing. (Also, I happen to need `@orient` support in an environment that uses AH.)

Also, I'm not quite sure how that approach should optimally work in cases where there are multiple consecutive short `@orient="land"` tables. Should each table like that start a new page sequence even though they can fit on a single page? Can it be done any other way?

drmacro:
There's no way in a one-pass FO process to know how deep (or wide) a table is, nor where you are in the current page, so there's no general way to know if a given table will fit in the currently-available space, so the only one-pass solution is to place the rotated table at the start of a new page, either within a rotated block-container, as this code is doing, or by starting a new page sequence with landscape pages.

With a two-pass process you can know the rendered table size and the current output position and therefore make appropriate decisions.

But stepping up to a two-pass process is again more than we can to in 2.x.

However, it wouldn't be that hard to add in 3.x as both FOP and AXF provide an area tree output option and there is an open-source project that provides Java code to make it relatively easy to work with either form of area tree.

A multi-pass approach of course can solve many more problems than just this one for tables.

eerohele:
Yeah, a multi-pass approach would work (albeit at the cost of increased processing time).

So I guess the question is whether it makes sense to take the simple option here for 2.x or leave the whole thing for 3.x (provided that there's someone with the time to implement it at that point). I would be inclined to go for the simple implementation for 2.x since it works fine for AHF (and is better than nothing for FOP, I think).

However, it's not a big deal for me to implement this in a custom plugin, either, if it comes to that. Or we could do it in the `org.dita.pdf2.axf` plugin instead, which wouldn't have the downside of potentially causing confusion for FOP/XEP users.

(Unfortunately, a quick Google search seems to indicate that [XEP doesn't support multi-page landscape tables as of 2014](http://services.renderx.com/lists/xep-support/7246.html).)

drmacro:
If this approach was only activated by a runtime parameter I think that would be fine. That would force people to be at least be aware that this is an optional feature and give them the opportunity to read about the limitations.

robander:
I do not really care for the idea of a runtime parameter - this seems like an extremely specific piece of markup to get its own parameter. 

I think I'd be OK with it being confined to AXF, or with the "better than nothing" approach that includes FOP. Of course the only way to disable it in FOP, if it really gives you bad output, is to remove the attribute from the source. [Curious - if the table does run off the page for this in FOP, do you get any warning about that?]

robander:
Hoping to freeze any code changes for 2.3 on Wednesday - @jelovirt any opinion on this? I admit to nervousness unless it is limited to AXF, it could be viewed as either good or both good and bad for the default FOP.

jelovirt:
I'm ok adding support for AXF only.

I've implemented landscape tables using before-floats, and FOP doesn't support that either.

jelovirt:
WRT the PDF2 implementation, the default PDF2 plugin implements XSL 1.1. FOP and XEP extensions to base PDF2 either disable features that the formatter doesn't support or replaces with base implementation with a workaround. So IMO this should go to base PDF, not just AXF. If the output completely breaks the output on FOP and XEP, then it can be disabled for them.

raducoravu:
I would vote also for making this available for FOP. Something is better than nothing. At least we'll see the interest in the new feature.

jelovirt:
For FOP, this feature will be in 3.x.

jelovirt:
LGTM

tommyr1962:
Is the orient attribute currently supported at all? My apologies; was just sent this post this week.

raducoravu:
:+1: from one of our users:

      When I try to rotate a table, nothing happens. That is, I set the orient attribute to land and it still comes out as portrait.

kirkilj:
What's the current status of multi-page landscape tables. I gather it's still limited to one page, correct? Is there an OT plugin that solves this or guidance on the best way to inject a new page sequence dynamically for such cases?

raducoravu:
@kirkilj we have this working well in the Oxygen CSS-based PDF publishing plugin. Making this properly work was complex, it implied creating a new page sequence, so this is not something I would know how to port back to the XSL-FO based publishing. I see a mention in the thread above that Antenna House might also be able to rotate the table and have it displayed on multiple pages.

Issue #2303, opened 2016-04-14
Fatal Error When Related Link Text Includes <tm>


shaneataylor:
This worked fine as late as OT 2.1.2. When I moved to 2.2, however, my builds broke with the following message in my log:

> /Users/staylor/Dev/dita_source_files/build/deps/DITA-OT2.2.4/build.xml:41: The following error occurred while executing this line:&#xa;/Users/staylor/Dev/dita_source_files/build/deps/DITA-OT2.2.4/plugins/org.dita.xhtml/build_general.xml:162: Fatal error during transformation using /Users/staylor/Dev/dita_source_files/build/deps/DITA-OT2.2.4/plugins/org.dita.html5/xsl/dita2html5.xsl: In the key() function, the node supplied in the third argument (or the context node if absent) must be in a tree whose root is a document node; SystemID: file:/Users/staylor/Dev/dita_source_files/build/deps/DITA-OT2.2.4/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl; Line#: 1128; Column#: -1

After much experimenting, I discovered this was breaking when a `<tm>` tag was present in the link text of a related link, as in this example.

``` xml
<relrow>
    <relcell>
        <topicref href="c_a_allow_cookies.xml"/>
    </relcell>
    <relcell>
        <topicref href="c_a_blackboard_3rd_party_cookies.xml">
            <topicmeta>
                <linktext>Enable Third-Party Cookies for <ph><tm tmtype="reg" tmowner="Blackboard, Inc."
                            trademark="blackboard">Blackboard</tm></ph></linktext>
            </topicmeta>
        </topicref>
    </relcell>
</relrow>
```

infotexture:
Looks like this was resolved in #2394.

PR #2394, opened 2016-05-27
Fix error when <linktext> contains <tm> #2303

eerohele:
Regression introduced in ad0e1d8. An intermediate tree is assigned to the `$unordered-links` variable, which is then applied. The root node of an intermediate tree is not a document node, which causes any calls to the `key()` function to fail.

The solution is to pass the original root node as a tunnel parameter and use that as the context node for any calls to `key()`.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
@@ -179,7 +179,9 @@
                                               [generate-id(.) = generate-id(key('hideduplicates', related-links:hideduplicates(.))[1])]"/>
        </xsl:apply-templates>
       </xsl:variable>
-      <xsl:apply-templates select="$unordered-links"/>
+      <xsl:apply-templates select="$unordered-links">
+        <xsl:with-param name="root" select="root()" as="document-node()" tunnel="yes"/>
+      </xsl:apply-templates>
       <!--linklists - last but not least, create all the linklists and their links, with no sorting or re-ordering-->
       <xsl:apply-templates select="*[contains(@class, ' topic/linklist ')]"/>
     </nav>
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -898,6 +898,7 @@
   <!-- process the TM tag -->
   <!-- removed priority 1 : should not be needed -->
   <xsl:template match="*[contains(@class, ' topic/tm ')]" name="topic.tm">
+    <xsl:param name="root" select="root()" as="document-node()" tunnel="yes"/>
   
     <xsl:apply-templates select="*[contains(@class, ' ditaot-d/ditaval-startprop ')]" mode="out-of-line"/>
     <xsl:apply-templates/> <!-- output the TM content -->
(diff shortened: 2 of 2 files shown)


src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl
```diff
@@ -1078,7 +1079,7 @@
         <xsl:if test="ancestor::*[contains(@class, ' topic/title ')]/parent::*[contains(@class, ' topic/topic ')]">
           <xsl:choose>
             <!-- Not the first one in a title -->
-            <xsl:when test="generate-id(.) != generate-id(key('tm', .)[1])">skip</xsl:when>
+            <xsl:when test="generate-id(.) != generate-id($root/key('tm', .)[1])">skip</xsl:when>
```
jelovirt: IMO passing the context as the third argument of `key()` is a bit more familiar to people.

jelovirt:
Needs to be ported to HTML5.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl
eerohele: It probably is, but I'm in the habit of using this form because if the third argument to `key()` is an empty sequence, it is a static error, whereas `$el/key('foo', 'bar')` will return an empty sequence if `$el` is an empty sequence.

In this case, though, `$root` can't be empty, so it's not a problem to change.

jelovirt: I'm ok with this then. Could you rebase to fix the merge conflict?

eerohele: Yep, I'll try to get to it during the weekend.

eerohele:
I tried porting this on top of the current `develop` code, but it looks like the HTML5 transtype currently fails for any document:

```
[xslt] /opt/dita-ot/src/main/plugins/org.dita.xhtml/xsl/xslhtml/taskdisplay.xsl:44:40: Fatal Error! XPath syntax error at char 8 on line 44 in {$newline}:
```

eerohele:
Sorry, it's the XHTML transtype that fails with that error — I was using the wrong transtype earlier in a rush.

I've now rebased the change on top of `develop`.

jelovirt:
Yeah, I was wondering how it was able to fail like that.

jelovirt:
@eerohele are you able to port this to HTML5?

eerohele:
That's actually what I did. I can also submit a new PR for XHTML.

jelovirt:
Yes, a PR for XHTML would be good too.

eerohele:
All right, can do, but it might be better if you could take a look at the XHTML transtype failures first since you've been refactoring that part quite heavily? I could also fix it I think, but you probably have an idea regarding how you wanted the XHTML transtype to work in 3.0.

At the moment, it looks like there's quite many named templates that the XHTML plugin needs that were moved into the HTML5 plugin. For example:

```
src/main/plugins/org.dita.xhtml/xsl/xslhtml/taskdisplay.xsl:362:49: Fatal Error! No template exists named commonattributes
src/main/plugins/org.dita.xhtml/xsl/xslhtml/taskdisplay.xsl:363:38: Fatal Error! No template exists named setid
src/main/plugins/org.dita.xhtml/xsl/xslhtml/taskdisplay.xsl:374:41: Fatal Error! No template exists named setaname
src/main/plugins/org.dita.xhtml/xsl/xslhtml/taskdisplay.xsl:381:61: Fatal Error! No template exists named dita2html:simpletable-cols
src/main/plugins/org.dita.xhtml/xsl/xslhtml/taskdisplay.xsl:446:37: Fatal Error! No template exists named style
src/main/plugins/org.dita.xhtml/xsl/xslhtml/taskdisplay.xsl:449:45: Fatal Error! No template exists named th-align
```

Issue #2378, opened 2016-05-22
PDF2 pagination: "This page is intentionally blank"


DJBHollis:
#2335, `layout-masters.xsl`, `vars`

The last proposed change to pagination. When the pagination requires a blank page between sections, the header would carry the message, "This page is intentionally blank".

So, this would be a new page design, but it would be generic, and would be used for chapters, appendices, ToC, index, glossary, etc. as required.

The drawback is that it introduces a new string, that would need translation.

The page design would not, necessarily, need a footer.

Issue #2377, opened 2016-05-22
PDF2 pagination: Map type determination


DJBHollis:
#2335, `root-processing.xsl`

`$mapType` currently has two values: `'bookmap'` or `'ditamap'`.

There is actually a third possibility which could be either `'noMap'` or `'topicOnly'`.

If this third option is recognised, it might be useful for processing topics to PDF.

It should, at least, make it easier to follow the logic.

robander:
I can see how this might be useful. A new value would not cause the default processing to change in any way with today's code, though it's possible that a customization expecting only the current values `bookmap` and `ditamap` could fail when a new value is presented.

I think a value for topic as input would just be `topic`. This would also match the `$mapType` variable in the ODT code, though the ODT code is in no way a shining example of best practices.

Issue #2376, opened 2016-05-22
PDF2 pagination: Default page designs


DJBHollis:
#2335, #2374, #2375, `basic-settings.xsl`

I suggest the defaults for `$mirror-page-margins`, `$add-first-page` and `$toc-use-body-pages` would be map based:

```
     <xsl:choose>
          <xsl:when test="$mapType = 'bookmap'">
               <xsl:variable name="mirror-page-margins" select="true()"/>
               <xsl:variable name="add-first-page" select="true()"/>
               <xsl:variable name="toc-use-body-pages" select="false()"/>
          <xsl:otherwise>
               <xsl:variable name="mirror-page-margins" select="false()"/>
               <xsl:variable name="add-first-page" select="false()"/>
               <xsl:variable name="toc-use-body-pages" select="true()"/>
          </xsl:otherwise>
     </xsl:choose>
```

This would put the basic pagination settings in `basic-settings.xsl`, so they are easy to find, and should be easy to understand.

This could also do away with the need for `ditamap-body-sequence`. Certainly its use in `layout-masters.xsl` is just to set defaults, and would not be necessary. However, it is also used in `root-processing.xsl`, so I'm not entirely sure.

`ditamap-body-sequence` only does half a job for ditamap as it leaves the ditamap using the standard ToC. The two new proposed variables, `$add-first-page` and `$toc-use-body-pages`, in conjunction with $mirror-page-margins` would be a much better solution.

Issue #2375, opened 2016-05-22
PDF2 pagination: Add variable for ToC to use body page designs


DJBHollis:
#2335, `layout-masters.xsl`

I suggest the variable could be `$toc-use-body-pages`.

A ditamap only has a ToC and body sections. This new variable would mean that a ditamap could use one set of page designs throughout.

If `$mirror-page-margins` is false, and the proposed `$add-first-page` is also false, then the page designs would reduce to a single page design.

It could also apply to a simple bookmap that doesn't have appendices, index, glossary, etc.

Current:

```
    <!--TOC simple masters-->
    <xsl:if test="$mirror-page-margins">
      <fo:simple-page-master master-name="toc-even" xsl:use-attribute-sets="simple-page-master">
        <fo:region-body xsl:use-attribute-sets="region-body.even"/>
        <fo:region-before region-name="even-toc-header" xsl:use-attribute-sets="region-before"/>
        <fo:region-after region-name="even-toc-footer" xsl:use-attribute-sets="region-after"/>
      </fo:simple-page-master>
    </xsl:if>

    <fo:simple-page-master master-name="toc-odd" xsl:use-attribute-sets="simple-page-master">
      <fo:region-body xsl:use-attribute-sets="region-body.odd"/>
      <fo:region-before region-name="odd-toc-header" xsl:use-attribute-sets="region-before"/>
      <fo:region-after region-name="odd-toc-footer" xsl:use-attribute-sets="region-after"/>
    </fo:simple-page-master>

    <fo:simple-page-master master-name="toc-last" xsl:use-attribute-sets="simple-page-master">
      <fo:region-body xsl:use-attribute-sets="region-body.even"/>
      <fo:region-before region-name="even-toc-header" xsl:use-attribute-sets="region-before"/>
      <fo:region-after region-name="even-toc-footer" xsl:use-attribute-sets="region-after"/>
    </fo:simple-page-master>

    <fo:simple-page-master master-name="toc-first" xsl:use-attribute-sets="simple-page-master">
      <fo:region-body xsl:use-attribute-sets="region-body.odd"/>
      <fo:region-before region-name="odd-toc-header" xsl:use-attribute-sets="region-before"/>
      <fo:region-after region-name="odd-toc-footer" xsl:use-attribute-sets="region-after"/>
    </fo:simple-page-master>
```

I suggest it becomes, in conjunction with #2373 and #2374:

```
  <!--TOC simple masters-->
[...]

PR #2397, opened 2016-05-30
Issue 2372 change to force-page-count to use $mirror-page-margins

DJBHollis:
First pull request. Hope I've got it right!

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/commons-attr.xsl
+++ b/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/commons-attr.xsl
@@ -692,7 +692,7 @@ See the accompanying license.txt file for applicable licenses.
     <xsl:attribute-set name="__force__page__count">
         <xsl:attribute name="force-page-count">
             <xsl:choose>
-                <xsl:when test="/*[contains(@class, ' bookmap/bookmap ')]">
+                <xsl:when test="$mirror-page-margins">
                     <xsl:value-of select="'even'"/>
                 </xsl:when>
                 <xsl:otherwise>


robander:
I'll admit I don't know this code well enough that the change has obvious meaning to me - can you explain _why_ this change is warranted? What about the current output is wrong, and how does this change make the output better?

DJBHollis:
#2372

I suspect that `$mirror-page-margins` is a more recent addition, that is it wasn't part of the code when bookmap was added.

The original code states that if there is a bookmap, the page count should be `even`. Otherwise, for ditamap or no map, the page count should be whatever it is, `auto`.

An even page count ensures that the start of a new section is on an odd page. By section this would be a chapter, ToC, appendix, etc. in a bookmap, or a top level topic in a ditamap.

But, this is what `$mirror-page-margins` defines, If true, `$mirror-page-margins` introduces odd and even pages, and a start on an odd page.

So, the assumption in the code is that bookmap always has this style of pagination, and a ditamap not.

This is a poor assumption. There is no reason why ditamap should not also use `$mirror-page-margins`.

So, this change removes poor assumptions, and tests for `$mirror-page-margins` instead. This is a more appropriate test.

robander:
Thanks, that explanation is very helpful. [For what it's worth, that kind of description + the simple code change is great for a first pull request.]

jelovirt:
Not sure this is correct either. While forcing even page count with mirror margins, it's not the same thing. One is about page margins and the other one is how many blank pages are generated and where do new sections start. This change would force bookmap users who don't use mirror margins to override the code to get chapters to start on odd pages. So it just shifts the functionality from one combination to another, and IMO this doesn't improve the code, it just changes it. As a default, this is not better or more right.

If we want to add easier configuration, this should be done as part of page sequence refactoring, that would also enable e.g. using a single page sequence for all chapers. There we could add a simple configuration option for "book-like page sequences", where every new section/chapter/part start on an odd page.

DJBHollis:
The way things are at the present leads to a bug: #2318.

It's also why I raised these issues: #2335, #2372, #2373, #2374, #2375, #2376, #2377, #2378.

The problem with PDF2 is that it makes assumptions based on map type, both about bookmap and also about ditamap. Those assumptions might be correct perhaps 80% of the time, but if the assumption is wrong, it can be awkward to figure out where the issue lies, and then over ride it in a customisation.

So, my next stage was to add the code for #2373.

IMO, `$mirror-page-margins` should definitely lead to "book-like page sequences". There's no need for two variables.

There's only two ways to layout a publication: plain pages or facing pages. That is `$mirror-page-margins` is either false or true. Everything else then follows from that.

BTW, are there any test documents anywhere, bookmap and ditamap?

jelovirt:
I disagree with trying to force page sequence generation to follow page margin configuration. They are two different things, no point trying to force them together.

drmacro:
I agree with Jarno--mirroring page margins does not necessarily imply right-hand start for chapters. There definitely book designs that are mirrored pages with next-page start for chapters.

So there clearly need to be two controls: mirror pages yes/no, chapter start odd/even/any 

But I also agree with David that the current behavior is wrong and making assumptions about what Bookmap users want is inappropriate.

DJBHollis:
> chapter start odd/even/any

Chapter start on an even page - really? I can't recall seeing that.

>There definitely book designs that are mirrored pages with next-page start for chapters.

Just to confirm: odd and even page designs, but a chapter start on 'any'?

Jarno mentioned odd page design only, but force a chapter start on 'odd'.

jelovirt:
Closing as won't fix

Issue #1958, opened 2015-06-30
Update DITA-OT to ship FOP 2.0


Simalot:
FOP 2.0 was released at the beginning of June. It contains a lot of updates; for me currently, the most important being OpenType CFF font support. I replaced the OT 2.01 and 2.1.0 fop directories with FOP 2.0, and in my limited testing have not found any issues.

Here are the release notes for FOP 2.0: https://xmlgraphics.apache.org/fop/2.0/releaseNotes_2.0.html

raducoravu:
I played with FOP 2.0 producing PDF from our user's manual about 1500 DITA topics and it threw a NullPointerException at some point:

https://issues.apache.org/jira/browse/FOP-2479

Users on the Docbook users list have also encountered this exception when producing PDF from Docbook.

So either we ship also a patch for the problem or we might consider waiting for FOP 2.1.

geewrites:
Issue FOR-2479 is closed as a duplicate of this issue:

   https://issues.apache.org/jira/browse/FOP-2461

FOR-2461 is still open but has a workaround. FWIW.

raducoravu:
Fop 2.1 just came out:

http://apache-fop.1065347.n5.nabble.com/ANN-Apache-FOP-2-1-Released-td43379.html

From what I tested seems to have resolved my NPE problem.

jelovirt:
@robander do you want to risk updating FOP for 2.3 release this close to the release?

jelovirt:
Skip to upgrade to FOP 2.1 #2399

PR #2399, opened 2016-05-31
Upgrade to FOP 2.1 #1958

jelovirt:
Update bundled FOP version to 2.1.


jelovirt:
@raducoravu Have you found any issues with FOP 2.1 that we need to create workarounds for?

raducoravu:
No problems so far. 
I heard on the users lists that FOP 2.1 does not work anymore with the PDF Box libraries which would probably mean that you can no longer embed PDF images. But I have not tested this myself.
There have been cases of using Hebrew with DITA topics which work with FOP 2.1 but threw NullPointerExceptions in older FOPs.
Also there seemed to be some problems regarding relative paths in the fop.xconf and the base to which they were made absolute but I did not follow that.

raducoravu:
Problem with FOP not properly interpreting SVG image dx and dy positions (multiple text runs in embedded SVGs are painted one over the other):
https://issues.apache.org/jira/browse/FOP-2617

Issue #1245, opened 2012-05-06
TM output


jelovirt:
Converted from [SourceForge issue 3311121](http://sourceforge.net/support/tracker.php?aid=3311121), submitted by julio_vazquez

Currently, the `<TM>` element marks all occurrences of a trademark in the output. The element should work similar to the HTML output in that the first occurrence is the only one marked. It makes sense to make this a configurable option in case there is a requirement for all occurrences to be marked.

robander:
@jelovirt Guessing you haven't looked at this in a while. I was thinking of implementing for 3.0 but now wondering if you have a preference as to whether this sort of option belongs in the base code. Current processing is set up to mark every `<tm>` element; I've never had this processing requirement, having only needed the first-in-PDF behavior requested above. I'm assuming others need this as well if it was requested from outside IBM, but guessing some people also expect the current behavior.

jelovirt:
It's fair feature request. I'd implement this by adding a mode that returns `xsl:boolean` to signal whether trademark symbol is output or not. This makes it easy to write code that e.g. always outputs tm in frontmatter, and first occurrence per chapter. I've used that approach with some other element before, but can't remember which transtype it was.

jelovirt:
XHTML seems to have something like that, but it doesn't use a mode, only `xsl:choose`. Mode is more extensible.

robander:
Agreed. Any preference on naming for an external parameter to control this? Or were you thinking just the function, and using it requires extension?

jelovirt:
I'd prefer _not_ to add a parameter for this, because we don't want parameter per feature.

robander:
OK - will just work on the function. Thanks.

robander:
Think that #2404 covers any likely changes to come from this - code is now optimized with a mode template that can configure when/if to mark trademarks.

PR #2404, opened 2016-06-01
Add support for configuring tm scope in PDF #1245

jelovirt:
Add mode to control whether trademark symbol is created.

Uses "℠" Unicode character instead of "SM".


robander:
:+1:

PR #2406, opened 2016-06-02
Fixes failure when topic and subelement have same ID.

drmacro:
Integration tests pass locally.

Refactor is to use a key to resolve topic references, replacing all places where code was like $doc//*[@id = $topicid']

This avoids the error and should make topic reference resolution more efficient.

By using a key table that only includes topics, avoid the need to check if the resolved reference is or is not a topic.


robander:
From the description, this sounds good, but I'm concerned by the failing test and this when I try to view the commit:
> 1,346 additions, 1,320 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

Wondering if something went wrong in pushing the code change?

drmacro:
Yes, I must have done something odd.

I'll retry making a pull request using the 2.3 hotfix branch. I only changed one file and made one commit so it should be a very simple pull request.

drmacro:
Closing as being based on the wrong starting branch. Will open a new, hopefully correct request

PR #2407, opened 2016-06-02
Fix for topic and element with same ID failure

drmacro:
This adds the use of a key table to look up the target topic, replacing various inefficient paths while also correcting the lookup bug.

The key table can only include topics so the need for separate checks that the target is in fact a topic are eliminated.

This code may be noticeably more efficient, specially for references to topics in documents with lots of topics.


drmacro:
The failures are due to small changes in the output, such as extra attributes or two nodes where three were expected. I'll figure out which code change caused the failures and determine if the code is at fault or the test cases are at fault.

My expectation is that these changes should not affect the result at all so likely some subtle code bug.

drmacro:
Found my bug. Integration tests now really pass locally.

PR #2412, opened 2016-06-10
Replace named template with mode for easy override of static content

robander:
I have a need to override static content generation for some types of topics - specifically, if referencing "Notices" or a "Preface" topic, when the referenced topic has its own title, I want to skip generating any default title. 

Currently, I have to override the entire `insertChapterFirstpageStaticContent` template, which contains a lot of processing for different conditions:
```xml
<xsl:template name="insertChapterFirstpageStaticContent">
  <fo:block>
    ...generate ID...
    if type=chapter, do a lot
    else if type=appendix, do a lot
    else if type=appendices, do a lot
    else if type=part, do a lot
    else if type=preface, create a title
    else if type=notices, create a title
  </fo:block>
</xsl:template>
```

Converting this to a mode template allows me to override the template, change the 2 simple conditions, and use `<xsl:next-match/>` to keep default processing for chapters / parts / appendices. In pseudocode:
```xml
<xsl:template match="*" mode="insertChapterFirstpageStaticContent">
  if (type="notices" or type="preface") and empty title, then
    just generate TOC id
  otherwise
    <xsl:next-match/>
</xsl:template>
```

After I typed this all out, I realized - an alternate way to do this would be to make each of the "if/then/else" templates into a new call to a mode template, allowing control of any individual `type` without affecting the others. I could add that as well, but I think either way it makes sense to convert this one to a mode.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -246,9 +246,9 @@ See the accompanying license.txt file for applicable licenses.
 
                     <xsl:apply-templates select="*[contains(@class,' topic/prolog ')]"/>
 
-                    <xsl:call-template name="insertChapterFirstpageStaticContent">
+                    <xsl:apply-templates select="." mode="insertChapterFirstpageStaticContent">
                         <xsl:with-param name="type" select="'chapter'"/>
-                    </xsl:call-template>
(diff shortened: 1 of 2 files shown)


drmacro:
As a rule, any template with lots of if/then/else applied to elements should be rewritten as templates in appropriate modes.

jelovirt:
LGTM

PR #2419, opened 2016-06-15
Drop empty fo:inline elements

robander:
There are quite a few instances of this in the PDF code:

``` xml
<fo:inline>
</fo:inline>
```

These are mostly added as a way to preserve `@id` anchors, but could potentially add other attributes.

When working on a document with a lot of `<xref>` elements, I noticed this resulted in `<fo:inline/>` for every link or cross reference. With an overly-linked document like the DITA specification, this results in over 100,000 unnecessary `<fo:inline/>` elements.

With a bit of searching I noticed there were 20+ instances of this same construct, so I've created a single mode template to handle each of them. It processes the `commonattributes` template as above. If any attributes are returned, it creates the `<fo:inline>` element with those attributes. Otherwise, it skips generating the empty element.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -104,7 +104,7 @@ See the accompanying LICENSE file for applicable license.
             </xsl:apply-templates>
         </fo:block>
     </xsl:template>
-
+ 
     <xsl:template name="startPageNumbering" as="attribute()*">
         <!--BS: uncomment if you need reset page numbering at first chapter-->
 <!--
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/links.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/links.xsl
@@ -247,9 +247,6 @@ See the accompanying LICENSE file for applicable license.
     </xsl:template>
 
     <xsl:template match="*[contains(@class,' topic/xref ')]" name="topic.xref">
-        <fo:inline>
-            <xsl:call-template name="commonattributes"/>
-        </fo:inline>
 
     <xsl:variable name="destination" select="opentopic-func:getDestinationId(@href)"/>
     <xsl:variable name="element" select="key('key_anchor',$destination, $root)[1]"/>
@@ -264,6 +261,7 @@ See the accompanying LICENSE file for applicable license.
     </xsl:variable>
 
     <fo:basic-link xsl:use-attribute-sets="xref">
+      <xsl:call-template name="commonattributes"/>
       <xsl:call-template name="buildBasicLinkDestination">
         <xsl:with-param name="scope" select="@scope"/>
         <xsl:with-param name="format" select="@format"/>
@@ -309,7 +307,7 @@ See the accompanying LICENSE file for applicable license.
                       <xsl:with-param name="element" select="$element"/>
                   </xsl:call-template>
             </xsl:if>
-    </xsl:if>
+      </xsl:if>
 
     </xsl:template>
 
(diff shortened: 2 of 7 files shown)


robander:
Looks like this shows 2 commits, the first one is already open as a pull request #2412

robander:
@jelovirt pointed out another way of doing the logic, but I don't know if one is better or worse:
```xml
<xsl:variable name="holdAttributes" as="element()">
  <fo:inline>
  </fo:inline>
</xsl:variable>
<xsl:if test="exists($holdAttributes/@*)">
  <xsl:sequence select="$holdAttributes"/>
</xsl:if>
```

src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
```diff
@@ -101,6 +101,13 @@ See the accompanying license.txt file for applicable licenses.
         </fo:block>
     </xsl:template>
 
+    <xsl:template match="*" mode="EmptyInlineWithCommonAttributes">
+      <xsl:variable name="holdAttributes" as="attribute()*"><xsl:call-template name="commonattributes"/></xsl:variable>
+      <xsl:if test="not(empty($holdAttributes))">
```
jelovirt: Replace with `exists(@holdAttributes)`.

src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
```diff
@@ -101,6 +101,13 @@ See the accompanying license.txt file for applicable licenses.
         </fo:block>
     </xsl:template>
 
+    <xsl:template match="*" mode="EmptyInlineWithCommonAttributes">
+      <xsl:variable name="holdAttributes" as="attribute()*"><xsl:call-template name="commonattributes"/></xsl:variable>
```
jelovirt: IMO the variable name could be just `attributes`.

src/main/plugins/org.dita.pdf2/xsl/fo/links.xsl
```diff
@@ -407,9 +406,7 @@ See the accompanying license.txt file for applicable licenses.
           <fo:list-item xsl:use-attribute-sets="related-links.ul.li">
             <fo:list-item-label xsl:use-attribute-sets="related-links.ul.li__label">
               <fo:block xsl:use-attribute-sets="related-links.ul.li__label__content">
-                <fo:inline>
```
jelovirt: Is there a reason why we generate an empty `fo:inline` with common attributes, instead of just adding those attributes to the parent `fo:block`. IMO we should stick ID etc. into the parent `fo:block`, if there is a parent element.

robander: I believe in many cases there is not a parent element. I initially noticed this on `<xref>` and `<link>` where there is no parent block, but the `<fo:inline>` is generated before the `<fo:basic-link>` container. I don't know if there is a reason to do it that way - if maybe at some point `@id` on `<fo:basic-link>` wouldn't create a valid link?

robander: For all of the syntax groups below, there is no containing element (though I'm guessing those are rarely used). Today, this construct is probably (most often) just resulting in the extra markup for link elements.

jelovirt:
@robander could you rebase and resolve conflicts-

jelovirt:
@robander could you rebase and resolve conflicts if you want this to be included in 2.4?

src/main/plugins/org.dita.pdf2/xsl/fo/links.xsl
robander: Returning to this so I can bring it up to date and close it out.

I don't see any reason we generate `fo:inline` in at least some of these case. I originally left those alone because I wanted to make the least invasive change I could. Now testing and moving to the appropriate containing block where that makes sense.

robander: In the end - there were convenient + appropriate group elements for almost everything. Syntax groups did not have group elements but it was easy / not disruptive to add them. 

The exceptions were `indexterm` and `imagemap`. For index term: with all the other processing that goes on, IDs are already added, so the `<fo:inline>` was useless, and I've removed it.

For image maps, I just moved the test into the `imagemap` template. Creating a new `fo:inline` would be wrong, and a new `fo:block` wrapper might mess up formatting for smaller images that appear inline.

robander:
All conflicts resolved, and updated for better design per the discussion here https://github.com/dita-ot/dita-ot/pull/2419#discussion-diff-67460588L410 ... should be ready for 2.5.

Issue #2017, opened 2015-08-26
Reltables do not work if using any keys outside map directory


dagoss:
When key definitions are includes from outside the current map directory, links that should be generated by `<reltable>` elements do not appear in PDF2 output. This happens even if the `<reltable>` doesn't use any keys from the external map. 

Consider the following map:

```
<map>
    <mapref href="../common/keydefs.ditamap"/>
    <keydef keys="foo" href="foo.dita"/>
    <keydef keys="bar" href="bar.dita"/>

    <topicref keyref="foo"/>
    <topicref keyref="bar"/>

    <reltable>
        <relcell>
            <topicref keyref="foo"/>
        </relcell>
        <relcell>
            <topicref keyref="bar"/>
        </relcell>
    </reltable>
</map>
```

In this example, the keys foo and bar are defined locally, but the map also has keys from another map a level above the current maps's directory.

The links that should be appearing with topics foo and bar do not appear in PDF2 output. If the reference to the external map is removed, the `<reltable>` links are generated as expected.

eerohele:
This is actually not related to keyrefs.

This is enough to reproduce the issue:

```xml
<?xml version="1.0" encoding="utf-8"?>

<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "map.dtd">

<map id="map" title="DITA Map">
  <topicref href="topic1.dita"/>
  <topicref href="../topic2.dita"/>

  <reltable>
    <relrow>
      <relcell>
        <topicref href="topic1.dita"/>
      </relcell>
      <relcell>
        <topicref href="../topic2.dita"/>
      </relcell>
    </relrow>
  </reltable>
</map>
```

The issue is similar to #2190 in that the path to the file to which related links are inserted is resolved against the root of the temp directory instead of against the URI of the map that includes the file.

This should fix that issue:

```diff
diff --git a/src/main/java/org/dita/dost/module/MoveLinksModule.java b/src/main/java/org/dita/dost/module/MoveLinksModule.java
index 012909f..8d28e11 100644
--- a/src/main/java/org/dita/dost/module/MoveLinksModule.java
+++ b/src/main/java/org/dita/dost/module/MoveLinksModule.java
@@ -86,11 +86,11 @@ final class MoveLinksModule extends AbstractPipelineModuleImpl {
             linkInserter.setLogger(logger);
             linkInserter.setJob(job);
[...]

PR #2444, opened 2016-08-08
Fix rellink path resolution #2017

eerohele:
The change in `MoveLinksModule` should be relatively straightforward: resolve the file URI against the input DITA map file instead of the root of the temp directory.

When `MergeTopicParser` creates a map from topic URLs to `unique_n` IDs, it uses relative URIs as map keys, which doesn't always yield the correct result. Using the absolute URI of the topic instead makes things more stable.

The code in the `MergeTopicParser` class isn't very easy to understand (at least for me), so a review wouldn't hurt.


eerohele:
Hum. I guess this PR didn't make it to 2.3.2. Is this made obsolete by some other change, or should I recreate it against the `develop` branch, or…?

Issue #2448, opened 2016-08-11
Bogus row header @id generated for <stentry> elements


shaneataylor:
For simpletables with a specified @keycol, the <th> element for the row is now (2.3.1 and possibly others) generated with @scope="row" and no automatic @id (which is valid). However, when the @headers attribute is added to elements in that row, the code creates an @id for the row header which is never applied to the row header. So you get a table like this:

``` html
           <table class="simpletable"><col style="width:30%"><col style="width:30%"><col style="width:40%"><thead><tr class="sthead">
                    <th class="stentry" id="r_a_units__stentry__28">Unit</th>
                    <th class="stentry" id="r_a_units__stentry__29">Abbreviation</th>
                    <th class="stentry" id="r_a_units__stentry__30">Notes</th>
                </tr></thead><tbody><tr class="strow">
                    <th scope="row" class="stentry" headers="r_a_units__stentry__28">ampere</th>
                    <td class="stentry" headers="r_a_units__stentry__29 d88454e275">A</td>
                    <td class="stentry" headers="r_a_units__stentry__30 d88454e275"></td>
                </tr><tr class="strow">
                    <th scope="row" class="stentry" headers="r_a_units__stentry__28">angstrom</th>
                    <td class="stentry" headers="r_a_units__stentry__29 d88454e286">Å, ångström</td>
                    <td class="stentry" headers="r_a_units__stentry__30 d88454e286"></td>
                </tr>
```

Note the headers values like d88454e275 don't refer to anything (and aren't needed); they just create noise in the change history for the generated HTML (see #2276).

I think I've identified the fix and I'll submit a PR later today.

robander:
I can see the `scope="row"` setting that presumably renders those header values obsolete. 

I don't know the answer to this, but - does the use of `<thead>` allow accessibility tools to associate cells with the headers in the first row? Or should those be using `@scope="col"`?

shaneataylor:
My understanding is that `@scope="col"` isn't a solution, because while a table cell belongs to a parent construct called a row, its association with a column is less determinate. Using the `@headers` attribute to refer to an `@id` on a column heading is the solution for that problem.

I realize that my PR doesn't address choicetables or tables that are not simpletables. 

- For choicetables, I'll plan to add the needed changes. 
- For tabIes, I'm not sure the value of making this change is worth its risk, given the additional complexity and the fact that a non-regular table structure (that is, with cells that span multiple rows or columns) poses some accessibility issues anyway.

[...]

PR #2449, opened 2016-08-11
2448 Don't generate a random value for the @headers attribute

shaneataylor:
If no header `@id` is found, generating a new value in `@headers` doesn't do anything for accessibility; instead, it merely adds noise to the change history of the generated HTML.


shaneataylor:
This addresses issue #2448

robander:
This is now obsolete -- I included the suggested fix as part of #2811, which also covered generated IDs in the `<properties>` and `<choicetable>` specializations.

Issue #2439, opened 2016-08-04
PDF2 build target dependency "depend.org.dita.pdf2.init.pre" seems to be expanded twice


raducoravu:
If you download the latest DITA OT 2.3.1 and look in the build file `dita-ot-2.3\plugins\org.dita.pdf2\build.xml`, it has a target like this:
```xml
<target name="transform.topic2fo.init"
  depends="transform.fo2pdf.xep.init,
    transform.fo2pdf.fop.init,
    transform.fo2pdf.ah.init,
    transform.fo2pdf.fop.init,
    transform.fo2pdf.xep.init">
```
which seems to call targets like `transform.fo2pdf.xep.init` two times. The same happens if I have a custom plugin which uses this extension point:
```xml
<feature extension="depend.org.dita.pdf2.init.pre" value="process-reviews"/>
```
The name of my target will appear twice in the `transform.topic2fo.init`  list of dependencies.

raducoravu:
Fortunately ANT does not perform the same task twice if it's listed two times in the dependencies list.

jelovirt:
Looking at the code, I can't see where it's coming from. We have
``xml
<target name="transform.topic2fo.init"
          dita:extension="depends org.dita.dost.platform.InsertDependsAction"
          dita:depends="{depend.org.dita.pdf2.init.pre}">
```

And that extension point is used by three plugins and the XEP one is not duplicated anywhere. Debugging aid would be appreciated.

raducoravu:
But do you get the same result when running the integrator on your side? Is for example the dependency to " transform.fo2pdf.fop.init" added twice in the @depends attribute value?

jelovirt:
Yes, I get the two dependencies. I just wasn't able to find the cause for this.

raducoravu:
The problem is in this method:

     org.dita.dost.platform.Integrator.loadPlugin(String)

at some point it does this:

      featureTable.put(currentFeature.getKey(), currentFeature.getValue());

which should probably be modified to create a shallow copy of the list:

    featureTable.put(currentFeature.getKey(), currentFeature.getValue() != null ? new ArrayList<String>(currentFeature.getValue()) : null);

because the "currentFeature.getValue()" list is also used inside a certain Features object but sometimes in the same loadPlugin method you do this:

                   final List<String> value = featureTable.get(currentFeature.getKey());
                    value.addAll(currentFeature.getValue());
                    featureTable.put(currentFeature.getKey(), value);

[...]

PR #2451, opened 2016-08-19
Make shallow clone of feature values list

raducoravu:
Fix for https://github.com/dita-ot/dita-ot/issues/2439, make shallow clone of feature values list


jelovirt:
Tests failed with the same cause that Travis fails often. Rerunning tests at CI.

PR #2460, opened 2016-08-25
Update Java version mentioned in README

xephon2:
The `build.gradle` mentions 1.8

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -8,7 +8,7 @@ Prerequisites
 
 To build and use DITA-OT, you’ll need:
 
-* Java Development Kit 7 or newer
+* Java Development Kit 8 or newer
 
 Building
 --------


infotexture:
The version listed in `build.gradle` was changed on the `develop` branch in 4af5d4e, but these changes have not yet been released, so to avoid confusion, the ReadMe should still use JDK 7 until the next version of DITA-OT is officially released.

infotexture:
To clarify my comment above: while it's certainly correct to update the JDK version in the ReadMe on the `develop` branch to reflect the requirement listed in `build.gradle` on the same branch, if we do that users might think 2.3.x requires JDK 8, since  the `develop` version of the ReadMe is shown by default when users visit the project.

Switching the **Branch** list to `master` would reveal the JDK 7 requirement for the latest released version, but I'm not sure how many casual users are aware of that.

xephon2:
This is surely small-minded, but the `README` on `develop` should reflect the `build.gradle` on `develop` and the `README` on `master` should reflect the `build.gradle` on `master`. If you want the `master` `README` to be displayed on Github by default, you have to switch the default branch in the repo settings. Whatever, just close the PR.

jelovirt:
It would make sense to use `master` as the default branch, because that's where the latest stable release comes from. But, that would mean `master` would be used as the PR target branch by default. Because we're using Gitflow, we want all PRs to `develop`. So not sure what's the best solution.

jelovirt:
Fixed with other changes in 5a231dc

PR #2462, opened 2016-08-26
Map temporary files to source/result files using job configuration

jelovirt:
Use job configuration to track temporary files and their source/result URI. This allows using e.g. hashed temporary file names, or temporary paths that contain the absolute path to source file. Using a hashed/full path for temporary files makes uplevels processing redundant and allows constructing full temporary file structure before reading all input files.


robander:
I'm probably not the best to review all of the code, but the function looks great. I've been trying to extend new transform types based on HTML5 with files all returned in a zip file, and the uplevels differences (for maps / topics) have been a pain.

PR #2466, opened 2016-09-02
Mappull fix topic id same as element id

drmacro:
This pr fixes the bug in mappullImpl.xsl where the code would fail at runtime if a topic had the same ID as one of its subelements.

I've refactored the code to only get the target topic once and removed redundancies in the code. I did not attempt to refactor the handling of all the attributes as it's too twisty and dangerous to touch at this time.

Added new test case, mappull_topic_element_same_id, through separate PR to the testsuite submodule. 

This test case (and all other test cases except already-failing conref test) pass for me locally.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "src/test/testsuite"]
 	path = src/test/testsuite
-	url = https://github.com/dita-ot/testsuite.git
+	url = https://github.com/drmacro/testsuite.git
 [submodule "src/main/docsrc"]
 	path = src/main/docsrc
 	url = https://github.com/dita-ot/docs.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ jdk:
 - oraclejdk8
 env:
   global:
-    - JAVA_OPTS="-Xmx1024M"
+    - JAVA_OPTS="-Xmx2024M"
 before_script:
 - git submodule update --init --recursive
 - "./gradlew"
--- a/build.gradle
+++ b/build.gradle
@@ -13,20 +13,20 @@ repositories {
      mavenCentral()
 }
 dependencies {
-    compile group: 'commons-io', name: 'commons-io', version: '2.4'
-    compile group: 'commons-codec', name: 'commons-codec', version:'1.9'
+    compile group: 'commons-io', name: 'commons-io', version: '2.5'
+    compile group: 'commons-codec', name: 'commons-codec', version:'1.10'
     compile group: 'xerces', name: 'xercesImpl', version:'2.11.0'
     compile group: 'xml-apis', name: 'xml-apis', version:'1.4.01'
     compile group: 'xml-resolver', name: 'xml-resolver', version:'1.2'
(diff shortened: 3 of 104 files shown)


robander:
Not sure about the failing test cases - those make me nervous, though I wouldn't expect this to cause the failures from conref. 

Generally speaking - it would be easier to review if there were separate commits for the fix versus the refactoring.

I'd like to take a look through this one, and possibly try it out with some additional test cases (I know our test cases don't have as much coverage as we'd like, and I'd like to try it out with some documents I've found troublesome, particularly to make sure there's no impact on files that use `<dita>` as a root element.

drmacro:
The fix requires the refactor since the fix is to redo the broken way in which the target topics were being fetched.

The test case that I added includes topics that are within composite documents.

The refactor is relatively minor in that it merely replaces duplicate business logic with a single function call. It should be possible to verify the correctness as much by inspection as anything else.

I ran my tests locally against the unmodified develop branch and got the conref failure, then ran against the pr branch and the test results were the same.

If there are specific combinations of elements you want to have tested I'm happy to work up a test case for those but it may be that the test case I added already covers those cases.

robander:
Ha ... that's what I get for talking about test cases without looking at your new test cases first. :-)

drmacro:
I'm also concerned about the degree of test case coverage and I'm happy to take direction on additional test cases to add.

jelovirt:
> The fix requires the refactor since the fix is to redo the broken way in which the target topics were being fetched.

The commit contains changes like

```diff
- <xsl:when test="$print='no' and ($FINALOUTPUTTYPE='PDF' or $FINALOUTPUTTYPE='IDD')"/>
+ <xsl:when test="$print='no' and ($FINALOUTPUTTYPE= ('PDF', 'IDD'))"/>
```

Which are clean-up changes and have nothing to do with the refactoring. These are the changes @robander mean with "separate commits for the fix versus the refactoring".

drmacro:
So when you say "separate commits" do you mean separate PRs or simply separate commits in the commit history? I've been squashing the commits in the PR to keep the commit history clean.

drmacro:
I think I understand what Robert's asking for. I'm not sure it's worth the time to do it for mappull because the the number of changes is relatively small.

For the topicpull module, I am doing the work now. Once I have all the tests passing I'll redo the work in such a way that the commit history will make a clear distinction between simple code cleanup and more involved refactorings so that the progression through the refactorings is clear.

drmacro:
I've added a small fix. The test case /conref_push/exp/preprocess/pushAfter/pushAfter_with_crossRef/pushAfter.ditamap caused a failure when I ran the map through the full OT process but doesn't fail in the integration test.

It looks like the conref process is not setting the @class value on an element where it's expected, not sure.

jelovirt:
Separate commits, _not_ separate PRs. We can squash on merge later on, but for review the code should have multiple commits. Preferable first cleanup commits, then refactorings. If you later on notice you should clean something up, create a new commit for it and rebase it to cleanup commit groups.

jelovirt:
Closing as inactive

PR #2467, opened 2016-09-03
Add copyright based on Git history

jelovirt:
- [x] Add missing copyright statements to files, based on Git history
- [x] Clean boilerplate language and formatting
- [x] Fix author names for IBM contributions
- [x] Fix IBM and other existing copyright statements to follow the convention


jelovirt:
@robander how do you want to handle contributions from IBM and yourself? Use "Copyright IBM Corp.", or should we use names of individuals who made those commits?

robander:
For anybody else scanning this - I answered @jelovirt through slack chat that the copyright for new files from me, or from other IBMers, should be "IBM Corporation".

robander:
A bit hard to review the change over that many files, but the pattern looks good, I'm happy with it.

infotexture:
Late to the conversation here, but I notice we mention the `license.txt` file as found in the distribution package rather than `LICENSE.md` at the repo root, which makes sense in this context. 

In the docs repo, we refer to `LICENSE.md` since there is no `license.txt` in the repo. 

Not sure if we should align that.

jelovirt:
Yes, I noticed this too. It might be easies to just bundle Apache 2 license as `LICENSE` plain text file.

infotexture:
That would work, then we do the same for docs. Windows users might cry without a file name extension though.

jelovirt:
Given how rarely someone actually opens that file, I think it's not a problem.

robander:
I've seen quite a few open source packages on Windows with a `LICENSE` file name (no extension). I also think it's not a problem.

infotexture:
I'm fine with the extensionless `LICENSE` file. 

If I understand correctly, we'd remove `src/main/license.txt` and `src/main/plugins/com.sophos.tocjs/license.txt` and update `build.gradle` to copy `LICENSE.md` to the distribution package as `LICENSE`, right?

After 9065ba6, some files in the `dita-ot` repo mention `LICENSE.md` while others point to `license.txt`. Sounds like we'd want to unify the boilerplate in both `dita-ot` & `docs` repos to refer to the extensionless variant:

>  See the accompanying LICENSE file for applicable licenses.

Do we need to add copyright statements to the docs files too?

Not sure how easy it would be to reconstruct that information from Git logs, as the history in the  `docs` repo only goes back to April 2013 when documentation source and output were split into different folders prior to the 1.8 release.

PR #2469, opened 2016-09-06
Add range by ID extension to coderef

jelovirt:
Add an extension to `<coderef>` processing to allow selecting ranges by ID.

Optional ID range is defined using:

```
uri ("#token=" start? ("," end)? )?
```
Lines identified using start and end IDs are exclusive. If the start ID is omitted, range starts from the first line; if end ID is omitted, range ends in last line.

Example
--------

With source `fact.hs`

```
-- START-FACT
fact :: Int -> Int
fact 0 = 1
fact n = n * fact (n-1)
-- END-FACT
main = print $ fact 7
```
and it's used with

```xml
<codeblock><coderef href="fact.hs#token=START-FACT,END-FACT"/></codeblock>
```

The result will equivalent to

```xml
<codeblock>fact :: Int -> Int
fact 0 = 1
fact n = n * fact (n-1)</codeblock>
```

### DIFF
--- a/src/main/java/org/dita/dost/writer/CoderefResolver.java
+++ b/src/main/java/org/dita/dost/writer/CoderefResolver.java
@@ -11,11 +11,7 @@ package org.dita.dost.writer;
 import static org.dita.dost.util.Constants.*;
 import static org.dita.dost.util.URLUtils.*;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStreamReader;
+import java.io.*;
 import java.net.URI;
 import java.nio.charset.Charset;
 import java.util.regex.Matcher;
@@ -23,10 +19,10 @@ import java.util.regex.Pattern;
 
 import org.dita.dost.util.Job;
 import org.xml.sax.Attributes;
+import org.xml.sax.ContentHandler;
 import org.xml.sax.SAXException;
 import org.dita.dost.exception.DITAOTException;
 import org.dita.dost.log.MessageUtils;
-import org.dita.dost.util.FileUtils;
 
 /**
  * Coderef element resolver filter.
@@ -44,6 +40,12 @@ import org.dita.dost.util.FileUtils;
  * <pre>uri ("#line-range(" start ("," end)? ")" )?</pre>
  * 
  * <p>Start and end line numbers start from 1 and are inclusive. If end range
+ * is omitted, range ends in last line.</p>
+ * <p>Optional id range is defined using:</p>
+ *
+ * <pre>uri ("#id-range(" start ("," end)? ")" )?</pre>
+ *
+ * <p>Lines idenfified using start and end IDs are exclusive. If end range
  * is omitted, range ends in last line.</p>
  */
 public final class CoderefResolver extends AbstractXMLFilter {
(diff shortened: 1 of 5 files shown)


jelovirt:
Some markup language processor has this same support, but haven't found it again. The basic idea was the same, you have markers in the source file and those are used to define the range. I can't remember if that was language agnostic, so that it would recognize that comments in e.g. Java are marked with `//` and in Ruby with `#`. In the end, I think it may be safer to not try to add language specific processing, but rather just use the ID and check if that exists in the line. It's up to the authors to use IDs that do not appear in other context in the same file.

jelovirt:
Looking for feedback on the syntax too. The current one is

```ebnf
uri ("#id-range(" start ("," end)? ")" )?
```

And I don't feel strongly about the `id-range` scheme name. I don't want to go for syntax like `fact.hs#START-FACT,END-FACT` because IMO it's too terse.

xephon2:
I don't like its syntax, but maybe it would be more DITA like to use the syntax of `@conref` and [`@conrefend`](https://docs.oasis-open.org/dita/v1.2/os/spec/common/theconrefendattribute.html), even if the technical implementation is different, because the concept (what the author wants to do) is the same. As an author, I don't care how (and by whom (e.g. TC or DITA-OT dev)) a functionality is implemented. But if I want to conref a range of a file, I expect to use `@conref` and `@conrefend`

So this:
```xml
<codeblock>
  <coderef href="fact.hs#line-id(START-FACT,END-FACT)"/>
</codeblock>
```

Could be:
```xml
<codeblock>
  <coderef conref="fact.hs#topic-id/START-FACT" conrefend="fact.hs#topic-id/END-FACT"/>
</codeblock>
```

jelovirt:
The point is, the `fact.sh` is not an XML DITA file. Thus, we can't use conref with it. We already have supported `line-range` extension for a while now, thus would basically be just an alternative range selector.

xephon2:
Why not overriding the mechanism, e.g. with an additional `@format`?

```xml
<codeblock>
  <coderef conref="fact.hs#topic-id/START-FACT" conrefend="fact.hs#topic-id/END-FACT" format="plain-text"/>
</codeblock>
```

jelovirt:
That would be very heavy overloading of base DITA functionality. There's a reason why `<coderef>` exists, instead of DITA TC trying to make conref work with plain text. Using the `line` extension is already defined in an existing specification RFC 5147; this uses a similar approach.

jelovirt:
Also, using something like `conref="fact.hs#topic-id/START-FACT"` requires that that Haskell file has two IDs `topic-id` and `START-FACT`. There is no syntax or concept for IDs in Haskell, thus we would have to invent that syntax. Same goes for every other possible target file type.

drmacro:
Why not use two query parameters: fact.hs?startline=x&endline=y

jelovirt:
@drmacro if the target URI is an absolute URI to an external resource, those parameters will be sent to the server and the results may not be the same as without them. The URI fragment is never sent to the server.

drmacro:
Good point. Then I think your function syntax is probably as good as anything. It's reasonably consistent with XPointer's precedent.

shaneataylor:
Since the use most closely parallels [RFC5147](https://tools.ietf.org/html/rfc5147), I'd parallel that syntax. So, instead of:

`<coderef href="Parser.scala#line=4,10"/>`

you'd specify

`<coderef href="Parser.scala#token=START-FACT,END-FACT"/>`

Since the toolkit isn't aware of the syntax of the code it's including, we don't know that the IDs are in fact IDs as defined in the code. All we know is that a matching token was found somewhere in a line. I think using `token` instead of `id` helps to communicate that.

jelovirt:
I like @shaneataylor's `token` idea. That's clean and simple. The value space for IDs is pretty much all characters except comman and whitespace.

robander:
FWIW, I also like @shaneataylor 's idea - clean and simple. (And I think @jelovirt had a pretty clear / appropriate reasoning for why we can't overload `@conref` here).

jelovirt:
Update syntax to `token=start,end` for consistency with RFC 5147.

robander:
:+1: LGTM, I'm OK to merge as-is now, pending any other comments.

PR #2471, opened 2016-09-06
Add sign-off requirement to contribution

jelovirt:
Mandate [Developer Certificate of Origin](http://developercertificate.org/) sign-off for contributions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contribution Guidelines
 
-We welcome contributions to the DITA-OT. 
+We welcome contributions to the DITA-OT.
 
 ## Create an Issue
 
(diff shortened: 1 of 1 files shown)


robander:
LGTM - the DCO here is the same one suggested by one of the folks I work with in IBM who is heavily involved with open source usage.

infotexture:
Looks like we'd need to update the following pages on the website once this is adopted:

* http://www.dita-ot.org/participate
* http://www.dita-ot.org/management
* http://www.dita-ot.org/contribution_policy.html

These are a bit stale anyway _(some still mention SourceForge in some places)_ and could probably be combined for dita-ot/dita-ot.github.io#20 – [Review & consolidate contribution information](https://github.com/dita-ot/dita-ot.github.io/issues/20).

jelovirt:
While adding a sign-off to every commit is very Git centric, I feel it's still acceptable. An alternative would be to add the sign-off to only the PR description, which is equivalent to email patch's description that Linux project uses. However, that would store the sign-offs only to GitHub and since we cannot backup PRs like you could backup email, the information could be lost if we moved away from GitHub for some reason or when GitHub goes under. Thus IMO the best approach is to just mandate that all commits in a PR contain the sign-off and we just have to enforce that. We may be able to automate the check with Travis to off-load some of the burden from PR reviewers.

raducoravu:
> Thus IMO the best approach is to just mandate that all commits in a PR contain the sign-off and we just have to enforce that. 

So all I would need to do extra would be to add this line at the end of my commit:

?

Does that have any legal value?

jelovirt:
Yes, or just use `git commit -s` and that will add it automatically after the commit message.

It has legal value because "we say so". Basically, our contribution policy will state that all commits must have that sign-off to be accepted and when you sign-off a commit, you agree to  http://developercertificate.org/

PR #2473, opened 2016-09-08
Topicpull refactor and bug fix

drmacro:
This is my refactor of topicpullImpl.xsl to use XSLT2 as fully as possible.

The commits reflect 5 refactors, done in stages to make review as easy as possible.

All the integration tests pass for me locally except for the test[lang] test, which was already failing on the develop branch (so not related to or affected by this refactor).

I tried to change the business logic as little as possible so that the overall organization of the templates is the same and to minimize the affect on any potential overriders (should there be any).


jelovirt:
Please remove all merge commits. Just rebase on top of `develop` and force push.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -77,8 +83,8 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
            match="*[contains(@class, ' topic/table ')][*[contains(@class, ' topic/title ')]]"
            use="'include'"/>
   
-  <xsl:key name="id" match="*[@id]" use="@id"/>
-  <xsl:key name="topic" match="*[@id][contains(@class, ' topic/topic ')]" use="@id"/>
+  <xsl:key name="elementsById" match="*[@id]" use="@id"/>
```
jelovirt: This key will also match topics. Either the name should be the old `id` or then the match pattern should be changed to ignore topics.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -100,31 +187,36 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
+      <xsl:variable name="type" as="xs:string?" select="dita-ot:get-link-target-type(.)"/>
+      
+      <xsl:variable name="format" as="xs:string?" select="dita-ot:get-link-format(.)"/>
+      <xsl:variable name="scope" as="xs:string?" select="dita-ot:get-link-scope(.)"/>
+      
+      <xsl:if test="not(@type) and $type">
         <xsl:attribute name="type"><xsl:value-of select="$type"/></xsl:attribute>
       </xsl:if>
-      <xsl:if test="not(@format) and $format!='#none#'">
+      <xsl:if test="not(@format) and $format">
```
jelovirt: IMO we should use `exists` and `empty` to make code a little more explicit. So this could be `empty(@format) and exists($format)`, but I don't feel very strongly about it being in this PR.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -100,31 +187,36 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
+        />
+        <xsl:if test="$importance">
+          <xsl:attribute name="importance" select="$importance"/>
+        </xsl:if>
       </xsl:if>
       <xsl:if test="not(@role)">
-        <xsl:apply-templates select="." mode="topicpull:inherit-and-set-attribute"><xsl:with-param name="attrib">role</xsl:with-param></xsl:apply-templates>
+        <xsl:variable name="role" as="xs:string?"
+          select="dita-ot:get-inherited-attribute-value(., 'role', ())"
+        />
```
jelovirt: The formatting style of pushing `/>` to a new line is something that's not used in almost at all on this code base. Thus please stick to the common convention of `/>` on the same line as the last attribute or element name.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -169,16 +261,20 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
   <!-- If an attribute is specified locally, set it. Otherwise, try to inherit from ancestors. -->
   <xsl:template match="*" mode="topicpull:inherit-from-self-then-ancestor">
     <xsl:param name="attrib"/>
-    <xsl:variable name="attrib-here">
-      <xsl:if test="@*[local-name()=$attrib]"><xsl:value-of select="@*[local-name()=$attrib]"/></xsl:if>
+    <xsl:variable name="attrib-here" as="xs:string?">
+      <xsl:if test="@*[local-name()=$attrib]">
```
jelovirt: The `xsl:if` is redundant, the nested `xsl:value-of` already contains the condition.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -191,33 +287,37 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
     <xsl:choose>
       <!-- Any time the attribute is specified on this element, use it -->
-      <xsl:when test="$attrib-here!=''"><xsl:value-of select="$attrib-here"/></xsl:when>
-      <!-- No ancestors left to check, so the value is not available. -->
-      <xsl:when test="contains(@class,' topic/related-links ')">#none#</xsl:when>
+      <xsl:when test="$attrib-here">
+        <xsl:value-of select="$attrib-here"/>
+      </xsl:when>
+      <xsl:when test="contains(@class,' topic/related-links ')">
+        <xsl:sequence select="()"></xsl:sequence>
```
jelovirt: Use empty-element tag.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -253,35 +353,41 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
         <xsl:copy>
           <xsl:apply-templates select="@*"/>
           <!--create variables for attributes that will be passed by parameter to the getstuff template (which is shared with link, which needs the attributes in variables to save doing inheritance checks for each one)-->
-          <xsl:variable name="type" as="xs:string">
+          <xsl:variable name="type" as="xs:string?">
```
jelovirt: Since we now return `()` for the else-branch, we can just remove the whole `xsl:choose` and just use `<xsl:variable name="type" select="@type" as="xs:string?"/>`

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -475,15 +588,15 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
     </xsl:variable>
 
     <!--now, create the type attribute, if the type attribute didn't exist locally but was retrieved successfully-->
-    <xsl:if test="$localtype='#none#' and not($type='#none#')">
-      <xsl:attribute name="type"><xsl:value-of select="$type"/></xsl:attribute>
+    <xsl:if test="not($localtype) and $type">
```
jelovirt: Preferably use `empty` instead of `not` here.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -167,6 +173,183 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
       ($node/@href and not(contains($node/@class,' delay-d/anchorkey ')) and (some $c in $link-classes satisfies contains($node/@class, $c)))"/>
   </xsl:function>
   
+  <!-- FIXME: Move this to dita utils. -->
+  <xsl:function name="dita-ot:encodeUri" as="xs:string">
```
jelovirt: Why do we need this? Which DITA source will cause the direct URI usage to fail?

drmacro:
The encodeUri function replaces original code that was only escaping spaces. This is more complete and presumably should be  used in other XSLT code that is handling potentially-unescaped URIs. If that case doesn't occur then the function isn't needed. But it's certainly the case that some source documents will not have correctly escaped their URIs and most processors will silently do the escaping.

src/main/xsl/preprocess/topicpullImpl.xsl
drmacro: by "elements" I mean "all elements", which is what the original "id" table matched.

But there is no need to be able to find topics and non-topic elements by ID in the same lookup so I think a better name would be "nontopicElementsByID" and have it exclude topics.

src/main/xsl/preprocess/topicpullImpl.xsl
drmacro: Consistency is good. I don't mind making this change--it's probably a practice I should adopt generally in any case.

drmacro:
I just realized that there's one more refactor I wanted to do: remove the use of "#none#" values for testing non-existence. I'll do that after I do all of Jarno's current comments.

drmacro:
Found two no-longer-used templates from the old get-stuff set of templates and removed those. That change will be included in the comment-response commit.

jelovirt:
I think we should just remove `encodeUri` because the assumption is pretty much that Saxon will handle the URI encoding before using it. It we do actually find a case where manual URI clean-up is needed (and hasn't already been done by `debug-filter`step), then we can add this to *all* XSLTs.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -170,6 +176,183 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
+        <xsl:variable name="resourcePart" as="xs:string?" 
+          select="tokenize($targetURI, '#')[1]"
+        />
+        <xsl:variable name="scope" select="dita-ot:get-link-scope($linkElement, 'local')" as="xs:string"/>
+        <xsl:variable name="format" select="dita-ot:get-link-format($linkElement, 'dita')" as="xs:string"/>
+        <xsl:choose>
+          <xsl:when test="not($scope = ('local'))">
+            <!-- Not a local-scope link, don't resolve it.
+              
+                 FIXME: For peer-scope references may need to actually do resolution
```
jelovirt: Remove

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -170,6 +176,183 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
+  <!-- Given a linking element that has a non-empty @href value,
+       a local scope, and a dita format, attempt to resolve it to 
+       a target element.
+       
+       This function reports any failures to resolve the reference.
+    -->
+  <xsl:function name="dita-ot:getTargetElement" as="element()?">
+    <xsl:param name="linkElement" as="element()"/>
+    
+    <!--    <xsl:message> + [DEBUG] dita-ot:getTargetElement(): linkElement:
```
jelovirt: Remove

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -45,34 +45,8 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
-                 the choice groups that act on the class value. Replaces named
-                 templates with more-normal match templates.
-                 
-                 This makes the code more XSLTish and makes it easier to override
-                 the link text generation.
-                 
-   - Refactor 5: Use @select on xsl:attribute wherever possible                 
-
-  -->
+<!-- 20160908 WEK: Refactored to use XSLT2 approaches: Use functions for getting link target, match templates
```
jelovirt: Remove, we have Git for revision history.

src/main/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -170,6 +176,183 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
+  <xsl:function name="dita-ot:getTargetDoc" as="document-node()?">
+    <xsl:param name="linkElement" as="element()"/>
+    
+    <xsl:variable name="targetURI" as="xs:string?" select="$linkElement/@href"/>
+    <xsl:choose>
+      <xsl:when test="normalize-space($targetURI) = ''">
+        <xsl:sequence select="()"/>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:variable name="resourcePart" as="xs:string?"
```
jelovirt: Use `dita-ot:strip-fragment` function.

jelovirt:
LGTM. I'll merge this manually by squashing some of the commits and will handle the testsuite submodule changes.

jelovirt:
@drmacro could you rebase on top of dita-ot/dita-ot develop branch, remove the commit that changes the testsuite submodule, and add the test into the current integration test set (that is in the dita-ot/dita-ot repo).

drmacro:
I'm not sure which commit you mean by "the commit that changes the testsuite submodule". I didn't intentionally commit any change to the submodule or the reference to it. Looking through the commit history for the PR branch I'm not seeing a commit for the submodule but I'm not sure I know what to look for.

jelovirt:
@drmacro https://github.com/dita-ot/dita-ot/pull/2473/commits/89134bc2a609d3a5f6a0e69cb8064136526e1e81

drmacro:
Removing the commit is beyond my feeble git skills--can you provide guidance?

I've updated my fork to remove the testsuite submodule, synced with the upstream develop branch, and rebase the PR branch to that version of develop.

The commit in question predates the creation of the PR branch so I'm not sure what it means to remove that commit or precisely how to do it.

jelovirt:
Maybe I should just cherry-pick the commits that contain the changes we do need.

You can remove a commit by rebasing interactively, and then remove the commit there.

drmacro:
OK. Let me try doing the interactive rebase since it's something I need to know how to do properly. I was trying to delete the commit from my develop branch but that seemed dicey at best.

drmacro:
OK, I think I did what was necessary to remove the commit. Please confirm and then I'll add the test case (although note that the test case is actually for verify the mappull refactor, I just needed to have it present locally so my test doc that also demonstrated the mappull bug would run)

jelovirt:
Merged as 6b2b53990ef5c7677db62e629b7102a974fa7acd after commit cleanup.

xephon2:
> :information_source: **Useful Tip** 

It may be helpful to compare `develop` and the feature branch to list all changed files with:

```git diff --name-only develop <branch>```

If a file has changed (but should not) you can easily grab the original file from `develop` with:

```git checkout develop -- file.xsl```

This may be helpful before doing a rebase.

PR #2474, opened 2016-09-09
Use arabic numbering for notices in back matter

robander:
We noticed that if a book map has `<notices>` in the back matter, the page numbering switches to roman numerals (under the assumption that this is the Notices section in the front matter). This is normally the correct default for any topic under `<notices>`, which is most commonly located in the front matter, and thus the page number format is picked up from the attribute set `page-sequence.notice`.

By default the body and back matter use arabic numerals; this fix checks to see if the notices is inside `<backmatter>`, and if so, switches format back to arabic.

Feel free to suggest a better way to do this; the current fix uses the same key from later in the template to pick up `<notices>` and check ancestry, which seems straightforward but looks cumbersome somehow.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -459,6 +459,9 @@ See the accompanying license.txt file for applicable licenses.
 
     <xsl:template name="processTopicNotices">
         <fo:page-sequence master-reference="body-sequence" xsl:use-attribute-sets="page-sequence.notice">
+            <xsl:if test="key('map-id', ancestor-or-self::*[contains(@class, ' topic/topic ')][1]/@id)/ancestor::*[contains(@class,' bookmap/backmatter ')]">
+              <xsl:attribute name="format">1</xsl:attribute>
+            </xsl:if>
             <xsl:call-template name="startPageNumbering"/>
             <xsl:call-template name="insertPrefaceStaticContents"/>
             <fo:flow flow-name="xsl-region-body">


src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
```diff
@@ -459,6 +459,9 @@ See the accompanying license.txt file for applicable licenses.
 
     <xsl:template name="processTopicNotices">
         <fo:page-sequence master-reference="body-sequence" xsl:use-attribute-sets="page-sequence.notice">
+            <xsl:if test="key('map-id', ancestor-or-self::*[contains(@class, ' topic/topic ')][1]/@id)/ancestor::*[contains(@class,' bookmap/backmatter ')]">
```
jelovirt: Preferably, move into `page-sequence.notice` attribute-set.

robander: I was thinking about how you couldn't have `<xsl:if>` inside of the attribute set, but hadn't processed that it could be within the `<xsl:attribute>` element inside of the attribute set. Must be tired.

The only concern I have here with that approach is that this means removing the reference to attribute set `page-sequence.frontmatter`, which could have been used to override numbering for the entire front matter. You would now also need to override `page-sequence.notices` on its own. Unless the logic here moved into `page-sequence.frontmatter`, which just feels ... off ... that we would be checking for a back matter ancestor within the front matter attribute set:
```
  <xsl:attribute-set name="page-sequence.notice" use-attribute-sets="__force__page__count">
    <xsl:attribute name="format">
      <xsl:choose>
        <xsl:when test="key('map-id', ancestor-or-self::*[contains(@class, ' topic/topic ')][1]/@id)/ancestor::*[contains(@class,' bookmap/backmatter ')]">1</xsl:when>
        <xsl:otherwise>i</xsl:otherwise>
      </xsl:choose>
    </xsl:attribute>
  </xsl:attribute-set>
```

robander: Another option would be to duplicate the `fo:page-sequence` element, using different attribute sets for back matter vs front matter. That would probably be the cleanest way to handle this while 1) keeping all values in attribute sets, 2) not affecting existing configurations, and 3) fixing the bug.

jelovirt: A separate `fo:page-sequence` is a good alternative too, but you can't add that in a bug fix. My preference is for the `xsl:attribute-set` [solution](#discussion_r78255155)

robander: This works to set it in the attribute set:
```
<xsl:attribute-set name="page-sequence.notice" use-attribute-sets="__force__page__count page-sequence.frontmatter">
    <xsl:attribute name="format">
      <xsl:choose>
        <xsl:when test="key('map-id', ancestor-or-self::*[contains(@class, ' topic/topic ')][1]/@id)/ancestor::*[contains(@class,' bookmap/backmatter ')]">1</xsl:when>
        <xsl:otherwise>i</xsl:otherwise>
      </xsl:choose>
    </xsl:attribute>
  </xsl:attribute-set>
```

But I do wonder if this is going to break some customizations - anybody who overrides `page-sequence.frontmatter` to use format="1" will now get roman numerals for the notices in the front matter, unless / until they override `page-sequence.notice`. 

I don't see a problem with using the `fo:page-sequence` solution in a patch (it's still just a fix), but if you prefer not to, it may be better to go with that solution and do this for 2.4?

jelovirt: The problem with fixing the issue in `processTopicNotices` template instead of the  `page-sequence.notice` attribute set that if someone wants to override that, they actually have to override the whole template, instead of just adding an attribute set. So my preference is using the attribute set in this case.

robander: Ah, I was assuming two attribute sets - so anybody who has overridden the current one would be good, but we'd still have an attribute set that lets people override the back matter version. Something like:
```
<xsl:choose>
  <xsl:when test="back matter notices">
    <fo:page-sequence master-reference="body-sequence" xsl:use-attribute-sets="page-sequence.notice.backmatter">
      ...call template for everything else...
    </fo:page-sequence>
  </xsl:when>
   <xsl:otherwise>
     <fo:page-sequence master-reference="body-sequence" xsl:use-attribute-sets="page-sequence.notice">
      ...call template for everything else...
    </fo:page-sequence>
  </xsl:when>
</xsl:choose>
```

Advantage there is that no current customizations would be affected, because the front matter attribute set is untouched. We still fix the back matter one, but it can also be updated via attribute set. The only code duplication is that we have two `<fo:page-sequence>` elements, with different attribute sets.

jelovirt: I would just go with one attribute set, since this is a bug. For 2.4, we can either add additional attribute sets for frontmatter/backmatter notices. Though, we already have quite a few of those, IMO bordering on too many.

jelovirt: @robander what's the plan for this? Move to attribute set?

robander: I don't want to move to one attribute set - there are clear cases where it will break existing configurations. We could do that in 2.4, but I don't see a good justification for breaking it, since there are equally good ways to handle without that.

I can see a few options:

- The duplicate `<fo:page-sequence>`, with different attribute sets. Doesn't break anybody, fixes the problem for everybody.
- I was playing with tables earlier this week and noted the `new-attr-set-reflection` named template, used to apply a new attribute set. That template seems kind of ugly but it's used a lot already for tables, and maybe that's a normal way to conditionally add attribute sets. So, instead of my original "If back matter, reset format", we could use "If back matter, add this back.matter.notices attribute set"
- Now wondering if I was over-complicating your original attribute set suggestion. My idea was to have all the logic in the notices attribute set -- "When back matter use 1, otherwise i". But I think perhaps it would work to inherit "i" from front matter -- thus preserving any customization -- and just use "if back matter reset to 1" inside the notices set. Hmm. That seems like it would work and not cause compatibility issues, right? Is that what you were thinking, rather than my `<xsl:choose>`?

jelovirt: The attribute-set reflection is just awful, let's not used it for anything anymore.

drmacro: OMG it is so bad--I was just looking at it yesterday in the context of a customer's overrides and I never realized just how awful that is.

Personally I think the use of attribute sets should be removed entirely and replaced with something easier to understand and override. But that's easy for me to say.

robander: > The attribute-set reflection is just awful

OK, good not just me...

> Personally I think the use of attribute sets should be removed entirely...

Can't argue with that, but talk about destroying peoples' customizations...

Considering both of those comments, I think my preference would still be the first bullet above. It's a slight bit more verbose, but not out of line with the rest of the code. It remains consistent with everything else (using attribute sets to set the format), fixes the problem, and cannot interfere with existing customizations to front matter. I also don't have a strong preference for 2.3.3 versus 2.4 -- my choice would be 2.3.3 to get the fix out sooner, but if a new attribute set is seen as a new feature, I'm fine with 2.4.

robander: Specifically, I was thinking this - when in back matter use a different attribute set, otherwise no change:

```
<xsl:template name="processTopicNotices">
  <xsl:choose>
    <xsl:when test="back matter">   
      <fo:page-sequence master-reference="body-sequence" xsl:use-attribute-sets="page-sequence.backmatter.notice">
        <xsl:apply-templates select="." mode="processTopicNoticesContents"/>
      </fo:page-sequence>
    </xsl:when>
    <xsl:otherwise>
      <fo:page-sequence master-reference="body-sequence" xsl:use-attribute-sets="page-sequence.notice">
        <xsl:apply-templates select="." mode="processTopicNoticesContents"/>
      </fo:page-sequence>
    </xsl:otherwise>
  </xsl:choose>
</xsl:template>

<xsl:template match="*" mode="processTopicNoticesContents">
  <!-- everything currently in the fo:page-sequence -->
</xsl:template>
```

jelovirt: My preference is that we don't introduce a new mode in the patch release. Instead, use the approach IMO should be used to replace attribute reflection. (Attribute reflection is a result of XSLT 1 limitations, with XSLT 2 we no longer need it):

```
<xsl:template name="processTopicNotices">
  <xsl:variable name="atts" as="element()">
    <xsl:choose>
      <xsl:when test="back matter">
        <dummy xsl:use-attribute-sets="page-sequence.backmatter.notice"/> 
      </xsl:when>
      <xsl:otherwise>
        <dummy xsl:use-attribute-sets="page-sequence.notice"/> 
      </xsl:otherwise>
    </xsl:choose>
  </xsl:variable>
  <fo:page-sequence master-reference="body-sequence">
    <xsl:copy-of select="$atts/@*"/>
    ...
```

It's a bit verbose, but unfortunately that's the only way we can decide between attribute sets.

jelovirt: In 2.4 or 3.0 we can move to replacing those named templates with moded templates, or come up with an alternative way to process the page sequence.

jelovirt:
Cherry-picked as 9ecf200b23178f89b8bc8ace6dd025d02d688896

Issue #2380, opened 2016-05-23
Disable topicpull for transtypes that don't need it


jelovirt:
Disable topicpull for transtypes that handle their own link text resolution to remove unneeded processing. This will apply to e.g. PDF that will resolve links inside the FO processing. It should be easy to disable topicpull and topicpull should not contain processing that has other side-effects.

PR #2484, opened 2016-09-17
Disable preprocessing topicpull in PDF

jelovirt:
PDF will construct link text and preprocessed link texts are not needed.

Implements PDF part of  #2380.

Signed-off-by: Jarno Elovirta jarno@elovirta.com


jelovirt:
This may break related link grouping because types will be missing.

jelovirt:
This will also break `<fn>` links.

jelovirt:
Closing as inactive

PR #2487, opened 2016-09-19
Fix missing "Continued" variable marked as TODO

robander:
While working on support for tables and index terms continued across pages, we noticed that 4 of the original PDF2 languages contained a "TODO" comment for the Index Continued variable (FI, RO, RU, and SV). I've put in a translation based on variable files we've used internally for many years.

While working on those, I noticed that both Russian and Romanian had a TODO comment for "Table of Contents Preface", even though both already contain a translation of Preface. I replaced that TODO with the existing translation and a colon. This uses the same pattern as existing languages that correctly set both variables.


robander:
(Not expecting any controversy here, just replacing a few TODO items with the intended content, so I think it's ready for merge.)

raducoravu:
I confirm the Romanian ones are proper :+1:

PR #2492, opened 2016-09-25
Consistent CLI argument format and validation

jelovirt:
Adds support for

* setting long form options using double hyphen
* defining properties with double hyphen addition to `-D`
* validate properties based on plugin configuration


jelovirt:
@dita-ot/docs the documentation should use `--long=value` instead of other alternatives. We still support the old formats for backwards compatibility, but the documentation should point to the common *NIX CLI format.

For passing properties, the format `--<property>=<value>` should be used. An error message is thrown if the property is not recognized or if the value is invalid (for enumerated property value space).

infotexture:
OK, I'll add a new docs issue to ensure it's on the radar for the 2.4 release.

PR #2497, opened 2016-10-05
Map-first preprocessing

jelovirt:
Change preprocessing to first process maps as far as possible and only then start processing topics. This simplifies the processing and allows combining modules, due to fewer process-order constraints.
- [x] combine `gen-list` and `debug-filter` into single target
- [x] separate map and topic processing
- [x] add new preprocess target
- [x] support topic start file
- [ ] split `maplink` to collect map links during map-phase and push in topic-phase
- [x] support filtering
- [ ] support subject schemes
- [x] support uplevels after cleanup

Signed-off-by: Jarno Elovirta jarno@elovirta.com


jelovirt:
Plan is to first break everything, then make it pass tests, and then clean up. It would be nice to have very little code duplication, but in the end some code will probably be duplicated. Since this code is intended to be used to replace the old preprocessing pipe, the old code will be removed on the next major release.

jelovirt:
Moved enhancements to #2625 and #2626

mironovalexey:
I suppose that map-first preprocessing will [exclude topics, exploded by `@chunk="by-topic",`from reltables processing](https://github.com/dita-ot/dita-ot/issues/2660) too. 
Is this the expected result?

PR #2499, opened 2016-10-06
Change HTML5 to use libsass instead of Compass

jelovirt:
Change HTML5 to use libsass instead of Compass. This will not use JRuby and is a lot faster as a result.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ task copyInstall(type: Copy) {
     }
     destinationDir = file("src/main/lib")
 }
-task buildLocal(type: JavaExec, dependsOn: [copyInstall, ":fo:copyInstall", ":axf:copyInstall", ":xep:copyInstall", ":html5:compassCompile"]) {
+task buildLocal(type: JavaExec, dependsOn: [copyInstall, ":fo:copyInstall", ":axf:copyInstall", ":xep:copyInstall", ":html5:compileSass"]) {
     description "Build archives and install all plugins with dependencies"
     main = "org.apache.tools.ant.launch.Launcher"
     classpath = sourceSets.main.runtimeClasspath + files("${projectDir}/src/main", "${projectDir}/src/main/lib", "${projectDir}/src/main/resources")
--- a/src/main/plugins/org.dita.html5/build.gradle
+++ b/src/main/plugins/org.dita.html5/build.gradle
@@ -1,27 +1,26 @@
 /*
  * This file is part of the DITA Open Toolkit project.
  *
- * Copyright 2015 Jarno Elovirta
+ * Copyright 2016 Jarno Elovirta
  *
  * See the accompanying LICENSE file for applicable license.
  */
-buildscript {
-    dependencies {
-        classpath "com.github.jruby-gradle:jruby-gradle-plugin:0.1.12"
-    }
+apply plugin: 'java'
+
+repositories {
+    mavenCentral()
+}
+dependencies {
+    compile group: 'com.cathive.sass', name: 'sass-java', version: '4.0.0'
 }
 
-plugins {
-  id "com.github.robfletcher.compass" version "2.0.6"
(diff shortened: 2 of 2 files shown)


jelovirt:
This needs to be tested on Windows. Works on OS X.

jelovirt:
Fails on Travis CI because libsass cannot be found.

jelovirt:
@eerohele are you able to help out with this?

eerohele:
I get this error on Windows:

https://discuss.gradle.org/t/ant-task-with-native-libs/19527

jelovirt:
That's the same error message my customer got on Windows and one that Travis gets on Linux.

I wonder if it would be easier to just go to Node.js for SASS compilation.

eerohele:
You could just forgo using the Ant task and do something like this instead (rough, incomplete, untested, illustrative example):

```groovy
/*
 * This file is part of the DITA Open Toolkit project.
 *
 * Copyright 2016 Jarno Elovirta
 *
 * See the accompanying LICENSE file for applicable license.
 */
buildscript {
    dependencies {
        classpath 'com.cathive.sass:sass-java:4.0.0'
    }

    repositories {
        mavenCentral()
    }
}

apply plugin: 'java'

task deleteCss(type: Delete) {
    delete file('css')
}

import com.cathive.sass.SassCompilationException
import com.cathive.sass.SassContext
import com.cathive.sass.SassFileContext
import com.cathive.sass.SassOptions
import com.cathive.sass.SassOutputStyle

import java.io.IOException
import java.nio.file.Path
import java.nio.file.Paths
import java.io.FileOutputStream

def compile(String file) {
    new File(project.projectDir, "css").mkdir()

    Path srcRoot = Paths.get(new File(project.projectDir, "sass").getAbsolutePath())
    SassContext ctx = SassFileContext.create(srcRoot.resolve(file + ".scss"))

    SassOptions options = ctx.getOptions()

    options.setOutputStyle(SassOutputStyle.NESTED)

    try {
        ctx.compile(new FileOutputStream(new File(project.projectDir, "css/${file}.css")))
    } catch (SassCompilationException e) {
        System.err.println(e.getMessage())
    } catch (IOException e) {
        System.err.println(String.format("Compilation failed: %s", e.getMessage()))
    }
}

task compileSass(dependsOn: [deleteCss]) << {
    compile("commonltr")
    compile("commonrtl")
}
```

That seems to work for me on Windows otherwise, except that it doens't work when the daemon is enabled. Didn't have time to look into that yet.

jelovirt:
I thought about that, using the Java code directory via Groovy. If you're able to test how to make it work on Windows, please push to the PR branch directly, you should have access.

eerohele:
Using Gradle 3.1 [resolves the issue on Windows](https://issues.gradle.org/browse/GRADLE-3104).

If that's fine by you, I can try to polish the code a bit and push the changes.

jelovirt:
@eerohele Yep, please push the changes.

infotexture:
So we'll update the Gradle wrapper to 3.1 then?

infotexture:
If we're updating the wrapper for other reasons, we might want to switch from the `-all` distribution to the smaller (default) binary [distribution type](https://docs.gradle.org/current/release-notes#the-distribution-type-can-be-selected-by-the-wrapper-task).

eerohele:
See bdb778d4fa8353c656474e8629a09f49bd3cfab5.

eerohele:
Actually, I think the `deleteCss` task can be removed, since deleting the output folder on every SASS compilation breaks the up-to-date check. I did remove the dependency to it, but the task is still there.

infotexture:
I tested locally on OS X 10.11.6, works for me — and also seems to be much faster. 👍

eerohele:
Works fine for me on Windows, too.

dfemia:
Thank you for fixing this issue so quickly!

PR #2511, opened 2016-10-26
Move XHTML parameters & remove `common-html` transtype

infotexture:
XHTML parameters are no longer shared after HTML5 split from XHTML for dita-ot/dita-ot#2405.

This PR ensures that the generated docs topics list these parameters as specific to XHTML rather than as shared [HTML5 and XHTML parameters](http://www.dita-ot.org/2.3/parameters/parameters-common-html.html).


jelovirt:
Sign-off missing.

Issue #2512, opened 2016-10-30
Bug in "dita-utilities.xsl" getVariable when called from string context


hscherzer:
There seems to be a bug in the official DITA-OT2

If **getVariable** is called as follows ...

```
    <xsl:variable name="$myNode" select="." as="node()"/>
    <xsl:analyse-string select="..." regex="{'something'}">
        <xsl:matching-substring>
           <xsl:call-template name="getVariable">
               <xsl:with-param name="id" select="'Chp'"/>
               <xsl:with-param name="params" select="$myNode""/>
           </xsl:call-template>
        </xsl:matching-substring>
    </xsl:analyse-string>
```

then the compile breaks because in xsl\common\dita-utilities.xsl

```
    <xsl:template name="getVariable">
```

... issues ...

```
        <xsl:sequence select="dita-ot:get-variable(., $id, $params)"/>
```

... which issues (.) as the first parameter. But (.) in the context 'matching-substring' is of **xs:string** and the compile breaks because dita-ot:get-variable expects a **node** as first parameter

The fast fix here could be

```
<xsl:template name="getVariable">
    <xsl:param name="id" as="xs:string"/>    
    <xsl:param name="params" as="node()*"/>
        <xsl:choose>
            <xsl:when test=". instance of xs:string">
                <xsl:sequence select="dita-ot:get-variable($params, $id, $params)"/>
            </xsl:when>
            <xsl:otherwise>
                <xsl:sequence select="dita-ot:get-variable(., $id, $params)"/>
            </xsl:otherwise>
        </xsl:choose>
  </xsl:template>
```

but a better fix might be thought of

robander:
We wouldn't want the getVariable parameters themselves to be the context. That said, I could see the usefulness adding a parameter that provides the context (which is used to determine the language). 

That is, we could add a `$ctx` parameter that defaults to the current node, but allows you to pass in a context node if calling `getVariable` from an unsupported context. The `$ctx` parameter in the function call would become the `$ctx` variable in the `dita-ot:get-variable` function, so that lines up nicely. The two uses I can see for this:

- When calling from an unsupported context (such as a string), this provides a way to set the context so the function can still find a language and find a string
- From any context, I think this would provide a mechanism to override the language, by passing in something like `<node xml:lang="fi"/>` (though I'm not sure how useful this would be outside of a test scenario).

robander:
This seems to work for me - language retrieval still works from current node and ancestor node:

```xml
[...]

PR #2513, opened 2016-10-31
Allow getVariable from string context #2512

robander:
getVariable is currently designed to work from the context of an XML node. This enhancement allows getVariable to be called from any context; it also allows somebody to override the context, which could be used to override the language (if that's ever needed).


jelovirt:
LGTM

PR #2515, opened 2016-11-03
Revert to FOP 1.1

jelovirt:
During testing it was discovered that FOP 2.x comes with a major performance degradation. This PR reverts back to FOP 1.1 that was originally bundled, but still uses Gradle dependency management for FOP JAR files.


jelovirt:
This PR is redundant. There seems to be no performance hit from using FOP 2.1. The original reports were invalid.

Issue #2446, opened 2016-08-08
Menu cascade accessibility


wintrymix:
Common practice is to use a right angle bracket as part of a menu cascade: `File > New`.  In this case, the angle bracket is considered ASCII art, and per Web Content Accessibility Guidelines 2.0, should be avoided.

As a workaround for HTML, it is acceptable to wrap an abbreviate tag around the the angle bracket so screen readers will detect and read the content of the abbreviation tag: 

``` html
Choose File <abbr title="and then"> > </abbr> New

```

This issue is to request that we add processing to the HTML processing in the DITA OT to detect right angle brackets in menu cascades and wrap in abbreviation tags when publishing HTML.

jelovirt:
Created #2516 to implement this.

robander:
#2516 will be in release 2.5.

At the moment DITA-OT only ships with English text, but the variable is part of other strings, so when a DITA-OT user gets custom translations it will be part of the set.

PR #2516, opened 2016-11-05
Add wrapper to menucascade separator for accessibility #2446

jelovirt:
Add wrapper element to `<menucascade>` separator for accessibility in HTML5.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/ui-d.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/ui-d.xsl
@@ -26,7 +26,14 @@ See the accompanying LICENSE file for applicable license.
   <xsl:if test="ancestor::*[contains(@class,' ui-d/menucascade ')]">
    <xsl:variable name="uicontrolcount"><xsl:number count="*[contains(@class,' ui-d/uicontrol ')]"/></xsl:variable>
     <xsl:if test="$uicontrolcount&gt;'1'">
-      <xsl:text> > </xsl:text>
+      <abbr>
+        <xsl:attribute name="title">
+          <xsl:call-template name="getVariable">
+            <xsl:with-param name="id" select="'a11y.and-then'"/>
+          </xsl:call-template>
+        </xsl:attribute>
+        <xsl:text> > </xsl:text>
+      </abbr>
     </xsl:if>
   </xsl:if>
    <span class="uicontrol">
--- a/src/main/plugins/org.dita.html5/xsl/ui-d.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/ui-d.xsl
@@ -26,13 +26,18 @@ See the accompanying LICENSE file for applicable license.
   <xsl:if test="ancestor::*[contains(@class,' ui-d/menucascade ')]">
    <xsl:variable name="uicontrolcount"><xsl:number count="*[contains(@class,' ui-d/uicontrol ')]"/></xsl:variable>
     <xsl:if test="$uicontrolcount&gt;'1'">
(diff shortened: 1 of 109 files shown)


jelovirt:
Translations of "and then" needed.

robander:
Looks good. I don't think we've established a policy for what to do with languages that don't yet have a translation, though we've talked about it several times. I don't see us getting quick translations for these, so it might be a good time to come up with that policy. Worth trying to figure that out here, or is there a better spot?

eelle:
Consider also that in translation, Hebrew and Arabic read right-to-left.  The solution with that uses text instead of the right angle bracket is a better solution.

robander:
I've been looking at this to see if we could get into 2.5 and noticed a couple of things.

- We already have a variable for the ` > ` separator, in the PDF variable files. We should be using that in our HTML, XHTML, and `text-only` processing, rather than hard coding the character in the code.
- At this point - with release planned next week - I don't want to switch the Arabic / Hebrew variables from `>` to `<` ... I've had too many issues in the past with those characters displaying wrong when menus are in English or in Arabic / Hebrew. But, using the variable everywhere now allows any local configuration to put in text or a different separator.
- In HTML, when you add `@title` to `<abbr>`, the default presentation switches to underline. Because that only surrounds the `>` character, it makes it appear as `≥`, which looks wrong -- `A ≥ B ≥ C`. So, I've tweaked the CSS so that `abbr` inside of menu cascade removes the underline.
- I know these elements exist everywhere in non-English documents. I don't want to surprise people with a new English "and then" pop-up in non-English docs. So, I've made empty translations as the default. We can add them as translations are available.

robander:
The simple test file I used, to test default language, English, Finnish, and a fake sub-group of Russian:
```
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN"
 "topic.dtd">
<topic id="ui">
<title>Testing UI Control</title>
<body>
<ul>
<li>Default: <menucascade><uicontrol>a</uicontrol><uicontrol>b</uicontrol><uicontrol>c</uicontrol></menucascade></li>
<li xml:lang="en">En: <menucascade><uicontrol>a</uicontrol><uicontrol>b</uicontrol><uicontrol>c</uicontrol></menucascade></li>
<li xml:lang="fi">Fi: <menucascade><uicontrol>a</uicontrol><uicontrol>b</uicontrol><uicontrol>c</uicontrol></menucascade></li>
<li xml:lang="ru-fake">ru-fake: <menucascade><uicontrol>a</uicontrol><uicontrol>b</uicontrol><uicontrol>c</uicontrol></menucascade></li>
</ul>
</body>
</topic>
```

src/main/plugins/org.dita.html5/xsl/ui-d.xsl
```diff
@@ -26,7 +26,19 @@ See the accompanying LICENSE file for applicable license.
...
+        <xsl:call-template name="getVariable">
+          <xsl:with-param name="id" select="'a11y.and-then'"/>
+        </xsl:call-template>
+      </xsl:variable>
+      <abbr>
+        <xsl:if test="exists($a11y.text)">
+          <xsl:attribute name="title" select="$a11y.text"/>          
+        </xsl:if>
+        <xsl:call-template name="getVariable">
+          <xsl:with-param name="id" select="'#menucascade-separator'"/>
```
jelovirt: To nitpick, `#menucascade-separator` is defined in PDF2 variable files. This means if you don't have PDF2 plugin installed, this will fail. Maybe it would be easier to move it from PDF2 to common variable files?

PR #2519, opened 2016-11-15
Clean PDF2 styling and move legacy styling plugin

jelovirt:
Clean PDF2 to remove styling that is almost always overridden or are ugly, and move legacy styling to a separate plugin to reproduce them.

* [Legacy plugin repository][1]

[1]: https://github.com/dita-ot/org.dita.pdf2.legacy

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/commons-attr.xsl
+++ b/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/commons-attr.xsl
@@ -85,7 +85,6 @@ See the accompanying LICENSE file for applicable license.
 
   <xsl:attribute-set name="common.link">
     <xsl:attribute name="color">blue</xsl:attribute>
-    <xsl:attribute name="font-style">italic</xsl:attribute>
   </xsl:attribute-set>
 
     <xsl:attribute-set name="__unresolved__conref">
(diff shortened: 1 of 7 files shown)


jelovirt:
We should remove descriptions from related links. Right now the output is

> **Related reference**
> JavaHelp parameters on page 82
> Certain parameters are specific to the JavaHelp transformation.

The problem may unsuitable `<shortdesc>` descriptions in our documentation, but IMO they make the flow of the page much worse and don't improve the content that much.

robander:
:+1: This looks like a good direction to me. Cleaning up this code has been needed for a long time.

FWIW we discussed this item in last week's contributor call -- particularly to get feedback from Leigh White, who has written so much about this transform, and everybody on the call seemed to be on board with this move. Not sure if any others want to chime in. Minutes from that call are here: https://github.com/dita-ot/dita-ot/wiki/Meeting-minutes-2017-02-02

infotexture:
> We should remove descriptions from related links. 

👍 I think this is worth doing, especially since they are currently rendered inconsistently. 

Some links in a collection are augmented with the `<shortdesc>` content, whereas others in the same collection are not — even though the target topics include short descriptions. Upon closer inspection, it appears as though link descriptions are currently dropped if the short descriptions of the target topics contain any inline markup.

infotexture:
Since this PR removes indents and borders for `<example>`, we might also want to consider doing the same for long quote `<lq>` elements.

Issue #2538, opened 2016-12-05
Link to keyscoped topic not properly resolved.


raducoravu:
I'm attaching a ZIP sample.
[improperLinkToKeyScopedTopic.zip](https://github.com/dita-ot/dita-ot/files/631027/improperLinkToKeyScopedTopic.zip)

So, the DITA Map looks like this:
```xml
<topicgroup keyscope="product1_scope">
  ...........
 <topicref keys="archiving" href="product1/c_archiving_mod.dita"/>
</topicgroup>
<topicgroup keyscope="product2_scope">
 ...............
 <topicref keys="archiving" href="product1/c_archiving_mod.dita"/>
</topicgroup>
<topicref href="linking.dita"/>
```

Because "c_archiving_mod.dita" is referenced in two key scopes, we will have in the output file two HTMLs for it: "c_archiving_mod.html" and "c_archiving_mod-1.html". And that's good.
The "linking.dita" should generate links to both those HTML files using fully qualified key scoped:
```xml
<p>Links <xref keyref="product1_scope.archiving"
/> to <xref keyref="product2_scope.archiving"
 />.</p>
```
But in the generated "linking.html" you will have both links point to the same HTML file (the one generated from the first context).

I looked a little bit in the code, the method `org.dita.dost.writer.KeyrefPaser.processElement(Attributes)` resolves key references based on the current "KeyScope" map. Unfortunately the map contains these mappings:
```
product1_scope.archiving=product1_scope.archiving=product1/c_archiving_mod.dita
product2_scope.archiving=product2_scope.archiving=product1/c_archiving_mod.dita
```
although in my opinion the "product2_scope.archiving" should have been re-written (possibly in the org.dita.dost.module.KeyrefModule.execute) to point to "product1/c_archiving_mod-1.dita" because the DITA Map has already been re-written to do so.

Might also be related with #2523

raducoravu:
Somehow when the "org.dita.dost.module.KeyrefModule.walkMap" updates the @href's in the DITA Map the key scopes would also need to be updated to point to the new hrefs... Not sure how this could be done though.

jelovirt:
I'm currently debugging this. It's kinda tricky to find the best stage when to rewrite the key targets. I think the only correct location is after all `ResolveTask` jobs have been generated. Then we know all the rewrites that are needed in for all scopes.

raducoravu:
[...]

PR #2539, opened 2016-12-07
Fix duplicate link targets in key scopes #2538

jelovirt:
Fix key link targets that have been rewritten because of duplicate use in multiple scopes.


jelovirt:
Key targets are adjusted for duplicates but scope prefixed keys are still broken.

PR #2552, opened 2016-12-27
Output error messages in red colour in Terminal

jelovirt:
Use [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) to colour error messages in `dita` command output. When log is written to a file, colour output is disabled. Colour output can be toggled using `cli.color=true|false` configuration option.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/DefaultLogger.java
+++ b/src/main/java/org/dita/dost/invoker/DefaultLogger.java
@@ -45,6 +45,17 @@ import org.apache.tools.ant.util.FileUtils;
  * 
  */
 class DefaultLogger implements BuildLogger {
+
+    public static final String ANSI_RESET = "\u001B[0m";
+    public static final String ANSI_BLACK = "\u001B[30m";
+    public static final String ANSI_RED = "\u001B[31m";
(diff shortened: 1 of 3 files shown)


jelovirt:
Could someone on Windows test this? Works on OS X, and most likely on Linux, but haven't a clue if this will work on Windows.

eerohele:
Except that the colors don't work on basic Windows consoles (`cmd.exe` or PowerShell). So the colors should only be enabled on platforms other than Windows.

jelovirt:
@eerohele are you able to verify that colouring is disabled on Windows?

eerohele:
Yes, it works correctly: no colors on Windows, either with `cmd.exe` or a PowerShell prompt.

infotexture:
LGTM in `bash` & `fish` shells on both El Capitan & Sierra. 

> Warnings could also be displayed in yellow

👍 That would be a nice enhancement.

We might also consider displaying the `Error:`/`Warning:` portions of the strings and/or `[DOT…]` message numbers in bold or increased intensity to make them stand out in longer messages, but red text for errors is already a big improvement.

Issue #1885, opened 2015-04-02
NPE when generating Eclipse Help [DOT 2.0]


raducoravu:
A DITA composite like:

``` xml
<!DOCTYPE dita PUBLIC "-//OASIS//DTD DITA Composite//EN" "ditabase.dtd" []>
<dita>
 <glossentry id="abc123">
    <glossterm>Term 1</glossterm>
    <glossdef>Definition 1</glossdef>
  </glossentry>
</dita>
```

A bookmap with a glossarylist pointing to it:

``` xml
<backmatter>
    <booklists>
    <glossarylist><topicref href="Glossary.dita" print="yes" navtitle="Glossary" type="glossary"/> </glossarylist>
    <indexlist/>
    </booklists>
</backmatter>
```

The NPE:

```
C:\Users\radu_coravu\Desktop\DOT2.0.1\DITA-OT2.0.1\build.xml:41: The following error occurred while executing this line:
C:\Users\radu_coravu\Desktop\DOT2.0.1\DITA-OT2.0.1\plugins\org.dita.base\build_preprocess.xml:78: Failed to run pipeline: null
    at org.dita.dost.invoker.ExtensibleAntInvoker.execute(ExtensibleAntInvoker.java:199)
    at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
    at sun.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
    at java.lang.reflect.Method.invoke(Unknown Source)
    at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:106)
    at org.apache.tools.ant.Task.perform(Task.java:348)
    at org.apache.tools.ant.Target.execute(Target.java:435)
    at org.apache.tools.ant.Target.performTasks(Target.java:456)
    at org.apache.tools.ant.Project.executeSortedTargets(Project.java:1393)
    at org.apache.tools.ant.helper.SingleCheckExecutor.executeTargets(SingleCheckExecutor.java:38)
    at org.apache.tools.ant.Project.executeTargets(Project.java:1248)
    at org.apache.tools.ant.taskdefs.Ant.execute(Ant.java:441)
    at org.apache.tools.ant.taskdefs.CallTarget.execute(CallTarget.java:105)
    at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
[...]

PR #2553, opened 2017-01-03
Ignore <dita> when resolving export anchors #1885

eerohele:
https://github.com/eerohele/dita-ot-issues/tree/master/fixtures/1885


jelovirt:
This will not push missing classes to the class stack. Will that work correctly if some element other than the root will not have class attribute? My preference would be to keep the stack in sync with the parsed elements and just push a null in there when there's no class and handle the nulls.

jelovirt:
Fixed this in 1c57724492dc13037db551d14254e37e4cecccc1 with an approach that pushes to the class stack on every element.

PR #2562, opened 2017-01-12
Added missing dependencies for dost.jar in build.xml

caio2k:
The dita:extesion is not including the two dependencies for dost.jar, which renders build.xml useless.
Probably the way it is done in this PR is not state-of-art, but at least it raises a flag.


jelovirt:
The `dost.class.path` construct in Ant is legacy code that should not be used. All classpath related declarations should be handled _outside_ the Ant process. The path is still there for backwards compatibility reasons, but it will be removed in DITA-OT 3.0.

jelovirt:
Rejecting because we don't want to maintain classpath inside Ant.

Issue #2561, opened 2017-01-12
Support for mathmlref element


robander:
Right now I'm working on support for `<mathmlref>`. This is a subset of #1774, but I don't expect to implement all of it in DITA-OT at this point so I've opened an issue specific to that element. (Once complete, whatever is done to make it work should be easily redone to make `<svgref>` work, a small subset of #1123)

The DITA specification says:
> Processors _SHOULD_ process the MathML as though the `<m:math>` element had occurred directly in the content of the containing `<mathml>` element.

http://docs.oasis-open.org/dita/dita/v1.3/os/part2-tech-content/langRef/technicalContent/mathmlref.html

I've got this part of the way done but have encountered a few issues:
1. `<mathmlref>` appears inside of a specialization of `<foreign>`, so the preprocess ignores it - the reference never registers. At this point I've set `MATHML_D_MATHMLREF` to the class value, as with other known class values, and added exceptions in `DebugFilter.java` so that it's processed. This feels like a hack. I wouldn't think this is appropriate for my own specializations. But this is an OASIS one, with an OASIS defined processing behavior, so it seems like we really should extend things to support it. For example: `if (foreignLevel <= 1 || MATHML_D_MATHMLREF.matches(atts.getValue(ATTRIBUTE_NAME_CLASS))){`
1. This feels very much like the `<coderef>` element. So I've used that debug filter to get the math targets added into `.job.xml`, because I expected a later step to pull in the math, much like with code references. But, I'm not sure this is the correct approach.
1.  Again like coderef, I tried to add a SAX filter, in the same `topic-fragment` step -- but it doesn't use the XML catalog, so couldn't parse the Math files (they have a DTD reference). I later heard from @jelovirt that the step which processes image metadata might be a good (catalog-aware) spot to add the filter, but I haven't tried this yet.
1. I'm more comfortable in XSL anyway so I tried to do that instead - but the `.job.xml` file isn't set up to retrieve only topics that use math, and throwing every DITA topic into a new XSL step is wasteful. Plus, so far, the files are processed in the source directory like code targets, and don't exist in the temp directory.

At this point I decided to seek advice. What would the ideal approach be here?

robander:
[...]

PR #2569, opened 2017-01-13
Support processing DITA elements within foreign / unknown

robander:
This is coming out of discussion in #2561 but distinct from that issue.

The `foreign` and `unknown` elements are outlets for non-DITA content. As such, DITA-OT more or less ignores all content during preprocessing. However, they will often contain DITA elements, possibly as a way to provide alternate versions for processors that cannot recognize or work with the actual foreign / unknown material. (In the math context, the standard uses an `xref` specialization for other reasons.) DITA-OT preprocess should recognize DITA elements -- those with valid DITA `@class` attributes -- and handle them as it would handle those same elements outside of a foreign context.

This is not intended to cover rendering the elements, which will always vary based on transform type and based on what else is part of the `foreign` or `unknown` elements.

### DIFF
--- a/src/main/java/org/dita/dost/reader/CopyToReader.java
+++ b/src/main/java/org/dita/dost/reader/CopyToReader.java
@@ -8,6 +8,7 @@
 package org.dita.dost.reader;
 
 import org.dita.dost.log.MessageUtils;
+import org.dita.dost.util.DitaClass;
 import org.dita.dost.writer.AbstractXMLFilter;
 import org.xml.sax.Attributes;
 import org.xml.sax.SAXException;
@@ -30,10 +31,10 @@ import static org.dita.dost.util.URLUtils.*;
  */
 public final class CopyToReader extends AbstractXMLFilter {
 
-    /** Map of copy-to target to souce */
+    /** Map of copy-to target to source */
     private final Map<URI, URI> copyToMap = new HashMap<>(16);
-    /** foreign/unknown nesting level */
-    private int foreignLevel = 0;
+    /** DITA class values for open elements **/
+    private final Deque<DitaClass> classes = new LinkedList<>();
     /** chunk nesting level */
     private int chunkLevel = 0;
     /** Stack for @processing-role value */
@@ -63,7 +64,7 @@ public final class CopyToReader extends AbstractXMLFilter {
      * Reset the internal variables.
      */
     public void reset() {
-        foreignLevel = 0;
+        classes.clear();
         chunkLevel = 0;
         copyToMap.clear();
         processRoleStack.clear();
@@ -86,11 +87,12 @@ public final class CopyToReader extends AbstractXMLFilter {
         processRoleStack.push(processingRole);
 
         final String classValue = atts.getValue(ATTRIBUTE_NAME_CLASS);
-
-        if (foreignLevel > 0) {
-            foreignLevel++;
-        } else if (TOPIC_FOREIGN.matches(classValue) || TOPIC_UNKNOWN.matches(classValue)) {
-            foreignLevel++;
+        
+        final DitaClass cls = atts.getValue(ATTRIBUTE_NAME_CLASS) != null ? new DitaClass(atts.getValue(ATTRIBUTE_NAME_CLASS)) : new DitaClass("");
+        if (cls.isValid()) {
+        	classes.addFirst(cls);
+        }else {
+        	classes.addFirst(null);
         }
 
         if (chunkLevel > 0) {
@@ -109,10 +111,8 @@ public final class CopyToReader extends AbstractXMLFilter {
     @Override
     public void endElement(final String uri, final String localName, final String qName) throws SAXException {
         processRoleStack.pop();
-
-        if (foreignLevel > 0) {
-            foreignLevel--;
-        }
+        classes.pop();
+        
         if (chunkLevel > 0) {
             chunkLevel--;
         }
--- a/src/main/java/org/dita/dost/reader/GenListModuleReader.java
+++ b/src/main/java/org/dita/dost/reader/GenListModuleReader.java
@@ -12,12 +12,15 @@ import static org.dita.dost.util.Configuration.*;
 import static org.dita.dost.util.Constants.*;
 import static org.dita.dost.util.URLUtils.*;
 import static org.dita.dost.reader.ChunkMapReader.*;
+import static org.dita.dost.util.XMLUtils.nonDitaContext;
 
 import java.net.URI;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.Map;
 import java.util.Set;
 import java.util.Stack;
@@ -73,8 +76,8 @@ public final class GenListModuleReader extends AbstractXMLFilter {
     private final Map<URI, URI> copytoMap = new HashMap<>(16);
     /** Flag for conrefpush */
     private boolean hasconaction = false;
-    /** foreign/unknown nesting level */
-    private int foreignLevel = 0;
+    /** DITA class values for open elements **/
+    private final Deque<DitaClass> classes = new LinkedList<>();
     /** Flag used to mark if current file is still valid after filtering */
     private boolean isValidInput = false;
     /** Set of outer dita files */
@@ -343,7 +346,7 @@ public final class GenListModuleReader extends AbstractXMLFilter {
         hasHref = false;
         hasCodeRef = false;
         currentDir = null;
-        foreignLevel = 0;
+        classes.clear();
         isValidInput = false;
         hasconaction = false;
         coderefTargetSet.clear();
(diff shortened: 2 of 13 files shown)


robander:
One question already: in `GenListModuleReader.java`, within `endElement`, the function returns early if inside the `foreign` context. Attributes are not available in this context to test `@class` and continue. This means it won't run `getContentHandler().endElement(uri, localName, qName);` -- I don't know if that's an issue or not, given that `startElement` now goes on to do all normal DITA processing?

robander:
I see there are failing tests as well - I had those fail locally but thought they were the "always seem to fail on Windows" tests. I'll update those but would appreciate feedback on the code / previous question first, as any changes will probably hit more tests.

robander:
Latest commit refactors original code (and my original request) to use a stack for tracking `<foreign>` and `<unknown>` elements, allowing them to nest without messing up the tracking. I haven't yet tried to change the failing test cases (or add new ones for this refactor), hoping to get feedback on the approach first.

There are a couple of side effects of this change I'm not sure about:

1. The latest commit makes it impossible to get message `DOTJ030I`, which previously appeared for elements that did not set `@class`. Those elements are now assumed to be (and are processed as) non-DITA content inside of the `foreign`/`unknown` context. This message has been useful on a couple rare occasions; should we iterate over the list of open elements, and continue to generate that when the closest DITA ancestor is not `foreign`/`unknown`?
1. Original commit added the method `isValidDitaClass`, which still feels like it should be useful, but is no longer used. I'll remove it and the pattern matcher unless somebody sees a use for them:
```
    private static final Pattern VALID_DITA_CLASS = Pattern.compile("(\\+|-)\\s+(topic|map)/\\S+\\s+" +
                                                         "([\\S[^/]]+/\\S+\\s+)*");
```

drmacro:
If an element's nearest DITA parent is not foreign/unknown I would expect to get a missing-class message just as I would in normal DITA processing contexts.

So:

```
<topic>
 <body>
 <foreign>
   <garble>
     <p class="- topic/p ">
       <oops>..</oops>
    </p>
 </garble>
</body>
</topic>
```

Should result in a message that "oops" is missing a @class attribute.

robander:
> If an element's nearest DITA parent
...yeah, that's what I was thinking, to match your example. A valid DITA grammar might only allow them as child of `foreign`/`unknown` (or as child of other non-DITA), but only there. So if you go up and the first DITA class you hit isn't one of those, something is wrong.

It looks like the previous message actually only came out if `@class` was missing, which wouldn't report the case where somebody edits their `<p>` class to something `class="paragraph"` ... and thus lost all processing for that element. I've run into that a few times over the years.

robander:
I have the message working in `GenListModuleReader.java` but need to clean it up a bit, and update the two other locations that previously worked with `DOTJ030I`. Was hoping to get to those today but kids aren't cooperating - should have something to review tomorrow.

robander:
After restoring the `DOTJ030I` message to the three locations where it previously appeared, I'm getting multiple copies of the message for a single bad element. I'll see if this happened previously (wouldn't surprise me), it feels like we should only need one instance of that message. In other words, not quite ready for review.

robander:
Question for @jelovirt if you have time, I'm not sure of the right course of action here.

Today, we have code to log `DOTJ030I` in three spots. That message appears when an element outside of `foreign`/`unknown` doesn't have `@class` specified:
* https://github.com/dita-ot/dita-ot/blob/d69bd806704340e6d08ed325a43aaf9126ab98e9/src/main/java/org/dita/dost/writer/DitaWriterFilter.java#L130
* https://github.com/dita-ot/dita-ot/blob/d69bd806704340e6d08ed325a43aaf9126ab98e9/src/main/java/org/dita/dost/writer/ProfilingFilter.java#L76
* https://github.com/dita-ot/dita-ot/blob/d69bd806704340e6d08ed325a43aaf9126ab98e9/src/main/java/org/dita/dost/reader/GenListModuleReader.java#L405

When I have one instance of an element without `@class`, the message appears four times. One comes from `GenListModuleReader`, one from `DitaWriterFilter`, and two from `ProfilingFilter`. To test - I just added a `<b>` element to my topic, and then deleted that element's class attribute from `highlightDomain.mod`, which is the condition needed for `DOTJ030I`. With that change, the one element results in 4 messages:
```
gen-list:
 [gen-list] Using Xerces grammar pool for DTD and schema caching.
 [gen-list] Processing file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/math/batmath.dita
 [gen-list] [DOTJ030I][INFO] No 'class' attribute for was found for element '<b>'. This generally indicates that your DTD or Schema was not developed properly according to the DITA specification.
 [gen-list] file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/math/batmath.dita:19:7: [DOTJ030I][INFO] No 'class' attribute for was found for element '<b>'. This generally indicates that your DTD or Schema was not developed properly according to the DITA specification.
 [gen-list] Serializing job specification

debug-filter:
   [filter] Using Xerces grammar pool for DTD and schema caching.
   [filter] Processing file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/math/batmath.dita to file:/C:/Users/IBM_AD~1/AppData/Local/Temp/temp20170303101911432/batmath.dita
   [filter] [DOTJ030I][INFO] No 'class' attribute for was found for element '<b>'. This generally indicates that your DTD or Schema was not developed properly according to the DITA specification.
   [filter] [DOTJ030I][INFO] No 'class' attribute for was found for element '<b>'. This generally indicates that your DTD or Schema was not developed properly according to the DITA specification.
```

With my refactoring -
* If I only keep a call to `DOTJ030I` in `GenListModuleReader` - I only get one message, in `gen-list`
* If I only keep it in `DitaWriterFilter` - I only get one message, in `debug-filter`
* The copy in `ProfilingFilter` gives me two messages for every instance of the element (one in `gen-list`, another in `debug-filter`

My preference now is to just call the message once, probably from `GenListModuleReader`, but wonder if there is a reason to prefer or keep the others.

jelovirt:
We can take the approach that we log everything and it's the log consumer who can remove the duplicates. If we added duplicate removal, we'd need to retain the state in the logger.

IIRC we already have the same warning message twice to warn that validation is disabled. Also, when map-first processing lands, we will not get separate messages from gen-list and debug-filter because those will be removed.

My preference here would be least code, even if we get duplicate warnings.

robander:
> My preference here would be least code, even if we get duplicate warnings.

I think I agree here ... the issue today is that we have the same code (well, variants of the same code) in three places, resulting in four messages. I was hoping to reduce that so we only test the condition once. Depending on which of the three is retained, that would mean either one message or two messages.

I know there are conflicts for the merge now, so I'll clean that up (and squash the commits) while cutting this down to one instance of the test, and see what we end up with.

robander:
Taking another run at this, and looking at the error messages again:
> My preference here would be least code, even if we get duplicate warnings.

We've currently got code in 3 files that will test for and issue error condition `DOTJ030I`. Those 3 test result in 4 messages for every missing class attribute. So to be clear - my primary goal isn't really to clean up the log - it's to clean up the code (no reason to test this 4 times), with a side effect of cleaning up the log.

Right now, the test for `DOTJ030I` in `ProfilingFilter.java` results in 2 messages, so it seems like we should remove that one and one other. Not sure if there is a reason to prefer having the test in `GenListModuleReader.java` or in `DitaWriterFilter.java`, or if upcoming changes to merge steps means we should actually keep the one in `ProfilingFilter` - @jelovirt do you have a preference?

robander:
Not quite ready for review, still working on test for `XMLUtils.nonDitaContext` and need to tidy up the commits.

jelovirt:
Rebase on `develop`, the integration tests there have been fixed.

src/main/resources/messages_template.xml
```diff
@@ -333,6 +333,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ070I" type="INFO">
+    <reason>Invalid 'class' attribute '%1' for was found for element '&lt;%2&gt;'.</reason>
+    <response>This generally indicates that your DTD or Schema was not developed properly according to the DITA specification.</response>
```
robander: Agreed. FWIW I made this by copying `DOTJ030I` which is a similar condition and has same response. Should probably fix both since this PR already works with both.

src/main/java/org/dita/dost/reader/CopyToReader.java
```diff
@@ -86,12 +87,9 @@ public void startElement(final String uri, final String localName, final String
...
 
         final String classValue = atts.getValue(ATTRIBUTE_NAME_CLASS);
-
-        if (foreignLevel > 0) {
-            foreignLevel++;
-        } else if (TOPIC_FOREIGN.matches(classValue) || TOPIC_UNKNOWN.matches(classValue)) {
-            foreignLevel++;
-        }
+        
+        final DitaClass cls = atts.getValue(ATTRIBUTE_NAME_CLASS) != null ? new DitaClass(atts.getValue(ATTRIBUTE_NAME_CLASS)) : null;
```
robander: Want to make sure I understand it correctly ... current code is 
```
 final DitaClass cls = atts.getValue(ATTRIBUTE_NAME_CLASS) != null ? new DitaClass(atts.getValue(ATTRIBUTE_NAME_CLASS)) : null;
 classes.addFirst(cls);
```
So -- if the attribute `@class` exists, create a new `DitaClass` and add to the stack (allowing tests for validity as needed), otherwise add `null` ... is the idea that if the class isn't DITA in the first place, we should add `null`? That sounds logical, just wanted to double check

src/main/java/org/dita/dost/writer/ProfilingFilter.java
```diff
@@ -68,31 +72,30 @@ public boolean hasElementOutput() {
...
+				props = StringUtils.getExtProps(domains);
 			}
 		}
 
 		if (exclude) {
 			// If it is the start of a child of an excluded tag, level increase
 			level++;
 		} else { // exclude shows whether it's excluded by filtering
-			if (foreignLevel <= 1 && filterUtils.needExclude(atts, props)) {
+			if (cls!=null && filterUtils.needExclude(atts, props)) {
```
robander: In this case -- changing from `cls!=null` to `cls.isValidDitaClass()` is what we need to prevent filtering non-DITA content inside the foreign / unknown context.

src/test/resources/ProfilingFilterTest/exp/topic1.dita
```diff
@@ -11,7 +11,8 @@
     <p class="- topic/p " platform="osx">OS X platform</p>
     <p class="- topic/p " platform="windows osx">Windows and OS X platform</p>
     <foreign class="- topic/foreign ">
-      <p platform="windows" class="- topic/p ">Foreign Windows</p>
+      <para>non-DITA para</para>
```
robander: Updated, adding one without class, and one with non-DITA class, both with `platform="windows"`

src/main/java/org/dita/dost/reader/GenListModuleReader.java
```diff
@@ -396,7 +396,9 @@ public void startElement(final String uri, final String localName, final String
 
         if (foreignLevel > 0) {
             foreignLevel++;
-            return;
+            if (!DitaClass.isValidDitaClass(classValue)) {
```
robander: Missed this (comment was made on an old commit but only showed up today) -- renaming now.

src/main/java/org/dita/dost/reader/GenListModuleReader.java
```diff
@@ -513,10 +520,10 @@ private void handleRootElement(final Attributes atts) {
...
         // @processing-role
         processRoleStack.pop();
-
-        if (foreignLevel > 0) {
-            foreignLevel--;
-            assert foreignLevel >= 0;
+        
+        final DitaClass cls = classes.pop();
+        if ((cls == null || !cls.isValid()) && !ELEMENT_NAME_DITA.equals(localName)) {
+            return;
```
jelovirt: This is a filter. If we return before passing the end element event to the next content handler, we'll end up with unbalanced tree. Is this really correct?

src/main/java/org/dita/dost/util/DitaClass.java
```diff
@@ -54,6 +58,7 @@ public DitaClass(final String cls) {
             sb.append(s).append(' ');
         }
         stringValue = sb.toString();
+        validDitaClass = (cls == null ? false : VALID_DITA_CLASS.matcher(stringValue).matches());
```
jelovirt: No need to check for null, `cls` cannot be null and will throw a `NullPointerException` already on row 52 if it is.

src/main/java/org/dita/dost/reader/GenListModuleReader.java
robander: Don't think it's correct - will fix.

src/main/java/org/dita/dost/util/DitaClass.java
robander: Makes sense - will update

src/main/java/org/dita/dost/reader/CopyToReader.java
```diff
@@ -86,11 +87,12 @@ public void startElement(final String uri, final String localName, final String
...
         final String classValue = atts.getValue(ATTRIBUTE_NAME_CLASS);
-
-        if (foreignLevel > 0) {
-            foreignLevel++;
-        } else if (TOPIC_FOREIGN.matches(classValue) || TOPIC_UNKNOWN.matches(classValue)) {
-            foreignLevel++;
+        
+        final DitaClass cls = atts.getValue(ATTRIBUTE_NAME_CLASS) != null ? new DitaClass(atts.getValue(ATTRIBUTE_NAME_CLASS)) : new DitaClass("");
+        if (cls.isValid()) {
+        	classes.addFirst(cls);
```
jelovirt: `addFirst` adds to the head of the Deque, whereas the  `add` in the else branch adds to the tails of the Deque and is equivalent to `addLast`. These should both be `addFirst`

robander: Fixed same issue in both classes.

jelovirt:
There were tabs in the Java code and the coding conventions for the project say we use four spaces. I'll fix those.

robander:
> There were tabs in the Java code and the coding conventions for the project say we use four spaces. I'll fix those.

[sigh] -- not intentional. I set up a new version of Eclipse and did the work there, must be the default. Thanks for noting it, I'll watch out for that.

Issue #2577, opened 2017-01-19
empty nav toc in topic with @chunk="to-content" 


lionelmoi:
When @chunk="to-content" is specified on the topicref in and nav-toc set to partial the `<nav/>` element in the html5 output is empty.

tomorales:
Ran into something similar to this. The problem seems to be this code in nav.xsl:

  <xsl:variable name="current-file" select="translate(if ($FILEDIR = '.') then $FILENAME else concat($FILEDIR, '/', $FILENAME), '\', '/')"/>
  <xsl:variable name="current-topicrefs" select="$input.map//*[contains(@class, ' map/topicref ')][dita-ot:get-path($PATH2PROJ, .) = $current-file]"/>

The current-topicrefs variable is supposed to be set to any topicrefs in the map whose @href attribute matches the value of the current-file variable. For example, if you're processing foo.dita, then <topicref href="foo.dita"> would match.

However, if the topicref to foo.dita has @chunk="to-content", then the DITA-OT rewrites the topicref so that it has the form <topicref chunk="to-content" href="foo.dita#topicid">, preventing the @href attribute from matching the current-file variable.

A solution to this problem is to first concatenate the first topic @id with the current-file variable, determine if there are any matching topicrefs, and (if not) then to fall back and try to find any topicrefs that match the current-file variable. I've verified that the following additions to nav.xsl fix this problem:

  <xsl:variable name="current-id" select="(//*[contains(@class, ' topic/topic ')])[1]/@id"/>
  <xsl:variable name="current-topicrefs" select="if ($current-id and $input.map//*[contains(@class, ' map/topicref ')][dita-ot:get-path($PATH2PROJ, .) = concat($current-file, '#', $current-id)]) then $input.map//*[contains(@class, ' map/topicref ')][dita-ot:get-path($PATH2PROJ, .) = concat($current-file, '#', $current-id)] else $input.map//*[contains(@class, ' map/topicref ')][dita-ot:get-path($PATH2PROJ, .) = $current-file]"/>

lionelmoi:
I made a pull request with my fix : https://github.com/dita-ot/dita-ot/pull/2578

Maybe you could comment there?

PR #2578, opened 2017-01-19
Fix empty nav ToC in topic with to-content chunk #2577

lionelmoi:
When a topic is chunked to-content a #id is added at the end of the file name in the @href.

When creating the `current-topicref variable` :

```
<xsl:variable name="current-topicrefs" input.map//*[contains(@class, ' map/topicref ')][dita-ot:get-path($PATH2PROJ, .) = $current-file]"/> 
```
The result of dita-ot:get-path contains the #id therefore the condition is never met.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/nav.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/nav.xsl
@@ -238,6 +238,9 @@ See the accompanying LICENSE file for applicable license.
                         (not(@format) or @format = 'dita' or @format = 'ditamap') ">
           <xsl:value-of select="@copy-to"/>
         </xsl:when>
+        <xsl:when test="contains(@chunk, 'to-content')">
+          <xsl:value-of select="substring-before(@href,'#')"/>
+        </xsl:when>
         <xsl:otherwise>
           <xsl:value-of select="@href"/>
         </xsl:otherwise>


src/main/plugins/org.dita.html5/xsl/nav.xsl
```diff
@@ -238,6 +238,9 @@ See the accompanying LICENSE file for applicable license.
                         (not(@format) or @format = 'dita' or @format = 'ditamap') ">
           <xsl:value-of select="@copy-to"/>
         </xsl:when>
+        <xsl:when test="@chunk = 'to-content'">
```
eerohele: This should probably be `contains(@chunk, 'to-content')` instead?

src/main/plugins/org.dita.html5/xsl/nav.xsl
```diff
@@ -238,6 +238,9 @@ See the accompanying LICENSE file for applicable license.
...
+                    <xsl:value-of select="substring-before(@href,'#')"/>
```
eerohele: I think this should use the [`dita-ot:resolve-href-path`](https://github.com/dita-ot/dita-ot/blob/325337c9a087da363e881366e5ef58e134ba4c1a/src/main/xsl/common/functions.xsl#L19) function.

eerohele: Actually, now that I reread it, probably not without refactoring other parts of the code.

src/main/plugins/org.dita.html5/xsl/nav.xsl
lionelmoi: Agreed, changed in commit f50ac8303d67d6f5748e28df7b417150f086c2d6

lionelmoi:
Could we mark this for the next maintenance release? 

It's a real problem that we cannot use the nav toc along with chunking.

robander:
Hi @lionelmoi -- are you able to squash these into a single commit? Our general policy is to consolidate them, particularly for small changes like this one. With that I think we can merge it into 2.4.4.

lionelmoi:
@robander I just need the hotfix/2.4.4 to be created so I can create the pull request against it.

infotexture:
@lionelmoi Done: https://github.com/dita-ot/dita-ot/tree/hotfix/2.4.4.

infotexture:
@lionelmoi You should now be able to [change the base branch](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) of this pull request.

infotexture:
Commit message amended.

infotexture:
@lionelmoi You need to [sign your work](http://www.dita-ot.org/DCO) to give us permission to use  your contribution.

1. Since the remote branch on your fork has diverged from the copy in your local repository, you'll need to delete the local copy of the `patch-2` branch with:

        git branch -D patch-2

2. Then get the latest version of that branch from your fork. 

    Assuming the remote entry for your fork is called `origin` in your local repository, use:

        git checkout -t origin/patch-2

3. To sign off your last commit without changing the commit message or associated files, use:

        git commit -s --amend --no-edit

4. Then force-push to update the branch on your fork with:

        git push --force

jelovirt:
Preferably

```
git push --force-with-lease
```

In this case it's the same really, but generally it's safer to use the `--force-with-lease` option

lionelmoi:
It should be okay now.

Issue #2584, opened 2017-02-01
Index term is rendered twice in the PDF output when it is declared in the DITA map


radu-pisoi:
If an index term is declared in the DITA map, then it will be displayed twice in the Index section of the PDF output. 

You can see a preview in the next screenshot:
![image](https://cloud.githubusercontent.com/assets/6837947/22509312/3f8a6e38-e896-11e6-88ee-3af5a1913a2f.png)

I have attached a sample DITA map to reproduce the problem [index-term-issue.zip](https://github.com/dita-ot/dita-ot/files/744962/index-term-issue.zip).

robander:
I can reproduce this with the attached sample using 2.4.3 -- the index term shows up with page numbers of ", 3"

robander:
I expanded the samples a bit and other terms added in the map also result in the empty entry - wondering if this has always been the case, for all terms in the map. Trying to find a fix now.

robander:
Found a fix for this, not sure if it's the best. 

First I should note - I believe this is only an issue with FOP. It works properly with Antenna House, and the markup in XEP output leads me to believe it should work there.

In `index_fop.xsl` I've added the final condition to the end of this test 
https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2.fop/xsl/fo/index_fop.xsl#L65
```
            <xsl:apply-templates mode="make-index-links"
              select="key('refid-by-value', @value)
                      [empty(ancestor-or-self::opentopic-index:index.entry[@end-range])]
                      [empty(ancestor::opentopic-index:index.groups)]
                      [empty(ancestor::*[@no-page eq 'true'])]
                      [ancestor::*[contains(@class,' topic/topic ')]]"/>
```

With this, FOP indexing only links to terms that appear inside topics. Basically it assumes that preprocess has pushed index terms from the map into topics, so that the terms still in maps can be ignored. 

There's still a known issue here -- index terms that are part of a `<topichead>` in the map do not generate links. The fix above is still an improvement though -- `topichead` did not previously generate working index links, and _still_ does not do so in AH / XEP, so that's really a separate issue.

robander:
Fixed for 2.4.4 with commit e0b8806

PR #2588, opened 2017-02-03
Fix empty link citation in FOP index #2584

robander:
This drops FOP index links to map entries, which currently do not generate anchors in the content anyway. This results in two problems:
* If anchors existed, we'd get duplicate page citations (one for the entry in the map, another for the entry after it is pushed into the topic)
* Because the anchor does not exist, we get an empty citation followed by a comma


robander:
Sigh. Didn't intend to create this as a branch in dita-ot, it was supposed to be in `robander/dita-ot`.

robander:
Deleting this pull request and the `dita-ot` branch - should be coming from my fork.

raducoravu:
👍 Agree with the fix, just to double check will try to get the entire set of user DITA sources and try the fix on it.

PR #2602, opened 2017-02-15
Feature/remove keep with next from list label

drmacro:
The attribute keep-with-next.* is not allowed on list-item-label. 

https://www.w3.org/TR/xsl/#fo_list-item-label


jelovirt:
@drmacro sign-off is needed. Amend the commit and force push to your branch.

PR #2604, opened 2017-02-17
Add simple API to invoke DITA-OT from Java

jelovirt:
A Java API can be use invoke DITA-OT from Java code without needing to have knowledge about the internals of DITA-OT, like Ant.

# Usage

```java
final ProcessorFactory pf = ProcessorFactory.newInstance(ditaDir);
pf.setTempDir(absoluteBaseTempDir);
final Processor p = pf.newProcessor("html5");
try {
  p.setInput(absoluteInputmapFile)
   .setOutputDir(absoluteOutputDir)
   .cleanOnFailure(false)
   .setMode(Mode.LAX)
   .setProperty("brand", "acme")
   .setLogger(logger)
   .run();
} catch (DITAOTException e) {
  logger.error("Failed to process: " + e.getMessage());
}
```

The `Processor` object is reusable but not thread-safe.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,12 +32,11 @@ dependencies {
     compile group: 'org.apache.ant', name: 'ant-launcher', version:'1.9.7'
     compile group: 'com.google.guava', name: 'guava', version: '19.0'
     compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.23'
-    runtime group: 'org.apache.ant', name: 'ant-apache-resolver', version:'1.9.7'
     compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.1'
+    runtime group: 'org.apache.ant', name: 'ant-apache-resolver', version:'1.9.7'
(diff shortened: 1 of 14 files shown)


jelovirt:
Should we provide setters for the same properties as we have options for in the `dita` command? That is, should we add remaining setters for

* `setFilter(File)`
* `setTempDir(File)` for the exact temporary directory, not the base directory for timestampped directories

jelovirt:
Should we have a toggle to _also_ write log file into the temporary directory, with e.g. name `log.txt`. We still want to keep the normal logger setter, but this would help in debugging. The log file in temporary directory would be written with verbose logging level. This could be combined with e.g. `setCleanOnFailure(boolean)` method to retain the temporary directory on failure.

robander:
Looks good to me.

For the questions above - I think it would be good to have setters for the same options available to `dita`. Those seem to be the most common, particularly the filter parameter that's really a critical set of info about the content. I don't think they're required, but good idea.

Overall -- looks very useful and I'm wondering if some of my team will want to make use of this as well.

PR #2605, opened 2017-02-18
Fix dita --install syntax help

infotexture:

See dita-ot/docs#129.


infotexture:
@jelovirt Can you verify whether these are the only long-form options that take a value but do **not** use the equals-sign syntax?

Issue #2559, opened 2017-01-11
Error thrown when using args.filter for HTML5 output


shaneataylor:
When generating HTML5 output, if you specify the ditaval file on the dita command line or via the  args.filter property, an error is thrown (but the output appears to be correct).

Applies to 2.4.2+, tested only for HTML5.

```
     [xslt] Recoverable error on line 114 of topic.xsl:
     [xslt]   FODC0002: java.io.FileNotFoundException: /Users/shtaylor/Dev/samples/${dita.input.valfile}
     [xslt]   (No such file or directory)
```

I have started work on a fix in my fork, setting dita.input.valfile (which is still used by other steps) based on args.filter.

tomorales:
The build-init task sets up the dita.input.valfile property from the filter file specified.

In DITA-OT 2.3.3 the html5 and xhtml plugins do the following:

    <makeurl file="${dita.input.valfile}" property="dita.input.valfile.url" validate="no"/>

This happens in the xhtml.init task which is executed during topic transformation which occurs after build-init and preprocess. The execution order for xhtml is something like dita2xhtml.init, build-init, preprocess, xhtml.topics, xhtml.init, ...

The execution order for html5 is something like html5.init, build-init, preprocess, html5.topic, xhtml.init, ...

In both cases, the dita.input.valfile.url property is initialized correctly.

In DITA-OT 2.4.2, the xhtml plugin maintains the task execution order present in DITA-OT 2.3.3 so dita.input.valfile is initialized correctly at the time dita.input.valfile.url is initialized.

However the html5 plugin attempts to initialize dita.input.valfile.url in the html5.init task which is executed before build-init and before the dita.input.valfile property is initialized.

A suitable fix would be to have a dita2html5.init task which is executed before build-init for all the properties that don't depend on build-init or preprocess, followed by an html5.init task to be executed as a dependency of html5.topic to properly initialize dita.input.valfile.url. Basically copy the execution order of the xhtml plugin and initialize the equivalent properties before and after build-init/preprocess.

infotexture:
@shaneataylor Is your work-in-progress on this ready for submission for 2.4.4?

jelovirt:
Fixed in 27a59d649487d95c3952a49b2a78e662e9f5f20f

infotexture:
For related discussion, see #2598 & #2627.

PR #2611, opened 2017-02-21
Fix valfile not found #2559

shaneataylor:
Signed-off-by: Shane Taylor <shane@taylortext.com>


jelovirt:
`DOTA012W` is explicitly intended to warn about deprecated properties. By ignoring the warning when `args.filter` is set and defaulting `dita.input.valfile` to that value effectively removes the warning.

Issue #2598, opened 2017-02-13
DITAVAL passthrough does not preserve attributes in HTML5 TOC


raducoravu:
In my DITA Map I have a topicref like this:

           <topicref href="topics/introduction.dita" product="product1">

and my DITAVAL looks like this:

            <val>
                <prop action="passthrough" att="product" val="product1"/>
            </val>

but the index.html generated by applying the HTML5 output looks like this:

     <li class="topicref"><a href="topics/introduction.html">Introduction</a></li>

so the "product1" value is not propagated in the TOC.

raducoravu:
The original report is here:
https://www.oxygenxml.com/forum/viewtopic.php?f=20&t=13954&p=41508#p41508

PR #2627, opened 2017-03-15
DITAVAL passthrough does not preserve attributes in HTML5 TOC #2598

vdanylyuk:
Moved creating of property named "dita.input.valfile.url" from target "html5.init" to macrodef "html5.topics".

Signed-off-by: Vitaliy Danylyuk <vdanylyuk@intelliarts.com>


robander:
This looks right to me - just made the change in `develop` and it fixed (currently) broken processing of filter conditions for HTML5.

jelovirt:
This doesn't fix passthrough in map, only in topics. Fixed in 27a59d649487d95c3952a49b2a78e662e9f5f20f for 2.4.5 hotfix.

PR #2628, opened 2017-03-15
Fix: args.outext not working for HTML5

shaneataylor:
Signed-off-by: Shane Taylor <shane@taylortext.com>

### DIFF
--- a/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
+++ b/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
@@ -102,7 +102,7 @@ See the accompanying LICENSE file for applicable license.
     <!-- end to check and init css relevant parameters -->
     <condition property="out.ext" value=".html">
       <not>
-        <isset property="out.ext"/>
+        <isset property="args.outext"/>
       </not>
     </condition>
     <makeurl file="${dita.input.valfile}" property="dita.input.valfile.url" validate="no"/>


shaneataylor:
@jarno, I'm pretty sure these lines in org.dita.base/build_init.xml already handle your request. My fix just addressed the problem that args.outext was always set to .html.

```
    <!-- append a '.' to out.ext if args.outext not contains '.' -->
    <condition property="out.ext" value=".${args.outext}">
      <and>
        <isset property="args.outext" />
        <not>
          <contains string="${args.outext}" substring="." />
        </not>
      </and>
    </condition>
    <condition property="out.ext" value="${args.outext}">
      <and>
        <isset property="args.outext" />
        <contains string="${args.outext}" substring="." />
      </and>
    </condition>
```

robander:
I'm not sure why I had to look at this so long, but it seems the current logic is:
1. `html5.init`: if `out.ext` isn't set, the code sets `out.ext` to `.html`  ... generally `out.ext`, meaning we never get another value
2. `build.init`: if `args.outext` _is_ set, use that value ... but we don't get here because `out.ext` is already set

The new logic is better:
1. `html5.init`: if `args.outext` isn't set, then `out.ext` to `.html`
2. `build.init`: if `args.outext` _is_ set, use that value

So - I think the logic is right. But it's easy to miss what's happening because they're in different targets in different files, and not everybody knows the order in which they are called. Would be nice to have this all set in `html5.init`...

jelovirt:
I misunderstood the case. The correct way to handle this is to modify the code to

```xml
<condition property="out.ext" value="${args.outext}" else=".html">
  <isset property="args.outext"/>
</condition>
```
So that `html5.init` will take care of `out.ext` initialization completely, not do part of the initialization in HTML5 plugin and part in base plugin.

Issue #2622, opened 2017-03-07
The 'ol/@class' attribute is missing in the generated HTML when 'steps' contains a 'stepsection' element 


radu-pisoi:
One of our users is complaining that the `ol/@class` attribute is missing when `steps` element has a `stepsection` child element.

For the case when the `steps/stepsection` element is missing, the generated HTML looks like:
```
<ol class="ol steps">
```
while for the case when it is present, the `class` attribute is omitted.

You can simply reproduce this problem with the following DITA snippet:
```
<steps>
    <step>
        <cmd>Step1</cmd>
    </step>
    <stepsection>The next two steps are very important!</stepsection>
    <step>
        <cmd>Step2</cmd>
    </step>
</steps>
```
I've tested with **DITA-OT 2.4.4**, transtype **html5**.

infotexture:
Looks like this has been resolved by the changes in 306a3f4 and amendments in c45268c.

PR #2631, opened 2017-03-20
The 'ol/@class' attribute is missing in the generated HTML when 'steps' contains a 'stepsection' element #2622

vdanylyuk:
Add commonattributes for generated lists from step elements with step-section.
 
Signed-off-by: Vitaliy Danylyuk <vdanylyuk@intelliarts.com>

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/task.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/task.xsl
@@ -184,6 +184,7 @@ See the accompanying LICENSE file for applicable license.
         <xsl:otherwise>
           <!-- First step in a series of steps -->
           <xsl:element name="{$list-type}">
+            <xsl:call-template name="commonattributes"/>
             <xsl:if test="$list-type='ol' and preceding-sibling::*[contains(@class,' task/step ')]">
               <!-- Restart numbering for ordered steps that were interrupted by stepsection.
                    The start attribute is valid in XHTML 1.0 Transitional, but not for XHTML 1.0 Strict.


robander:
Looks good to me but I think it should go into `hotfix/2.4.5` for the next patch, rather than into `develop`...

jelovirt:
Don't merge, only the single commit should be in the PR

jelovirt:
Cherry-picked to 2.4.5 as 306a3f41780bced66267d3b618b82d7dd3825501

robander:
Was suggesting to @jelovirt that we should merge this as I was in the process of testing, and now seeing an issue.

The context when calling this (to add `@xml:lang`, `@class`, etc to the `<ol>` or `<ul>`) is actually the step element rather than the list. So the commit here will result in `<ol class="li step">` for an ordered step, rather than adding something based on the steps element like `<ol class="ol steps">`. The `li step` class is repeated on the step itself already. (Probably much lower priority, but on the off chance that `@xml:lang` or `@dir` are set on the parent steps, those are lost today for the same reason that `@class` is lost.

robander:
When I modify the change to
```
<xsl:element name="{$list-type}">
  <xsl:for-each select=".."><xsl:call-template name="commonattributes"/></xsl:for-each>
```

I get the correct results (`xml:lang`, `class`, etc based on the parent steps, each time it restarts).

robander:
Updated with https://github.com/dita-ot/dita-ot/commit/c45268c4f126061b47036c852787b042ee246672 to set the parent context - think this is all the way done now.

PR #2637, opened 2017-03-23
Support list of DITAVAL filter files

robander:
Opening this early in order to get feedback on parameter name, design, etc.

The user goal / use case here is to allow a build to specify a list of DITAVAL files to be used instead of the existing `args.filter` parameter. Currently, the single parameter means that for every unique set of conditions, a separate build file must be used that also includes all common conditions. Supporting a list allows the maintenance of common conditions where they make sense, while storing unique conditions where they make sense. 

I've seen this most often with teams that want to store a large set of revision conditions separately from other conditions. As a real world example, a list of 4 files for a single PDF build might include:
- A set of audience conditions that applies to all products
- A set of revisions that applies to everything within a single product
- A set of filter conditions that applies to all printed material
- A final set of conditions used within this single document

# Design plans [now revised! see below]

Looking for review / input:

*   Parameter named `args.filter.list` specifies a text file, with one filter file listed per line. This parameter is ignored if `args.filter` is also specified.
*   Listed files are evaluated in order, and concatenated into a generated file within a single `<val>` root
    *   Conflicting conditions do not present a new issue here, so should not need to be addressed separately. That is - anything you can do with 4 or 10 files, you can do in a single filter file today. You can define a single file with 10 conditions for `otherprops="testy"`. Whether that is done in one file, or results from using 10 files, our processing should respond the same way.
*   The resulting single file is generated in the temp directory, and the existing `args.filter` property is set to that value; this allows all remaining processing to function as it does today without change

I generally don't like "text file with one file per line" but other options (an XML file, comma separated list, etc) seem worse.

Comments welcome, as I start playing with a larger implementation. Goal is to have this completed for 2.5.

# Updated design

- `args.filter` updated to handle a list of files separated by path separator `;` character
- For any file(s) specified with `args.filter`, a generated version is placed in the temp dir, named based on a configuration option (but maybe this should be hard coded?)
- For any image that uses a relative path, a `basedir` attribute is added with the name of the ditaval directory; `DitaValReader.java` is updated to use this when finding images for copying to [wherever]
- Because the image reference itself isn't changed, any steps that process a relative path entry continue to work as they did (whether this is appropriate or not)
- `gen-list`, `debug-filter`, and `flag-module` now take in the name of the generated file, rather than taking in `args.filter` or a URL version of `args.filter`; parameters to those steps do not need to change


jelovirt:
Since `args.filter` value is a system path, we can just add support for a list of paths, separated by the platform path separator character. That way we would not need to add a new property that works exactly like `args.filter` but takes a list/array as a value.

shaneataylor:
@robander This would be hugely useful. The method I have thought of in the past to address this problem is for ditaval files to support including other ditaval files, but that involves extending the spec. In practice, I've used XML entities and ditaval fragment files to implement this kind of inclusion, but the resulting ditaval file is kind of equivalent to your text file with one per line (except more complicated).

What about — instead of creating a list of ditaval files in a separate (text or XML) file, we simply support `args.filter._XX_` where _XX_ is a two-digit number specifying the sequence in which the specified ditaval file should be applied? 

My sense is that even 0-9 would cover 99% of use cases.

shaneataylor:
@jelovirt I like the simplicity of your suggestion, but for even a handful of chained ditaval files, the resulting property value could be quite long & harder for users to troubleshoot (I see the main issues being filters applied out of order or missing/unwanted filters). Including each as a separate property (especially in a .properties file) makes them easier to manage.

robander:
@jelovirt I haven't used parameters as a list that way ... feels a bit painful to author but does feel a bit cleaner than having a separate file to reference other files (I've had pain with that in the past). Although, as @shaneataylor points out, the ones we'd be using will have long paths from our CMS systems, so that one property would probably be a few hundred characters long. That's based on some lists we have today.

It does seem like doing it as a separate parameter would still be preferable, after which we set `args.filter` to the merged file. Otherwise, we (and everybody who has special processing for ditaval) will need to change scripts to use a new parameter. If starting from scratch, I could buy an argument that one parameter is sufficient. At this point, the changes it requires + plus the fact that it's a different type of value seem like enough to use a new parameter.

(FWIW we have an internal implementation for this, but it's part of a wrapper for DITA-OT, not integrated. It follows the rough logic above. Teams that use it often have several in the list, and each path is quite long. We did use a single parameter for both, which proved endlessly confusing, but partly that's because we had to read the file to figure out if it was a list or a DITAVAL.)

I ... really couldn't go for parameters named `something1`, `something2`, ... `somethingN`. It immediately feels like a hack with an artificial limit, that limit being "only as many as we decided you should need". I remember having the same reaction when somebody said that they already used `@otherprops` for a specific thing in DITA topics, so could OASIS please add `@otherprops2` and `@otherprops3`.

shaneataylor:
@robander My first impulse was actually to suggest _not_ limiting the parameter's numeric component, but to parse all `args.filter.X` parameters (where _X_ matches [0-9]+) in numerical order so users could chain an arbitrary number of ditaval files, but I figured that introduced complexity that just wasn't useful. 

Maybe the single `args.filter.list` parameter would become more usable if we added an optional `args.filter.basedir` against which relative paths to each listed `.ditaval` would be resolved? 

For example:
```
args.filter.basedir = /my/path/to/my/cms/and/repo/ditaval/dir
args.filter.list = first.ditaval second.ditaval third.ditaval special_subdir/fourth.ditaval
```

I think that would simplify _most_ use cases in that at least the base CMS path could be specified with `args.filter.basedir`.

If we did add `args.filter.basedir`, should `args.filter` also be resolved against that? It's a change, but only if you specify `args.filter.basedir`, and I think it would be expected behavior.

robander:
`*.basedir` generally make me nervous just because I've had so many problems over the years (DITA-OT and elsewhere) dealing with and explaining paths -- "OK this image will be located relative to DITAVAL, which you referenced with a relative parameter, which is evaluated against this other specified directory, and if `*.basedir` is relative, then all of this is relative to your `properties` file when calling the build."

That said ... I can understand the use here, and it would make passing in a long list of ditaval files easier. 
- Common case: they're all located either in the same directory, or very near each other, so base dir gets rid of most of the path
- Edge case: they're located all over the drive so basedir isn't helpful ... but in that case don't use it. And I think this is really an edge case, at least with my customers, the common case is to store them together.

robander:
RE: one parameter versus two (regardless of path length) ... we currently set up internal parameters `dita.input.valfile` and `dita.input.valfile.url` from the existing `args.filter` parameter. (Arguably, `dita.input.valfile.url` should have been based on the calculated internal `dita.input.valfile` rather than going back to the original.) Processing extensions from that point on probably _should_ be working with the validated internal property rather than the original.

So I've reconsidered, and I as long as we properly set up the internal property, extending `args.filter` to support a list shouldn't cause compatibility issues or any real changes to later steps. (This is true whether or not a `*.basedir` parameter is created.)

robander:
Discussed this just a bit at the contributor call today, and followed up with @jelovirt with some design questions - sharing a summary (and question) here now.

Just to see what was possible - I initially got this working by updating both `gen-list` and `debug-filter` so that they took in a list of files (from a parameter), processed them in sequence, and continued. This was mostly to play around because I haven't played much in that code.

As confirmed with Jarno - a better design is really to merge the ditaval files before `gen-list`. To keep it simple, this process should run whenever the parameter is used (whether just one file or multiple). The merged result will be placed in the temp directory, with some adjustment of any relative image paths.

Today -- the code takes in `args.filter`, validates it, and uses that to set `dita.input.valfile`, which is used for the remainder of preprocess. A warning is generated whenever `dita.input.valfile` is passed into the build ("This is deprecated, use `args.filter`), but it still works. As part of this pull request, passing in that parameter will become a fatal error. The new process will thus be able to reliably set `dita.input.valfile` to a generated file in the temp dir, meaning that all remaining steps (including any customizations outside of DITA-OT core) can continue to rely on `dita.input.valfile` to reflect all conditions.

To merge the files, I need to create a new Java module that runs before `gen-list`. As I haven't done that in a long time, I have a question (probably for @jelovirt). I started by tweaking a module that sounded similar (`TopicMergeModule`). It extends `AbstractPipelineModuleImpl`, while a lot of the newer code extends `SourceReaderModule`. Is there a reason to pick one over the other for merging DITAVAL files?

robander:
Update: still making progress on this and still planning to have it in 2.5.

Have not pushed code up to github yet because it's still about 30% code / 70% debug logging. I've now got a standalone preprocess step that reads in the DITAVAL files before `gen-list`, and sets up a generated file to contain the conditions from each file; next up is to write them back out, with any image paths adjusted as needed.

robander:
[Sigh]. I was using "default Java coding conventions" as laid out by Eclipse, modified to replace tabs with spaces. I did not expect it to reflow comments and longer lines.

robander:
Curious ... the test case `testsubjectschema_case` is failing in Travis but does not fail for me locally. When running on my system, the only failing tests are the same 14 Chunk test cases that always fail on Windows.

robander:
Working on the `subjectschema_case` test case, it looks like this is the only test case in our IntegrationTest which:
- has a DITAVAL
- has `<startflag>` within the ditaval (not just include/exclude)
- has an `imageref` attribute to specify a flag (other tests just use text for flagging)

The test works fine on Windows - filter conditions are properly evaluated, flags are evaluated, and the flag image is copied to the output directory. Sorry for all the noise with the various commits - easiest way to test if / why it fails in Travis is to add more debug info and push up another commit.

robander:
Found the issues and merged my 20 or 25 debug commits.

The merge processing adds `@basedir` attribute to any filter element that uses `imageref`. it's supposed to be the path to the DITAVAL.

On Windows, that value (retrieved running`toURI(currentDitaval.getParent())`) started with `file:/` and ended with a file separator. In the travis build, it did not start with `file:/` (causing the `isAbsolute()` test to fail later). Once that was fixed, the missing slash caused `basedir` + `imageref` to return an invalid path.

Latest commit fixes both of those, but it's ... not very pretty so another one is coming. (Not that the rest of the code is exactly pretty yet, but ... :roll_eyes:

src/main/java/org/dita/dost/util/Constants.java
```diff
@@ -941,7 +945,7 @@
     public static final String ATTRIBUTE_NAME_NAME = "name";
     /**type attribute value subjectScheme.*/
     public static final String ATTR_TYPE_VALUE_SUBJECT_SCHEME = "subjectScheme";
-    /**store how many scheme files a ditamap file used in form of {@code Map&lt;String, Set&lt;String>>}.*/
+    /**store how many scheme files a ditamap file used in form of Map<String, Set<String>>.*/
```
robander: Not sure what this was - thought I was *undoing* a change that had been introduced with the Eclipse auto-formatting of Java.

src/main/plugins/org.dita.base/plugin.xml
```diff
@@ -91,6 +91,7 @@ See the accompanying LICENSE file for applicable license.
       <val>NUMTITLE</val>
     </param>
     <param name="args.filter" desc="Specifies a filter file to be used to include, exclude, or flag content." type="file"/>
+    <param name="args.filter.list" desc="Specifies a list of filter files to be used to include, exclude, or flag content." type="file"/>
```
robander: Still need to remove this, left over from original commit that started the pull request.

src/main/java/org/dita/dost/module/MergeDitavalModule.java
```diff
@@ -0,0 +1,315 @@
+/*
+ * This file is part of the DITA Open Toolkit project.
+ *
+ * Copyright 2004, 2005 IBM Corporation
```
jelovirt: Copyright year should be 2017

src/main/java/org/dita/dost/module/MergeDitavalModule.java
```diff
@@ -0,0 +1,315 @@
...
+import org.dita.dost.pipeline.AbstractPipelineInput;
+import org.dita.dost.pipeline.AbstractPipelineOutput;
+import org.dita.dost.util.*;
+import org.xml.sax.*;
+
+/**
+ * This class reads a list of DITAVAL files, and merges 
+ * conditions into a single file
+ * 
+ * @version 1.0 20
```
jelovirt: We don't use `@version`. I've inconsistenly used `@since` where the version number is the DITA-OT version number when the class/method was added.

src/main/java/org/dita/dost/reader/DitaValReader.java
```diff
@@ -54,6 +54,8 @@
 
     private boolean setSystemid = true;
 
+    private static final String DITA_OT_NS = "http://dita-ot.sourceforge.net";
```
jelovirt: Instead of redeclaring this in multiple classes as a private field, it should be a public constant in `Constants` class and then imported with a static import.

src/main/java/org/dita/dost/reader/DitaValReader.java
```diff
@@ -147,6 +149,10 @@ public void startElement(final String uri, final String localName, final String
             if (f.isAbsolute()) {
                 imageList.add(f);
                 relFlagImageList.add(getRelativePath(ditaVal, f));
+            } else if (atts.getValue(DITA_OT_NS, ATTRIBUTE_NAME_IMAGEREF_URI) != null) {
+                final URI imageuri = toURI(atts.getValue(DITA_OT_NS, ATTRIBUTE_NAME_IMAGEREF_URI));
```
jelovirt: `toURI` is meant for cases where the input may be invalid URI or system path. In cases where you know the value is a valid URI, you should just use `URI.create(String)` method. If you know the value is a URI but not sure if it's valid, then use `new URI(String)` and catch exception.

src/main/lib/configuration.properties
```diff
@@ -9,6 +9,7 @@ otversion = 2.4.6
 default.cascade = merge
 temp-file-name-scheme = org.dita.dost.module.GenMapAndTopicListModule$DefaultTempFileScheme
 cli.color = true
+preprocessed.ditaval.name = ditaot.generated.ditaval
```
jelovirt: IMO this doesn't have to be configurable and the name can be just a constant in Java code.

src/main/plugins/org.dita.base/build_preprocess_template.xml
```diff
@@ -487,7 +506,7 @@ See the accompanying LICENSE file for applicable license.
           filedirparameter="FILEDIR">
         <ditaFileset format="dita" processingRole="normal"/>
         <param name="TRANSTYPE" expression="${transtype}"/>
-        <param name="FILTERFILEURL" expression="${dita.input.filterfile.url}"/>
+        <param name="FILTERFILEURL" expression="${mergedditaval.url}"/>
```
jelovirt: This is not backwards compatible in cases where users still assume `dita.input.filterfile.url`is available. Why not just keep on using the same old property, just point it to the merged filter file?

src/main/plugins/org.dita.base/build_preprocess_template.xml
```diff
@@ -20,6 +20,7 @@ See the accompanying LICENSE file for applicable license.
   <target name="preprocess"
     dita:depends="{depend.preprocess.pre},
                   preprocess.init,
+                  preprocess-ditavals,
```
jelovirt: IMO the `preprocess-` prefix is redundant and we don't use it in any other target name. (The init target doesn't count.)

src/main/java/org/dita/dost/module/MergeDitavalModule.java
```diff
@@ -0,0 +1,315 @@
...
+                }
+            }
+        }
+
+    }
+
+    private void writeConditions(XMLStreamWriter export, NodeList conditionElements, URI ditavalDirectory) {
+        try {
+            for (int i = 0; i < conditionElements.getLength(); i++) {
+                if (conditionElements.item(i).getNodeType() == Node.ELEMENT_NODE) {
```
jelovirt: Use switch-statement for node type tests, rest of the Java code does that too.

src/main/java/org/dita/dost/module/MergeDitavalModule.java
```diff
@@ -0,0 +1,315 @@
...
+
+    private void writeMergedDitaval() throws DITAOTException {
+        XMLStreamWriter export = null;
+        try (OutputStream exportStream = new FileOutputStream(new File(job.tempDir, mergedDitavalFilename))) {
+            export = XMLOutputFactory.newInstance().createXMLStreamWriter(exportStream, "UTF-8");
+            export.writeStartDocument();
+            export.writeStartElement("val");
+            export.writeNamespace(DITA_OT_NS_PREFIX, DITA_OT_NS);
+            if (!ditavalFiles.isEmpty()) {
+                final Iterator<File> dvi = ditavalFiles.iterator();
```
jelovirt: Use for-loop instead of while-loop with an iterator

src/main/plugins/org.dita.base/build_preprocess_template.xml
robander: Sure - that makes sense, and we can keep using the old one. The reason I changed it is that previously we didn't create a URL property for this until `flag-module`  -- now that I'm creating one earlier on, I reused the early version. I'm not sure whether it's best to create it early as `dita.input.filterfile.url` or to use two different properties for the two different cases.

src/main/java/org/dita/dost/module/MergeDitavalModule.java
```diff
@@ -211,55 +207,29 @@ private void updateBaseDirectory() {
...
-                while (dvi.hasNext()) {
-                    final File curDitaVal = dvi.next();
-                    final DocumentBuilder ditavalbuilder = XMLUtils.getDocumentBuilder();
-                    final Document doc = ditavalbuilder.parse(curDitaVal);
-                    final Element ditavalRoot = doc.getDocumentElement();
-                    final NodeList rootChildren = ditavalRoot.getChildNodes();
-                    logger.debug("Writing conditions from ditaval: " + curDitaVal);
-                    writeConditions(export, rootChildren, curDitaVal.toURI().resolve("."));
-                }
+            for (int i = 0; i < ditavalFiles.size(); i++) {
```
jelovirt: Since `ditavalFiles` is a `List`, this should not use a [for statement](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/for.html), but rather [for-each loop](https://docs.oracle.com/javase/1.5.0/docs/guide/language/foreach.html).

src/main/java/org/dita/dost/invoker/Main.java
```diff
@@ -135,6 +135,22 @@ String getValue(final String value) {
...
     }
 
+    private static class AbsoluteFileListArgument extends Argument {
+        AbsoluteFileListArgument(final String property) {
+            super(property);
+        }
+
+        @Override
+        String getValue(final String value) {
+            String allPaths = "";
```
jelovirt: This should really use a stream that's mapped and then collected with a joiner, but I'm ok with rewriting it later on.

src/main/plugins/org.dita.base/build_preprocess_template.xml
```diff
@@ -67,6 +68,26 @@ See the accompanying LICENSE file for applicable license.
...
+      <module class="org.dita.dost.module.MergeDitavalModule">
+        <param name="inputdir" location="${args.input.dir}" if="args.input.dir"/>
+        <param name="ditadir" location="${dita.dir}"/>
+        <param name="ditaval" location="${args.filter}" if="args.filter"/>
+        <param name="outputdir" location="${output.dir}"/>
+        <param name="generate-debug-attributes" value="${generate-debug-attributes}" if="generate-debug-attributes"/>
+        <param name="processing-mode" value="${processing-mode}" if="processing-mode"/>
+      </module>
+    </pipeline>
+    <property name="ditaotmerg" value="testingnngngngngng"/>
```
jelovirt: Remove test code

robander: [ouch]. Yeah.

src/main/java/org/dita/dost/module/MergeDitavalModule.java
```diff
@@ -0,0 +1,277 @@
...
+                    export.close();
+                } catch (final XMLStreamException e) {
+                    logger.error("Failed to close merged ditaval file: " + e.getMessage(), e);
+                }
+            }
+        }
+
+    }
+
+    private void writeConditions(XMLStreamWriter export, NodeList conditionElements, URI ditavalDirectory) {
```
jelovirt: TODO: This should be generalized to be a DOM to XML stream utility function and moved to XMLUtils.

src/main/java/org/dita/dost/invoker/Main.java
robander: Adding a TODO note so it's not forgotten.

SanFanDocu:
We are using DITAVAL modules that are transformed in a separate XSLT to txt files which are included in turn as entities in the final publishing DITAVALs.  
The benefit of this solution: The final publishing view for each set is available within the authoring environment (Oxygen) and authors are able to preview the final filtering situation.
If I understand your solution correctly the final DITAVAL sets are only temporarily available during publishing procedure, right? 
So there should be a transformation that publishes the final DITAVALS to a selected location for permanently storage from where authors can import them e.g. as profiling sets in Oxygen.

robander:
Hi @SanFanDocu - I want to make sure I understand what you're saying.

* Today - you've got multiple DITAVAL files.
* You use an XSLT process to create one "merged" file by importing each individual DITAVAL (done by importing them as entities)
* That merged file is what you use in Oxygen when editing, and what you pass into DITA-OT today for processing

Does that all sound right?

If so - I can see how that would be useful and why you'd like to get a copy from the temp directory. I don't really think that should be part of the core feature though. For the core feature - it's really just writing the file to temp as a convenience that works well with our current process that reads/writes a lot of files in temp. Ideally, it would only exist in memory and no file would be written out.

I'd be nervous to write this back to my own source directory, because I think it would get out of sync with the actual source DITAVAL files. Clearly this hasn't been an issue for you, though it might once you no longer _need_ to pass the merged version in as a DITA-OT parameter.

For what it's worth ... with the code as posted at this point, it would be trivial to add a post-process step in any transform type that automatically copied this file out:
```
<target name="return-merged-ditaval" unless="skip.ditaval.preprocess">
  <copy 
    file="${dita.input.valfile}"
    tofile="${output.dir}/ditaot.generated.ditaval"/>
</target>
```

robander:
Going back to a comment from @jelovirt well up the chain:
> Since `args.filter` value is a system path, we can just add support for a list of paths, separated by the platform path separator character. That way we would not need to add a new property that works exactly like `args.filter` but takes a list/array as a value.

This is how the latest commits work, and I'm getting nervous about it.

We've got sets of build properties that are transferred around between users and build systems - Windows, Mac, and Linux. The different systems all have different separator characters. Currently, the properties work fine whichever system pulls them down. I don't think that can work with the platform path separator - my properties file on Windows uses semi-colons, but would have to change to colons to run on a Mac ... at which point I couldn't use it anymore.

Are there common ways to resolve this, or are we stuck with one of either 1) property values that can't work across platforms or 2) defining our own separator?

mironovalexey:
@robander Robert, this is a very useful feature. Thank you. Do I understand correctly, now I can pass multiple filters in the following syntax (ant)?
```xml
<property name="args.filter" value="${first.dir}/first.ditaval;${second.dir}/second.ditaval;"/>
```

robander:
@mironovalexey yes, that is correct, though the path separator character is platform-dependent so the semicolon ';' will only work on Windows. For Mac/Linux, use a colon ':' instead.

raducoravu:
Ideally we should have the same path separator work, no matter for what platform the publishing is done. In this way the value of the args.filter would be portable no matter on what OS.

robander:
@raducoravu that would be my preference as well, I'm already thinking about how to get the same properties file to work that my customers create on Windows but then run on Linux or Mac (when transferred to another user or to a build system). If I understand correctly @jelovirt was pushing for the path character because it's More Proper...?

xephon2:
This is probably not the correct solution, but this works with pure Ant.

```xml
<project basedir="." default="fix-delimiters">
  <target name="fix-delimiters" depends="-check-delimiter, -fix-windows-delimiter, -fix-unix-delimiter">
    <echo>BEFORE: ${ditavals}</echo>
    <echo>AFTER:  ${ditavals.normalized}</echo>
  </target>
  <target name="-check-delimiter">
    <property name="ditavals" value="file:/Users/eike/DITA/test1.property,file:/Users/eike/DITA/test2.property"/>
    <condition property="commaDelimiter">
      <contains string="${ditavals}" substring=","/>
    </condition>
    <condition property="semicolonDelimiter">
      <contains string="${ditavals}" substring=";"/>
    </condition>
    <condition property="isWindows">
      <os family="windows"/>
    </condition>
    <condition property="isUnix">
      <os family="unix"/>
    </condition>
    <condition property="replaceCommaWithSemicolon">
      <not>
        <and>
          <istrue value="${commaDelimiter}"/>
          <istrue value="${isWindows}"/>
        </and>
      </not>
    </condition>
    <condition property="replaceSemicolonWithComma">
      <not>
        <and>
          <istrue value="${semicolonDelimiter}"/>
          <istrue value="${isUnix}"/>
        </and>
      </not>
    </condition>
    <echo>${commaDelimiter}</echo>
  </target>
  <target name="-fix-windows-delimiter" if="replaceCommaWithSemicolon">
    <echo>Is Unix OS but uses Windos delimiter: Delimiter should be replaced</echo>
    <loadresource property="ditavals.normalized">
      <propertyresource name="ditavals"/>
      <filterchain>
        <tokenfilter>
          <filetokenizer/>
          <replacestring from=";" to=","/>
        </tokenfilter>
      </filterchain>
    </loadresource>
  </target>
  <target name="-fix-unix-delimiter" if="replaceSemicolonWithComma">
    <echo>Is Windows OS but uses Unix delimiter: Delimiter should be replaced</echo>
    <loadresource property="ditavals.normalized">
      <propertyresource name="ditavals"/>
      <filterchain>
        <tokenfilter>
          <filetokenizer/>
          <replacestring from="," to=";"/>
        </tokenfilter>
      </filterchain>
    </loadresource>
  </target>
</project>
```

robander:
I think that's right, but we should be able to do something simpler, making use of `${path.separator}`. Because we don't really care what OS we're on or what OS the property was written for, we just care about converting *to* a value that will be recognized on the current OS. So regardless of input value, if we converted `;` and `:` to `${path.separator}`, we'd end up with a normalized value that could be passed into the merge module.

robander:
Ah, reminding myself -- if you call the build with the `dita` command, the `--filter` parameter parses any value expecting the platform separator, and then puts the value into `args.filter`. It's not possible to set `args.filter` directly using the `dita` command. So, with `dita`, the only way to fix this would be to update the Java code that parses the property; I'm guessing @jelovirt doesn't like that idea:
https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/invoker/Main.java#L145

I guess realistically that's lower priority -- if you're calling `dita` and specifying `--format`, you're doing it on a specific OS and can use the right value.

If you're specifying `args.filter` directly in a property file or by calling `ant`, it looks like this variant of @xephon2 's suggestion works (at least on Windows, it lets me specify either `;` or `:`):
```
<loadresource property="args.filter.normalized">
      <propertyresource name="args.filter"/>
      <filterchain>
        <tokenfilter>
          <replacestring from=";" to="${path.separator}"/>
          <replacestring from=":" to="${path.separator}"/>
        </tokenfilter>
      </filterchain>
    </loadresource>
    <pipeline message="Preprocess and merge ditavals" taskname="ditaval-merge">
      <module class="org.dita.dost.module.MergeDitavalModule">
        <param name="ditaval" value="${args.filter.normalized}" if="args.filter"/>
      </module>
    </pipeline>
```

xephon2:
@robander, this would also replace the colon in `file:` or `http:`. If fixed the code snippet.

This would work, but is still ugly. I'm curious about the final solution. 😄 

```xml
<tokenfilter>
  <replacestring from="property;" to="property${path.separator}"/>
  <replacestring from="property:" to="property${path.separator}"/>
</tokenfilter>
```

robander:
> this would also replace the colon in `file:` or `http:`.

Details, details. ;-) But yeah ... seems the general feeling is that we need a property that can work across platforms. I'm not sure of the ideal solution.

drmacro:
Surely a list of URIs is fundamentally different from an OS-specific path--the two are not directly interchangeable.

jelovirt:
The Ant Way of doing this is to use filesets to create a path and combining them with pathconvert into a single property.

I understand the wish for multi-OS solution, but if you have eg just a single filter path C:\foo\bar.ditaval, the path itself will never work on Linux. The problem is not the separator character.

robander:
RE: the Ant way of doing things ... I think that would only work (??) if we have a different property for each ditaval, and I don't want to do that.

I get that an absolute path is never going to work on different systems. That's the reason we never use absolute paths in our `args.filter` parameter today -- our DITA builds always have to be portable between machines (Win/AIX/MacOS), because any user on a team might be running the build on their local system (either Win or MacOS), and eventually they will be running on an AIX server. We will always be setting up our property files to use something like `args.filter=first.ditaval;second.ditaval`. In that case the separator character does end up causing problems, unless I'm missing something.

Thinking through simple workarounds in my environment, I might end up using my own property and tell people to always use `;` or `,`, and then convert to the system separator in my own init step, but I'd love to hear better suggestions.

jelovirt:
The root issue here is that some users want to use Java properties file to define a list of filter files. Since properties file does not have support for array or list types, the values need to be separated with some delimiter. The is no character that could be used for URIs, Windows paths and UNIX paths.

PR #2651, opened 2017-04-07
Fix build failures from bad simpletable markup

robander:
A very large document from one of my users went into a new DITA-OT build today for the first time, and failed due to two markup issues:

- Some tables had an invalid `@relcolwidth` attribute of `relcolwidth="*4"`. (No idea how that ended up in the doc.) In XHTML output this kills the build with an XSLT failure; in HTML5 the build continues but it still results in a long series of XSLT errors on the console.
- Another table had a header with only one `<stentry/>` while the table itself had two columns. HTML5 doesn't have a problem with this but an XHTML build fails when adding accessibility tagging, it assumes that if you have a header, there will be one `<stentry>` in the header for each column.

The fix ignores any individual values in `relcolwidth` that can't be interpreted as numbers. (It makes use of a tip I saw on stackoverflow -- `number($x) = number($x)` fails when the input is not a number, because `NaN != NaN`.) The header issue was simpler, we already test "is there a header", I updated that test to say "is there a header with enough columns".

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/simpletable.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/simpletable.xsl
@@ -27,9 +27,16 @@ See the accompanying LICENSE file for applicable license.
       <xsl:apply-templates select="." mode="dita2html:get-max-entry-count"/>
     </xsl:variable>
     <xsl:variable name="col-widths" as="xs:double*">
-      <xsl:variable name="widths" select="tokenize(normalize-space(@relcolwidth), '\s+')" as="xs:string*"/>
+      <xsl:variable name="widths" select="tokenize(normalize-space(translate(@relcolwidth, '*', '')), '\s+')" as="xs:string*"/>
       <xsl:for-each select="$widths">
-        <xsl:sequence select="xs:double(substring(., 1, string-length(.) - 1))"/>
+        <xsl:choose>
+          <xsl:when test=". castable as xs:double">
+            <xsl:sequence select="xs:double(.)"/>
+          </xsl:when>
+          <xsl:otherwise>
+            <xsl:sequence select="xs:double(1)"/>
+          </xsl:otherwise>
+        </xsl:choose>
       </xsl:for-each>
       <xsl:for-each select="1 to ($col-count - count($widths))">
         <xsl:sequence select="xs:double(1)"/>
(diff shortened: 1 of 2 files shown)


robander:
I guess one question here would be - do we generate a message for the bad `relcolwidth` attribute that we're ignoring here? I think probably not, just because there are so many potential issues like that we could (but won't) warn about.

shaneataylor:
Realizing that we can't warn about every possible issue, I'd go with log warnings when we know the processing is affected by bad input. Mistakes happen, and logging warnings helps in two ways:

- identify changes that affected the build and (as this one) might not be immediately obvious on a quick inspection of the output
- help users know why they are not seeing what they expected — it's (sometimes) not the DITA-OT, but your bad input

src/main/plugins/org.dita.html5/xsl/simpletable.xsl
```diff
@@ -29,7 +29,14 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="col-widths" as="xs:double*">
       <xsl:variable name="widths" select="tokenize(normalize-space(@relcolwidth), '\s+')" as="xs:string*"/>
       <xsl:for-each select="$widths">
-        <xsl:sequence select="xs:double(substring(., 1, string-length(.) - 1))"/>
+        <xsl:choose>
```
jelovirt: Since the only allowed value space is `\d+(\.\d+)\*(\w+\d+(\.\d+)\*)*`, this could be refactored so that we do a `translate(@relcolwidth, '*', '')` in variable `$widths`, that way those values would already be doubles as strings. Then this part could just do casting to double.

Also, XPath 2.0 `. castable as xs:double` would IMO be more readable than the XPath 1.0 solutions to testing for NaN.

robander: :+1: makes sense, seems obvious once you point it out...

robander: Updating to
```
<xsl:variable name="widths" select="tokenize(normalize-space(translate(@relcolwidth, '*', '')), '\s+')" as="xs:string*"/>
      <xsl:for-each select="$widths">
        <xsl:choose>
          <xsl:when test=". castable as xs:double">
            <xsl:sequence select="xs:double(.)"/>
          </xsl:when>
          <xsl:otherwise>
            <xsl:sequence select="xs:double(1)"/>
          </xsl:otherwise>
        </xsl:choose>
      </xsl:for-each>
```

Worth noting that this has a side effect of evaluating some invalid values --- by converting the `*` to an empty string, we end up treating all of these as columns with proportional widths of 1, 3 5:

- `relcolwidth="1* 3* 5*"`
- `relcolwidth="1 3 5"`
- `relcolwidth="*1 3* *5"` (this is the sort of typo that caused me to make this change initially)

I think that's probably fine - better than the current crash, and means we're doing what the author wanted even if the syntax is a bit off. But, just wanted to highlight it.

drmacro:
You could replace the for-each and choice with:

<xs:sequence
select="for $wid in $widths 
                   return if ($wid castable as xs:double) 
                              then xs:double($wid) 
                              else 1.0"
/>

Assuming you want a sequence of double values.

Issue #2261, opened 2016-03-15
Eclipse Help - automatically build the contexts.xml + JAR


raducoravu:
Looking through some DITA OT patches we have for Eclipse Help, the transformation can be configured to automatically build the contexts.xml (useful if you want to point to certain topics from dialogs for example) and to create the JAR containig the entire help.
What it would take:

1) Additional targets to generate the contexts and build the JAR:

```
       <target name="generate.contexts">
           <echo>Generate contexts.xml file from topic output files.</echo>
           <xslt in="${output.dir}/${dita.map.filename.root}.xml" out="${output.dir}/contexts.xml" style="${dita.plugin.org.dita.eclipsehelp.dir}/xsl/contexts.xsl">
               <xmlcatalog>
                   <catalogpath>
                       <pathelement location="${dita.dir}/catalog-dita.xml"/>
                   </catalogpath>
               </xmlcatalog>
           </xslt>
       </target>

       <target name="build.eclipsehelp.jar" depends="generate.contexts">           
           <property name="help.jar"  value="${dita.map.filename.root}.jar" />
           <echo>Building Eclipse help jar file: ${help.jar}</echo>
           <jar destfile="${dita.map.output.dir}/${help.jar}" basedir="${dita.map.output.dir}">
               <include name="**/*"/>
               <exclude name="${help.jar}"/>
               <exclude name="plugin.*"/>
           </jar>
       </target>
```

2) A contexts.xsl which would have content like:

```
  <?xml version="1.0" encoding="UTF-8"?>
  <!--OXYGEN PATCH FOR EXM-18224-->
  <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" 
    xmlns:xhtml="http://www.w3.org/1999/xhtml"  
    exclude-result-prefixes="xhtml" 
    version="2.0">

    <xsl:output method="xml" indent="yes" encoding="UTF-8"/>

    <!-- The path of the dir where the Table of Contents file is located (relative to the base dir of
      the Eclipse plugin). -->
    <xsl:param name="prefixHelpInstallPath" select="''"/>

    <xsl:template match="/">
      <xsl:element name="contexts">
        <xsl:apply-templates/>
      </xsl:element>
    </xsl:template>

      <xsl:template match="topic[@href]
                          [ends-with(@href, '.html') or ends-with(@href, '.xhtml')]">
      <xsl:apply-templates select="document(@href)" mode="addContext">
[...]

PR #2670, opened 2017-05-02
Support alternate rendering directory with internal property

robander:
I have several custom transform types that override the `html5`, `xhtml`, and `eclipsehelp` transform types. In each case, my custom transform type is _almost_ the same as the default, but I need to return a zip of the result to the output directory.

Currently, this is very difficult (particularly with `eclipsehelp`). The default `output.dir` property is used in many targets that either copy or render files. My custom transform type - which again is _almost_ exactly the same as `html5` - has to override all of preprocess so that I can redirect targets that copy images. It has to override the main HTML5 steps for the same reason.

This pull request allows custom transform types to use an `init` style parameter that specifies a relative path within the temporary directory. When specified, all steps that render or copy files (apart from the log) will be placed in that directory within temp. This allows a final custom target to zip the files or do any other custom post-processing before they are returned. For example, an ebook transform type could use this to render all HTML5 / CSS / images in a temporary directory, add a step to generate the TOC and manifest, and return a complete zipped EPUB.

The new internal property is (for now) named `ditaot.render.dir`, and defaults to the output directory. This is fully backwards compatible; any custom transform type that uses `output.dir` will still work, and any value specified for `output.dir` will still work. This will only change processing for transform types that choose to override the render directory before running `build-init`.

For myself - this property is most important for `html5`, `xhtml`, and `eclipsehelp` (I override all of these to create a zip). It's less urgent for other transform types, but potentially useful for the same reasons, so this pull request supports the new property for all default transform types. For that same reason, I've added a new parameter `dita.map.render.dir` that could deprecate the existing `dita.map.output.dir`. This wasn't 100% necessary, but particularly with Eclipse Help, it felt wrong to have all the topics go to a `render` directory while map based files all went to a `output` directory.

The property used to do this is currently `temp.render.dir.name`. (I tried a lot of different property names and wasn't 100% happy with any of them.) It must be a relative path that will be created within the temporary directory. For example, if the value is `zip_dir`, all output files will be placed inside `temp/zip_dir`.

Whether or not that is the final parameter name, I'd like to create some initial doc for the parameter, but would like input from @infotexture on where that should go. (It's not really something that should be set on the command line, rather it's there for anybody doing custom transform types. If you set it on the command line with a default transform type, you won't get any output.)

### DIFF
--- a/src/main/plugins/com.sophos.tocjs/integrator.xml
+++ b/src/main/plugins/com.sophos.tocjs/integrator.xml
@@ -29,7 +29,7 @@ See the accompanying LICENSE file for applicable license.
             <isset property="output.file" />
         </not>
     </condition>
-    <property name="output.file" value="${output.dir}${file.separator}toctree.js"/>
+    <property name="output.file" value="${dita.output.dir}${file.separator}toctree.js"/>
     <property name="out.ext" value=".html"/>
   </target>
 
@@ -61,7 +61,7 @@ See the accompanying LICENSE file for applicable license.
       <antcall target="tocjsDefaultXhtmlOutput"/>
       <antcall target="tocjsDefaultFrameset"/>
       <!-- Copy default JS and related files -->
-      <copy todir="${output.dir}"><fileset
+    <copy todir="${dita.output.dir}"><fileset
           dir="${dita.plugin.com.sophos.tocjs.dir}${file.separator}basefiles"><include name="**/*"/></fileset></copy>
   </target>
 
@@ -77,7 +77,7 @@ See the accompanying LICENSE file for applicable license.
       <!-- Create the default frameset itself -->
       <xslt
 		basedir="${dita.temp.dir}"
-		destdir="${output.dir}"
+		destdir="${dita.output.dir}"
         includesfile="${dita.temp.dir}${file.separator}${user.input.file.listfile}"
         extension="${out.ext}"
 		style="${dita.plugin.com.sophos.tocjs.dir}${file.separator}xsl${file.separator}frameset.xsl">
--- a/src/main/plugins/org.dita.base/build_init.xml
+++ b/src/main/plugins/org.dita.base/build_init.xml
@@ -85,6 +85,9 @@ See the accompanying LICENSE file for applicable license.
       </condition>
     </fail>
     <property name="output.dir" location="${basedir}/out" />
+    <condition property="dita.output.dir" value="${dita.temp.dir}${file.separator}${temp.output.dir.name}" else="${output.dir}">
+      <isset property="temp.output.dir.name"/>
+    </condition>
     <property environment="env" />
   </target>
 
--- a/src/main/plugins/org.dita.base/build_preprocess2_template.xml
+++ b/src/main/plugins/org.dita.base/build_preprocess2_template.xml
@@ -103,7 +103,7 @@ See the accompanying LICENSE file for applicable license.
         <param name="generatecopyouter" value="${generate.copy.outer}"/>
         <param name="outercontrol" value="${outer.control}"/>
         <param name="onlytopicinmap" value="${onlytopic.in.map}"/>
-        <param name="outputdir" location="${output.dir}"/>
+        <param name="outputdir" location="${dita.output.dir}"/>
         <param name="transtype" value="${transtype}"/>
         <param name="gramcache" value="${args.grammar.cache}"/>
         <param name="setsystemid" value="${args.xml.systemid.set}"/>
@@ -239,7 +239,7 @@ See the accompanying LICENSE file for applicable license.
         <param name="generatecopyouter" value="${generate.copy.outer}"/>
         <param name="outercontrol" value="${outer.control}"/>
         <param name="onlytopicinmap" value="${onlytmopic.in.map}"/>
-        <param name="outputdir" location="${output.dir}"/>
+        <param name="outputdir" location="${dita.output.dir}"/>
         <param name="transtype" value="${transtype}"/>
         <param name="gramcache" value="${args.grammar.cache}"/>
         <param name="setsystemid" value="${args.xml.systemid.set}"/>
@@ -436,7 +436,7 @@ See the accompanying LICENSE file for applicable license.
         <param name="FILTERFILEURL" expression="${dita.input.filterfile.url}"/>
         <param name="DRAFT" expression="${args.draft}" if="args.draft"/>
         <param name="BASEDIR" expression="${basedir}"/>
-        <param name="OUTPUTDIR" expression="${output.dir}"/>
+        <param name="OUTPUTDIR" expression="${dita.output.dir}"/>
         <param name="defaultLanguage" expression="${default.language}"/>
         <dita:extension id="dita.preprocess.flag-module.param" behavior="org.dita.dost.platform.InsertAction"/>
         <xmlcatalog refid="dita.catalog"/>
@@ -476,7 +476,7 @@ See the accompanying LICENSE file for applicable license.
           dita:depends="{depend.preprocess.copy-image.pre}"
           dita:extension="depends org.dita.dost.platform.InsertDependsAction"
           description="Copy image files">
-    <copy todir="${output.dir}" failonerror="false">
+    <copy todir="${dita.output.dir}" failonerror="false">
       <dita-fileset format="image" />
     </copy>
   </target>
--- a/src/main/plugins/org.dita.base/build_preprocess_template.xml
+++ b/src/main/plugins/org.dita.base/build_preprocess_template.xml
@@ -96,7 +96,7 @@ See the accompanying LICENSE file for applicable license.
         <param name="generatecopyouter" value="${generate.copy.outer}"/>
         <param name="outercontrol" value="${outer.control}"/>
         <param name="onlytopicinmap" value="${onlytopic.in.map}"/>
-        <param name="outputdir" location="${output.dir}"/>
+        <param name="outputdir" location="${dita.output.dir}"/>
         <param name="transtype" value="${transtype}"/>
         <param name="gramcache" value="${args.grammar.cache}"/>
         <param name="setsystemid" value="${args.xml.systemid.set}"/>
@@ -167,7 +167,7 @@ See the accompanying LICENSE file for applicable license.
     <property name="copytosourcefile" value="copytosource.list"/>
     <property name="subtargetsfile" value="subtargets.list"/>
     <property name="resourceonlyfile" value="resourceonly.list"/>
-    <dirname property="_dita.map.output.dir" file="${output.dir}/${user.input.file}" />
+    <dirname property="_dita.map.output.dir" file="${dita.output.dir}/${user.input.file}" />
     <dirname property="_dita.map.temp.dir" file="${dita.temp.dir}/${user.input.file}" />
     <property name="dita.map.output.dir" location="${_dita.map.output.dir}/${uplevels}"/>
     <condition property="noMap">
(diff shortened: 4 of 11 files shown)


robander:
I added a test case for this to `IntegrationTest.java` but it doesn't seem to work.

In the real world - this new property will only be set by custom transform types during initialization. I tried creating a test case that forces the property to be set, which _should_ result in zero files in the output directory. This is how I tested every default transform type locally, and that test passed locally. But, on Travis it gives a Null pointer error when comparing an empty output directory to an empty `exp` directory.

The test case I created, which I'll remove with a new commit, was:
```
    @Test
    public void testrenderdir() throws Throwable {
        builder().name("renderdir")
                .transtype(xhtml)
                .input(Paths.get("simplemap.ditamap"))
                .put("temp.render.dir.name", "hidden")
                .put("clean.temp", "no")
                .test();
    }
```

robander:
FWIW -- here is the batch file I used to test this on Windows with every default transform type. For each transform type, I ran a normal build which places everything in `out/transtypenormal`. Then I ran a transform that set `temp.render.dir.name=testInTemp` on the command line; the output directory was set to `out/transtypeempty` (which should be empty), and temp directory was saved in `out/transtypetemp` (which should have a valid copy of all output in `out/transtypetemp/testInTemp`).

```
call bin\dita -install

set TRANSTYPE=tocjs
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp

set TRANSTYPE=eclipsehelp
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp

set TRANSTYPE=html5
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp

set TRANSTYPE=htmlhelp
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp

set TRANSTYPE=javahelp
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp

set TRANSTYPE=pdf2
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp

set TRANSTYPE=troff
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp

set TRANSTYPE=xhtml
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%normal -f %TRANSTYPE% -temp temp -Dclean.temp=no
call bin\dita -i docsrc\samples\hierarchy.ditamap -o out\%TRANSTYPE%empty -f %TRANSTYPE% -temp out\%TRANSTYPE%temp -Dclean.temp=no -Dtemp.render.dir.name=testInTemp
```

xephon2:
Could you please explain, why "a second temp dir" is necessary? I developed something similar, which works a little different. I set the `out.dir` to a `processing` dir and then add a copy step to copy everything from that dir to a calculated output dir. I calculated the dir by reading metadata after the preprocessing and then save it to properties. The copy step is, of course, called after the publication. This was a little tricky to implement, because there is no after-publication extension point for any reason...

src/main/plugins/com.sophos.tocjs/integrator.xml
```diff
@@ -19,7 +19,10 @@ See the accompanying LICENSE file for applicable license.
   *-->
 <project>
 
-  <target name="dita2tocjs" depends="build-init, preprocess, tocjsInit, map2tocjs, topic2tocjs, tocjsDefaultOutput">
+  <target name="dita2tocjs" depends="test-init, build-init, preprocess, tocjsInit, map2tocjs, topic2tocjs, tocjsDefaultOutput">
+  </target>
+  <target name="test-init">
+    <property name="render.path.within.temp" value="flerp"/>
```
jelovirt: Debug code?

src/main/plugins/com.sophos.tocjs/integrator.xml
```diff
@@ -29,7 +32,7 @@ See the accompanying LICENSE file for applicable license.
             <isset property="output.file" />
         </not>
     </condition>
-    <property name="output.file" value="${output.dir}${file.separator}toctree.js"/>
+    <property name="output.file" value="${ditaot.render.dir}${file.separator}toctree.js"/>
```
jelovirt: The `ditaot.` prefix for the property doesn't follow any of the assorted conventions we currently have. The only real consistent naming scheme we have is prefixing internal properties with `dita.`. It's not really good, but I don't want to introduce `ditaot.` since with would be yet another convention.

src/main/plugins/org.dita.base/build_init.xml
```diff
@@ -85,6 +85,10 @@ See the accompanying LICENSE file for applicable license.
       </condition>
     </fail>
     <property name="output.dir" location="${basedir}/out" />
+    <condition property="ditaot.render.dir" value="${dita.temp.dir}${file.separator}${temp.render.dir.name}" else="${output.dir}">
+      <isset property="temp.render.dir.name"/>
+    </condition>
+    <property name="ditaot.render.dir" location="${output.dir}"/>
```
jelovirt: This is not needed because the else-branch of the preceding `condition` will set the property value.

src/main/plugins/com.sophos.tocjs/integrator.xml
robander: Sigh. Yes.

robander:
> Could you please explain, why "a second temp dir" is necessary?

@xephon2 Of course your method works. TL;DR version: I can't really control `output.dir` itself in my environment - and many targets put results there directly. I'm currently duplicating each of those targets just to avoid this. I'm looking for something easier that works across various transform types. 

My user community is already setting `output.dir` today as part of their command (they've been doing this for a long time now, and the value cannot be automated). I think this rules out my doing anything new with the `output.dir` parameter itself -- I'd either add a whole lot of wrapper code to call a new transform with a new `output.dir`, or break build files by forcing a change to `actual.output.dir`.

For my HTML5 output, I basically need the default HTML5 output, plus a navigation file unique to our platform, and I need to return all of that as a zip. In other words, the zip is the output file, rather than the HTML5. To make this work in 2.3 or 2.4, I had to override all of `preprocess`, so that I could insert my own version of `copy-image`, `copy-flag`, and `copy-html`. I also had to duplicate all of the major HTML5 targets (`html5.topics`, `html5.map`, and `html5.css`) so that the rendered HTML and CSS did not go straight to `output.dir`. All 6 of these steps are exactly the same as the core OT steps, except that they substitute `internal.zip.dir` for `output.dir`.

With this new feature, I'm able to use a single property in my `ibmhtml5.init` to redirect all of the standard HTML5 output to a temporary location. I no longer need to duplicate any core targets. I can run `preprocess`, the core HTML5 targets, and then add my own two targets to generate a navigation file + place a zip in `output.dir`.

I did this once last summer for HTML5. In the fall, I had to extend `eclipsehelp` so that it produced the same thing, but returned everything as `doc.zip`. Eclipse has 11 targets that explicitly use `output.dir`, so I had to duplicate each of those + targets that call them.

Now I need to create another variant of `eclipsehelp` that tweaks a couple of the 11 steps. I was looking with horror on the idea of duplicating those 11 _again_, when I realized others might have this problem. I discussed it at the March and April DITA-OT contributor calls and several others found this idea very useful. And that's the long version of 'why the pull request'...

src/main/plugins/com.sophos.tocjs/integrator.xml
robander: I liked `render` just because it's not the actual "output" as we've used the term -- it's not the file(s) returned to the user, rather it's the place we put everything we're rendering. But I really don't care too much so I'm happy to change.

robander:
@infotexture thanks - that makes sense.

src/main/plugins/com.sophos.tocjs/integrator.xml
robander: Along with this - I think it's best to change the override property from `temp.render.dir.name` to `temp.output.dir.name`.

xephon2:
@robander thank you so much for your time and explanation 👍 😄

robander:
@jelovirt Any issues with this one? I think I've addressed all of the code review comments - once integrated, I'm ready to add a couple more fixes that would otherwise create conflicts.

mironovalexey:
Robert, I'm sorry, I still do not understand the need for a second parameter that works just like the output.folder.
Plugin developer can override `output.folder` and any other DITA variable in this way:
```xml
<project name="ibmhtml5">
...
<target name="ibmhtml5">
    <antcall target="ibmhtml5-makehtml">
        <param name="output.dir" value="out/transtypetemp"/>
    </antcall>
    <antcall target="ibmhtml5-makejar"/>
</target>

<target name="ibmhtml5-makehtml"
        depends="ibmhtml5.init,
                 html5.init,
                 build-init,
                 preprocess,
                 html5.topic,
                 html5.map,
                 ...">

    <echo>Temp output folder: ${output.folder}</echo>
</target>

<target name="ibmhtml5-makejar">
    <echo>Final output folder: ${output.folder}</echo>
    <jar ...>...</jar>
</target>

</project>
```

It seems to me that there's no need to introduce new variable, as it complicates the code and increases the number of parameters.
Your approach implies introducing new "core" properties for each similar task.

robander:
@mironovalexey I don't think this implies new core properties for similar tasks. (I'd also argue that this one parameter doesn't significantly complicate the code; it also does not introduce a build parameter, as this is something that should only be seen or set by a plugin developer.)

Initially I didn't consider this as something to go back into the core; I implemented workarounds for one transform type last summer, then a new one last fall, and only started thinking about this as a useful core property when I was called on to do it for a third. At that point I saw others expressing needs that would also be greatly simplified by this property. For example, creating a JAR file in #2261, which is one of the things I had to do with my workaround.

Before starting a pull request, I also brought up the idea in our monthly contributor call to see if others thought it useful, and got quite a bit of positive feedback. Others have also run into the need for post-processing and zipping, mostly for variations of HTML5 or EPUB. Allowing all of that to happen within the temp directory simplifies this (apparently) common task, as opposed to creating + deleting a second temp directory outside of the normal build process.

mironovalexey:
@robander Robert, I just suggest an approach to get the same result without modifying the base code,  like in the above example. We can simply divide the full sequence of Ant targets into number of parts (targets as well) and call them sequentially. Each part can use its own overrode (or not) Ant properties like `output.folder`.

PR #2672, opened 2017-05-03
Let's keep the map valid after the branch filtering

mironovalexey:
According to the schema ([map](http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part3-all-inclusive/contentmodels/cmltd.html#cmltd__ditavalref), [topicref](http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part3-all-inclusive/contentmodels/cmltt.html#cmltt__topicref), ...) the ditavalref element should be placed after topicmeta. But the [BranchFilterModule doesn't keep it in mind](https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/module/BranchFilterModule.java#L481).

This microscopic fix [moves ditavalref to the bottom](https://github.com/mironovalexey/dita-ot/blob/fix/ditavalref-branchfilter-validmap/src/main/java/org/dita/dost/module/BranchFilterModule.java#L481) of the filtering branch (instead of its top). It restores the validation and should not break the filtering. Then the preprocessed ditamap can be processed again. E.g. this is very useful when the document is preprocessed for translation purposes.

The  [expected transformation result](https://github.com/mironovalexey/dita-ot/blob/fix/ditavalref-branchfilter-validmap/src/test/resources/BranchFilterModuleTest/exp/input.ditamap) of the [corresponding test](https://github.com/dita-ot/dita-ot/blob/develop/src/test/java/org/dita/dost/module/BranchFilterModuleTest.java#L107) is also to be changed.

[test.tar.gz](https://github.com/dita-ot/dita-ot/files/972629/test.tar.gz)

### DIFF
--- a/src/test/resources/BranchFilterModuleTest/exp/input.ditamap
+++ b/src/test/resources/BranchFilterModuleTest/exp/input.ditamap
@@ -3,7 +3,6 @@
   ditaarch:DITAArchVersion="1.3">
   <title class="- topic/title ">DITA Topic Map</title>
   <topicref class="- map/topicref " href="install.dita" keyscope="install">
-    <ditavalref class="+ map/topicref ditavalref-d/ditavalref " format="ditaval" href="linux.ditaval" processing-role="resource-only"/>
     <topicref class="- map/topicref " href="perform-install.dita" keyscope="perform-install"/>
     <topicref class="- map/topicref " href="perform-install.dita" copy-to="installation-procedure.dita" keyscope="perform-install"/>
     <topicref class="- map/topicref " href="configure-novice.dita" keyscope="novice-configure">
@@ -23,14 +22,9 @@
       </ditavalref>
     </topicref>
     <topicref class="- map/topicref " href="http://example.com/install.dita" scope="external"/>
+    <ditavalref class="+ map/topicref ditavalref-d/ditavalref " format="ditaval" href="linux.ditaval" processing-role="resource-only"/>
   </topicref>
   <topicref class="- map/topicref " href="install-mac.dita" keyscope="mac-install">
-    <ditavalref class="+ map/topicref ditavalref-d/ditavalref " format="ditaval" href="mac.ditaval" processing-role="resource-only">
-      <ditavalmeta class="+ map/topicmeta ditavalref-d/ditavalmeta ">
-        <dvrResourceSuffix class="+ topic/data ditavalref-d/dvrResourceSuffix " name="dvrResourceSuffix">-mac</dvrResourceSuffix>
-        <dvrKeyscopePrefix class="+ topic/data ditavalref-d/dvrKeyscopePrefix " name="dvrKeyscopePrefix">mac-</dvrKeyscopePrefix>
-      </ditavalmeta>
-    </ditavalref>
     <topicref class="- map/topicref " href="perform-install-mac.dita" keyscope="perform-install"/>
     <topicref class="- map/topicref " href="installation-procedure-mac.dita" keyscope="perform-install"/>
     <topicref class="- map/topicref " href="configure-novice-mac.dita" keyscope="novice-configure">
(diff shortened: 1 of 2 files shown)


jelovirt:
This is something @robander and I have been discussing over the years. In some cases the order of elements is not actually important, e.g. metadata elements. They are defined to be in a given order in the schemas, but that's only because the schema languages aren't able to constrain ordinality in unordered content model. The question is, is there really any benefit in making the code more complex by trying to retain the redundant order during DITA-OT processing, or could we just allow them to be unordered?

src/test/resources/BranchFilterModuleTest/exp/input.ditamap
```diff
@@ -50,14 +44,14 @@
...
     </topicref>
     <topicref class="- map/topicref " href="http://example.com/install.dita" scope="external"/>
-  </topicref>
-  <topicref class="- map/topicref " href="install-win.dita" keyscope="win-install">
-    <ditavalref class="+ map/topicref ditavalref-d/ditavalref " format="ditaval" href="win.ditaval" processing-role="resource-only">
+    <ditavalref class="+ map/topicref ditavalref-d/ditavalref " format="ditaval" href="mac.ditaval" processing-role="resource-only">
       <ditavalmeta class="+ map/topicmeta ditavalref-d/ditavalmeta ">
-        <dvrResourceSuffix class="+ topic/data ditavalref-d/dvrResourceSuffix " name="dvrResourceSuffix">-win</dvrResourceSuffix>
-        <dvrKeyscopePrefix class="+ topic/data ditavalref-d/dvrKeyscopePrefix " name="dvrKeyscopePrefix">win-</dvrKeyscopePrefix>
+          <dvrResourceSuffix class="+ topic/data ditavalref-d/dvrResourceSuffix " name="dvrResourceSuffix">-mac</dvrResourceSuffix>
```
xephon2: Is this correct? `win` has been replaced with `mac`, further on, there are leading spaces added.

src/test/resources/BranchFilterModuleTest/exp/input.ditamap
```diff
@@ -77,6 +71,12 @@
       </ditavalref>
     </topicref>
     <topicref class="- map/topicref " href="http://example.com/install.dita" scope="external"/>
+      <ditavalref class="+ map/topicref ditavalref-d/ditavalref " format="ditaval" href="win.ditaval" processing-role="resource-only">
+          <ditavalmeta class="+ map/topicmeta ditavalref-d/ditavalmeta ">
+              <dvrResourceSuffix class="+ topic/data ditavalref-d/dvrResourceSuffix " name="dvrResourceSuffix">-win</dvrResourceSuffix>
```
xephon2: Indentation with 4 spaces

mironovalexey:
@jelovirt Preprocessing is highly demanded for the translation procedure, especially when CAT tools are used. The lack of preprocessing seriously slows the translation process. Working with a highly parameterized text, containing unresolved keyrefs and xref's text, is very inconvenient for the translators. Conrefs must be resolved as well, otherwise we face the risk of not having the document translated in its full. It is also very convenient to filter the text prior starting the translation, since this reduces the amount of text be translated. The translation process is seriously simplified if the DITA files remain valid after being preprocessed. In this sense it is better to have a little bit more of code.
 
I would even suggest of making a rule to have the DITA files valid after each preprocessing step. In this case the result of this steps can be used for other purposes, not just for the translation.

robander:
I can understand the need to have valid files after the preprocess - I've had that need before, for tools that take (key- and conref-resolved) DITA as input and are written based on the spec. But overall, in the context of DITA-OT, I know this is tricky.

For this specific fix - I think it's a good one. When we can easily keep validity, we should do so. Along with that, if we're generating something invalid and it would be easy / backwards compatible to change it, I'm probably in favor.

I don't think we can or should aim for validity after every individual step of the preprocess. Very generally speaking, the goal of DITA-OT is to resolve DITA features + publish DITA content within the context of the toolkit itself. We already make this easier by calculating some things once (like flagging or table column information), and passing it along with attributes that are not valid against the original grammar. This works well and is good for all non-DITA outputs; without it, HTML5, XHTML, PDF, and every other format would have to re-implement the same thing. 

That doesn't even get into the restrictions that come with specialization. We have to make some assumptions when resolving DITA features - for example, that metadata will cascade as described in the spec. However, you could easily create a `<topic`> specialization that disallows `<prolog`>, or a `<topicref>` specialization that disallows `<topicmeta>`. The toolkit itself will never be able to recognize that in advance and avoid common features so that the file remains valid.

For better or worse - generating a resolved form of "source DITA" is effectively a new transform type, and will probably have to go through some work to make sure it is still valid. Mostly this will mean cleaning up DITA-OT artifacts like the table column attributes, but it might also require more advanced work to comply with future specializations.

All that said - I still think this is a good thing to fix, and we should try to keep the files "valid" whenever it is easy to do so.

PR #2673, opened 2017-05-03
Add support for troubleshooting elements

shudson310:
Add support for troubleshooting elements

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/common/vars/en.xml
+++ b/src/main/plugins/org.dita.pdf2/cfg/common/vars/en.xml
@@ -29,6 +29,9 @@ with those set forth herein.
 This file is part of the DITA Open Toolkit project. 
 See the accompanying LICENSE file for applicable license.
 -->
+
+<!-- 20170503 SCH: Add support for troubleshooting elements. -->
+
 <vars xmlns="http://www.idiominc.com/opentopic/vars">
 
     <!-- 
@@ -367,6 +370,14 @@ See the accompanying LICENSE file for applicable license.
     <variable id="Task Result"></variable>
     <variable id="Task Example"></variable>
     <variable id="Task Postreq"></variable>
+    
+    <!--Labels for troubleshooting sections, now reused from common-->
+    <variable id="Troubleshooting Condition">Condition or error</variable>
+    <variable id="Troubleshooting Cause">Cause</variable>
+    <variable id="Troubleshooting Remedy">Resolution</variable>
+    <variable id="Troubleshooting responsibleParty">Required personnel</variable>
+    <variable id="Troubleshooting Step">Trouble?</variable>
+    <variable id="Troubleshooting Task">Trouble with this task?</variable>
     
     <!--Glossary Variables-->
     <!-- The footer that appears on the odd-numbered glossary pages. -->
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -31,6 +31,8 @@ This file is part of the DITA Open Toolkit project.
 See the accompanying LICENSE file for applicable license.
 -->
 
+<!-- 20170503 SCH: Add support for troubleshooting elements. -->
+
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
     xmlns:xs="http://www.w3.org/2001/XMLSchema"
     xmlns:fo="http://www.w3.org/1999/XSL/Format"
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/hazard-domain.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/hazard-domain.xsl
@@ -1,0 +1,286 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- 20170503 SCH: Add support for troubleshooting elements. -->
+
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xmlns:fo="http://www.w3.org/1999/XSL/Format"
+  xmlns:opentopic="http://www.idiominc.com/opentopic"
+  xmlns:opentopic-index="http://www.idiominc.com/opentopic/index"
+  xmlns:opentopic-func="http://www.idiominc.com/opentopic/exsl/function"
+  xmlns:dita2xslfo="http://dita-ot.sourceforge.net/ns/200910/dita2xslfo"
+  xmlns:ot-placeholder="http://suite-sol.com/namespaces/ot-placeholder"
+  exclude-result-prefixes="ot-placeholder opentopic opentopic-index opentopic-func dita2xslfo xs"
+  version="2.0">
+
+  <!-- hazardstatement =================================================== -->
+
+  <xsl:template match="*[contains(@class,' hazard-d/hazardstatement ')]">
+    <fo:table xsl:use-attribute-sets="hazardstatement">
+      <xsl:call-template name="commonattributes"/>
+
+      <fo:table-body>
+        <fo:table-row>
+          <fo:table-cell width="100px">
+            <xsl:apply-templates select="./*[contains(@class,' hazard-d/hazardsymbol ')]"/>
+          </fo:table-cell>
+          <fo:table-cell text-align="left">
+            <xsl:call-template name="hazardStatementHead" />
+            <xsl:apply-templates select="*[contains(@class,' hazard-d/messagepanel ')]" />
+          </fo:table-cell>
+        </fo:table-row>
+      </fo:table-body>
+    </fo:table>
+  </xsl:template>
+
+  <xsl:template name="hazardStatementHead">
+    <xsl:choose>
+      <xsl:when test="@type eq 'note'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-note-head" >
+          <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'tip'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-tip-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'fastpath'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-fastpath-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'restriction'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-restriction-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'important'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-important-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'remember'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-remember-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'attention'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-attention-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'caution'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-caution-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'notice'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-notice-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'danger'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-danger-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:when test="@type eq 'warning'">
+        <fo:block xsl:use-attribute-sets="hazardstatement-warning-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:when>
+
+      <xsl:otherwise>
+        <fo:block xsl:use-attribute-sets="hazardstatement-other-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:otherwise>
+    </xsl:choose>
+  </xsl:template>
+  
+  <xsl:template match="*[contains(@class,' hazard-d/hazardstatement ')]" mode="setIconImagePath">
+    <xsl:variable name="noteType" as="xs:string">
+      <xsl:choose>
+        <xsl:when test="@type">
+          <xsl:value-of select="@type"/>
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:value-of select="'note'"/>
+        </xsl:otherwise>
+      </xsl:choose>
+    </xsl:variable>
+    <xsl:call-template name="getVariable">
+      <xsl:with-param name="id" select="concat($noteType, ' Note Image Path')"/>
+    </xsl:call-template>
+  </xsl:template>
+
+  <xsl:template name="hazardStatementHeadContent">
+      <xsl:variable name="type">
+      <xsl:choose>
+        <xsl:when test="exists(@type)">
+          <xsl:value-of select="@type"/>
+        </xsl:when>
+        <xsl:otherwise>other</xsl:otherwise>
+      </xsl:choose>
+    </xsl:variable>
+
+    <xsl:variable name="iconImagePath">
+      <xsl:apply-templates select="." mode="setIconImagePath"/>
+    </xsl:variable>
+    <fo:inline xsl:use-attribute-sets="hazardstatement-head">
+    <xsl:if test="not($iconImagePath = '')">
+      <fo:external-graphic src="url('{concat($artworkPrefix, $iconImagePath)}')" xsl:use-attribute-sets="hazardstatement-head-icon"/>
+    </xsl:if>
+    <xsl:choose>
+        <xsl:when test="@type='note' or not(@type)">
+          <xsl:call-template name="getVariable">
+              <xsl:with-param name="id" select="'Note'"/>
+            </xsl:call-template>
+        </xsl:when>
+        <xsl:when test="@type='notice'">
+          <xsl:call-template name="getVariable">
+              <xsl:with-param name="id" select="'Notice'"/>
+            </xsl:call-template>
+        </xsl:when>
+        <xsl:when test="@type='tip'">
+          <xsl:call-template name="getVariable">
+              <xsl:with-param name="id" select="'Tip'"/>
+            </xsl:call-template>
+        </xsl:when>
+        <xsl:when test="@type='fastpath'">
+          <xsl:call-template name="getVariable">
+              <xsl:with-param name="id" select="'Fastpath'"/>
+            </xsl:call-template>
+        </xsl:when>
+        <xsl:when test="@type='restriction'">
+         <xsl:call-template name="getVariable">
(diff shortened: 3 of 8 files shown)


infotexture:
The DITA-OT project uses the [Git Flow](http://nvie.com/posts/a-successful-git-branching-model/) branching strategy, so *feature enhancement*​ pull requests should target the `develop` branch, not `master` (see the [Contribution Guidelines](https://github.com/dita-ot/dita-ot/blob/develop/CONTRIBUTING.md) for details).


Please [change the base branch](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) of this pull request to `develop`.

src/main/plugins/org.dita.pdf2/cfg/common/vars/en.xml
```diff
@@ -368,6 +371,14 @@ See the accompanying LICENSE file for applicable license.
...
     <variable id="Task Postreq"></variable>
     
+    <!--Labels for troubleshooting sections, now reused from common-->
+    <variable id="Troubleshooting Condition">Condition or error</variable>
+    <variable id="Troubleshooting Cause">Cause</variable>
+    <variable id="Troubleshooting Remedy">Resolution</variable>
+    <variable id="Troubleshooting responsibleParty">Required personnel</variable>
+    <variable id="Troubleshooting Step">Trouble?</variable>
+    <variable id="Troubleshooting Task">Trouble with this task?</variable>
+
```
robander: Right now, having these only appear in the English file will result in "Cannot resolve variable" errors for other languages. I think that at least until our overall string design changes, new variables have to add a stub element for each language. @jelovirt may have additional thoughts though.

src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
```diff
@@ -31,6 +31,8 @@ This file is part of the DITA Open Toolkit project.
 See the accompanying LICENSE file for applicable license.
 -->
 
+<!-- 20170503 SCH: Add support for troubleshooting elements. -->
+
```
robander: Another comment to remove - it doesn't look like this file is touched apart from the comment.

I'll skip commenting on the same comment for other XSL files.

src/main/plugins/org.dita.pdf2/xsl/fo/hazard-domain.xsl
```diff
@@ -0,0 +1,286 @@
...
+  xmlns:fo="http://www.w3.org/1999/XSL/Format"
+  xmlns:opentopic="http://www.idiominc.com/opentopic"
+  xmlns:opentopic-index="http://www.idiominc.com/opentopic/index"
+  xmlns:opentopic-func="http://www.idiominc.com/opentopic/exsl/function"
+  xmlns:dita2xslfo="http://dita-ot.sourceforge.net/ns/200910/dita2xslfo"
+  xmlns:ot-placeholder="http://suite-sol.com/namespaces/ot-placeholder"
+  exclude-result-prefixes="ot-placeholder opentopic opentopic-index opentopic-func dita2xslfo xs"
+  version="2.0">
+
+  <!-- hazardstatement =================================================== -->
```
robander: This is actually outside of the troubleshooting domain - not sure if it should be located in its own Pull Request, or if we should just broaden the scope of this one so that it's actually "troubleshooting + hazard statement".

src/main/plugins/org.dita.pdf2/xsl/fo/hazard-domain.xsl
```diff
@@ -0,0 +1,286 @@
...
+
+      <xsl:otherwise>
+        <fo:block xsl:use-attribute-sets="hazardstatement-other-head">
+            <xsl:call-template name="hazardStatementHeadContent" />
+        </fo:block>
+      </xsl:otherwise>
+    </xsl:choose>
+  </xsl:template>
+  
+  <xsl:template match="*[contains(@class,' hazard-d/hazardstatement ')]" mode="setIconImagePath">
```
robander: This looks like an exact duplicate of the current `<note>` processing, except with a new mode - I'm not sure why we would not use the existing lookup? Anybody can still override the original, but overriding core code inside the core code, without any change, does not seem to buy anything.

shudson310:
Thx @infotexture ! Still getting used to github...

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
```diff
@@ -0,0 +1,205 @@
...
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+    xmlns:fo="http://www.w3.org/1999/XSL/Format"
+    xmlns:dita2xslfo="http://dita-ot.sourceforge.net/ns/200910/dita2xslfo"
+    xmlns:opentopic="http://www.idiominc.com/opentopic"
+    xmlns:opentopic-index="http://www.idiominc.com/opentopic/index"
+    exclude-result-prefixes="opentopic opentopic-index dita2xslfo"
+    version="2.0">
+
+    <!-- Determines whether to generate titles for task sections. Values are YES and NO. -->
+    <xsl:param name="GENERATE-TASK-LABELS">
```
robander: This parameter should not be in the troubleshooting override, looks like it's leftover from starting with a task override as the basis.

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
```diff
@@ -0,0 +1,205 @@
...
+        <fo:block xsl:use-attribute-sets="troubleSolution">
+            <xsl:call-template name="commonattributes"/>
+            <xsl:apply-templates/>
+        </fo:block>
+    </xsl:template>
+
+    <xsl:template match="*[contains(@class, ' troubleshooting/condition ')]">
+        <fo:block xsl:use-attribute-sets="condition">
+            <xsl:call-template name="commonattributes"/>
+            <xsl:apply-templates select="." mode="dita2xslfo:task-heading">
```
robander: The `dita2xslfo:retrieve-task-heading` mode was put in place specifically for tasks, as a way to handle the bad original design of task labels defined twice (once in PDF, once for common, with different IDs). It should not be used here - the single variable ID should be used to retrieve the variable. Same comment for other uses below.

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
```diff
@@ -0,0 +1,205 @@
...
+                    </xsl:apply-templates>
+                </xsl:with-param>
+            </xsl:apply-templates>
+            <fo:block xsl:use-attribute-sets="responsibleParty__content">
+                <xsl:apply-templates/>
+            </fo:block>
+        </fo:block>
+    </xsl:template>
+
+    <xsl:template match="*[contains(@class, ' task/steptroubleshooting ')]">
```
robander: A rule for task elements should not be part of the troubleshooting override, unless it's made specific to the `troubleshooting` topic context. This looks like it should be in the task override.

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
```diff
@@ -0,0 +1,205 @@
...
+                </xsl:apply-templates>
+            </xsl:with-param>
+        </xsl:apply-templates>
+        <fo:block xsl:use-attribute-sets="steptroubleshooting">
+            <xsl:call-template name="commonattributes"/>
+            <xsl:apply-templates/>
+        </fo:block>
+    </xsl:template>
+
+    <xsl:template match="*[contains(@class, ' task/tasktroubleshooting ')]">
```
robander: Same as above - custom processing for task elements should be part of the task module.

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
```diff
@@ -0,0 +1,205 @@
...
+                </xsl:apply-templates>
+              </xsl:otherwise>
+            </xsl:choose>
+            <fo:block xsl:use-attribute-sets="task.example__content">
+              <xsl:apply-templates select="*[not(contains(@class, ' topic/title '))]|text()|processing-instruction()"/>
+            </fo:block>
+        </fo:block>
+    </xsl:template>
+
+    <xsl:template match="*" mode="dita2xslfo:task-heading">
```
robander: This should either be removed, or modified so that it is specific to troubleshooting headings. Currently, this template will override the template already in place for task elements.

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
```diff
@@ -0,0 +1,205 @@
...
+        <xsl:if test="$GENERATE-TASK-LABELS='YES'">
+            <fo:block xsl:use-attribute-sets="section.title">
+                <fo:inline><xsl:copy-of select="$use-label"/></fo:inline>
+            </fo:block>
+        </xsl:if>
+    </xsl:template>
+
+    <!-- Set up to allow string retrieval based on the original PDF2 string;
+         if not found, fall back to the common string -->
+    <xsl:template match="*" mode="dita2xslfo:retrieve-task-heading">
```
robander: Remove this, it is only for tasks and the odd situation of 2 variables for task elements.

src/main/plugins/org.dita.xhtml/xsl/dita2html-base_template.xsl
```diff
@@ -28,6 +28,9 @@ See the accompanying LICENSE file for applicable license.
 
 <!-- the dita to xhtml converter for task documents -->
 <xsl:import href="xslhtml/taskdisplay.xsl"/>
+  
+  <!-- the dita to xhtml converter for troubleshooting documents -->
+  <xsl:import href="xslhtml/troubleshootingdisplay.xsl"/>
```
robander: This is only changing XHTML; I think we'll want equivalent changes in HTML5. The two have now branched, so once you've got the troubleshooting override finalized for one of them, you should copy it into the other and make equivalent updates in the other template.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+<xsl:stylesheet version="2.0"
+     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+     xmlns:related-links="http://dita-ot.sourceforge.net/ns/200709/related-links"
+     xmlns:dita2html="http://dita-ot.sourceforge.net/ns/200801/dita2html"
+     xmlns:ditamsg="http://dita-ot.sourceforge.net/ns/200704/ditamsg"
+     xmlns:xs="http://www.w3.org/2001/XMLSchema"
+     exclude-result-prefixes="related-links dita2html ditamsg xs">
+
+<!-- Determines whether to generate titles for task sections. Values are YES and NO. -->
+<xsl:param name="GENERATE-TASK-LABELS" select="'NO'"/>
```
robander: Again - don't want to duplicate parameters that are already defined for task.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+
+<!-- == TASK UNIQUE SUBSTRUCTURES == -->
+
+  <xsl:template match="*[contains(@class,' troubleshooting/troublebody ')]" name="topic.troubleshooting">
```
robander: I haven't double checked to be sure, but I don't think this does anything that the core rule for `<body>` doesn't do already - if that's the case, we shouldn't override it, and should use the fallback behavior.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+
+  <!-- get the short descr para -->
+  <xsl:apply-templates select="preceding-sibling::*[contains(@class,' topic/shortdesc ')]" mode="outofline"/>
+
+  <xsl:apply-templates/>
+  <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-endprop ')]" mode="out-of-line"/>
+</div><xsl:value-of select="$newline"/>
+</xsl:template>
+
+<xsl:template match="*[contains(@class,' task/steptroubleshooting ')]" name="topic.task.steptroubleshooting">
```
robander: As with the comments on FO -- task elements should be handled in the task module.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+  <xsl:call-template name="generateItemGroupTaskElement"/>
+</xsl:template>
+
+  <xsl:template match="*[contains(@class,' troubleshooting/troubleSolution ')]" name="topic.troubleshooting.troubleSolution">
+  <xsl:call-template name="generateItemGroupTaskElement"/>
+</xsl:template>
+
+<xsl:template name="generateItemGroupTaskElement">
```
robander: This code is simple enough that I don't think it's buying anything by moving it to a named template - particularly given that the `steptroubleshooting` override should really be handled in the task module. That said ... as with the troubleshooting body element ... if this isn't changing the default behavior, the override should just be removed entirely.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+    <xsl:call-template name="setidaname"/>
+    <xsl:apply-templates/>
+  </div>
+</xsl:template>
+
+  <xsl:template match="*[contains(@class,' troubleshooting/condition ')]" mode="dita2html:section-heading">
+  <xsl:apply-templates select="." mode="generate-task-label">
+    <xsl:with-param name="use-label">
+      <xsl:call-template name="getVariable">
+        <xsl:with-param name="id" select="'trouble_condition'"/>
```
robander: This variable ID doesn't exist, at least within this pull request. I think basing this override on the task modules has caused extra confusion here. There should just be one ID for any new variable, and that one ID should be used by both PDF and X/HTML overrides. Same goes for all of the new strings in this file.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+<xsl:template name="generateItemGroupTaskElement">
+  <div>
+    <xsl:call-template name="commonattributes"/>
+    <xsl:call-template name="setidaname"/>
+    <xsl:apply-templates/>
+  </div>
+</xsl:template>
+
+  <xsl:template match="*[contains(@class,' troubleshooting/condition ')]" mode="dita2html:section-heading">
+  <xsl:apply-templates select="." mode="generate-task-label">
```
robander: By using the `generate-task-label` mode, these headings will only appear when task headings are also active. I'm not sure if that's what is intended, so just calling it out.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+    <xsl:apply-templates select="." mode="generate-task-label">
+      <xsl:with-param name="use-label">
+        <xsl:call-template name="getVariable">
+          <xsl:with-param name="id" select="'trouble_remedy'"/>
+        </xsl:call-template>
+      </xsl:with-param>
+    </xsl:apply-templates>
+  </xsl:template>
+  
+  <xsl:template match="*[contains(@class,' troubleshooting/responsibleParty ')]" mode="dita2html:section-heading">
```
robander: I don't think this can ever be active -- `<responsibleParty>` is a paragraph, not a section, so it won't ever be processed with this mode. That also means treating it with the same formatting as a task heading is probably not correct.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+    <xsl:apply-templates select="." mode="generate-task-label">
+      <xsl:with-param name="use-label">
+        <xsl:call-template name="getVariable">
+          <xsl:with-param name="id" select="'trouble_responsibleParty'"/>
+        </xsl:call-template>
+      </xsl:with-param>
+    </xsl:apply-templates>
+  </xsl:template>
+  
+  <xsl:template match="*[contains(@class,' task/tasktroubleshooting ')]" mode="dita2html:section-heading">
```
robander: This one should be moved into the task override.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+      </xsl:with-param>
+    </xsl:apply-templates>
+  </xsl:template>
+
+<!-- 
+     To override the task label for a specific element, match that element with this mode. 
+     For example, you can turn off labels for <context> with this rule:
+     <xsl:template match="*[contains(@class,' task/context ')]" mode="generate-task-label"/>
+-->
+<xsl:template match="*" mode="generate-task-label">
```
robander: This should be removed, or else modified so that it uses a new mode name; `generate-task-label` is already part of the task module, so having it here means the original is ignored.

I now see that this adds an extra `troubleshootinglabel` class -- so having this here is currently adding that class to every task heading as well.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/troubleshootingdisplay.xsl
```diff
@@ -0,0 +1,162 @@
...
+    <div class="troubleshootinglabel">
+      <xsl:element name="{$headLevel}">
+        <xsl:attribute name="class">sectiontitle troubleshootinglabel</xsl:attribute>
+        <xsl:value-of select="$use-label"/>
+      </xsl:element>
+    </div>
+  </xsl:if>
+</xsl:template>
+
+  <!-- Tasks have their own group. -->
```
robander: I think the rest of this can be removed - we should not be modifying task links within the troubleshooting override. I'm not sure if you intended to create special processing for troubleshooting links.

src/main/plugins/org.dita.pdf2/cfg/common/vars/en.xml
```diff
@@ -29,6 +29,9 @@ with those set forth herein.
 This file is part of the DITA Open Toolkit project. 
 See the accompanying LICENSE file for applicable license.
 -->
+
+<!-- 20170503 SCH: Add support for troubleshooting elements. -->
+
```
shudson310: Hm. Saw similar entries in other files, so was following example (taskdisplay.xsl)


I will remove them.

src/main/plugins/org.dita.pdf2/cfg/common/vars/en.xml
shudson310: Good point. I'll add an entry to all of the language files.

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
shudson310: There has been a discussion about creating specific label overrides, but I don't know where we ended up with it. Should I create a trouble label variable or have we decided on a generic one for all labels?

src/main/plugins/org.dita.xhtml/xsl/dita2html-base_template.xsl
shudson310: Indeed. Will update HTML5, too.

src/main/plugins/org.dita.pdf2/xsl/fo/troubleshooting-elements.xsl
robander: I'm not sure we have a definite design here.

Most of our generated text Just Appears -- if you don't want it (or want it to show up differently), you have to customize. Honestly - the reason we have the strange situation with task labels is that originally we didn't have the translations, so didn't generate anything. A lot of people put their own titles in. So when we got translations and added support, there had to be a fairly easy way to turn it off, or else people expecting the old behavior might be broken. (I've seen thousands of topics that still have pseudo-titles in those sections from the early years, and still turn off task headings because of it.)

With this essentially being the first support for troubleshooting, I would probably just generate the headings, without adding a parameter to turn them off.

src/main/plugins/org.dita.pdf2/cfg/common/vars/en.xml
robander: Yeah, those still linger. FWIW those are leftover from ... wow, almost 8 years ago ... years before we were in github, and when the overall source control was mostly handled by a rotating set of folks in IBM China. We just haven't cleaned those out.

jelovirt:
Closing as inactive

PR #2678, opened 2017-05-08
Fix chunking a document loses @xml:lang from root topic #2022

### LINKED ISSUE #2022: Chunking a document loses @xml:lang from root topic
I've edited `garagetaskoverview.xml` to add `xml:lang="fi-fi"` to the root, add a `<note>` that generates text, and add a nested topic that doesn't set the language. I then added `chunk="by-topic"` to the map reference. I get 2 output files as expected. The root file uses the expected generated text for the note: Huomautus. The nested topic uses the default English, losing the `@xml:lang` value set on the parent element.

Edited version of the file:

``` xml
<!DOCTYPE concept PUBLIC "-//OASIS//DTD DITA Concept//EN"
 "../../dtd/technicalContent/dtd/concept.dtd">
<concept id="taskconcept" xml:lang="fi-fi">
  <title>Garage Tasks</title>
  <shortdesc>When you go into the garage, be prepared to get your hands dirty!</shortdesc>
  <conbody>
    <note>This is a note in the root topic</note>
  </conbody>
  <concept id="concept_hg1_xl2_2t">
    <title>Nested Garage Tasks</title>
    <shortdesc>When you go into the garage, be prepared to get your hands dirty!</shortdesc>
    <conbody>
      <note>This is a note in the nested topic</note>
    </conbody>
  </concept>
</concept>

```

mironovalexey:
[test.tar.gz](https://github.com/dita-ot/dita-ot/files/984180/test.tar.gz)
Proposed solution to resolve #2022.

### DIFF
--- a/src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
+++ b/src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
@@ -32,7 +32,6 @@ import static org.apache.commons.io.FileUtils.moveFile;
 import static org.dita.dost.module.GenMapAndTopicListModule.ELEMENT_STUB;
 import static org.dita.dost.reader.ChunkMapReader.*;
 import static org.dita.dost.util.Constants.*;
-import static org.dita.dost.util.StringUtils.isEmptyString;
 import static org.dita.dost.util.StringUtils.split;
 import static org.dita.dost.util.URLUtils.*;
 import static org.dita.dost.util.XMLUtils.*;
@@ -307,13 +306,10 @@ public final class SeparateChunkTopicParser extends AbstractChunkTopicParser {
         if (xmlLang != null) {
             currentLang = xmlLang;
         }
-        else {
-            if (lang.size() > 0) {
-                currentLang = lang.peek();
-            }
-            else {
-                currentLang = "";
-            }
+        else if (!lang.isEmpty()) {
+            currentLang = lang.peek();
+        } else {
+            currentLang = "";
         }
         lang.push(currentLang);
 
@@ -329,8 +325,8 @@ public final class SeparateChunkTopicParser extends AbstractChunkTopicParser {
                     outputFile = generateOutputFilename(id);
                     output = new OutputStreamWriter(new FileOutputStream(new File(outputFile)), UTF8);
 
-                    if(atts.getIndex(ATTRIBUTE_NAME_XML_LANG) < 0 && !isEmptyString(currentLang)) {
-                        attsMod.addAttribute("", ATTRIBUTE_NAME_LANG, ATTRIBUTE_NAME_XML_LANG, "NMTOKEN", currentLang );
+                    if(atts.getIndex(ATTRIBUTE_NAME_XML_LANG) < 0 && !currentLang.isEmpty()) {
+                        attsMod.addAttribute("", ATTRIBUTE_NAME_LANG, ATTRIBUTE_NAME_XML_LANG, "CDATA", currentLang );
                     }
 //                    final FileInfo fi = generateFileInfo(outputFile);
 //                    job.add(fi);
--- a/src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
+++ b/src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
@@ -26,6 +26,7 @@ import java.net.URI;
 import java.util.ArrayDeque;
 import java.util.Collection;
 import java.util.Deque;
+import java.util.LinkedList;
 
 import static org.apache.commons.io.FileUtils.deleteQuietly;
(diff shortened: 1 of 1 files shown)


robander:
I'm having some system issues today that are preventing me from adding comments on the code as an actual "code review". Generally speaking though:

- This is the same approach I thought to take when I planned to fix this myself, so the design seems right to me
- The 3 commits should be squashed to one
- I think it needs just a bit of cleanup around line 307 of `SeparateChunkTopicParser.java`, I'm not sure why it has `if () else {   if () else()  }` instead of `if () else if () else ()` (or maybe @jelovirt will suggest that there's a more straightforward way to write it).

src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
```diff
@@ -293,11 +295,28 @@ public void startElement(final String uri, final String localName, final String
...
 
         if (skip && skipLevel > 0) {
             skipLevel++;
         }
 
+        if (xmlLang != null) {
+            currentLang = xmlLang;
+        }
+        else {
+            if (lang.size() > 0) {
```
jelovirt: Use `isEmpty` instead of `size`.

src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
```diff
@@ -293,11 +295,28 @@ public void startElement(final String uri, final String localName, final String
...
+                currentLang = lang.peek();
+            }
+            else {
+                currentLang = "";
```
jelovirt: Use `null` instead of empty string for empty value

src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
```diff
@@ -310,6 +329,9 @@ public void startElement(final String uri, final String localName, final String
                     outputFile = generateOutputFilename(id);
                     output = new OutputStreamWriter(new FileOutputStream(new File(outputFile)), UTF8);
 
+                    if(atts.getIndex(ATTRIBUTE_NAME_XML_LANG) < 0 && !isEmptyString(currentLang)) {
```
jelovirt: Don't use `isEmptyString`, it should be deprecated. Instead, just test for `null`

src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
```diff
@@ -293,11 +295,28 @@ public void startElement(final String uri, final String localName, final String
...
+
 
         if (skip && skipLevel > 0) {
             skipLevel++;
         }
 
+        if (xmlLang != null) {
+            currentLang = xmlLang;
+        }
+        else {
```
jelovirt: Use `if () {} else if () {} else {}`, ie. no brackets on the first else.

src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
```diff
@@ -310,6 +329,9 @@ public void startElement(final String uri, final String localName, final String
                     outputFile = generateOutputFilename(id);
                     output = new OutputStreamWriter(new FileOutputStream(new File(outputFile)), UTF8);
 
+                    if(atts.getIndex(ATTRIBUTE_NAME_XML_LANG) < 0 && !isEmptyString(currentLang)) {
+                        attsMod.addAttribute("", ATTRIBUTE_NAME_LANG, ATTRIBUTE_NAME_XML_LANG, "NMTOKEN", currentLang );
```
jelovirt: `@xml:lang` type should be `CDATA`.

src/main/java/org/dita/dost/writer/SeparateChunkTopicParser.java
mironovalexey: [Null elements are prohibited](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayDeque.html) in ArrayDeque. Using null is is likely to complicate the code.

jelovirt: Use a `LinkedList`. `null` in this case is the correct solution, because the value of `xml:lang` can also be an empty string to denote undetermined language and that's different from language not being defined.

mironovalexey: > xml:lang can also be an empty string to denote undetermined language
Jarno, thank you. If one expects such behavior, I will fix it

mironovalexey:
Fixed. But `xml:lang=""` (even without chunking) leads to
```bash
[test.tar.gz](https://github.com/dita-ot/dita-ot/files/989463/test.tar.gz)

[xslt] Failed to transform document: An empty sequence is not allowed as the value of variable $l
```

That's why I used String and checked its emptiness.

Issue #1077, opened 2012-04-25
PDF2 plug-in fails when processing bookmap with DITA composite


jelovirt:
Converted from [SourceForge issue 1803111](http://sourceforge.net/support/tracker.php?aid=1803111), submitted by chris-sf

Bookmap processing by PDF2 plug-in 1.4 fails or only renders first child of DITA composite file topicref.

For a bookmap with following structure:

```
bookmap
  chapter
    topicref A (DITA composite)
    topicref B (DITA topic)
```

Topicref A points to a DITA composite file with this structure:

```
dita
  topic (A1)
  concept (A2)
  reference (A3)
  task (A4)
```

When the bookmap is rendered using the PDF2 plug-in, the process fails or you get the following PDF:

```
book
 chapter 1
  topic A1
  topic B
```

A2, A3, and A4 are missing.

robander:
Taking another look at this one. Still not quite sure where to make the modification.
* Update topicmerge? If here, we end up with a merged map that has out of date info in the `<map>` portion, unless topicmerge re-implements title / link text pulling, which seems like poor design.
* In preprocess, after chunking? This seems the best spot. The map can be updated to refer to each topic. But this might result in unexpected output for some formats - not sure about it. Other formats could probably use the update, but like PDF, could still need an update to some sort of merged format.
* Before chunking, update references to these topics to add `chunk="by-topic"`? This would work for formats like PDF or EPUB, but could be undesirable for formats like HTML that have always produced a single output document.

Sample map attached below.
[1077.zip](https://github.com/dita-ot/dita-ot/files/232587/1077.zip)

robander:
Had one other thought over the weekend. The `job.xml` file could be updated to note which files are composite files (root element `<dita>`) ... but I'm not sure if that would really improve efficiency of processing on those documents.

robander:
Further work on this, playing in topicmerge today, a few notes:

- The Java portion of topicmerge drops `<dita>` wrappers with a comment calling them "redundant": https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/reader/MergeTopicParser.java#L204
[...]

PR #2679, opened 2017-05-09
Support DITA composite docs in topicmerge #1077

robander:

This fix addresses #1077 as well as an unreported issue I found with a test case for #1077.

Background: if you have a DITA document that uses the `<dita>` root element, the topicmerge process will only ever preserve the first `<topic>` in that document. This is true whether you reference the document (`href="composite.dita"`), the first topic (`href="composite.dita#first"`), another child of the root element (`href="composite.dita#thirdChildOfRootDita"`), or a sub-topic anywhere in the document (`href="composite.dita#subTopicSomewhere"`). This is clearly wrong; no matter what topic you reference, you get the first (and only the first) topic with its children, which may not even include the one you referenced.

Additional background: with a "normal" DITA document that has a root `topic`: regardless of how a map references the document (just the file, root topic, or sub-topic), you get the whole document in the PDF.

I think there's an argument to be made that if a map references the Nth child of `<dita>` you should only get the Nth child. But I've based this fix around the current processing for "normal" DITA topics. So: if a map references a topic that starts with `<dita>`, you get the whole file (this is the normal scenario). If a map references a sub-topic, you still get the whole file (I haven't actually seen this happen but it matches "normal" topics). This is the behavior we've supported internally for 13+ years, and it's the one that's made the most sense for our users.

To make this fix: `topicmerge` preserves the `<dita>` element and adds an ID so that maps can reference it. In the XSL post-process, if I reference a `<dita>` element, I split the reference up - preserving metadata for each reference, but only putting child topics in the end. [I had hoped to mostly handle this in the actual PDF2 code. But it's so heavily based on topic-id keys that any fix would have to add "If parent is `<dita>` then X else Y" around 20 or 100 times, and would effectively 1) make this a backwards incompatible change and 2) mean nobody with a customization gets the fix.

*Unreported issue that's fixed*

I realized that if I have a "normal" DITA document with a root topic, but only have the map reference a sub-topic, all links are broken. Basically, the map merge process (lines [214](https://github.com/dita-ot/dita-ot/pull/2679/files#diff-e43908434e07462fa522d4e253d607dbR214) and [215](https://github.com/dita-ot/dita-ot/pull/2679/files#diff-e43908434e07462fa522d4e253d607dbR215) ) assume that if you reference a topic ID in the map, that can only be the root topic; so they re-write any references to be to the first topic. That is: if my map has `topics.dita#subtopic`, lines [214](https://github.com/dita-ot/dita-ot/pull/2679/files#diff-e43908434e07462fa522d4e253d607dbR214) and [215](https://github.com/dita-ot/dita-ot/pull/2679/files#diff-e43908434e07462fa522d4e253d607dbR215) ensure that any references to `#subtopic` are rewritten to go to `firstTopicId`. This is also the reason that even if a map references the 3rd child of `<dita>`, you only get the first child of `<dita>` in your PDF.

### DIFF
--- a/src/main/java/org/dita/dost/reader/MergeTopicParser.java
+++ b/src/main/java/org/dita/dost/reader/MergeTopicParser.java
@@ -45,6 +45,7 @@ public final class MergeTopicParser extends XMLFilterImpl {
     private final MergeUtils util;
     private DITAOTLogger logger;
     private File output;
+    private final String GENERATED_DITA_ELEMENT_ID = "GENERATED-DITA-ID";
     
     /**
      * Default Constructor.
@@ -93,10 +94,6 @@ public final class MergeTopicParser extends XMLFilterImpl {
 
     @Override
     public void endElement(final String uri, final String localName, final String qName) throws SAXException {
-        // Skip redundant <dita> tags.
-        if (ELEMENT_NAME_DITA.equals(qName)) {
-            return;
-        }
         getContentHandler().endElement(uri, localName, qName);
     }
 
@@ -202,15 +199,16 @@ public final class MergeTopicParser extends XMLFilterImpl {
     @Override
     public void startElement(final String uri, final String localName, final String qName, final Attributes attributes)
             throws SAXException {
-        // Skip redundant <dita> tags.
-        if (ELEMENT_NAME_DITA.equals(qName)) {
-            rootLang = attributes.getValue(XML_NS_URI, "lang");
-            return;
-        }
         final AttributesImpl atts = new AttributesImpl(attributes);
         final String classValue = atts.getValue(ATTRIBUTE_NAME_CLASS);
 
-        if (TOPIC_TOPIC.matches(classValue)) {
+        if (ELEMENT_NAME_DITA.equals(qName)) {
+            rootLang = attributes.getValue(XML_NS_URI, "lang");
+            if (atts.getValue(ATTRIBUTE_NAME_ID) == null) {
+            	XMLUtils.addOrSetAttribute(atts, ATTRIBUTE_NAME_ID, GENERATED_DITA_ELEMENT_ID);
+            }
+        }
+        if (TOPIC_TOPIC.matches(classValue) || ELEMENT_NAME_DITA.equals(qName)) {
             handleID(atts);
             if (firstTopicId == null) {
                 firstTopicId = atts.getValue(ATTRIBUTE_NAME_ID);
--- a/src/main/plugins/org.dita.pdf2/xsl/common/topicmergeImpl.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/common/topicmergeImpl.xsl
@@ -47,9 +47,8 @@ See the accompanying LICENSE file for applicable license.
   <xsl:variable name="msgprefix" select="'PDFX'"/>
   <xsl:variable name="separator" select="'_Connect_42_'"/>
     
-    <xsl:variable name="originalMap" as="element()">
-        <xsl:sequence select="/dita-merge/*[contains(@class,' map/map ')][1]"/>
-    </xsl:variable>
+  <xsl:variable name="originalMap" as="element()"
+    select="/dita-merge/*[contains(@class,' map/map ')][1]"/>
 
     <xsl:output indent="no"/>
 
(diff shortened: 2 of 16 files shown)


robander:
Attaching test case `map.ditamap` with the following conditions:

- Reference a topic document
- Reference root topic in a topic document
- Reference a sub-topic in topic document
- Reference a composite `<dita>` document
- Reference first topic in a composite `<dita>` document
- Reference middle topic a composite `<dita>` document
- Reference sub-topic in middle topic of a composite `<dita>` document
- Reference last topic in a composite `<dita>` document
- Reference a composite `<dita>` document
- Include 2 topics that use the same topic IDs
- Link to all of those using `<xref>`; also link to a figure within the referenced sub-topic, and to a figure within the `<dita>` document that isn't in the first topic (both of these figure links previously failed).

src/main/plugins/org.dita.pdf2/xsl/common/topicmergeImpl.xsl
```diff
@@ -46,26 +46,108 @@ See the accompanying LICENSE file for applicable license.
...
+        <xsl:choose>
+            <xsl:when test="$ditaDocs/*[@id=$topicId]">
+                <xsl:variable name="topicref" select="." as="element()"/>
+                <xsl:for-each select="$ditaDocs/*[@id=$topicId]/*[contains(@class,' topic/topic ')]">
+                    <xsl:variable name="updateTopicTarget" select="concat('#',@id)"/>
+                    <xsl:variable name="lastTopic" select="if (position()=last()) then true() else false()" as="xs:boolean"/>
+                    <xsl:for-each select="$topicref">
+                        <xsl:copy>
+                            <xsl:copy-of select="@*[not(name()='href' or name()='first_topic_id')]"/>
+                            <xsl:attribute name="first_topic_id"><xsl:value-of select="$updateTopicTarget"/></xsl:attribute>
```
jelovirt: Use `<xsl:attribute name="first_topic_id" select="$updateTopicTarget"/>` instead since it's more simple.

src/main/plugins/org.dita.pdf2/xsl/common/topicmergeImpl.xsl
```diff
@@ -46,26 +46,108 @@ See the accompanying LICENSE file for applicable license.
...
     </xsl:template>
+    
+    <xsl:template match="*[contains(@class,' map/topicref ')][@first_topic_id]" mode="resolve-root-dita">
+        <xsl:param name="ditaDocs" as="element()+"/>
+        <xsl:variable name="topicId" select="substring-after(@first_topic_id,'#')"/>
+        <xsl:choose>
+            <xsl:when test="$ditaDocs/*[@id=$topicId]">
+                <xsl:variable name="topicref" select="." as="element()"/>
+                <xsl:for-each select="$ditaDocs/*[@id=$topicId]/*[contains(@class,' topic/topic ')]">
+                    <xsl:variable name="updateTopicTarget" select="concat('#',@id)"/>
```
jelovirt: While it's clear that the value of `updateTopicTarget` is a string, an explicit `as="xs:string"` should be added.

src/main/plugins/org.dita.pdf2/xsl/common/topicmergeImpl.xsl
```diff
@@ -46,26 +46,108 @@ See the accompanying LICENSE file for applicable license.
...
+                                    <xsl:sequence select="*[contains(@class,' map/topicmeta ')]"/>
+                                </xsl:otherwise>
+                            </xsl:choose>
+                        </xsl:copy>
+                    </xsl:for-each>
+                </xsl:for-each>
+            </xsl:when>
+            <xsl:otherwise>
+                <xsl:copy>
+                    <xsl:apply-templates select="@*|*|text()|comment()|processing-instruction()" mode="resolve-root-dita">
```
jelovirt: Instead of `*|text()|comment()|processing-instruction()` just use `node()`

src/main/plugins/org.dita.pdf2/xsl/common/topicmergeImpl.xsl
```diff
@@ -46,26 +46,108 @@ See the accompanying LICENSE file for applicable license.
...
+        </xsl:choose>
+    </xsl:template>
+    
+    <xsl:template match="@href[starts-with(.,'#')]" mode="resolve-root-dita">
+        <xsl:param name="ditaDocs" as="element()+"/>
+        <xsl:variable name="topicId" select="substring-after(.,'#')"/>
+        <xsl:choose>
+            <xsl:when test="$ditaDocs/*[@id=$topicId]">
+                <xsl:attribute name="href">
+                    <xsl:value-of select="concat('#',$ditaDocs/*[@id=$topicId]/*[1]/@id)"/>
```
jelovirt: Set value of attribute directly with `xsl:attribute/@select` instead of nested `xsl:value-of`

src/main/java/org/dita/dost/reader/MergeTopicParser.java
```diff
@@ -202,15 +200,16 @@ public void startDocument() throws SAXException {
...
-            return;
-        }
         final AttributesImpl atts = new AttributesImpl(attributes);
         final String classValue = atts.getValue(ATTRIBUTE_NAME_CLASS);
 
-        if (TOPIC_TOPIC.matches(classValue)) {
+        if (ELEMENT_NAME_DITA.equals(qName)) {
+            rootLang = attributes.getValue(XML_NS_URI, "lang");
+            if (atts.getValue(ATTRIBUTE_NAME_ID) == null) {
+            	XMLUtils.addOrSetAttribute(atts, ATTRIBUTE_NAME_ID, GENERATED_DITA_ELEMENT_ID);
```
robander: It's adding it into `@oid` -- the original ID -- which is not unique. (If every topic in your map has `id="topic"` then every topic in the merged file will have `oid="topic"`.) 

Strictly speaking - it probably shouldn't be necessary. But code today can reasonably assume that any DITA element with `@id` also has `@oid`, so I wanted to avoid any errors from code that tries to use it.

Even then I thought about making them unique by adding a counter, but _if_ somebody is making use of the value, it might actually be better to have the standard token?

robander: Oh looking at the code again I see why this is confusing.

`<dita>` shouldn't have an ID (it's not legal in source). I double check that just in case it's added by an earlier preprocess step, but assuming it's not there, I add the generated `@id`.

The next portion of the code previously only worked with topics (which *must* have an ID), but I've added in support for `<dita>`:
```
if (TOPIC_TOPIC.matches(classValue) || ELEMENT_NAME_DITA.equals(qName)) {
            handleID(atts);
```

That `handleID` method assumes the original ID exists, which it does for topics, and now exists as that common token for `<dita>`. `handleID` moves `@id` to `@oid`, and is where we actually generate the unique-within-merged-context new `@id`.

jelovirt:
Refactored the code and added test automation. @robander could you review and merge if you're ok with the changes?

Issue #2675, opened 2017-05-05
Something about Ant


mironovalexey:
1. Perhaps it makes sense to add Ant [if/unless namespaces](https://ant.apache.org/manual/ifunless.html) to the [build project root](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/build_preprocess_template.xml#L9).
Since DITA-OT Ant version is 1.9.7, these namespaces automacally open the possibility to use if and unless attributes on tasks and inner elements.
Also Ant itself can be upgraded (http://ant.apache.org/antnews.html)  (release notes https://archive.apache.org/dist/ant/).

1. Despite the fact that there is [preprocess.mapref.skip property](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/build_preprocess_template.xml#L336), we have no chance to skip mapref resolution. Because the keyref task contains its own mapref resolution pipeline (do we really need to resolve them twice?), and that one does not respond to `preprocess.mapref.skip`.
It doesn't affect the result of full preprocessing, but mapref resolution is likely to break ditamap validation in some cases. So for partial preprocessing it may be useful to have a possibilty to switch it off.
With (1) it is quite simple:
```xml
<pipeline message="Resolve mapref in ditamap" taskname="mapref" unless:set="preprocess.mapref.skip">
```

3. The BranchFilterModule gives birth to a new (filtered) topics, which are [added to the .job.xml](https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/module/BranchFilterModule.java#L223). But the [branch-filter task](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/build_preprocess_template.xml#L219) doesn't rewrite file lists, they are [rewritten later - in the chunk task](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/build_preprocess_template.xml#L409). The chunk task execution depends on `preprocess.chunk.skip` property and can be skipped, then these new topics will never be exploded and processed.
I think it would be useful to add the following lines to the end of the branch-filter task.
```xml
<job-helper file="fullditatopic.list" property="fullditatopiclist"/>
<job-helper file="fullditamapandtopic.list" property="fullditamapandtopiclist"/>
```

[...]

PR #2681, opened 2017-05-09
Support override of args.filter

robander:
This request builds on #2637 to make the `args.filter` processing just a bit more flexible.

Currently (in 2.4 and with the update in #2637) the `build-init` task checks to see if a build defines the `args.filter` parameter. If not, it sets up a property that will skip all future ditaval processing. The problem here is that there is no way for a plugin to insert processing for default transforms before the `build-init` task; you either need a new transform with your own init step, or you can plug in before the bigger `preprocess`.

This request moves the check for `args.filter` into `preprocess` so that we do not set `skip.ditaval.preprocess` until it's needed. I see three benefits for this:

1. A plugin can supply one or more common ditaval files. By inserting a target before preprocess, it can set `args.filter` to use those common properties for any or all existing transform types.
1. A plugin can use its own property with a custom platform-independent separator, and then convert that property to an `args.filter` value that is appropriate for the system running the build. This is probably how I will handle the platform questions that came up after #2637 was merged -- use a property like `custom.args.filter` that always uses a comma or semi-colon, and then converts that separator to `;` or `:` in `args.filter`.
1. Even if those use cases are not broadly applicable, it's just a better design to avoid setting `skip.ditaval.preprocess` until we actually need it.

### DIFF
--- a/src/main/plugins/org.dita.base/build_preprocess_template.xml
+++ b/src/main/plugins/org.dita.base/build_preprocess_template.xml
@@ -6,7 +6,7 @@ Copyright 2006 IBM Corporation
 
 See the accompanying LICENSE file for applicable license.
 -->
-<project xmlns:dita="http://dita-ot.sourceforge.net" name="ditaot-preprocess">
+<project xmlns:dita="http://dita-ot.sourceforge.net" name="ditaot-preprocess" xmlns:if="ant:if">
 
   <!-- Set to "true" if you get out-of-memory errors during preprocess
        while processing very large (thousands of files) document sets. -->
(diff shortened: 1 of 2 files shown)


robander:
This picks up a change in the parameter passed to Java:
`<param name="ditaval" value="${args.filter}" if="args.filter"/>`

The attribute `@location` switched to `@value`. I'm not sure which is preferred here, but it's not a single location so that seemed incorrect.

src/main/plugins/org.dita.base/build_preprocess_template.xml
```diff
@@ -69,13 +69,21 @@ See the accompanying LICENSE file for applicable license.
     </delete>
   </target>
   
-  <target name="ditaval-merge" unless="skip.ditaval.preprocess"
-    dita:depends="{depend.preprocess.ditaval-merge.pre}"
+  <target name="ditaval-merge.init">
+    <condition property="skip.ditaval.preprocess" value="true">
```
robander: That's ... a good point. It's there because we've always had it as part of DITAVAL. Right now, if you try to call the merge step without having an `args.filter` parameter, I think it fails, but that could be fixed by updating the code to do nothing when `args.filter` isn't specified.

The main thing it would activate, apart from filtering, is that it can turn off flagging. But I'm not sure if that makes sense. And I'm not sure it's valid to say "I want to use this for filtering but not flagging", or the reverse.

Possibly - you could want to turn off flagging pre-process for a specific transform type. But I don't know that this should be grouped in with *all* the preprocess.

robander: Thinking more on it, the only thing it does is say "If there is no list of ditavals, don't bother with this step". Using the `skip.` property like with other steps is rather misleading.

robander:
Per code review comments from @jelovirt -- there was no need for the property originally set up to allow skipping the merge process. Instead, making use of the `if:set` attribute suggested in #2675 to simplify the merge code.

jelovirt:
@robander My question was really about whether we should provide skip functionality at all. Now it's implemented with `@if:set`, but what's the use case for skipping the the merge? Why would a user want to skip it? The module will do nothing when there's no filter defined and the cost of creating the module and then immediately returning is very small.

robander:
Eh, seemed worth avoiding the very small cost, but I agree it's very small. My thought was -- this module has a very specific purpose, which won't ever do anything when `args.filter` isn't set, so why run the module if it's not set?

I don't have a strong opinion either way.

Issue #2648, opened 2017-04-05
Incorrect topicset/topicref processing in FO template


mironovalexey:
First level topicrefs within topiset container are excluded from FO/PDF processing.

As far as I understand [this selector](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2/xsl/fo/root-processing.xsl#L225) should be more accurate. Something like this
`<xsl:for-each
select="opentopic:map/*[contains(@class, ' map/topicref ') and not(contains(@class, ' mapgroup-d/topicset '))] |
opentopic:map/*[contains(@class, ' mapgroup-d/topicset ')]/*[contains(@class, ' map/topicref ')]">`

Otherwise 
`<map>
<topicset id="xxx">
  <topicref href="..."/>
</topiset>
</map>` 
will lead to incorrect FO tree and broken PDF.

robander:
This really isn't related to the `<topicset>` element, and the fix shouldn't make use of the `topicset/@class` token. That element is really just ... a `topicref` that has a required ID, with no other special processing (it just has the rather subjective extra meaning of "a set of topics"). I can get the same failure with a single root `<topicref>` element that does not have `@href`; I can get fix the failure with the root `<topicset>` by adding `@href` to it.

So, I think all it takes to break the PDF is:
```
<map>
  <topicref>
    <topicref href="goodtopic.dita"/>
  </topicref>
</map>
```

Attaching a version of that test case here with one root `<topicref>` (no `@href`) and 2 simple sub-topics.
[2648.zip](https://github.com/dita-ot/dita-ot/files/988007/2648.zip)

robander:
Related (and same root cause): when I have a map like this, I don't get the build failure, but only the first topic ends up in the PDF -- the second topic is in the TOC / Bookmarks but nowhere else:
```
<map xml:lang="en-us"><title>Grouped group group</title>
  <topicref href="2648a.dita"/>
  <topicref>
    <topicref href="2648b.dita"/>
  </topicref>
</map>
```

jelovirt:
Fixed by #2683

Issue #2614, opened 2017-02-24
Further topicrefs in empty <preface> are ignored in PDF output


raducoravu:
See here for the initial report, it is a regression from DOT 1.8:

https://groups.yahoo.com/neo/groups/dita-users/conversations/messages/41869

For a bookmap with the following preface structure:

        <preface>
            <topicref href="topics/task_preface.dita"  toc="no"/>
        </preface>

the PDF no longer contains the preface content.
It might be connected with: https://github.com/dita-ot/dita-ot/issues/1930

chrizzbee:
Hi Jarno
Was there any progress on this bug in 2.4.x? Topicrefs under Preface and Notice in the Frontmatter are not available in the PDF outoput.
Thanks.

robander:
I've got a fix for this as part of https://github.com/dita-ot/dita-ot/pull/2683 -- should restore topics that are missing after inclusion in empty `<preface>` or empty `<topicref>` inside of the front / back matter.

jelovirt:
Fixed by #2683

Issue #2092, opened 2015-10-08
Using "appendices" wrapper causes PDF2 failure or error


robander:
Using the latest development code, I am trying to reuse a set of appendices from one bookmap inside of another, so I wrapped them in the `<appendices>` element:

``` xml
<appendices id="app">
<appendix href="bookapp1.dita"/>
<appendix href="bookapp2.dita"/>
</appendices>
```

When `@href` is specified on the `<appendices>` element, the PDF looks fine. When there is no `@href` as in the sample above, output is broken. AH returns a PDF without the two topics (though the titles appear in the contents / bookmarks). It also generates the following errors:

```
      [ahf] Error Code  : 10769 (2A11)
      [ahf] Unresolved id value: "_OPENTOPIC_TOC_PROCESSING_d68e2780".
      [ahf] Line 1, Col 20232, file:///C:/Users/IBM_AD~1/AppData/Local/Temp/temp20151008163148010/topic.fo
      [ahf] Error Level : 2
      [ahf] Error Code  : 10769 (2A11)
      [ahf] Unresolved id value: "_OPENTOPIC_TOC_PROCESSING_d68e2867".
      [ahf] Line 1, Col 20849, file:///C:/Users/IBM_AD~1/AppData/Local/Temp/temp20151008163148010/topic.fo
```

When I try with FOP, I get Java errors and no PDF.

robander:
Looks like #1428 from a few years back added some level of support for `<appendices>` that did not reference a topic, but that's missing now.

eerohele:
:+1: encountered this in a customer case.

adambhere:
Tried with Apache FOP and no attributes on the element. Got no errors, but the PDF output does not include the individual appendices (appendix elements and their content). In DITA OT 1.8.5, it was possible to wrap any number of <appendix> elements in <appendices>, with appendices including neither @navtitle nor @href, and the individual <appendix> elements were rendered in PDF. 

A work-around seems to be to add @navtitle or @href to the <appendices> element. This causes the nested <appendix> elements to be rendered, but results in an unwanted "Appendix" page, similar to a part front page. 

The ideal solution would be to render <appendices> only if a @navtitle or @href is provided, otherwise, render only the nested <appendix> elements. In other words the following:

<appendix href="appendix_topic1.dita" />
<appendix href="appendix_topic2.dita" />

... and this:
<appendices>
  <appendix href="appedix_topic1.dita" />
[...]

PR #2683, opened 2017-05-10
Support root level grouping topicrefs #2092 #2648 #2614

robander:
The current PDF2 process assumes that if a root branch of a map has any real content, the root `topicref` of the branch will have either a title or reference a topic (or both). When that is not the case, the PDF is broken. If there is only the one root branch, FOP will crash (as reported in #2648). If there are "valid" root branches, the branch with a root grouping element will appear in the TOC + contents but not in the actual content.

Related, the much older issue #2092 has the same root cause -- it assumes that in a bookmap, any `chapter`, `appendix`, `appendices`, `preface`, or `topicref` (as child of front/back matter container) will either have a title or reference a topic. When that is not the case, results are the same as above.

In all of these cases, the code assumes that the container elements will reference a topic, and try to process the topic with `<xsl:for-each select="key('topic-id', @id)">` (note that when there is only a title in the `topicref` / `chapter` / etc, the merge process generates a topic with that title, so the key-lookup still works). When the `topic-id` lookup fails, no page sequence is generated, and children are not processed.

The fix here checks to see if the `topic-id` returns anything. If the result is empty, it continues to any `topicref` children with the same mode. Additional grouping `topicref` elements do the same. If a topic is eventually found, a valid page sequence is generated, and normal processing continues at that point.

Just realized that this also addresses #2614.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/root-processing.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/root-processing.xsl
@@ -223,9 +223,7 @@ See the accompanying LICENSE file for applicable license.
           <xsl:call-template name="insertBodyStaticContents"/>
           <fo:flow flow-name="xsl-region-body">
             <xsl:for-each select="opentopic:map/*[contains(@class, ' map/topicref ')]">
-              <xsl:for-each select="key('topic-id', @id)">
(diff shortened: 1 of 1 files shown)


robander:
Now updated based on review comments from @jelovirt

PR #2684, opened 2017-05-11
Support creating Eclipse Help output as JAR file

robander:
In 15 years of supporting people using Eclipse, _nearly_ every user has wanted me to return the output as a single JAR rather than as many separate files. I've always managed this with complicated overrides + post-processing, but the new feature #2670 makes it simple.

I realized when I saw #2261 that this was probably common to every Eclipse Help user - the preferred output is a JAR file - so this pull request adds a parameter to make this automatic. When the new parameter specifies a JAR file name, the results will all be placed in that JAR file, and only that file is placed in the output directory.

If I was starting from scratch, returning a JAR would be the default behavior, and the default JAR name would be based on the map. At this point though it's likely many others have post-processing like mine, so we shouldn't change the default behavior.

### DIFF
--- a/src/main/plugins/org.dita.eclipsehelp/build_dita2eclipsehelp_template.xml
+++ b/src/main/plugins/org.dita.eclipsehelp/build_dita2eclipsehelp_template.xml
@@ -6,7 +6,7 @@ Copyright 2006 IBM Corporation
 
 See the accompanying LICENSE file for applicable license.
 -->
-<project xmlns:dita="http://dita-ot.sourceforge.net" name="dita2eclipsehelp">
+<project xmlns:dita="http://dita-ot.sourceforge.net" name="dita2eclipsehelp" xmlns:if="ant:if">
 	
   <target name="dita.eclipsehelp.init">
     <property name="html-version" value="xhtml"/>
@@ -14,6 +14,9 @@ See the accompanying LICENSE file for applicable license.
       <not>
         <isset property="args.xsl" />
       </not>
+    </condition>
+    <condition property="temp.output.dir.name" value="temp_jar_dir">
+      <isset property="args.eclipsehelp.jar.name"/>
     </condition>
   </target>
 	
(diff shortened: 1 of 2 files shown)


src/main/plugins/org.dita.eclipsehelp/build_dita2eclipsehelp_template.xml
```diff
@@ -26,14 +29,21 @@ See the accompanying LICENSE file for applicable license.
 	
     <target name="dita2eclipsehelp"
             unless="noMap"
-        depends="build-init, dita.eclipsehelp.init, preprocess, 
+        depends="dita.eclipsehelp.init, build-init, preprocess, 
                  xhtml.topics,
                  copy-css">
       <antcall target="dita.map.eclipse"></antcall>
+      <jar destfile="${output.dir}/${args.eclipsehelp.jar.name}" basedir="${dita.output.dir}" if:set="args.eclipsehelp.jar.name">
```
robander: FWIW I originally placed the `<jar>` element in a target of its own. But with the peculiar dependency chain here (5 dependency targets followed by nested `<antcall>`), the only way that works is to move the JAR target into the final spot in the final `<antcall>`. That is ... awkward and sort of hides what is happening, so I moved it in here right after `<antcall>`.

As to why the `<antcall>` is there, I'm not really sure, but I don't see much reason to refactor that at this point.

jelovirt: Avoid `<antcall>` if possible.

robander: Agreed, just didn't want to touch stuff that has been there for years.

If we remove that `<antcall>` and just move `dita.map.eclipse` into the `dita2eclipsehelp` dependency list, is there any chance of side effects? I wouldn't think so but I don't know why it was chosen in the first place.

jelovirt: Antcall was probably used because the original coder didn't know there was any difference between antcall and using a dependency (antcall creates a new Project and thus allows overriding property values).

In this case you can just move `dita.map.eclipse` to dependencies.

robander: Done, and made the `<jar>` action into a target that runs after `dita.map.eclipse`.

Issue #2688, opened 2017-05-12
HTML5 tables: missing table classes when args.html5.classattr=no


howlger:
In HTML5 the table CSS classes _frame-none_, _colsep-1_, _rowsep-0_, etc. are missing if and only if the corresponding XML element has no _outputclass_ and when `args.html5.classattr=no`.

## Steps to reproduce:
1. Create the file _table_sample.dita_:
```
<!DOCTYPE concept PUBLIC "-//OASIS//DTD DITA Concept//EN" "dtd/concept.dtd">
<concept id="t1">
   <title>Table sample</title>
   <conbody>
      <table frame="sides" rowsep="0" colsep="1">
         <tgroup cols="2">
            <tbody>
               <row>
                  <entry outputclass="foo">one</entry>
                  <entry>two</entry>
               </row>
            </tbody>
         </tgroup>
      </table>
   </conbody>
</concept>
```
2. Run dita-ot: `dita -f html5 -i table_sample.dita -Dargs.html5.classattr=no`
### Results:
- **Actual**: correct _class="..."_ in the first `<td>`, but missing _class="..."_ in `<table>`  and in the second `<td>`:
```
<table>
  ...
  <tr>
    <td class="foo colsep-1 rowsep-0">one</td>
    <td>two</td>
  </tr>
  ...
```
- **Expected**: with _class="..."_ as in the following:
```
<table class="frame-sides">
  ...
  <tr>
    <td class="foo colsep-1 rowsep-0">one</td>
    <td class="colsep-1 rowsep-0">two</td>
  </tr>
  ...
```

PR #2689, opened 2017-05-12
HTML5 tables: missing table classes when args.html5.classattr=no #2688

howlger:
Fix for https://github.com/dita-ot/dita-ot/issues/2688

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/css-class.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/css-class.xsl
@@ -39,19 +39,21 @@ See the accompanying LICENSE file for applicable license.
 
   <xsl:template match="*" mode="css-class" priority="100">
     <xsl:param name="default-output-class"/>
-    
+
     <xsl:variable name="outputclass" as="attribute(class)?">
       <xsl:apply-templates select="." mode="set-output-class">
         <xsl:with-param name="default" select="$default-output-class"/>
       </xsl:apply-templates>
     </xsl:variable>
-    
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.html5/xsl/css-class.xsl
```diff
@@ -39,19 +39,21 @@ See the accompanying LICENSE file for applicable license.
...
+      <xsl:if test="$outputclass">
         <xsl:sequence select="data($outputclass)"/>
-        <xsl:next-match/>
-      </xsl:variable>
-      
+      </xsl:if>
+      <xsl:next-match/>
+    </xsl:variable>
+
+    <xsl:if test="$class[text()]">
```
jelovirt: This should be `exists($class)`

src/main/plugins/org.dita.html5/xsl/css-class.xsl
```diff
@@ -39,19 +39,21 @@ See the accompanying LICENSE file for applicable license.
...
     <xsl:variable name="outputclass" as="attribute(class)?">
       <xsl:apply-templates select="." mode="set-output-class">
         <xsl:with-param name="default" select="$default-output-class"/>
       </xsl:apply-templates>
     </xsl:variable>
-    
-    <xsl:if test="$outputclass">
-      <xsl:variable name="class">
+
+    <xsl:variable name="class">
```
jelovirt: Add `as="xs:string*"` datatype

Issue #2691, opened 2017-05-16
Unresolved keyrefs to nested topics and elements


mironovalexey:
When keyref points to a non-root topic or element, [KeyrefParser](https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/writer/KeyrefPaser.java#L363) does not process it correctly.

There are some errors in calculation of topic and element IDs and, as a result, we get the wrong `@href`attribute.

[test.tar.gz](https://github.com/dita-ot/dita-ot/files/1003628/test.tar.gz)

jelovirt:
The [spec] says

>For references to topics, maps, and non-DITA resources, the value of the @keyref attribute is simply a key name (for example, keyref="topic-key").
>
> For references to non-topic elements within topics, the value of the @keyref attribute is a key name, a slash ("/"), and the ID of the target element (for example, keyref="topic-key/some-element-id".)

If we are the nitpick the spec, the latter case doesn't say it includes nested topics. However, I think the intent of the spec is to allow referencing nested topics with the slash syntax. @robander could you confirm?

The problem here is that the keyref doesn't contain the information whether the ID after the slash is a topic ID or an element ID (of the first topic that contains an element with that ID). In order to correctly process this, the `KeyrefPaser` would need to actually parse the target file and check which ID it is.

[spec]: http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part3-all-inclusive/archSpec/base/using-keys-for-addressing.html

mironovalexey:
Ok, let's postpone the link to the nested topic for now and take a look at the keyref that points to the non-topic element of the nested topic.
**ditamap**
```xml
<keydef href="topic.dita#nestedtopic" keys="topic-fragment"/>
```
**source topic**
```xml
<xref keyref="topic-fragment/nestedtopic-section"/>
```
**preprocessed topic**
```xml
<xref  ... href="topic.dita#roottopic/nestedtopic-section" .../>
```

But it should be

**preprocessed topic**
```xml
<xref  ... href="topic.dita#nestedtopic/nestedtopic-section" .../>
```

robander:
I see this was opened / closed before I even logged in today, but just wanted to follow up because much of this does depend on specification definitions.

@jelovirt was correct in the initial response that for any `<topic>` level element, the specification requires that you declare a key directly. With the `topickey/elementid` syntax, the `/elementid` portion can only be for a non-topic element referenced within a topic.

[...]

PR #2692, opened 2017-05-16
Suggested approach for #2691

mironovalexey:
It is likely to fix keyrefs to `nested-topic-id/element-id`.

Also it makes possible to reference a topic by the keyref of `a/b` form where `a` points to file and `b` points to topic id. For now this is not supported.  It seems that this change should not affect the current behavior, but rather expands it.

### DIFF
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -375,7 +375,6 @@ public final class KeyrefPaser extends AbstractXMLFilter {
                 String elementId = "";
                 if (slashIndex != -1) {
                     keyName = keyrefValue.substring(0, slashIndex);
-                    elementId = keyrefValue.substring(slashIndex);
                 }
 
                 keyDef = definitionMap.get(keyName);
@@ -385,7 +384,7 @@ public final class KeyrefPaser extends AbstractXMLFilter {
                 if (keyDef != null) {
                     if (currentElement != null) {
                         final NamedNodeMap attrs = elem.getAttributes();
-                        final URI href = keyDef.href;
+                        URI href = keyDef.href;
 
                         if (href != null && !href.toString().isEmpty()) {
                             if (TOPIC_IMAGE.matches(currentElement.type)) {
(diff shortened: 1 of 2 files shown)


jelovirt:
If we are to add support where `keyref="<topic-file>#<topic-id>"` is supported, this will be an extension to DITA-OT. Since it's not obvious that this is non-standard feature, I'm pretty sure it will be rejected. @robander what's your take?

jelovirt:
Cherry-picked eace90edc830858cefb82ab7f2c2c16ed5ca6b2a as aa3b0ff89232407424ac622d2bbe5b1ba9eefa0d because it's not related to this issue.

mironovalexey:
@jelovirt I understand. But it looks convenient. If the spec is not changed we can consider it as undocumented fallback for keyrefs to "file-key/topic-id".
Let's wait for @robander

jelovirt:
Fixed original issue in  	a3a8c42. I you want to send an enhancement PR that adds support for `keyref="<topic-file>#<topic-id>"`, please open a new PR.

robander:
We should not allow `keyref="a/b"` to refer to a nested topic `b` within a DITA document - it's pretty clearly a violation of the specification, and leads to ambiguous resolution for the `keyref` value. We spent a lot of time on specification language to lay out explicit rules for how to interpret a key / keyref value, so that there is only possible way to resolve any `keyref`.

Specifically - based on the specification rules today, `a/b` resolves to a non-topic element with `id="b"` inside of the topic represented by `a`. (If `a` refers to a document, then `a` is explicitly equivalent to the first or root topic within that document  - see below.) If we also allow `a/b` to refer to sub-topic `b` then there are two possible ways to interpret that key reference -- it would be either an element within a topic, or a topic, or both.

Here are the relevant specification details, based on this case:
`<keydef keys="a" href="topic.dita">`

The [specification](http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part1-base/langRef/attributes/thehrefattribute.html) is clear that for linking purposes the `@href` (and thus the key) resolves to the first topic in that document: 
> When an `@href` attribute references a DITA resource, an `@href` value that consists of a URI without a fragment identifier resolves to the document element in the referenced document. ... For the purpose of linking, the reference resolves to the first (or only) topic (or topic specialization) in the document. 

For the `keyref` attribute itself, the [spec](http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part1-base/archSpec/base/using-keys-for-addressing.html#using-keys-for-addressing__syntax) is also clear that the `/elementid` portion must refer to a sub-element, not to a topic itself:
> **Syntax**
> For references to topics, maps, and non-DITA resources, the value of the @keyref attribute is simply a key name (for example, keyref="topic-key").
> For references to non-topic elements within topics, the value of the @keyref attribute is a key name, a slash ("/"), and the ID of the target element (for example, keyref="topic-key/some-element-id".)

So, apart from setting up ambiguous resolution for `a/b`, allowing `/b` to refer to a topic with `id="b"` would contradict the syntax described in the specification.

Issue #2293, opened 2016-04-10
PDF2 drops ® trademark symbols in footers & MiniToCs


infotexture:
If a product name variable is defined in a `<keyword>` with a registered trademark ® symbol as follows, the product name appears correctly in body text when referenced via a key reference such as `<keyword keyref="productname_long"/>`:

``` xml
<keydef keys="productname_long">
  <topicmeta>
    <keywords>
      <keyword>A&amp;P<tm tmtype="reg"/>ProductName&#xA0;S</keyword>
    </keywords>
  </topicmeta>
</keydef>
```

However, the symbol is **dropped** when referenced in PDF footers via `<param ref-name="prodname"/>`, which picks up the content of the `<prodname>` element in `<bookmeta>`, defined via a reference to the same key as:

``` xml
<prodinfo>
  <!-- $productName in footers = <prodname> element in topicmeta or bookmeta -->
  <prodname><keyword keyref="productname_long"/></prodname>
</prodinfo>
```

With the product name defined in this fashion, the trademark symbol is _also_ dropped from any heading links in the chapter-level MiniTOCs when headings contain `<keyword keyref="productname_long"/>`.

If a literal registered trademark ® symbol is used in either `<keyword>` or `<prodname>`, the symbol renders as expected, but is not superscripted as it is when proper `<tm>` markup is used, which scales font size to 75% and applies a 20% baseline shift via the `tm__content` attribute set.

infotexture:
@dita-ot/tc: Since the `<tm>` element is not permitted within `<prodname>`, the client is using the key reference as a workaround.

Does anyone recall the rationale for excluding `<tm>` from the `<prodname>` content model? Seems like a natural candidate for inclusion…

infotexture:
@robander Do you think this is another case of “opinionated” trademark processing like #2065, or just  inconsistent text-only processing in some places, but not in others?

robander:
I suspect that with the content model not allowing trademarks, the processing assumes it's safe to treat as plain text. I'm not sure I've ever seen the contents of `<prodname>` used outside of the cover / front matter, and at least with typical use in our own products, we don't use the TM symbol along with the product name in those locations. 

For what it's worth - I've never seen the `<tm>` element used like this:

```A&amp;P<tm tmtype="reg"/>ProductName&#xA0;S```

[...]

PR #2693, opened 2017-05-16
Enable PDF product name overrides for #2293

robander:
The `$productName` variable is set globally in the PDF code, and then used in many places (headers, footers, etc).

If product name metadata is specified in the map, that name is used to set the variable with `xsl:value-of`. If you want to modify the product name in any way (typically by supporting keywords that should generate extra text, as with the ™ character in #2293), there is no way to do so without overriding every use of `$productName` (default code today uses it in 17 named templates).

This pull request uses the existing `text-only` mode to ensure elements that add text are supported. This doesn't directly address ™, but it does pick up changes from similar elements and means that a simple override can add ™ into those 17 contexts.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/root-processing.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/root-processing.xsl
@@ -64,8 +64,8 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="productName">
         <xsl:variable name="mapProdname" select="(/*/opentopic:map//*[contains(@class, ' topic/prodname ')])[1]" as="element()?"/>
         <xsl:choose>
-            <xsl:when test="$mapProdname">
-                <xsl:value-of select="$mapProdname"/>
+            <xsl:when test="exists($mapProdname)">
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/root-processing.xsl
```diff
@@ -64,8 +64,8 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="productName">
         <xsl:variable name="mapProdname" select="(/*/opentopic:map//*[contains(@class, ' topic/prodname ')])[1]" as="element()?"/>
         <xsl:choose>
-            <xsl:when test="$mapProdname">
-                <xsl:value-of select="$mapProdname"/>
+            <xsl:when test="exists($mapProdname)">
+                <xsl:apply-templates select="$mapProdname" mode="dita-ot:text-only"/>
```
jelovirt: Should this be `mode="set-product-name"`?

robander: Yes. Thanks.

Issue #2077, opened 2015-09-30
Topics with processing-role="resource-only" generate XHTML


robander:
Reported to me by a user (not sure which OT version), but I replicated with 2.1.1. If a map has nothing but key definitions (that is, nothing but references with `processing-role="resource-only"`), XHTML output still generates result files for each of the topics. Resource-only topics should be available to the transform for whatever reason, but should not be generating output.

To reproduce: in `hierarchy.ditamap`, add `processing-role="resource-only"` to every `<topicref>`. Build to XHTML; all topics are still generated. PDF output fails, I assume because no real content goes in the PDF. I tested this test both with the original `<topicref>` (add `@processing-role` manually) and with `<keydef>` (pick up default `@processing-role` and add `@keys`).

If I set the first topic in the map to `@processing-role="normal"`, but leave the others as `resource-only`, XHTML still generates all topics. PDF correctly includes only the one normal topic.

Sample set: 
[2077.zip](https://github.com/dita-ot/dita-ot/files/1005790/2077.zip)

eerohele:
#1734 sounds related.

robander:
Yes, you're right - it is. (I did a search on `resource-only` but that topic didn't sound like the same thing from the title.)

The other one adds in the complicating factor of `@conref` - I've always found it strange that if you just use `@conref` to refer to a topic, that other topic generates X/HTML. But, that's been the case since day 1, and the `onlytopic.in.map` parameter was added to get around it.

robander:
I looked into this more with 2.2.2 after another user reported similar issues. I don't think my original description was correct. 

If I use the sample map and set every topic to `resource-only`, only *some* of the topics are generated. I haven't verified with every topic and link, but it looks like if a `resource-only` topic links to another topic, that other topic is generated. So, when I have these:
```xml
<topicref processing-role="resource-only" href="tasks/changingtheoil.xml" type="task"/>
<topicref processing-role="resource-only" href="concepts/oil.xml" type="concept"/>
```
When I build to HTML5 or XHTML, the output file `concepts/oil.html` is generated because there is a link to it in `tasks/changingtheoil.xml`. The reverse is also true, a link from `oil.xml` results in the output file `tasks/changingtheoil.html`

[...]

Issue #1734, opened 2014-09-01
Too many files generated in XHTML output folder when using conref situation


raducoravu:
Let's say I have a DITA Map:

```
 <!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "map.dtd">
 <map>
  <title>DITA Topic Map</title>
  <topicref href="a.dita"/>
 </map>
```

and 3 topics. 

a.dita:

```
  <!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
  <topic id="topic_zlx_4gd_hp">
    <title>ABC</title>
    <body>
      <p>ADITA </p>
      <p conref="b.dita#topic_gwk_qgd_hp/p_k5d_tgd_hp"/>
    </body>
  </topic>
```

b.dita:

```
    <!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
    <topic id="topic_gwk_qgd_hp">
      <title>B</title>
      <body>
        <p id="p_k5d_tgd_hp">BDITA</p>
        <p><xref href="c.dita"></xref></p>
      </body>
    </topic>
```

c.dita:

```
  <!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
  <topic id="topic_snj_chd_hp">
    <title>CCC</title>
    <body>
      <p>CDITA</p>
    </body>
  </topic>
```

When publishing to XHTML, only the "a.html" should be present in the output folder. Right now in the output folder we also have the "c.html" which is referenced from a part of "b.dita" which should have been ignored by the processing.

raducoravu:
👍 Another one of our users encountered this problem.

eerohele:
[Test files](https://github.com/eerohele/dita-ot-issues/tree/master/fixtures/1734).

Probably also relevant for #2077.

`GenListModuleReader.java` goes through all elements in the input files. It sees `<topicref href="topic3.dita" processing-role="resource-only"/>` and marks `topic3.dita` as resource-only.

Later, it sees `<xref href="topic3.dita"/>`. The code doesn't check whether `topic3.dita` has already been marked as resource-only and assigns it the (implicit) normal processing role, overriding the previously set role.

So, a possible fix might be to check whether a topic (uniquely identified by its absolute URI) has already been assigned a processing role previously:

```diff
[...]

PR #2694, opened 2017-05-16
Stop generating extra linked or conref'ed files #2077

robander:
There are several problems with current processing of `resource-only` topics, and (related) problems with the `onlytopic.in.map` parameter.  In each case addressed here, the problem is that we generate output files for topics that should not be generated (in X/HTML or other formats that generate output by topic). This is because some topics marked `resource-only` are treated as normal, and topics not referenced by the map are (by default) always treated as normal.

In all of these cases, I recognize that switching to keys might resolve some issues. But, it remains legal to use this markup without keys; I've got millions of legacy topics without keys, many of which demonstrate these problems.

This request solves several issues in current `develop` level code. I think some are clearly bugs and the fix is appropriate:

1. I have `topicA.dita` marked `resource-only` in the map. In turn, `topicA.dita` has a `conref` reference to `topicZ.dita` which is not in the map. The resource only `topicA` does not generate output, but `topicZ` does. Instead, `topicZ` should default to `resource-only` as well, and should not generate output unless listed as normal in the map.
1. Similar: I have `topicB.dita` marked as `resource-only`. In turn, `topicB.dita` links to `topicZZ.dita` that does not appear in the map. The `topicZZ` target should also default to `resource-only`; currently it will generate output, *even if listed as `resource-only` in the map*.
1. When `onlytopic.in.map` is true: we should only generate output for topics referenced as `normal` in the map. We currently generate all local conref targets not listed in the map, whether the target is in normal or `resource-only` topics.
1. When `onlytopic.in.map` is true, we also generate output for any link targets that are only listed in the map as `resource-only`. 

I think these are also good corrections to current behavior but there _might_ be an argument for the current results:

1. I have a normal topic in my map. It has a `conref` reference to some other file not listed in the map. In my view - that `conref` target should _not_ generate output -- it should default to `resource-only` because it is only being used as a resource. If output is required, it should be listed in the map as a normal topic.
1. I have `topic.dita` in my map but only marked as `resource-only`. There are links to `topic.dita` from other normal topics. Currently this will generate output - and there is no way *not* to generate output, because links inside the topic cannot be marked `resource-only`. This pull request will not generate output for `topic.dita` unless a) the `resource-only` value is removed, b) it is added somewhere in the map as a "normal" topic, or c) it is removed from the map entirely. [Note: changes when using `onlytopic.in.map` are clearer and addressed above.]
1. Almost the same as previous: I have a normal `topic.dita` with link to `rOnly.dita`. The topic `rOnly.dita` appears in the map as `resource-only`. I think we should interpret this to mean the topic is `resource-only`, for the reasons above, and not generate output. Today, we will generate `rOnly.html`.

Remaining issue not resolved by this fix: If I have a `resource-only` topic, AND that topic has an `@href` reference to `edge.dita` (not `@conref`) to another topic, AND that other topic doesn't appear in the map, AND `onlytopic.in.map` is false or unspecified, we still generate `edge.html`. Ideally this link inside a `resource-only` topic would be defaulting to `resource-only` as well.

The overall code changes to do this are relatively small:
* Currently all references are added to a "normal" or "resourceOnly" set. This adds a third set for anything referenced by non-topicref elements (which _cannot_ specify an explicit role).
* When all references are scanned, if `onlytopic.in.map` is specified, anything that only appears in the non-topicref set (that is, anything never referenced by `topicref`) is added to the resourceOnly set. Otherwise, they remain as normal references (will generate output).
* When all references are scanned, any "pure Conref target" topics are also added to the resourceOnly set, regardless of `onlytopic.in.map`

### DIFF
--- a/src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
+++ b/src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
@@ -460,7 +460,7 @@ public final class GenMapAndTopicListModule extends SourceReaderModule {
     }
 
     /**
-     * Process results from parsing a single topic
+     * Process results from parsing a single topic or map
      *
      * @param currentFile absolute URI processes files
      */
@@ -684,6 +684,8 @@ public final class GenMapAndTopicListModule extends SourceReaderModule {
 
         // Remove pure conref targets from fullTopicSet
         fullTopicSet.removeAll(pureConrefTargets);
+        // Treat pure conref targets same as resource-only
+        resourceOnlySet.addAll(pureConrefTargets);
     }
 
     /**
@@ -718,6 +720,10 @@ public final class GenMapAndTopicListModule extends SourceReaderModule {
         job.setProperty("uplevels", getLevelsPath(rootTemp));
 
         resourceOnlySet.addAll(listFilter.getResourceOnlySet());
+        
+        if (job.getOnlyTopicInMap()) {
+            resourceOnlySet.addAll(listFilter.getNonTopicrefReferenceSet());
+        }
 
         for (final URI file: outDitaFilesSet) {
             getOrCreateFileInfo(fileinfos, file).isOutDita = true;
--- a/src/main/java/org/dita/dost/reader/GenListModuleReader.java
+++ b/src/main/java/org/dita/dost/reader/GenListModuleReader.java
@@ -90,6 +90,8 @@ public final class GenListModuleReader extends AbstractXMLFilter {
     private final Set<URI> resourceOnlySet = new HashSet<>(32);
     /** Topics with processing role of "normal" */
     private final Set<URI> normalProcessingRoleSet = new HashSet<>(32);
+    /** Topics referenced from something other than <topicref> */
+    private final Set<URI> nonTopicrefReferenceSet = new HashSet<>(32);
     /** Subject scheme relative file paths. */
     private final Set<URI> schemeRefSet = new HashSet<>(32);
     /** Relationship graph between subject schema. Keys are subject scheme map paths and values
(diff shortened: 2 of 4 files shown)


robander:
Attaching test case with normal/resource-only topics, with

- Conref and link targets not in the map
- Conref and link targets in the map
- Conref and link targets `resource-only` in the map

Also including the output with current `develop` code in `output-develop`, plus output with this pull request in `output-pr`.
[2077.zip](https://github.com/dita-ot/dita-ot/files/1015091/2077.zip)

robander:
This partially addresses the issue reported in #1734. In that case the map only has `a.dita`, which has conref to `b.dita`, which has link to `c.dita`. 

With latest `develop` code, building HTML5:
- Default HTML5 build generates `a.html`, `b.html`, and `c.html`
- HTML5 with `onlytopic.in.map=true` generates both `a.html` and `b.html`, plus an error that `c.dita` is unavailable to resolve link text, because it's not copied to temp dir.

With the code fix in this PR:
- Default HTML5 build generates `a.html` and `c.html` (this is the unaddressed edge case above)
- HTML5 with `onlytopic.in.map=true` gives the correct result, only generates `a.html`, with no errors

robander:
Playing further with the code to see if I could fix the edge case / the other aspect of #1734 and I'm not sure I can without a major overhaul (or with the map-first process).

Use case: map has `href="a.dita"`; `a` in turn has `conref="b.dita#a/b"`; `b` has `href="c.dita"`.

I process `a.dita` and find the reference to `b.dita`, add to conref targets, add to the wait list for processing.

I process `b.dita` and find the link. I can tell that it hasn't _yet_ been referenced by a `topicref`. So I could treat the link to `c` as resource only. But if `b.dita` later found as a normal reference, I don't have the ability to un-mark `c` as resource-only.

The ideal real way to resolve this is with fully map-first processing - whenever I encounter `conref` or `href` inside of a topic, I will always know if it's part of the map, and if so, if it's normal or resource-only. Outside of that ... given our default policy of `onlytopic.in.map=false`, I think the only option would be a lot more sets that associate the referencing file with the referenced file, so that we can go through later and say "File X referenced Y, and we now know that X was `resource-only` so treat Y as `resource-only`".

robander:
FWIW I think the code is working now. Still plan to create some simple integration tests as well, splitting up my big test case above into smaller very targeted tests.

jelovirt:
Refactored the code a bit, to not create `DitaClass` with empty value, but this was something that was done before this. Also, added tests to `develop` so we can see how this PR changes the expected output for those tests. Used test files from #2077 as the basis for the tests. Please feel free to split those into multiple smaller unit tests.

robander:
One thing on my list for a couple weeks out (so I don't disrupt 2.5 dev here at the end) is to work on cleaning up tests - think I'll leave the current tests as @jelovirt added them and can split them up / add more at a later time.

Issue #1298, opened 2012-06-18
Not all attributes defined on a DITA Composite are copied in the preprocessing stage


raducoravu:
The callback `org.dita.dost.reader.MergeTopicParser.startElement(String, String, String, Attributes)` by default copies all attributes defined on the element in the string buffer. But for a `<dita>` element it makes a fast return and does not copy the attributes defined on the <dita> element to the string buffer.

Why is this a problem?
1. The attributes xtrf and xtrc are not passed further.
2. The `<dita>` element (at least the one which is XML Schema based) could have extra proxy-namespace declarations which are not copied and the XML would become namespace not-wellformed.

By the way, maybe in the entire dost.jar StringBuffer could be replaced with StringBuilder which is faster.

robander:
Also related to #1077 - the fast return means that the `<dita>` root element is dropped, which makes it impossible to determine the original document boundaries in a merged file.

robander:
This should have been marked closed in 2.5 (it was noted in the release notes along with the #2698 pull request).

PR #2698, opened 2017-05-19
Preserve atts from <dita> when merging topics #1298

robander:
Addresses the issue in #1298 -- attributes on `<dita>` (from the source or from preprocessing) are lost during the topic merge step. This update copies those attributes onto child topics, with matching attributes from `topic` taking precedence.


jelovirt:
Fixed indentation to use two space characters.

PR #2714, opened 2017-06-01
Restore support for spectitle, specentry

robander:
The HTML5 refactoring for tables dropped support for two odd attributes, `simpletable/@spectitle` and `stentry/@specentry`.

This pull request restores the call to `spec-title` that supports `@spectitle` (that named template is used in XHTML and still exists in HTML5) It also restores the older processing for entries -- if content or PIs, use that, else if `@specentry`, use that, else use a non-breaking-space that prevents collapsing `<td/>`. The `nbsp` support is there because collapsed `<td/>` has caused a lot of browser issues in the past -- not sure how relevant that is for modern HTML5 browsers but doesn't hurt and seems like a good thing to restore.

[For the named `spec-title` template ... I think we really should have done apply-templates on `@spectitle` itself rather than using the named template, to allow by-element overrides, but `spec-title` is called in quite a few spots so I've left that. Seems a useful cleanup item for 3.0.]

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/simpletable.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/simpletable.xsl
@@ -164,6 +164,7 @@ See the accompanying LICENSE file for applicable license.
   <xsl:template match="*[contains(@class, ' topic/simpletable ')]" name="topic.simpletable">
     <xsl:apply-templates select="*[contains(@class, ' ditaot-d/ditaval-startprop ')]" mode="out-of-line"/>
 
+    <xsl:call-template name="spec-title"/>
     <table>
       <xsl:apply-templates select="." mode="table:common"/>
       <xsl:call-template name="dita2html:simpletable-cols"/>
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.html5/xsl/simpletable.xsl
```diff
@@ -216,7 +217,17 @@ See the accompanying LICENSE file for applicable license.
...
-    <xsl:apply-templates/>
+    <xsl:choose>
+      <xsl:when test="*|text()|processing-instruction()">
+        <xsl:apply-templates/>
+      </xsl:when>
+      <xsl:when test="@specentry">
+        <xsl:apply-templates select="@specentry"/>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:text>&#160;</xsl:text>
```
jelovirt: Why NBSP when entry is empty? This is not backwards compatible, because 2.5 didn't create it. Also, this _may_ break some styling when the element isn't empty anymore.

robander: Two reasons really, but I don't feel very strongly about either of them. 

First - it was there in the XHTML version where this originated - so technically it is restoring the function that was there in HTML5 before the code base split.

Second - it is there in XHTML because in the past, we saw a lot of browser issues where the overall formatting of the table and document were really messed up by `<td/>`, as opposd to `<td></td>` or `<td>&nbsp;</td>`. I don't know if that is relevant any more.

jelovirt: It's probably XSLT 1.0 era, when you had to use output method `xml` for XHTML. Now we have actual `xhtml`, so it will not result in empty elements in XHTML output. My preference is to not output those, unless you really can find a case where modern or semi-modern browsers have issues with the output without the NBSP.

robander: Updated. I'm fine with assuming modern browsers should accept this, particularly with HTML5.

PR #2716, opened 2017-06-02
Fix bug with duplicate IDs in topic

robander:

I have a map that references a topic using an ID:
`<topicref href="resource.dita#duplicate"/>`

In the mappull step, we have several sections that will try to put the target topic into a variable, which is then used to pull text / validate attributes / etc. They're all currently set like this:
`<xsl:variable name="target" select="$doc//*[@id=$topicid]" as="element()?"/>`

I ran into a topic where `resourceid/@id` had to match the topic's ID. When that variable is evaluated, this crashes with the following error because we end up with 2 elements in the variable:
```
[move-meta] Loading stylesheet C:\DITA-OT\dita-ot-2.5\xsl\preprocess\mappull.xsl
Error: Failed to run pipeline: Failed to transform document: A sequence of more than one item is not allowed as the value of variable $target (<concept/>, <resourceid/>)
```

The various templates that set up `$target` this way should only be checking for an ID match on a topic (Because a the map level, the DITA spec says we should only be referencing topics); additionally, adding `[1]` at the end ensures that if somebody ends up with conflicting topic IDs in one document we only grab the first. 

(An alternative might be to change `element()?` to `element()*` and then make tests within each location more complicated - to allow for more than one potential match - but this seems cleaner / better, and duplicate topic IDs is not valid to begin with.)

### DIFF
--- a/src/main/xsl/preprocess/mappullImpl.xsl
+++ b/src/main/xsl/preprocess/mappullImpl.xsl
@@ -63,6 +63,8 @@ Other modes can be found within the code, and may or may not prove useful for ov
       </xsl:when>
     </xsl:choose>
   </xsl:function>
+  
+  <xsl:key name="topic-by-id" match="*[contains(@class,' topic/topic ')]" use="@id"/>
 
   <!-- Find the relative path to another topic or map -->
   <xsl:template name="find-relative-path">
@@ -499,7 +501,7 @@ Other modes can be found within the code, and may or may not prove useful for ov
 
           <!--finding type based on name of the target element in a particular topic in another file-->
           <xsl:when test="$topicpos='otherfile'">
-            <xsl:variable name="target" select="$doc//*[@id=$topicid]" as="element()?"/>
+            <xsl:variable name="target" select="key('topic-by-id', $topicid, $doc)[1]" as="element()?"/>
             <xsl:choose>
               <xsl:when test="$target[contains(@class, $classval)]">
                 <xsl:attribute name="type">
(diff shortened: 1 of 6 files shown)


robander:
Adding sample files:
[2716.zip](https://github.com/dita-ot/dita-ot/files/1048490/2716.zip)

src/main/xsl/preprocess/mappullImpl.xsl
```diff
@@ -499,7 +499,7 @@ Other modes can be found within the code, and may or may not prove useful for ov
 
           <!--finding type based on name of the target element in a particular topic in another file-->
           <xsl:when test="$topicpos='otherfile'">
-            <xsl:variable name="target" select="$doc//*[@id=$topicid]" as="element()?"/>
+            <xsl:variable name="target" select="$doc//*[contains(@class,' topic/topic ')][@id=$topicid][1]" as="element()?"/>
```
jelovirt: Should we declare

```xml
<xsl:key name="topic-by-id" match="*[contains(@class,' topic/topic ')]" use="@id"/>
```

And replace these and other `target` variables with

```xml
<xsl:variable name="target" select="key('topic-by-id', $topicid, $doc)[1]" as="element()?"/>
```

robander: I think that sounds reasonable - did not occur to me since I was just tweaking the existing logic. I think the result should be the same, and slightly more efficient.

robander:
Updated to use suggested code change from @jelovirt and original test doc still passes without error (as with my original version). Should be ready to merge, unless there are further comments.

Issue #2014, opened 2015-08-24
Keyref text resolution not using link text unless on link element


robander:
Based on clarifications to the keyref processing rules in DITA 1.3 (content was there in 1.2 but difficult to interpret):
http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part1-base/archSpec/base/processing-keyref-for-text.html#processing_key_references

For _any element_, text can be pulled from `<linktext>`, which makes that a good default way to specify text with a key (number five in the list of rules). Otherwise (rule 6), normal rules apply as they would for xref, such as pulling from `<navtitle>`.

I've tested this with traditionally non-linking elements `<cite>` and `<keyword>`, then with `<xref>` and `<link>`. I tried keys that place text in `<keyword>`, `<linktext>`>, `@navtitle`, and `<navtitle>`. Results:
- For non-linking elements, text is only pulled from `<keyword>`. If a link target is added, references to the other three will use the URI as link text.
- For `<xref>`, text is pulled from the first three but not from `<navtitle>`. If a link target is added, the URI is used for that case.
- For `<link>`, cases without a link target are removed. With a link target, results match `<xref>`.

Including `test.ditamap` and `cit-test.dita`:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN"
 "map.dtd">
<map xml:lang="en-us">
 <keydef keys="citkeyword"><topicmeta><keywords><keyword>text in keyword</keyword></keywords></topicmeta></keydef>
 <keydef keys="citlinktext"><topicmeta><linktext>text in linktext</linktext></topicmeta></keydef>
 <keydef keys="citnavatt" navtitle="navtitle attribute"/>
 <keydef keys="citnavel"><topicmeta><navtitle>navtitle element</navtitle></topicmeta></keydef>
 <keydef keys="xrefkeyword" href="http://www.dita-ot.org" format="html" scope="external">
  <topicmeta><keywords><keyword>text in keyword</keyword></keywords></topicmeta>
 </keydef>
 <keydef keys="xreflinktext" href="http://www.dita-ot.org" format="html" scope="external">
  <topicmeta><linktext>text in linktext</linktext></topicmeta>
 </keydef>
 <keydef keys="xrefnavatt" navtitle="navtitle attribute" href="http://www.dita-ot.org" format="html" scope="external"/>
[...]

Issue #1590, opened 2013-10-23
cite with keyref to peer HTML broken in XHTML


robander:
Using the hierarchy.ditamap sample, I've added this key definition:

``` xml
<topicref 
   scope="peer" 
   format="html"
   keys="citekey" 
   href="../com.ibm.sample/sample.html#targetid" 
   navtitle="Peer HTML file"/>
```

In changingtheoil.xml, I've added these references:

``` xml
<p>Testing cite <cite keyref="citekey"/> and xref <xref keyref="citekey"/></p>
```

The xref element resolves correctly into a link to the HTML file, with the proper link text. The cite element generates `<a href="..%5C.html#targetid"><cite class="cite"/></a>` in the XHTML output, along with the following warning:

```
  [xslt] C:\DITA-OT1.8.M2\plugins\org.dita.xhtml\xsl\xslhtml\dita2htmlImpl.xsl:4243: Error! java.io.FileNotFoundException: C:\DITA-OT1.8.M2-\temp\temp20131023105911072\tasks\..\..\com.ibm.sample\sample.html (The system cannot find the path specified.) Cause: java.io.FileNotFoundException: C:\DITA-OT1.8.M2\temp\temp20131023105911072\tasks\..\..\com.ibm.sample\sample.html (The system cannot find the path specified.)
```

It looks like the keyref module does not do anything for cite, which makes some sense because cite does not have the scope/format/href attributes. It relies on the dita2html code, which seems to have some bad assumptions in the mode="find-keyref-target" -- it assumes the target is DITA, and assumes that a "." in the href is an extension to a local file.

jelovirt:
@robander #1837 removed keyref processing from HTML. The output is not the same with the current code, but at least no errors are output.

robander:
OK. "The output is not the same" - is it pulling link text now for cite? That was the primary issue - XSL error was almost secondary (though more obvious).

robander:
Tried this again, just working with link text (no `@href` values). Tried a few things that might be expected to work. Storing text as:
* keyword in keywords: works (as it did when I opened this)
* navtitle element (works in Oxygen, not in DITA-OT)
* navtitle attribute (doesn't work)
* linktext element (doesn't work)
* in a `<cite>` (hidden inside phrase inside shortdesc) (doesn't work but I'm kind of glad it doesn't)

Test map:
```xml
<map xml:lang="en-us">
  <keydef keys="citkeyword">
[...]

PR #2719, opened 2017-06-09
Fix key resolution text #2014 and more

robander:
The DITA 1.3 spec topic about [generated text based on key references](http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part1-base/archSpec/base/processing-keyref-for-text.html#processing_key_references) is clear that when the primary rules do not work, all elements with `@keyref` should look to `<linktext>`. This is not working today for elements without `@href`, as reported in #2014.

After starting with @raducoravu 's suggested fix, I extended my test case and found quite a few more edge cases that are either broken (wrong output) or bad (no output when we could provide something). This pull request covers all of the following updates:

- In rule number 2, the spec topic says non-linking elements should look for `<keyword>` inside of `<keywords>` and consider that as matching content. **Today: broken for all elements when keyword appears outside of keywords.** The code looks for `<keyword>` _anywhere_ within the key definition. So if I have `<linktext><keyword>book</keyword> title</linktext>`, matching text for all elements becomes "book" instead of "book title".
- The spec explicitly says that for linking elements, if there are multiple keyword elements in `<keywords>`, all should be used (rule 3 has an example). **Today: broken, we only use the first.** This pull request takes them all, though I consider this an edge case - the rule carried over from DITA 1.2.
- Getting to rule 4, this request adds support for the full content of `<linktext>` as the next fallback for all elements. **Today: broken for all non-linking elements, broken for all elements if keyword appears in linktext.**
- Based on rule 5 in the spec topic, linking elements today fall back to `@navtitle`. **Today: broken for navtitle element, sort of working for navtitle attribute on linking elements, and gives null result for non-linking elements.** There are a couple of problems here:
  - If using navtitle, it should prefer `<navtitle>` (today that is ignored), so I've added that.
  - Non linking elements never get here, and give bad output. Having reached the end of the specification algorithm, with no good output, I think it's foolish to ignore the available `navtitle`, so I've added that as another fallback for non-linking elements.
  - If the target is a local DITA topic, we should _not_ pull the `navtitle` unless that title is locked, because the true title will be pulled from that topic later
- Additionally in rule 5, the specification says that "normal rules apply", which is clearly implementation specific. In links without keys, we fall back to the URI as the last resort, so I've added that as well. Again this is only for non-DITA or non-local content, because local DITA should wait and retrieve link text from the target.

Basically, the focus of this report is on
1. Getting linktext to work so that key definitions have a universal fallback (this is specifically described in the spec as a best practice, and is broken today.
1. Fixing output that today is broken (or just bad), and can easily be made correct (or at least reasonable).

### DIFF
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -291,7 +291,7 @@ public final class KeyrefPaser extends AbstractXMLFilter {
                                 final NodeList linktext = elem.getElementsByTagName(TOPIC_LINKTEXT.localName);
                                 if (linktext.getLength() > 0) {
                                     domToSax((Element) linktext.item(0), true);
-                                } else if (fallbackToNavtitle(elem)) {
+                                } else if (fallbackToNavtitleOrHref(elem)) {
                                     final NodeList navtitleElement = elem.getElementsByTagName(TOPIC_NAVTITLE.localName);
                                     if (navtitleElement.getLength() > 0) {
                                         final AttributesImpl atts = new AttributesImpl();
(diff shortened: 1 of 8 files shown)


robander:
Looks like this also addresses #1590.

src/main/java/org/dita/dost/writer/KeyrefPaser.java
```diff
@@ -249,15 +249,23 @@ public void endElement(final String uri, final String localName, final String na
                     // grab keyword or term from key definition
                     if (!hasSubElem.peek() && currentElement != null) {
                         NodeList nodeList = elem.getElementsByTagName(TOPIC_KEYWORD.localName);
+                        final List<Element> keywordsInKeywords = new ArrayList<>();
+                        for (int index = 0; index < nodeList.getLength(); index++) {
+                            final Element keywordParent = (Element) nodeList.item(index).getParentNode();
+                            final DitaClass parentcls = keywordParent.getAttribute(ATTRIBUTE_NAME_CLASS) != null ? new DitaClass(keywordParent.getAttribute(ATTRIBUTE_NAME_CLASS)) : null;
+                            if (parentcls != null && parentcls.equals(TOPIC_KEYWORDS)) {
```
jelovirt: This should use `matches`, not `equals`.

jelovirt:
Converted for-loop with index to stream API, intent easier to read.

Extracted `writeLinktext` methods to separate methods, also for readability.

Issue #2721, opened 2017-06-12
Update to Saxon-HE 9.8


jelovirt:
Update bundled Saxon-HE to version 9.8 if the performance is approximately at the same level as 9.1.

jelovirt:
Currenly blocked by https://saxonica.plan.io/issues/3252. This can be worked around by extracting `categories.xml` to `src/main/resources/net/sf/saxon/data/` from Saxon-HE-9.8.0-1.jar.

jelovirt:
XMLUnit 1.6 uses XSLT 1.0 to document manipulation. Move to XMLUnit 2.x per #2232

jelovirt:
Depends on #2723

jelovirt:
XMLUnit 2.x also uses XSLT 1.0, thus that is blocking the move to Saxon 9.8.

jelovirt:
It seems only comment stripping uses XSLT in XMLUnit. Thus, if we disable that and clean out comments ourselves, this should be possible.

jelovirt:
Something seems to have changes with how Saxon handles either URI resolution or static base URI. Had to modify XSpec compiler stylesheet in order to get stylesheet loading to work from classpath. It might make sense to update the XSpec stylesheets we bundle.

jelovirt:
Pending https://saxonica.plan.io/issues/3284

jelovirt:
Pull request #2727

PR #2727, opened 2017-06-15
Upgrade to Saxon 9.8 #2721

jelovirt:
Update Saxon 9.1 to 9.8.


raducoravu:
Saxon 9.8.0.2 was just released.

jelovirt:
9.8.0-2 is already in this PR. A bug was fixed in that version that I reported while implementing this PR.

jelovirt:
Updated all code and tests to pass with Saxon-HE 9.8.0-2. Now we need to verify whether there's performance degradation that would be a show-stopper.

jelovirt:
Some simple statistics from our docs (6 x warmup, 10 x measure):

Transtype | Saxon 9.1 | Saxon 9.8
--------- | --------- | ---------
HTML | 6500ms | 5662ms
PDF | 9783ms | 9454ms

PR #2739, opened 2017-06-27
Preserve submap elements when merging

robander:
(I'm not sure the current design is the best for what I'm trying to accomplish, but opening this up to start discussion.)

Currently, the `mapref` process merges all maps into a single map, with a `<submap>` grouping element in place to mark boundaries as needed (such as for key scopes / precedence, or to retain chunk information). However, it loses a lot of information from the submaps:

- Most attributes from the submap itself (like `@outputclass`, `@audience`, `@xml:lang`, etc)
- Titles from the submap
- `<topicmeta>` from the submap

In most cases, this isn't a problem. For my use case, some of the submaps correspond to distinct output artifacts (similar to what `chunk="to-navigation"` was originally intended for). This is a straightforward step that runs after preprocess, with a tweak to the `clean-map` step so they're not removed -- I simply key off of the appropriate `<submap>` elements and chunk the maps.

However, in my case, losing the title means I cannot produce the output I need - the title is a critical part of the individual artifacts. 

This pull request preserves that information, using namespaced elements and attributes (the namespace is the same `dita-ot:` prefix used to preserve a small number of attributes today, like `dita-ot:orig-format` and `dita-ot:orig-href`). The information is preserved as follows:

1. All attributes from the original referenced map are added to the generated `<submap>` element, using the prefix `dita-ot:submap-XXX` where `XXX` is the local name of the attribute. So, the 3 I mentioned above would become `@dita-ot:submap-outputclass`, `@dita-ot:submap-audience`, and `@dita-ot:lang`. Some of these probably are not useful, but I saved them all because a) they won't cause any problems and b) I would rather not pick the ones I consider useful and miss ones others need.
1. Any title in the map is preserved as `<dita-ot:submap-title class="+ topic/foreign ditaot-d/submap-title ">`. This is where I'm not sure about the design. In a map, this would be the first child, and would be a title. `- topic/title ` is not allowed inside `<topicref>` so using it directly would break things. Making it a specialization of `foreign` means our processes should happily ignore it, and also means that whatever elements appear in that title would be legal as well. (Another thought - because the name of an element is not really relevant, this probably should be `<submap-title>` rather than `<dita-ot:submap-title>`?)
1. Any `topicmeta` in the map is preserved as `<dita-ot:submap-topicmeta class="+ topic/foreign ditaot-d/submap-topicmeta ">`. Again, making it foreign means this will be ignored by default (nothing existing today will break), and means the overall structure is valid against core DITA (2 `foreign` elements as a child of the container is good, while a `foreign` title followed by `topicmeta` would not be). The primary purpose here is that when needed, any general metadata for the submap is now _available_ for use by those who need it.
1. In the `clean-map` step today that strips out generated `<submap>` elements, the additional `dita-ot:submap-title` and `dita-ot:submap-topicmeta` are also stripped out. So, no customization or transform type that uses `preprocess` as a whole will not see any change.

The change has been tested against the `gradle test` and integration test with no impact (and I've verified that the test includes submaps with titles). Also tested with my own specialized maps, and it preserves both `map/@title` and `map/title` so that I can restore them later.

There is currently no documentation in our dev section about the `<submap>` element, so if you don't dig into the preprocess code, you never know it's there. I'm not sure if this is by design (probably not?), but it might be useful to have a topic in the developer docs that explain how `<submap>` is used, along with a description of these elements if/when the pull request is accepted.

### DIFF
--- a/src/main/xsl/preprocess/maprefImpl.xsl
+++ b/src/main/xsl/preprocess/maprefImpl.xsl
@@ -128,6 +128,9 @@ See the accompanying LICENSE file for applicable license.
                 </xsl:otherwise>
               </xsl:choose>
             </xsl:variable>
+            <xsl:variable name="targetTitleAndTopicmeta" as="element()*">
+              <xsl:sequence select="$file/*/*[contains(@class,' topic/title ') or contains(@class,' map/topicmeta ')]"/>
+            </xsl:variable>
             <xsl:variable name="contents" as="node()*">
               <xsl:choose>
                 <xsl:when test="not(contains($href,'://') or empty($element-id) or $file/*[contains(@class,' map/map ')][@id = $element-id])">
@@ -169,6 +172,8 @@ See the accompanying LICENSE file for applicable license.
               </xsl:if>
               <xsl:apply-templates select="$target/@chunk"/>
               <xsl:apply-templates select="@* except (@class, @href, @dita-ot:orig-href, @format, @dita-ot:orig-format, @keys, @keyscope, @type)"/>
+              <xsl:apply-templates select="$target/@*" mode="preserve-submap-attributes"/>
+              <xsl:apply-templates select="$targetTitleAndTopicmeta" mode="preserve-submap-title-and-topicmeta"/>
               <xsl:apply-templates select="*[contains(@class, ' ditavalref-d/ditavalref ')]"/>
               <xsl:apply-templates select="$contents">
                 <xsl:with-param name="refclass" select="$refclass"/>
@@ -387,6 +392,23 @@ See the accompanying LICENSE file for applicable license.
       <xsl:with-param name="relative-path" select="$relative-path"/>
       <xsl:with-param name="mapref-id-path" select="$mapref-id-path"/>
     </xsl:call-template>
+  </xsl:template>
+
+  <xsl:template match="@*" mode="preserve-submap-attributes">
+    <xsl:attribute name="dita-ot:submap-{local-name()}" select="."/>
+  </xsl:template>
+  
+  <xsl:template match="*[contains(@class,' topic/title ')]" mode="preserve-submap-title-and-topicmeta">
+    <dita-ot:submap-title class="+ topic/foreign ditaot-d/submap-title ">
+      <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
+      <xsl:apply-templates select="*|processing-instruction()|text()"/>
+    </dita-ot:submap-title>
+  </xsl:template>
+  <xsl:template match="*[contains(@class,' map/topicmeta ')]" mode="preserve-submap-title-and-topicmeta">
+    <dita-ot:submap-topicmeta class="+ topic/foreign ditaot-d/submap-topicmeta ">
+      <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
+      <xsl:apply-templates select="*|processing-instruction()|text()"/>
+    </dita-ot:submap-topicmeta>
   </xsl:template>
 
   <xsl:template match="@* | node()" mode="reltable-copy">
--- a/src/main/xsl/preprocess/maprefImpl.xsl
+++ b/src/main/xsl/preprocess/maprefImpl.xsl
@@ -399,16 +399,18 @@ See the accompanying LICENSE file for applicable license.
   </xsl:template>
   
   <xsl:template match="*[contains(@class,' topic/title ')]" mode="preserve-submap-title-and-topicmeta">
-    <dita-ot:submap-title class="+ topic/foreign ditaot-d/submap-title ">
-      <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
-      <xsl:apply-templates select="*|processing-instruction()|text()"/>
-    </dita-ot:submap-title>
+    <submap-topicmeta class="+ map/topicmeta ditaot-d/submap-topicmeta ">
+      <submap-title class="+ topic/navtitle ditaot-d/submap-title ">
+        <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
+        <xsl:apply-templates select="*|processing-instruction()|text()"/>
+      </submap-title>
+    </submap-topicmeta>
   </xsl:template>
   <xsl:template match="*[contains(@class,' map/topicmeta ')]" mode="preserve-submap-title-and-topicmeta">
-    <dita-ot:submap-topicmeta class="+ topic/foreign ditaot-d/submap-topicmeta ">
+    <submap-topicmeta-container class="+ topic/foreign ditaot-d/submap-topicmeta-container ">
       <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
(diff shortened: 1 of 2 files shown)


src/main/xsl/preprocess/maprefImpl.xsl
```diff
@@ -389,6 +394,23 @@ See the accompanying LICENSE file for applicable license.
     </xsl:call-template>
   </xsl:template>
 
+  <xsl:template match="@*" mode="preserve-submap-attributes">
+    <xsl:attribute name="dita-ot:submap-{local-name()}" select="."/>
+  </xsl:template>
+  
+  <xsl:template match="*[contains(@class,' topic/title ')]" mode="preserve-submap-title-and-topicmeta">
+    <dita-ot:submap-title class="+ topic/foreign ditaot-d/submap-title ">
```
jelovirt: Since DITA has at least traditionally been anti-namespace, I think we could just use `<submap-title class="+ topic/foreign ditaot-d/submap-title ">` here. That's what we've been doing in other places: elements in empty namespace and attribute in DITA-OT namespace.

jelovirt: What's the reason for not using `topicmeta/navtitle` instead of `foreign`? It would more closely map to base DITA and would semantically be pretty sound. A problem with that is that if the submap already has a navtitle, then that would be overridden or lost.

robander: Couple of reasons:
1. Main reason: `<navtitle>` is not allowed as a direct child of `topicref`. I could easily picture some step failing when a fallthrough rule intended for normal `navtitle` processing fires based on that element, when it should just be processing `topicref` elements.
1. If we're going to use / allow elements that would be out-of-context against the original grammar, I'd prefer to just stick with `topic/title` -- use the exact match rather than a close match. (Or ... were you suggesting that we actually create `<topicmeta><navtitle>Title here</></>` inside the `submap` element? I could see that... except based on the submap `@class` value, I bet the current `mappull` would then warn "you have a topic group with a title and that's wrong")

jelovirt: Yes, I meant we wrap navtitle into topicmeta, per the spec. But if we get an error about topicgroup having a title, that would need to be adjusted too.

drmacro:
It is not a DITA error for topicgroup to have title--the spec simply requires that topicgroup titles be ignored, just as submap titles are.

It can be useful to have titles on topicgroups, e.g.:

```
<topicgroup>
  <topicmeta><navtitle>Keydefs</navtitle><topicmeta>
  <keydef .../>
</topicgroup>
```

robander:
@drmacro I hadn't pictured that use, but can envision using it that way.

Testing just now, we don't generate a warning, but we do put out a message (which would definitely be annoying in this submap case):
` [DOTX072I][INFO]: Ignoring navtitle within topicgroup.`

lief-erickson:
@drmacro, I do exactly that in my maps.

jelovirt:
@robander how about we explictly skip throwing DOTX072I for submap wrapper?

robander:
@jelovirt sounds good to me. I'll update for that.

robander:
Discovered we didn't need to skip that message. The code that generates that message is in `clean-map`, and is done with normal fallthrough processing. By dropping the new generated topicmeta element during the `clean-map` stage, we never hit the message.

src/main/xsl/preprocess/maprefImpl.xsl
```diff
@@ -389,6 +394,25 @@ See the accompanying LICENSE file for applicable license.
...
+  
+  <xsl:template match="*[contains(@class,' topic/title ')]" mode="preserve-submap-title-and-topicmeta">
+    <submap-topicmeta class="+ map/topicmeta ditaot-d/submap-topicmeta ">
+      <submap-title class="+ topic/navtitle ditaot-d/submap-title ">
+        <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
+        <xsl:apply-templates select="*|processing-instruction()|text()"/>
+      </submap-title>
+    </submap-topicmeta>
+  </xsl:template>
+  <xsl:template match="*[contains(@class,' map/topicmeta ')]" mode="preserve-submap-title-and-topicmeta">
```
jelovirt: We still have the `<foreign>` element here. Not the output is

```xml
<submap>
  <submap-topicmeta>
    <submap-title>...</submap-title>
  </submap-topicmeta>
  <submap-topicmeta-container>
    ...
  </submap-topicmeta-container>
  ...
</submap>
```

What's the reason for not doing simply

```xml
<submap>
  <submap-topicmeta>
    <submap-title>...</submap-title>
    <submap-topicmeta-container>
      ...
    </submap-topicmeta-container>
  </submap-topicmeta>
  ...
</submap>
```

And if we do that, we don't really need `<submap-topicmeta-container>` anymore?

jelovirt: Or is `<submap-topicmeta-container>` used to retain original metadata elements from map? I'm ok keeping `<submap-topicmeta-container>` but would prefer to move it _inside_ `<submap-topicmeta>`.

robander: Yes, moving inside is much cleaner. I wanted to keep the container because doing otherwise would change some results -- putting previously ignored `topicmeta` elements into the inheritance hierarchy, so that it would suddenly appear. The spec is generally quiet about whether this is right or wrong, so my goal was to make the elements _available_ but not push them in at this point. (This also removes any conflict concern about maps that have both `title` and `topicmeta/navtitle`.)

robander: Also: yes, it's used to retain the original metadata from the referenced map.

jelovirt: What if we keep the original metadata as is, and convert `<title>` to be `<foreign>` specialization?

robander: I'm not sure, it raises another issue not covered by this yet.

Currently we don't save `<topicmeta>` from either the referencing map or the referenced map. 

My primary goal here was to save the title from the referenced map (it's the only thing I actually need for my current use case). I added in `topicmeta` because it's there and easy to save at the same time, using the hidden `foreign` element.

If we're keeping the original metadata from the referenced map as a literal `<topicmeta>`, it sort of highlights that we're dropping the metadata from the referencing map. If we save both, then either we have to use `foreign` for one, or we have to re-implement the metadata merge rules (I don't think we want to do that).

There seem to be a few ways we could proceed on this. In all cases I'm assuming we save the title somehow, and then:

1. Forget about topicmeta, just keep title because that's straightforward, can be put into a generated `topicmeta/navtitle`
1. Put the `topicmeta` contents from the _referenced_ map in as-is (latest suggestion above)
1. Put either the _referenced_ or the _referencing_ `topicmeta` in as-is, but store the other `topicmeta` as foreign
1. Put both the _referenced_ and _referencing_ `topicmeta` in as separate foreign elements, again just making them available for use later as needed
1. Merge the topicmeta using the normal referenced / referencing priority rules, which means re-implementing or re-using that algorithm in the `mapref` module, which I'm not excited about

PR #2740, opened 2017-06-28
Changes to support omit .xml extension in PDF filename

easirois:
Currently in the toolkit when you supply a .xml file to the PDF process it appends .xml extension to the pdf filename.  The pull request provides a fix to remove that extension from the filename.

- [] I have signed-off my commits per http://www.dita-ot.org/DCO.
 - didn't do the -s option since I committed the fix from a GUI tool, but I do sigh off on the commit.
- [x ] I read the [CONTRIBUTING][] file.

## Description

Check to see if the file has a .xml extansion and pass that as the suffix to the basename call in Ant.

## Motivation and Context

The issue is that the .xml file extension ends up in the output filename when the input is .xml file versus a ..dita file.

Before the fix
MapFileName.ditamap = MapFileName.pdf (The OT already does this)
TopicFileName.dita= TopicFileName.pdf (The OT already does this)
TopicFileName.xml= TopicFileName**.xml.**pdf
TopicFileName.anyotherextension= TopicFileName.anyotherextension.pdf

After the fix
MapFileName.ditamap = MapFileName.pdf (The OT already does this)
TopicFileName.dita= TopicFileName.pdf (The OT already does this)
TopicFileName.xml= TopicFileName.pdf (remove the xml from the output file name)
TopicFileName.anyotherextension= TopicFileName. Anyotherextension.pdf (leave it as is and just add PDF)

## How Has This Been Tested?

Yes, have verified the fix works.  Test it in an Windows environment.

## Type of Changes

- [ ] Bug fix _(non-breaking change which fixes an issue)_
- [x ] New feature _(non-breaking change which adds functionality)_
- [ ] Breaking change _(fix or feature that changes existing functionality)_

## Checklist

- [ ] Builds & tests completed successfully (`./gradlew test integrationTest`).
    There were some errors, but I don't believe these are related to the updated code.

org.dita.dost.platform.IntegratorTest > testExecute FAILED
    java.io.IOException at IntegratorTest.java:120 **-- looking for a file**

org.dita.dost.platform.IntegratorTest > testIdPattern FAILED
    java.io.IOException at IntegratorTest.java:120  **-- looking for a file**

org.dita.dost.ProcessorTest > testRun FAILED
    org.dita.dost.exception.DITAOTException at ProcessorTest.java:59
        Caused by: org.apache.tools.ant.BuildException at ProcessorTest.java:59
            Caused by: java.io.FileNotFoundException at ProcessorTest.java:59  **-- looking for a file**

org.dita.dost.util.TestFileUtils > testGetRelativePathFromMapFileFile FAILED
    java.lang.IllegalArgumentException at TestFileUtils.java:77 **-- looking for a file**

415 tests completed, 4 failed, 8 skipped
:test FAILED
    

- [ ] My code follows the code style of this project.
- [ ] I have updated the unit tests to reflect the changes in my code.
- [ ] My change requires a change to the documentation. @dita-ot/docs
- [ ] I will update the documentation accordingly or create a new [docs][] issue.

[CONTRIBUTING]: https://github.com/dita-ot/dita-ot/blob/develop/.github/CONTRIBUTING.md
[docs]: https://github.com/dita-ot/docs/issues/new

### DIFF
--- a/src/main/plugins/org.dita.base/build_preprocess_template.xml
+++ b/src/main/plugins/org.dita.base/build_preprocess_template.xml
@@ -54,7 +54,15 @@ See the accompanying LICENSE file for applicable license.
     
     <basename property="dita.input.filename" file="${args.input}" />
     <basename property="dita.map.filename.root" file="${dita.input.filename}" suffix=".ditamap" />
-    <basename property="dita.topic.filename.root" file="${dita.input.filename}" suffix=".dita" />
+    <condition property="dita.topic.filename.extension" value=".xml">
+      <matches pattern="\.xml$" string="${dita.input.filename}"/>
+    </condition>
+    <condition property="dita.topic.filename.extension" value=".dita">
+      <matches pattern="\.dita$" string="${dita.input.filename}"/>
+    </condition>
+    <property name="dita.topic.filename.extension" value="."/>
+    <basename property="dita.topic.filename.root" file="${dita.input.filename}" suffix="${dita.topic.filename.extension}"/>
+    
     
     <echo level="info">*****************************************************************</echo>
     <echo level="info">* input = ${args.input}</echo>


src/main/plugins/org.dita.base/build_preprocess_template.xml
```diff
@@ -54,7 +54,15 @@ See the accompanying LICENSE file for applicable license.
     
     <basename property="dita.input.filename" file="${args.input}" />
     <basename property="dita.map.filename.root" file="${dita.input.filename}" suffix=".ditamap" />
-    <basename property="dita.topic.filename.root" file="${dita.input.filename}" suffix=".dita" />
+    <condition property="dita.topic.filename.extension" value=".xml">
```
jelovirt: I don't like hard-coding common extensions for topics and then failing to remove the extension if some other extension is used. This should be handled with more generic approach, one which supports any extension.

jelovirt: I don't like hard-coding common extensions for topics and then failing to remove the extension if some other extension is used. This should be handled with more generic approach, one which supports any extension.

src/main/plugins/org.dita.base/build_preprocess_template.xml
```diff
@@ -54,7 +54,15 @@ See the accompanying LICENSE file for applicable license.
     
     <basename property="dita.input.filename" file="${args.input}" />
```
jelovirt: Sign-off missing

easirois: Is there a way to do a sign-off post pull request?

easirois: Nevermind...I'll need to submit another pull request for the changes. I'll talk to Martin at the office for a more generic solution. Thanks!

jelovirt: You can just modify the branch locally and then force push to your GitHub remote to update the PR.

src/main/plugins/org.dita.base/build_preprocess_template.xml
easirois: Without getting into ant-contrib, there isn't really an easy way to the extension without using regex.

easirois: That said, we only support .xml or .dita for the OT, no?

robander: Any topic extension should work (I just tried changing our old hierarchy sample file from `wwfluid.xml` to `wwfluid.beep` and that worked fine for HTML5 output).

robander: OK, yes, nobody is going to use a `.beep` extension. In the real world, I've only ever encountered `.xml` and `.dita`. But we should continue to allow other extensions. Also - I have encountered maps that included `.xml` files from one source and `.dita` files from another, so hard coding a single extension would not work for those.

jelovirt:
Implemented in 62a08c4 to support any file extension.

src/main/plugins/org.dita.base/build_preprocess_template.xml
easirois: What if someone submitted wwfuild as the filename, do we want to support that or just say you need an extension?  The regex solution assumes a "."  The current solution works fine, just wondering if this use case should be supported or not.  I'm okay either way.

PR #2748, opened 2017-07-12
Add flagging information during filtering

jelovirt:
Refactor preprocessing so that flagging information is not populated to topics in `flag-module` with XSLT, but instead it's pushed when files are filtered. The benefit is that subject scheme information is available at that stage and we can handle all DITAVAL related operations in one go, in both base and branch filtering.

- [x] preprocess gen-list and debug-filter
- [x] preprocess2 map-reader 
- [x] preprocess2 topic-reader 
- [x] preprocess branch filtering
- [x] preprocess2 map branch filtering
- [x] preprocess2 topic branch filtering
- [x] style conflict
- [x] preprocess2 branch filtering doesn't copy flag images to output


infotexture:
@jelovirt Looks like this refactoring will require docs changes in [Flagging (flag-module)](http://www.dita-ot.org/dev/dev_ref/preprocess-flagging.html).

Could you review & verify?

If significant changes are required, we may need to create a docs issue to track them.

jelovirt:
Yes. There is no longer a separate the flag module. Instead, the same flagging information has been pushed to topics during filtering. The information is still valid, the only thing that needs to be updated is _when_ the flags are processed.

hongyung:
Hi @jelovirt , is there any reason that the usage of URLUtil.toFile has been removed in the BranchFilterModule filterTopics method ?
I have href with fragment in my keydef and causing error during branch-filtering.

jelovirt:
@hongyung This is a two year old PR, I think it's better to open a new issue about the bug.

PR #2749, opened 2017-07-13
Fix for broken revprop processing

robander:
## Description

If I have a revision that is set to `flag` but does not specify an image, it results in this failure from the flag module:
```
[preprocess_flag] Failed to transform document: A value must be supplied for the
 parameter because the default value is not a valid instance of the required type
```

This occurs because when the flag module was updated to add typing to parameters and variables, two parameters for `default-rev-start` and `default-rev-end` were set to `xs:string`, but our code for handling default revisions calls those templates without any parameters (the 2 `with-param` elements are actually there but commented out).

## Motivation and Context

The default start/end templates are there so that a process can add default start/end formatting (we've typically added a start-rev and end-rev graphic). By default they do nothing. Because the default parameters do not match the expected parameters, the whole module fails, and no flagging is added.

## How Has This Been Tested?

This issue highlighted that we do not actually have any integration tests that use revision flags, so I've updated an existing test for `<prop>` flags so that it also tests `<revprop>` flags. The test now includes full default (flagged but no action), flag with color (no image) and flag with specified image.

## Type of Changes

In addition to the updated test case, this fix changes 4 instances of `xs:string` so that they are now `xs:string?`.

An alternate fix would be to un-comment the sections that pass `$lang` and `$bidi`. I'm not sure which fix is better, or if it matters. Those have been commented out for a long time, so not knowing the reason, I left them commented out.

The fix is intended for `hotfix/2.5.2` but opened against `master` because that branch doesn't yet exist.

### DIFF
--- a/src/main/xsl/preprocess/flagImpl.xsl
+++ b/src/main/xsl/preprocess/flagImpl.xsl
@@ -277,8 +277,8 @@ LOOK FOR FIXME TO FIX SCHEMEDEF STUFF
        Commented out section shows how to insert a default start flag of "delta.gif"
        by creating the proper ditaval syntax -->
   <xsl:template name="default-rev-start">
-    <xsl:param name="lang" as="xs:string"/>
-    <xsl:param name="biditest" as="xs:string"/>
+    <xsl:param name="lang" as="xs:string?"/>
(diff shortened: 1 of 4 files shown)


jelovirt:
Changed the base to `hotfix/2.5.2`

PR #2751, opened 2017-07-13
Remove obsolete test cases

robander:

We have a lot of test cases that never actually run. (I discovered with #2749 that our test directory contains several DITAVAL files that could have caught the failure, but did not run.) In addition to wasting space / wasting time copying these files during our test process, these are misleading because it makes our test coverage look more complete than it is.

The ignored tests should be removed - most of them are from early days of DITA-OT and do not fit modern test methods. They either duplicate existing tests that are running, or they test so many things at once that a failure does not clearly identify the problem. For many of these tests, the purpose of the test is no longer clear -- for example, a very large test document may contain one element that failed in DITA-OT 1.3.1, but there is no indication what caused the problem at that point.

In the rare case that the purpose of a test is clear, it should be cleaned up to test that single point of failure and added back in.

Opening this now with the first set of deleted tests, and will add more as I make my way through.


robander:
Lots of commits today, so far I'm trying to do only a few tests at a time to make it easier to track. Main goal of the day has been to go through old test cases with test or directory names that have no relation to the test -- such as `1.5.2_M4_BUG3052904`, which is based on old SourceForge bug report 3052904 (link no longer available) and was either found or fixed in the 4th milestone build of 1.5.2. I've renamed those tests to clarify the purpose, while cleaning up some markup and removing some files that are unrelated to the actual tested condition. 

It's likely (especially with keyref) that some of these duplicate other tests I haven't gotten to, but it should be easier to find those now that these tests all clearly indicate when they are related to keyref.

jelovirt:
Ready for merge?

robander:
Yes, I think it's ready for merge. I'd still like to add **new** tests but should be a different pull request -- leave this one for the clearing out.

Issue #2759, opened 2017-08-01
Using "onlytopic.in.map" gives errors, reads directory


robander:
## Expected Behavior
I have a map, which references a single topic, which references another topic with conref.

In the file that is the target of the conref, I have several cross references (marked with `scope="peer"`), more conref references (to the same file but using strange syntax to leave the file and return: `conref="../buildfiles/conref_api_overview.dita##conref_api_overview/test"`), and related links (also peer).

When I build this map using default options, the log is clean. I'd expect the same behavior when running with `onlytopic.in.map=true`, which is our local default so that we don't generate extra files.

## Actual Behavior

When I build with `onlytopic.in.map=true` (our local default), I get a bunch of errors. 1) I get errors about "ignoring copy-to" (this is incorrect, copy-to is not used; these are just peer HTML files or peer DITA topics). 2) I get `Content is not allowed in prolog` when a directory name is treated as a file and read by `debug-filter`, by `topicpull`, and more.

```
debug-filter:
   [filter] Using Xerces grammar pool for DTD and schema caching.
   [filter] Ignoring a copy-to file file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/com.ibm.1.doc/ssl.dita
   [filter] Processing file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/debug.ditamap to file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/temp/debug.ditamap
   [filter] Ignoring a copy-to file file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/com.ibm.2.doc/3.html
   [filter] Ignoring a copy-to file file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/com.ibm.1.doc/1.html
   [filter] Processing file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/com.ibm.abcd.install.doc/ to file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/temp/com.ibm.abcd.install.doc
   [filter] Content is not allowed in prolog.
   [filter] Ignoring a copy-to file file:/C:/Users/IBM_ADMIN/github/robander/dita-ot/src/main/test/gary/com.ibm.2.doc/2.html
[...]

PR #2761, opened 2017-08-03
Do not add peer references to resource-only collection #2759

robander:

Provides a fix for #2759.

With 2.5 we collect non-topicref references so that they can be avoid generating output (in combination with the `onlytopic.in.map` and `generate.copy.outer` parameters). 

When those references are marked with `scope="external"` we ignore them. If they are marked with `scope="peer"` we still add them to the list of referenced files, causing later steps to try reading them (even when they are not DITA).

In other cases we avoid adding / tracking both external and peer; adding the test for peer in this case prevents numerous errors from coming up later in the build / prevents later DITA-OT steps from trying to read peer or non-DITA files as if they are local DITA.

## How Has This Been Tested?

1. Tested against the original files that demonstrated the error
2. Tested against unit / integration tests

### DIFF
--- a/src/main/java/org/dita/dost/reader/GenListModuleReader.java
+++ b/src/main/java/org/dita/dost/reader/GenListModuleReader.java
@@ -405,7 +405,7 @@ public final class GenListModuleReader extends AbstractXMLFilter {
 
         final URI href = toURI(atts.getValue(ATTRIBUTE_NAME_HREF));
         final String scope = atts.getValue(ATTRIBUTE_NAME_SCOPE);
-        if (href != null && !ATTR_SCOPE_VALUE_EXTERNAL.equals(scope)) {
+        if (href != null && !ATTR_SCOPE_VALUE_EXTERNAL.equals(scope) && !ATTR_SCOPE_VALUE_PEER.equals(scope)) {
             if (!(MAP_TOPICREF.matches(cls))) {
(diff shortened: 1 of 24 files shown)


robander:
Note - this passed Travis with my original branch `robander/hotfix/onlytopic`, which was somewhat surprising, but failed here because 2.5.x builds are still running + usually failing on the doc build.

infotexture:
@robander I created the `hotfix/2.5.3` branch from the tip of `master` and updated this PR to use that as the base branch.

robander:
Just pushed up a test case that tests both `onlytopic.in.map=true` and `onlytopic.in.map=false`. We had one test with this parameter in the integration test, but it didn't actually do anything useful (the topics only referenced an image, rather than other topics).

From my testing, it looks like when the integration test compares XHTML, it only compares files that exist in both the expected result directory `exp/` and in the generated directory. So, it's not enough to simply have or not have a file in `exp/`. If a file is generated when it shouldn't be -- but doesn't exist in `exp/` -- the test passes because no comparison is performed. 

To get around this, the test I created for `onlytopic.in.map=true` has a couple of dummy HTML files to ensure linked files are _not_ generated. If the build changes so that they are generated, the actual generated file will not match the dummy file in `exp/`, and the test will fail. This feels odd but seems to work well. The test case for `onlytopic.in.map=true` has the dummy files, so it will fail if the option breaks again in the future, while the test case for `onlytopic.in.map=false` has the actual generated versions.

PR #2772, opened 2017-09-06
Fix variable failure from xref without format

robander:

I have an external link without a specified `@format`. This is to a DITA topic so defaulting to `dita` is fine for the value.

The topicpull step sets format with a variable, and expects an `xs:string` result:
```
<xsl:variable name="format" as="xs:string"
           select="dita-ot:get-link-format(.)"/>
```

However, the function call does not pass in a default, and the function itself (with or without the default) is set to return optional `xs:string?`. My test case below returns an empty value, and results in the build error:
`[topicpull] Failed to transform document: An empty sequence is not allowed as the value of variable $format`

The fix just updates the variable to expect an optional value, which is what the function already declares.

Test document:
```
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN"  "topic.dtd">
<topic id="topic" xml:lang="en-us">
<title>Test external DITA xref</title>
<body>
<p>Try this out: <xref
href="https://www.ibm.com/support/knowledgecenter/SSFHJY_1.0.6/using/text/document/d_changing_the_default_file_format.dita"
scope="external"/></p>
</body>
</topic>
```

### DIFF
--- a/src/main/xsl/preprocess/topicpullImpl.xsl
+++ b/src/main/xsl/preprocess/topicpullImpl.xsl
@@ -593,9 +593,9 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
       </xsl:when>
       <xsl:otherwise>
         <!-- Scope must be peer or external or format is not dita, use any local text. -->
-        <xsl:variable name="format" as="xs:string"
+        <xsl:variable name="format" as="xs:string?"
           select="dita-ot:get-link-format(.)"/>
-        <xsl:variable name="scope" as="xs:string"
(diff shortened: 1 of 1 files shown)


jelovirt:
The PR should go to either a bug fix branch or develop, not master. Please change the target branch.

robander:
Yeah, it was intended for 2.5.4 but we don't have that yet so I did it against master. I need to create the branch and fix that...

jelovirt:
Created `hotfix/2.5.4` and changed the PR to target that.

PR #2782, opened 2017-09-13
Warn about broken tables, and avoid FOP crash

robander:

## Description
We found an old DITA topic that included an invalid table -- the `<tgroup>` element used `@cols="3"`, indicating three columns, but somehow an author added a fourth column. HTML output tolerates the markup and everything appears. When processing the topic for PDF, FOP crashes with a Java error. Antenna House returns a PDF, but the text of the "extra" column appears to the right of the table, in a one-character column.

The fix adds an error message when PDF2 code detects an entry beyond the specified number of columns. It operates based off of the generated `@dita-ot:x` and `@dita-ot:morecols` attributes added during `preprocess`. It's possible (but difficult) to get a table without those, so the code falls back to current behavior if they are not specified. When the last column of the current entry is beyond the specified number of columns, a new error message is generated. In the FOP override, the entry is dropped after generating that message.

FWIW, having too *few* entries in a row does not break either Antenna House or FOP, and you end up with an empty cell in the PDF.

## How Has This Been Tested?

Integration tests pass. The table used to test the specific fix is a trimmed down version of the original topic. It tests a row with an extra cell (no spanning), and a row that has an extra cell due to spanned columns.

```xml
<!DOCTYPE task PUBLIC "-//OASIS//DTD DITA Task//EN"
 "task.dtd">
<task id="wonkywonk" xml:lang="en-us">
<title>Wonky Table</title>
<shortdesc>The tables are wonky. REALLY WONKY!</shortdesc>
<taskbody>
<result>
<table frame="all" id="table_vt3_ll5_r4">
<title>Extra cell</title>
<tgroup cols="2">
<colspec colname="c1" colnum="1" colwidth="1.0*"/>
<colspec colname="c2" colnum="2" colwidth="1.0*"/>
<thead>
<row>
<entry>head!</entry>
<entry>Head!</entry>
</row>
</thead>
<tbody>
<row>
<entry>a</entry>
<entry>b</entry>
</row>
<row>
<entry>a</entry>
<entry>b</entry>
<entry>THIS IS EXTRA</entry>
</row>
<row>
<entry>a</entry>
<entry>b</entry>
</row>
</tbody>
</tgroup>
</table>
<table frame="all" rowsep="1" colsep="1" id="table_qg3_kpb_fbb">
<title>Extra cell due to spans</title>
<tgroup cols="2">
<colspec colname="c1" colnum="1" colwidth="1*"/>
<colspec colname="c2" colnum="2" colwidth="1*"/>
<thead>
<row>
<entry>header1</entry>
<entry>header2</entry>
</row>
</thead>
<tbody>
<row>
<entry>normal</entry>
<entry>normal</entry>
</row>
<row>
<entry namest="c1" nameend="c2">spanned cols = next cell is invalid</entry>
<entry>This is invalid</entry>
</row>
<row>
<entry>normal</entry>
<entry>normal</entry>
</row>
</tbody>
</tgroup>
</table>
</result>
</taskbody>
</task>
```

Not sure how best to add to test suite due to general lack of PDF-specific test cases.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
@@ -216,6 +216,19 @@ See the accompanying LICENSE file for applicable license.
                                                     '+-0123456789.'))"/>
   </xsl:function>
   
+  <xsl:function name="dita-ot:get-entry-end-position" as="xs:integer">
+    <xsl:param name="thisentry" as="element()"/>
+    <xsl:choose>
+      <xsl:when test="$thisentry/@dita-ot:x and $thisentry/@dita-ot:morecols">
+        <xsl:sequence select="xs:integer($thisentry/@dita-ot:x) + xs:integer($thisentry/@dita-ot:morecols)"/>
+      </xsl:when>
+      <xsl:when test="$thisentry/@dita-ot:x">
+        <xsl:sequence select="xs:integer($thisentry/@dita-ot:x)"/>
+      </xsl:when>
+      <xsl:otherwise>0</xsl:otherwise>
+    </xsl:choose>
+  </xsl:function>
+  
   <xsl:function name="dita-ot:get-unit">
     <xsl:param name="width" as="xs:string"/>
     <xsl:value-of select="normalize-space(translate($width,
@@ -399,7 +412,16 @@ See the accompanying LICENSE file for applicable license.
         </fo:table-row>
     </xsl:template>
 
+    <xsl:template match="*[contains(@class, ' topic/entry ')]" mode="validate-entry-position">
+        <xsl:if test="dita-ot:get-entry-end-position(.) gt number(ancestor::*[contains(@class,' topic/tgroup ')][1]/@cols)">
+            <xsl:call-template name="output-message">
+                <xsl:with-param name="id" select="'PDFX012E'"/>
+            </xsl:call-template>
+        </xsl:if>
+    </xsl:template>
+
     <xsl:template match="*[contains(@class, ' topic/thead ')]/*[contains(@class, ' topic/row ')]/*[contains(@class, ' topic/entry ')]">
+        <xsl:apply-templates select="." mode="validate-entry-position"/>
         <fo:table-cell xsl:use-attribute-sets="thead.row.entry">
             <xsl:call-template name="commonattributes"/>
             <xsl:call-template name="applySpansAttrs"/>
@@ -414,6 +436,7 @@ See the accompanying LICENSE file for applicable license.
     </xsl:template>
 
     <xsl:template match="*[contains(@class, ' topic/tbody ')]/*[contains(@class, ' topic/row ')]/*[contains(@class, ' topic/entry ')]">
+        <xsl:apply-templates select="." mode="validate-entry-position"/>
         <xsl:choose>
             <xsl:when test="ancestor::*[contains(@class, ' topic/table ')][1]/@rowheader = 'firstcol'
                         and empty(preceding-sibling::*[contains(@class, ' topic/entry ')])">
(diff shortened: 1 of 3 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -399,7 +412,16 @@ See the accompanying LICENSE file for applicable license.
         </fo:table-row>
     </xsl:template>
 
+    <xsl:template match="*[contains(@class, ' topic/entry ')]" mode="validate-entry-position">
+        <xsl:if test="dita-ot:get-entry-end-position(.) &gt; number(ancestor::*[contains(@class,' topic/tgroup ')][1]/@cols)">
```
jelovirt: Use `gt` instead of `>`

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -216,6 +216,19 @@ See the accompanying LICENSE file for applicable license.
                                                     '+-0123456789.'))"/>
   </xsl:function>
   
+  <xsl:function name="dita-ot:get-entry-end-position">
```
jelovirt: Add `as="xs:integer"`

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -216,6 +216,19 @@ See the accompanying LICENSE file for applicable license.
...
+    <xsl:param name="thisentry" as="element()"/>
+    <xsl:choose>
+      <xsl:when test="$thisentry/@dita-ot:x and $thisentry/@dita-ot:morecols">
+        <xsl:value-of select="number($thisentry/@dita-ot:x) + number($thisentry/@dita-ot:morecols)"/>
```
jelovirt: Use `xsl:sequence` instead

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
robander: Not sure what device you're viewing this on, but I wonder if it's converting -- I've currently got the `&gt;` entity rather than the greater-than character `>` ...

xephon2: Just `gt` not `&gt;`

jelovirt: I mean use the `gt` operator from XPath 2.0 instead of `>` operator from XPath 1.0

robander: Ahh of course.

robander:
Made the update to add `xs:integer` and build now complains about item having type `xs:double`, so not yet ready for merge.

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -216,6 +216,19 @@ See the accompanying LICENSE file for applicable license.
...
+  <xsl:function name="dita-ot:get-entry-end-position" as="xs:integer">
+    <xsl:param name="thisentry" as="element()"/>
+    <xsl:choose>
+      <xsl:when test="$thisentry/@dita-ot:x and $thisentry/@dita-ot:morecols">
+        <xsl:sequence select="xs:integer($thisentry/@dita-ot:x) + xs:integer($thisentry/@dita-ot:morecols)"/>
+      </xsl:when>
+      <xsl:when test="$thisentry/@dita-ot:x">
+        <xsl:sequence select="xs:integer($thisentry/@dita-ot:x)"/>
+      </xsl:when>
+      <xsl:otherwise>0</xsl:otherwise>
```
jelovirt: Saxon probably doesn't care, but I would prefer to use

```xml
<xsl:sequence select="0"/>
```

To make it easier to read that we're returning an integer instead of text node with content "0".

Issue #2781, opened 2017-09-12
Separate configuration to own directory


jelovirt:
Currently configuration files that are modified by integration are located in `src/main/resources` and they will end up in `dost.jar`; some configuration files are also bundled into `dost-configuration.jar` by the integration process. This will cause unstable behaviour because classpath order will define which will take precedence.

All configuration files that are generated or modified by integration or user should be located in a separate `config` directory and only static configuration should be kept in `resources` directory.

robander:
We ran into this over the last few days and it took quite a while to debug. It appears that the class loader started picking up `resources/messages.xml` out of `dost.jar`, instead of from `dost-configuration.jar` or from the literal copy in the installed directory. As a result, when we tried to generate a message added by a plugin, the build failed due to an undefined message; only those messages defined+integrated at build time were part of the loaded copy in `dost.jar`.

robander:
Tangentially related -- the message task fails (build halts) if you try to look up a message ID that doesn't exist. Meaning if we try to load a meaningful message ("hey your build has problems with X") but can't get that message, we halt with "Cannot load message ID xyz". Is that ... the best design? Or should the exception at that point print out "Cannot load message xyz" but then allow the build to continue? In my case - it was actually just an info message, to add something useful in the log, so halting the build over a missing info message seemed pretty harsh.

So I'm wondering - is it worth opening another issue for that tangential issue, to change the design?

PR #2783, opened 2017-09-13
Separate generated and modifiable configuration into own directory #2781

jelovirt:
Add `config` directory for generated and modifiable configuration files.


jelovirt:
Fixed breaking test

PR #2784, opened 2017-09-16
Allow any attribute to be used for profiling

jelovirt:
Modify DITAVAL reader and processing to allow using any attribute for filtering and flagging.

Spec compliant DITA implementation will only allow filtering and flagging by profiling attributes or specializations of `@props`. However, in some cases users should be allowed to filter by e.g. `@importance` or namespaced extension attributes.

Because this adds functionality that's explicitly not allowed by DITA 1.3, this has been implemented as an opt-in functionality. Users must add `filter-attributes` and `flag-attributes` properties to configuration to specify which attributes can be used in filtering and flagging, respectively, in addition to the spec defined attributes. The value of the properties is a comma separated list of attribute QNames in Clark notation.

For example, the following is added to `config/configuration.properties`:

```properties
filter-attributes = importance, status
flag-attributes = {http://www.cms.com/}review
```

This will allow using `@importance` and `@status` in filtering, and custom `@cms:review` in flagging.

### DIFF
--- a/src/main/config/configuration.properties
+++ b/src/main/config/configuration.properties
@@ -6,7 +6,8 @@
 #DITA-OT configuration properties
 default.cascade = merge
 temp-file-name-scheme = org.dita.dost.module.GenMapAndTopicListModule$DefaultTempFileScheme
-filter-any-attribute = false
+#filter-attributes =
+#flag-attributes =
 cli.color = true
 
 # Integration
--- a/src/main/config/configuration.properties
+++ b/src/main/config/configuration.properties
@@ -6,6 +6,7 @@
 #DITA-OT configuration properties
 default.cascade = merge
 temp-file-name-scheme = org.dita.dost.module.GenMapAndTopicListModule$DefaultTempFileScheme
+filter-any-attribute = false
 cli.color = true
 
 # Integration
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -353,6 +353,11 @@ See the accompanying LICENSE file for applicable license.
     <response>Using 'scope' attribute value 'external'.</response>
   </message>
 
+  <message id="DOTJ074W" type="WARN">
+    <reason>Rev attribute cannot be used with prop filter.</reason>
+    <response></response>
+  </message>
+
   <!-- End of Java Messages -->
     
   <!-- Start of XSL Messages -->  
--- a/src/main/java/org/dita/dost/module/MergeDitavalModule.java
+++ b/src/main/java/org/dita/dost/module/MergeDitavalModule.java
@@ -136,16 +136,25 @@ public final class MergeDitavalModule extends AbstractPipelineModuleImpl {
 
     }
 
-    /**
-     * TODO: This should be generalized to be a DOM to XML stream utility function and moved to XMLUtils.
-     */
     private void writeConditions(XMLStreamWriter export, NodeList conditionElements, URI ditavalDirectory) {
         try {
             for (int i = 0; i < conditionElements.getLength(); i++) {
-                switch (conditionElements.item(i).getNodeType()) {
+                final Node node = conditionElements.item(i);
+                switch (node.getNodeType()) {
                 case Node.ELEMENT_NODE:
-                    export.writeStartElement(conditionElements.item(i).getNodeName());
-                    final NamedNodeMap atts = conditionElements.item(i).getAttributes();
+                    final Element elem = (Element) node;
+                    export.writeStartElement(node.getNodeName());
+                    final NamedNodeMap atts = node.getAttributes();
+                    for (int j = 0; j < atts.getLength(); j++) {
+                        final String value = atts.item(j).getNodeValue().trim();
+                        if (value.indexOf(':') != -1) {
+                            final String prefix = value.substring(0, value.indexOf(':'));
+                            final String ns = elem.lookupNamespaceURI(prefix);
+                            if (ns != null) {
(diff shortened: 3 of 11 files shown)


robander:
I will admit that I'm very nervous about allowing `@xml:lang` to filter. I don't like it. 

In DITA 1.0 the spec was clear -- only filter with `audience` / `product` / `platform` / `otherprops`. I din't question that much, and enforced it because the spec said so. 

For almost any other attribute in DITA, I've since come around to "if people want to try filtering on this, fine, who am I to judge". If somebody wants to filter out anything with `status="deprecated"`, go for it. If you want to filter on `outputclass="flashy"`, I'd question the design, but don't see a reason to disallow.

With language ... it's just a property of the content -- the content is / is not written in the specified language. If you decide you want a paragraph to stay in / stay out, the language doesn't change, while filtering values _generally_ tend to be turned on / off if the content becomes relevant.

In every case where I've actually seen the filtering request, the actual goal is to filter for a specific locale -- "This is going to be published in Japan, so I want to filter out everything else". From there is very easy to use the markup improperly -- marking English content as Japanese because you want it published in Japan, or vice versa. Given the difficulties I've seen so many writers have with translated content, this scares me. (Every couple of months I get a defect report because a language isn't properly tagged, they don't know how to fix it, and they can't figure out why Japanese tables are getting a "Table X" heading in English.)

So yeah ... that one makes me nervous. With `@rev` I understand the original rationale to disallow, but don't really care anymore. With `@cms:confidentiality`, I'm fine with it, even though the spec would push you to declare that as a specialization of `@props`. As mentioned above, if you want to turn on filtering for the existing `@status`, `@outputclass`, even `@collection-type`, I no longer have any argument.

drmacro:
I agree with Robert on xml:lang--the potential for abuse and getting tied into knots is very high. 

If you want to filter on locale then props="(locale jp)" will do it and not risk interference with normal use of xml:lang.

robander:
Slight update to Eliot's post -- and actually this might be a good tip in general -- you should be able to filter with 
`props="locale(jp)"`
That is the syntax for a generalized filter attribute -- effectively meaning that you have a new `locale` attribute, with value of `jp`. 

This will work today, by setting an include/exclude rule for `locale="jp"` in your DITAVAL file. I've just tested to be sure, using DITA-OT 2.5.3, and successfully removed both a branch of a map and a paragraph from a topic.

jelovirt:
The original goal was not to specifically enable `@xml:lang` filtering, but to generalize the code so that we don't hard-code the list of elements we use for filtering or flagging. In practice, we still have a list of DITA spec specified profiling base classes, but this will allow adding attributes to that list. This will allow e.g. the aforementioned filtering on `@outputclass` or `@rev`. While new profiling attributes should be added via attribute specializations, CMS occasionally insert their own attributes that are namespaced and not part of the specializations. This was the reason for enabling the example `@cms:confidentiality`.

This is a extension feature that is about pragmatic use of the tool in cases where a lone DITA writer needs to get the publication out without a lot of manual work for a one-off publishing need.

raducoravu:
As long as the end user has to manually set up the configuration to filter by xml:lang I do not see a problem with this. Maybe the example in the documentation could be for filtering using the rev attribute and the CMS specific attribute so that we do not give people the idea that filtering by xml:lang is some usual usecase.

drmacro:
If there are going to be examples they should not be for @rev or @xml:lang. 

Filtering on attributes in a namespaces seems perfectly reasonable, or filtering on @status or @importance.

The spec's prohibition against filtering on @rev and @xml:lang are directly related to the unique semantics of those two attributes and the potential for problems that stem from trying to use those particular attributes to do something they cannot do in the general case.

In the case of @rev it cannot replace a real version control system.

In the case of @xml:lang it cannot reflect both language and delivery locale or region. It is this likely and potential for conflation, especially by users in regions where language and locale are usually the same (e.g., U.S. users), that the spec is trying to avoid.

robander:
Agree with what @drmacro said above.

Specifically, `@rev` is intended to indicate a specific revision of the document, *not* (as would commonly be used for filtering) the product revision to which something applies. There's quite a difference here -- "This content was added in version X" is what the spec is based on today, along with the expectation that you might need to flag content that changes in a revision. With filtering, the intent is generally to say "This content applies to product version X, even if added it at some other point". Of course the spec cannot prevent people from using it that way, but it would conflict with the purpose of the attribute, and our docs should not highlight this (technically invalid) usage.

Same for `@xml:lang` -- the attribute explicitly means "This content is language X", not "This content should only be read when I am publishing for locale X". While they very often overlap, they will not always, which is the basis for our concern here. There is so much processing associated with `@xml:lang` (both in and out of DITA-OT) based on the explicit XML definition, that our docs should not advertise an alternate usage likely to conflict with it.

Basically I guess I'm just repeating what Eliot said for emphasis ... I'm also happy to see examples using `@status`, `@importance`, or namespaced local extensions.

jelovirt:
Changed the description to use `@importance` and namespaced extension attribute as an example.

robander:
FWIW discussed some of this further at the TC today. Several of us seemed inclined to (at least) relax the language for 2.0 to at least say implementations SHOULD support filtering for the core set of values (currently says to just support the base set), without explicitly ruling any others in or out. The SHOULD  leaves rooms for applications to support filtering on other attributes (currently language is a lower-case "must" which isn't normative but can be read that way).

raducoravu:
@infotexture is this documented anywhere?

raducoravu:
I added this issue to document the feature: https://github.com/dita-ot/docs/issues/420

PR #2789, opened 2017-09-19
Support language-independent variable files

robander:

## Description
Adds support for no-language / language-independent variable files. Intent is to use this for variables that do not actually use any localized text. This can be either:

- variables that are empty by default (such as the current `Glossary odd footer`, empty in all 48 default string files in the PDF dir)
- variables that are common to all languages, where the default variable does not actually contain localized text (such as `Step Number`, which defaults to `<param ref-name="number"/>. `, or `Glossary odd header`, which has 3 parameters separated by `|`)

The current design uses all of the current logic, which will look for more specific languages, falling back to more general (`fr-ca` falls back to `fr`), then to the configured default, and only then will look for a "null language" string. 

## Motivation and Context

- Today, we ship 48 variable files in the PDF directory, 55 in `xsl/common` (the same 48 plus slight variants). For common variables, we're maintaining 48 or 55 copies of the same thing. With this change, we only maintain one.
- Locally, across the default + my 4 custom PDF styles, I have 165 definitions of `<variable id="Glossary odd footer">` (and most other common variables). Within each style, these are common, so I really only need 5 definitions - one per style, in the empty-language set. This enhancement would significantly ease the pain of adding / modifying common variables.
- I've also added my own common variables, such as `<variable id="Chapter number with title">`, which just lay out passed-in variables. It would be much simpler if I could do this by adding one common variable.
- Some common variables have been overlooked in the core. For example, we actually maintain 47 copies of `Step Number`, because it was somehow lost from Serbian, which means Serbian tasks will warn you that they're falling back to English. Making these common variables truly common ensures that doesn't happen. For strings that actually need text, any missing variable still falls back to the default and generates a "using default language" error, as it should.

There are a few open questions with this design that I'd like to address before I go any further.

1. Current design expects that the no-language elements are identified with `xml:lang=""`. It could instead look for `empty(@xml:lang)`, or for both. I'm not sure which approach is best; I'd probably do both because people will try both, except that I generally don't like defining 2 ways to do something.
1. If we have a string that is common to 46 languages - maybe all LTR languages - but differs for 2 others, is it best to define in the null space + the 2 exceptions? Pro: only 3 total definitions. Con: if another exception comes along, it _might_ be missed because having the null value means you won't get a warning. (Note: I expected to find some of these in our default PDF2 strings but have not yet.)
1. Overall goal is to move existing common strings like `Step Number` to a common set of variables, but I'm holding off until I get feedback. Is that a wise goal? It has all the advantages above. But, it at least opens the possibility that you won't notice those common strings when it's time to localize. That is, if you always saw `Index even header` in the existing files, and updated it, you may no longer even notice that it exists. But ... I don't think this is a problem, because for strings like that, you probably have your own matching copies just like DITA-OT does, so should "localize" the null set as well.

## How Has This Been Tested?
No impact on current tests / integration tests because the null space is only checked _after_ all currently working strings are tested.

Tested locally by defining a string set with an empty language. Added the following to `xsl/common/common-strings.xml`:
`<lang xml:lang="" filename="strings-emptylang.xml"/>`

In that file, defined a variable:
```
<strings xml:lang="">
<str name="test-emptylang">Test if I can get a string with an empty xml:lang</str>
</strings>
```

And successfully retrieved that variable as part of an HTML build with:
```
<div>
<xsl:call-template name="getVariable">
<xsl:with-param name="id" select="'test-emptylang'"/>
</xsl:call-template>
</div>
```

Similarly, I tested adding a new variable set to the common PDF strings, with `xml:lang=""`, and in that set defined:
`<variable id="emptyvar">[EMPTY]<param ref-name="stringy"/>[EMPTY]</variable>`

I successfully retrieved that variable using:
```
<xsl:call-template name="getVariable">
  <xsl:with-param name="id" select="'emptyvar'"/>
  <xsl:with-param name="params" as="element()*">
    <stringy>This is a test of the string</stringy>
  </xsl:with-param>
</xsl:call-template>
```

## Type of Changes

This is a new enhancement to the variable localization code.

Existing custom code and existing variables will work without modification. (Edge case - anyone who overrides `getVariable` will need to update, but hopefully that is not done today.)

Those adjusting variable files will need to be aware that common variables have been consolidated.

## Checklist

## Still to do

Pending feedback - next step is to consolidate common PDF strings. I'm not sure exactly how many of these exist (guessing 20 or 30). In the `xsl/common/` string files, we only have a few common values, but they should probably be consolidated. Some depend on the answer to question 2 above ("ColonSymbol" is common to 52 string files, but 3 French variants differ).

Also need to verify whether existing tests catch one or more of the variables that move to the consolidated file; if not, one should be added.

### DIFF
--- a/src/main/xsl/common/dita-utilities.xsl
+++ b/src/main/xsl/common/dita-utilities.xsl
@@ -62,7 +62,7 @@ See the accompanying LICENSE file for applicable license.
     </xsl:choose>
   </xsl:template>
 
-  <xsl:variable name="stringFiles" select="document($variableFiles.url)/langlist/lang" as="element(lang)*"/>
+  <xsl:variable name="variableFiles" select="document($variableFiles.url)/langlist/lang" as="element(lang)*"/>
   
   <!-- Deprecated. Use getVariable template instead. -->
   <xsl:template name="getString">
@@ -94,15 +94,15 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="l" select="($ancestorlang, $defaultlang)[1]" as="xs:string?"/>
     <xsl:choose>
       <xsl:when test="exists($l)">
-        <xsl:variable name="stringfile" select="$stringFiles[lower-case(@xml:lang) = lower-case($l)]/@filename" as="xs:string*"/>
-        <xsl:variable name="str" as="element()*">
-          <xsl:for-each select="$stringfile">
-            <xsl:sequence select="document(., $stringFiles[1])/*/*[@name = $id or @id = $id]"/><!-- strings/str/@name opentopic-vars:vars/opentopic-vars:variable/@id -->
+        <xsl:variable name="variablefile" select="$variableFiles[lower-case(@xml:lang) = lower-case($l)]/@filename" as="xs:string*"/>
+        <xsl:variable name="variable" as="element()*">
+          <xsl:for-each select="$variablefile">
+            <xsl:sequence select="document(., $variableFiles[1])/*/*[@name = $id or @id = $id]"/><!-- strings/str/@name opentopic-vars:vars/opentopic-vars:variable/@id -->
           </xsl:for-each>
         </xsl:variable>
         <xsl:choose>
-          <xsl:when test="exists($str)">
-            <xsl:apply-templates select="$str[last()]" mode="processVariableBody">
+          <xsl:when test="exists($variable)">
+            <xsl:apply-templates select="$variable[last()]" mode="processVariableBody">
               <xsl:with-param name="params" select="$params"/>
             </xsl:apply-templates>
             <xsl:if test="empty($ancestorlang)">
@@ -124,11 +124,26 @@ See the accompanying LICENSE file for applicable license.
         </xsl:choose>
       </xsl:when>
       <xsl:otherwise>
-        <xsl:value-of select="$id"/>
-        <xsl:call-template name="output-message">
-          <xsl:with-param name="id" select="'DOTX052W'"/>
-          <xsl:with-param name="msgparams">%1=<xsl:value-of select="$id"/></xsl:with-param>
-        </xsl:call-template>
+        <xsl:variable name="variablefile" select="$variableFiles[@xml:lang='']/@filename" as="xs:string*"/>
+        <xsl:variable name="variable" as="element()*">
+          <xsl:for-each select="$variablefile">
+            <xsl:sequence select="document(., $variableFiles[1])/*/*[@name = $id or @id = $id]"/>
+          </xsl:for-each>
+        </xsl:variable>
+        <xsl:choose>
+          <xsl:when test="exists($variable)">
+            <xsl:apply-templates select="$variable[last()]" mode="processVariableBody">
+              <xsl:with-param name="params" select="$params"/>
+            </xsl:apply-templates>
+          </xsl:when>
+          <xsl:otherwise>
+            <xsl:value-of select="$id"/>
+            <xsl:call-template name="output-message">
+              <xsl:with-param name="id" select="'DOTX052W'"/>
+              <xsl:with-param name="msgparams">%1=<xsl:value-of select="$id"/></xsl:with-param>
+            </xsl:call-template>            
+          </xsl:otherwise>
+        </xsl:choose>
       </xsl:otherwise>
     </xsl:choose>
   </xsl:template>
--- a/src/main/plugins/org.dita.pdf2/cfg/common/vars/ar.xml
+++ b/src/main/plugins/org.dita.pdf2/cfg/common/vars/ar.xml
@@ -21,91 +21,11 @@ See the accompanying LICENSE file for applicable license.
        Strings used in the headers and footers of pages.
     -->
 
-    <!-- Product name to be placed inside headers etc. -->
-    <variable id="Product Name"/>
-
-    <!-- The footer that appears on odd-numbered pages. -->
-    <variable id="Body odd footer"/>
-
-    <!-- The footer that appears on even-numbered pages. -->
-    <variable id="Body even footer"/>
-
-    <!-- The footer that appears on the first page of a chapter. -->
-    <variable id="Body first footer"/>
-
-    <!-- The header that appears on the first page of a chapter. -->
-    <variable id="Body first header"/>
-
-    <!-- The header that appears on odd-numbered pages. -->
-    <variable id="Body odd header"><param ref-name="prodname"/> | <param ref-name="heading"/> | <param ref-name="pagenum"/></variable>
-
-    <!-- The header that appears on even-numbered pages. -->
-    <variable id="Body even header"><param ref-name="pagenum"/> | <param ref-name="heading"/> | <param ref-name="prodname"/></variable>
-
-    <!-- The footer that appears on Preface odd-numbered pages. -->
-    <variable id="Preface odd footer"/>
-
-    <!-- The footer that appears on Preface even-numbered pages. -->
-    <variable id="Preface even footer"/>
-
-    <!-- The footer that appears on the first page of a Preface -->
-    <variable id="Preface first footer"/>
-
     <!-- The header that appears on Preface odd-numbered pages. -->
     <variable id="Preface odd header"><param ref-name="prodname"/> | تمهﻳد | <param ref-name="pagenum"/></variable>
 
     <!-- The header that appears on Preface even-numbered pages. -->
     <variable id="Preface even header"><param ref-name="pagenum"/> | تمهﻳد | <param ref-name="prodname"/></variable>
-
-    <!-- The header that appears on Preface first page. -->
-    <variable id="Preface first header"/>
-
-    <!-- The footer that appears on the odd-numbered table of contents
-         pages. --> 
-    <variable id="Toc odd footer"/>
-
-    <!-- The footer that appears on the even-numbered table of contents
-         pages. --> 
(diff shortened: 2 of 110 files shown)


robander:
I know at some point @jelovirt pointed out that even the definitions of our variables are duplicated -- we have a whole bunch of variable files that all say 
```
<!-- TODO: Default title text for referenced list item. Used to generate
         reference text in case reference has no explicit content and 
         reference target is a list item. -->
```

They're not in every file; many languages cleaned all of those out when translations were added, but many did not. It probably makes sense to get rid of the extra definitions as part of this, leaving the full description in just the default English file. (There might be a better long term spot for them, but at least reducing to one copy is a good start.)

robander:
(Also currently working on a related pull request, which should show up tonight or tomorrow, to remove strings marked "deprecated" from the PDF variable files - but that's a distinct issue that should go in first.)

jelovirt:
Per [ 
Tagging text with no language](https://www.w3.org/International/questions/qa-no-language), `xml:lang=""` means text language is undetermined and `xml:lang="zxx"` means text is known to be not in any language. I don't know if we need to make this distinction in the variable configuration and just go with `xml:lang=""` because it's more simple.

robander:
I think I'll stick with the simple `xml:lang=""`. Using `zxx` sounds like it would be accurate, but feels a little strange for the common variables that will insert their own text values. For example, the following variable is currently identical in all 48 variable files -- it does not have any language text, but will after resolution: 
`<variable id="Index even header"><param ref-name="pagenum"/> | <param ref-name="heading"/> | <param ref-name="prodname"/></variable>`

src/main/xsl/common/dita-utilities.xsl
```diff
@@ -124,11 +124,26 @@ See the accompanying LICENSE file for applicable license.
         </xsl:choose>
       </xsl:when>
       <xsl:otherwise>
-        <xsl:value-of select="$id"/>
-        <xsl:call-template name="output-message">
-          <xsl:with-param name="id" select="'DOTX052W'"/>
-          <xsl:with-param name="msgparams">%1=<xsl:value-of select="$id"/></xsl:with-param>
-        </xsl:call-template>
+        <xsl:variable name="stringfile" select="$stringFiles[@xml:lang='']/@filename" as="xs:string*"/>
+        <xsl:variable name="str" as="element()*">
```
jelovirt: The variable name here is a bit misleading. Maybe rename to `variable` or `var`?

robander: Sure - though note that I was just using the same variable name we already use a few lines up, when pulling in the locale-specific variables :-)

jelovirt: The old variable names are misleading too :)

robander:
Added updated version of `dita-utilities.xsl` -- function is same, but based on @jelovirt 's comment:
> The variable name here is a bit misleading because it's not a string, the content's is a set of variable defining elements. Maybe rename to `variable` or `var`?
I've updated to use `variable`; also updated the code this was based on so that we consistently use `variableFiles` (instead of `stringFiles`), `variablefile` (instead of `stringfile`), and `variable` (instead of `str`).

robander:
Latest commits move all the duplicate PDF2 variables to the `commonvariables` file.

Worth noting that for people who customize, some of these common variables -- which don't contain any localized text _as shipped_  -- might end up being customized. For example, the default TOC header variables are
```
<variable id="Toc odd header"><param ref-name="prodname"/> | <param ref-name="heading"/> | <param ref-name="pagenum"/></variable>
<variable id="Toc even header"><param ref-name="pagenum"/> | <param ref-name="heading"/> | <param ref-name="prodname"/></variable>
```

These are currently common to all languages because any localized text comes in as a parameter. If you want them to contain some standard localized text, these would have to move back into individual variable files for each language.

Starting to think a new doc topic about this might be a good idea, if @infotexture would like me to draft one...

robander:
Also, I just want to point out -- out of all variables that moved from the 48 individual files the the common files, the only ones that were _not_ exactly the same in all 48 files were:

1. Several common variables were missing (forgotten) from the Serbian variable file. They are now found in the common file, so this fixes a minor bug with default Serbian PDFs.
1. The `#note-separator` variable has an extra space for French and Japanese. Those two retain the definition, which is found first; the other 46 languages pull from common.
1. For `Toc odd header` and `Toc even header`, the default has changed for Spanish. The other 47 languages used the usual default heading of product name / heading / pagenum, as Spanish did for other sections, but Spanish used pagenum / heading / pagenum for TOC. This appears to be a bug, so it will now use the common variable.

robander:
Hoping to get this done (ready for full review and/or merge) this week.

Current status of variables in the PDF2 directory:

- I want to remove explanatory comments from the locale-specific files, leaving comments only in the English file + the common variables file. 
- We have `<!--TODO-->` comments spread very inconsistently around for variables that are empty but should have localized text. Ideally I'd like this to be consistent, adding them where missing, to make it easier to fix / get help / easier for those customizing strings to find ones they need to add.
- A few languages generate ``[DOTX052W][WARN]` for missing translation of "Notices title" (es, fr, it), "Search title" (fi, he, ro, ru, sv). Most languages actually define these but it they are empty. I plan to add those missing empty variables to the few languages that don't have them.
- "Required Step" is missing for Italian, and can be added
- "Notice" is missing for Dutch, and can be added
- Serbian is broken for task headings, and should be fixed
- Vietnamese generates an error for variable `task_procedure_unordered`, which should be tracked down
- Every language except English is generating this for `<note type="troubleshooting">`:
`[DOTX052W][WARN]: No string named 'Trouble' was found when creating generated text; using the value 'Trouble' in your output file. `
This is true in 2.4, not a change related to this code, but I noticed it when testing. English pulls this from the common strings (not PDF2). Not sure if we should add this to other languages with a `TODO`; right now I don't plan to add it.

Remaining work outside of PDF2:

- Create a `commonstrings.xml` file in `xsl/common/` (these are all raw strings, rather than variables)
- Move the few common strings over. Currently only `intro-to-topic-link`, quote and colon characters, `figure-number-separator`, `#menucascade-separator`

infotexture:
Haven't reviewed the latest file states here in this branch, but one question/suggestion comes to mind:

I recall when customizing strings for earlier versions, there were a lot of unnecessary differences between strings files for the various languages, mostly w/r/t to indentation, line breaks and comments. To facilitate customization, it would be ideal if the only differences between these files were the actual localized content. 

I realize @robander may have resolved these differences with these _(or earlier)_ revisions, but if not, I just wanted to suggest that now might be a good time to harmonize things, since the files are changing anyway.

robander:
@infotexture  ... well, sort of. I'm working on fixing indentation / spacing across all the non-English ones (that'll be the next commit, I've made my way through a bunch but it's a tedious task).

One wish I've heard from @jelovirt in the past is that we shouldn't have repetitive comments in the files. I've been removing those as I go from non-English. They're already missing in the variable files Jarno added. For example, comments like this:
```
<!-- Template for generated page number of a reference
         in printed manual, when referenced element has no title. -->
```
For now that means the comments will only be in the English file -- so spacing / comments for English don't match the others. If we want English to match the others (which has a certain logic), I think we'd need to define the variables in the docs somewhere.

robander:
Just heard directly from a colleague following this thread, asking why we would remove the comments, because it seems like it would be easier on translators to keep them.

I'm not certain about that. My experience within IBM, along with what I've heard anecdotally outside of IBM, is:

- For adding a new language, normal process is to start with the English and translate each value, though for closely related languages the others may be used as an aid.
- I've also heard that the normal process for custom styles is to simply re-translate the English (along with any local tweaks made as part of that style).
- In both cases, it seems like having the comments available in the English version would be helpful - easier than having to look up a page in our docs to find out what each ID corresponds to.
- For entirely new languages added to DITA-OT, this would mean stripping out the comments / normalizing spacing, which is kind of a pain. But ... new languages are added rarely, and I think in fact I've been the one to add them all.
- Anybody starting with English and creating their own translations will likely preserve the comments in their own versions, but that's up to them and probably isn't a problem.

All of that sort of tips me towards keeping comments in the English variables, but I'd like to hear from others too...

infotexture:
I can see both sides here, but would lean toward keeping comments in English & removing elsewhere.

:+1: for aligning as much as possible.

robander:
Another sort-of-alignment issue ... in the non-English files, some variables are empty, while others still have English. So far, as I've removed comments / fixed spacing, if there is English text, I've put the `TODO` outside of the variable:
`<variable id="Search title">Search title</variable><!--TODO:Search title-->`
If the variable is empty, I've put the `TODO` inside:
`<variable id="Notices title"><!--TODO:Notices title--></variable>`

Not sure if that's the best plan or not but it seemed more consistent with the current variable files.

robander:
Test failure comes from the `zh-TW` test case. It previously defined the colon symbol as the common `:` character, but `zh-CN` used the wide colon `：` character. By making the normal colon ':' common, `zh-TW` processing now falls back to `zh`, which finds the wide colon in simplified Chinese.

Based on our long use, along with other plugins I see (like the pdf5 one from AH), the wide colon should also be used in `zh-TW` output; I'm going to update that string file and fix the test.

jelovirt:
@robander could you rebase to `develop` to resolve the conflicts?

robander:
Well, I tried the rebasing and clearly didn't get the results I wanted, and (bonus) still have the merge conflict. Should have been only one minor conflict file, `dita-utilities.xsl`.

robander:
@jelovirt fixed the rebase - this should be ready for merging now if you're OK with the current status.

Issue #2447, opened 2016-08-08
Relationship table accessibility


wintrymix:
When relationship tables are published to HTML, the links are formatted as a set of `div` elements. While this was acceptable under Web Content Accessibility Guidelines(WCAG) 1.0, WCAG 2.0 calls for lists of links to be formatted using list markup. HTML processing should be updated to conform to the recommendations of WCAG 2.0.

robander:
I've gotten the same guidance from accessibility experts I work with, and we've had to override the default output to meet WCAG. I'm working on getting this one into the base code for 3.0 for HTML5 and XHTML.

PR #2799, opened 2017-09-27
Generate list of links as list rather than div #2447

robander:

## Description
Converts lists of links from `<div>` markup to actual list markup for more accessible output, as described in #2447 

First commit only covers HTML5, which doesn't have a big hit on integration tests. I made the equivalent change in XHTML, but ... not surprisingly ... it hits a ton of integration tests, and I haven't had a chance to update them all.

In both cases, I've updated the CSS / SASS files so that generated link lists retain the same formatting in browsers.

## Motivation and Context

The web content accessibility guidelines from w3 state that groups of links should be encoded using list markup: https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H48.html

We already present generated links as lists under "Related concepts" and similar headings; manually grouped `<linklist>` contents are also presented as lists. The markup should match. I've encountered accessibility tools that flag the current output as incompatible with WCAG 2.0, and screen readers will perform better with the new markup.

## How Has This Been Tested?

Tested initially with a set of topics designed to generate links + include linklists, then against integration tests.

## Type of Changes

Somewhere between a bug fix / enhancement, improves output that doesn't pass WCAG validation today.

## Checklist

I haven't yet updated the integration tests so XHTML will pass. There's a lot to update, so if further changes are requested, I'd like to make those first before updating all of the `exp/` directories for XHTML.

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/_links.scss
+++ b/src/main/plugins/org.dita.html5/sass/_links.scss
@@ -26,6 +26,17 @@
   margin-bottom: 1em;
 }
 
+ul.linklist {
+  margin-top: 0;
+  list-style-type: none;
+  padding-left: 0;
+}
+
+li.linklist {
+  margin-top: 0;
+  margin-bottom: 0;
+}
+
 .linklistwithchild {
   margin-bottom: 1em;
   margin-left: 1.5em;
--- a/src/main/plugins/org.dita.html5/sass/commonrtl.scss
+++ b/src/main/plugins/org.dita.html5/sass/commonrtl.scss
@@ -15,6 +15,12 @@
   margin-bottom: 1em;
 }
 
+ul.linklist {
+  margin-top: 0;
+  list-style-type: none;
+  padding-right: 0;
+}
+
 .linklistwithchild {
   margin-right: 1.5em;
   margin-top: 1em;
--- a/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
@@ -428,14 +428,7 @@ Each child is indented, the linktext is bold, and the shortdesc appears in norma
 
   <!--basic child processing-->
   <xsl:template match="*[contains(@class, ' topic/link ')][@role = ('child', 'descendant')]" priority="2" name="topic.link_child">
-    <xsl:variable name="el-name">
-      <xsl:choose>
-        <xsl:when test="contains(../@class, ' topic/linklist ')">div</xsl:when>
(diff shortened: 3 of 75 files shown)


jelovirt:
I'm a bit worried to push this to XHTML (the old stable output) because it will likely cause breakage in custom plugins and expected output. For HTML5 (the new correct output) this definitely should be made.

For integration tests, if we start testing both HTML5 and XHTML the test run duration will again increase and it's way too long the way it is now. Should we have a separate test drivers for HTML5 and XHTML, and... maybe not run XHTML test all the time?

raducoravu:
Same as @jelovirt said, although the semantic of the HTML tags is much better this way this may break some XHTML CSS customizations although they would be easy to fix if noticed. Probably most people match the class attribute value directly.
One thing, if in the CSS you set "list-style-type:none" does the acessibility narrator (JAWS for example) interpret the structure as the list or does it not consider it a list anymore?

robander:
@raducoravu - I've always been told by our accessibility experts to use that list style to display a list that way, because it lets screen readers continue to identify it properly. The WCAG guidelines actually suggest that in Example 5 "[Using lists to group links](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H48.html#H48-ex5)".

robander:
For XHTML in general ... I understand the concern, as one of the people with customizations that will be affected. At the same time ... 

- I know a lot of people still use XHTML (it's still the most common build for many of my customers, despite a push to move to HTML5). That's why I've continued to maintain it, putting fixes in both plugins when they apply to both HTML5 and XHTML.
- The current output is not considered accessible, which is bad. (Given that it passed WCAG 1.0, I think it's probably not as significant as some other WCAG rules. But it's still bad.)
- Given that I'm otherwise continuing to maintain XHTML, it seems wrong to avoid fixing it to pass WCAG? Release notes can highlight the CSS change required to get style back as it was (for those who override / remove the default `commonltr` / `commonrtl` CSS files

Thoughts on that?

For integration tests ... I think we should probably switch the defaults to HTML given that it's clearly the more-supported / more-updated one by the team in general, so more likely to have problems worth catching during the development process. Also agree we shouldn't run both HTML and XHTML all the time, but it would be nice to be _able_ to run XHTML as well, so that as I personally maintain it I can still get reliable testing?

raducoravu:
If this improves accessibility then indeed it is worth doing also in the XHTML output especially that the backwards compatibility problem is a minor one.

robander:
@raducoravu thanks 👍 I'll work on updating tests.

PR #2800, opened 2017-09-29
In-topic links break when using onlytopic.in.map=true

robander:
## Description

I've got in-topic links like the following:
`<xref href="#roottopic"/>`
or
`<xref href="#topic/niftyFigure"/>`

Our custom x/html transform types set `onlytopic.in.map=true` as a default. When that is set, these in-topic links are added to the list of "non topicref resources", which eventually ends up with the directory itself treated as a file reference, and the following series of errors from the build:
```
   [filter] C:\Users\IBM_AD~1\AppData\Local\Temp\temp20170928202410586 (Access is denied.)
[topic-fragment] Failed to process XML filter: Failed to transform C:\Users\IBM_AD~1\AppData\Local\Temp\temp20170928202410586: C:\Users\IBM_AD~1\AppData\Local\Temp\temp20170928202410586 (Access is denied.)
[topicpull] Failed to transform document: org.xml.sax.SAXParseException: Content is not allowed in prolog.
```

The fix ensures that any `@href` beginning with `#` is not added to any reference list (if we're in a topic, that topic is already in a list, and this self-reference does not represent a new type of reference for tracking).

## How Has This Been Tested?

I added an in-topic reference to our existing integration tests for `onlytopic.in.map`, which caused the build to fail; the fix restores a passing build, and also fixes the problem in the files that originally encountered the issue.

## Type of Changes

Bug fix.

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/reader/GenListModuleReader.java
+++ b/src/main/java/org/dita/dost/reader/GenListModuleReader.java
@@ -405,7 +405,8 @@ public final class GenListModuleReader extends AbstractXMLFilter {
 
         final URI href = toURI(atts.getValue(ATTRIBUTE_NAME_HREF));
         final String scope = atts.getValue(ATTRIBUTE_NAME_SCOPE);
-        if (href != null && !ATTR_SCOPE_VALUE_EXTERNAL.equals(scope) && !ATTR_SCOPE_VALUE_PEER.equals(scope)) {
+        if (href != null && !href.toString().startsWith("#") &&
+                !ATTR_SCOPE_VALUE_EXTERNAL.equals(scope) && !ATTR_SCOPE_VALUE_PEER.equals(scope)) {
             if (!(MAP_TOPICREF.matches(cls))) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/dita/dost/reader/GenListModuleReader.java
```diff
@@ -405,7 +405,8 @@ public void startElement(final String uri, final String localName, final String
 
         final URI href = toURI(atts.getValue(ATTRIBUTE_NAME_HREF));
         final String scope = atts.getValue(ATTRIBUTE_NAME_SCOPE);
-        if (href != null && !ATTR_SCOPE_VALUE_EXTERNAL.equals(scope) && !ATTR_SCOPE_VALUE_PEER.equals(scope)) {
+        if (href != null && !href.toString().startsWith("#") &&
```
jelovirt: Instead of `!href.toString().startsWith("#")` this should be e.g. `href.getPath() != null`

Issue #2795, opened 2017-09-22
Related links to external web sites using scope="external" no longer shown in PDF


raducoravu:
Tested with 2.5.3, seems to occur also with 2.4, but with DITA OT 2.3 it seemed to work.
The topic has a related link section at the end:

    <related-links>
        <link href="http://www.oxygenxml.com" format="html" scope="external"/>
    </related-links>

I'm using "args.rellinks="all"". The generated PDF has in the topic related links section a title like "Related links" but it's not followed by any link.
Related links to internal topics still work though.

raducoravu:
Using scope="peer" on the link element starts showing the link in the PDF.

raducoravu:
It also works if you specify a linktext element inside the link.

robander:
I've reproduced with 2.5.3 (including the peer / linktext updates, not sure what is going on.

robander:
Fixed with #2802

PR #2802, opened 2017-09-30
Fix for #2795 missing external link

robander:
Bug fix for #2795 missing external links in PDF `<related-links>` section.

Current processing tries to determine a link text based on the DITA target; if that cannot be found, it uses the `<linktext>` element. 

For `<xref>` when the target is external / non-dita or cannot be accessed, link text is set using text/element content (if available), otherwise it uses `@href`.

This fix updates `<link>` processing to match that -- using `<linktext>` _if available_, otherwise using `@href`. This means links where text could not be retrieved still show up, using the supplied link text. (Note that the link was already there in the FO / PDF -- it just had no text).


raducoravu:
👍 Fix looks good.

PR #2806, opened 2017-10-03
Conref in a submap in a subdirectory is broken

robander:

## Description

I have a submap that uses conref to pull in a phrase. The submap is not in the same directory as the root map. When the `mapref` step runs and the maps are resolved, paths in `@conref` are not adjusted, so the relative path is no longer valid. When the conref step runs, each relative-path conref in the submap will throw an error.

Much less of a priority, but I also noticed that relative paths on non-topicref `@href` that appear in submaps are also not adjusted.

A similar fix works to fix each case, by ensuring that the `$relative-path` parameter is passed to non-topicref contexts.

## How Has This Been Tested?

Test case uses submap in a sub-dir. In the submap there is a relative `@conref` as well as `xref/@href`. Test case is now added to the integration tests with correct resolved markup from `mapref` + `conref`. Failed before the fix, works with the fix, no other tests affected.

## Type of Changes

Bug fix

## Checklist


src/main/xsl/preprocess/maprefImpl.xsl
```diff
@@ -30,17 +30,51 @@ See the accompanying LICENSE file for applicable license.
   <xsl:key name="reltable" match="//*[contains(@class, ' map/topicref ')]" use="@dita-ot:orig-format"/>
 
   <xsl:template match="@* | node()">
+    <xsl:param name="relative-path" as="xs:string">#none#</xsl:param>
```
jelovirt: We really should refactor the code to stop using these `#none#` magic strings for null values and just start supporting `()` instead.

src/main/xsl/preprocess/maprefImpl.xsl
```diff
@@ -30,17 +30,51 @@ See the accompanying LICENSE file for applicable license.
...
+        </xsl:otherwise>
+      </xsl:choose>
+    </xsl:attribute>
+  </xsl:template>
+
   <xsl:template match="*[contains(@class, ' map/topicref ')][(@format, @dita-ot:orig-format) = 'ditamap']
                         [empty(@href(: | @dita-ot:orig-href:)) or
                          (:@processing-role = 'resource-only' or:)
                          @scope = ('peer', 'external')]" priority="15">
+    <xsl:param name="relative-path" select="'#none#'" as="xs:string"/>
```
jelovirt: Since `$relative-path` is not modified when passing to children, use tunnelled parameter instead.

src/main/xsl/preprocess/maprefImpl.xsl
robander: Agreed, but probably not a good idea to do in the hotfix...

src/main/xsl/preprocess/maprefImpl.xsl
robander: I tried changing to use tunnelled parameters, but it broke other uses. The value changes each time a new submap is encountered, which I'm guessing is the issue. We currently have a mix of tunnelled and not tunnelled uses for `$relative-path` in this module ... seemed like it would be easy to update them all but it broke when I had submaps in 2 different directories.

Issue #1779, opened 2014-11-04
1.3: Support new table accessibility attributes


drmacro:
Proposal 62

This is most important for HTML output but there may be implications for PDF output as well.

robander:
HTML has a very large amount of code today that automatically sets up these values, so that you do not need to maintain them in your source. Thinking the correct approach is to use the values in the document when specified, otherwise, do the automated accessibility. 

FWIW I still have not encountered a table from any of my users where the automated accessibility was not enough, so I will continue to encourage people to use the automated processing for all but the most extreme table examples. (Those can exist, but again, I have not encountered them except in mock-up samples that exist only to show when these attributes can be useful). Maintaining the values manually in DITA topics will be extremely time consuming and error prone, unless your editor does it for you by implementing exactly the same sort of processing already done in DITA-OT.

drmacro:
Oracle insists that even the accessibility features in 1.3 are not sufficient to meet all accessibility requirements, although I am not convinced. One thing they pointed out was that the lack of a built-in table footer in the DITA table model is a problem. I pointed out that you can get the effect by using @outputclass on <row>. I think there may be some misunderstanding of either what the current OT does or what it can be made to do easily enough. They do have some gnarly tables in their docs (these are the docs for Oracle Database, after all) but it's hard to tell how much is a side effect of legacy practices, how much is failure to configure or extend the OT appropriately, and how much is actual missing features.

I see that the current code is generating cell header references--how does it handle row headers?

robander:
Current code handles row headers based on `@rowheaders="firstcol"` (for complex tables) or based on `@keycol` (for `<simpletable>`). Entries in the designated column are marked up with generated IDs using `<th id="gen-id">`, and entries to the right/left in the same row use `@headers` to reference that cell. Spanning cells are supported, just as with tables that use spanned cells in `<thead>`.

This is clearly a case where you can construct tables that need more - maybe the first two rows are headers (impossible with 1.2 in each model), or maybe the last column is a header (impossible in 1.2 with the complex table). That said - the tables I've most often seen to illustrate this usage are less usable than they would be if flipped so that the two header columns ran across the top. That's why I agree it's possible you will need this markup - maybe you have a table where that is in fact the best way to present information - but I haven't run across one in the millions of topics that we've published.

[...]

PR #2810, opened 2017-10-11
Update grammar files for coming DITA 1.3 errata

robander:
The DITA 1.3 grammar files ship support for the `colspec/@rowheader` attribute. They mistakenly set a default of `rowheader="headers"`, which would designate every cell in every column as a header (for those tables which use `<colspec>` elements).

This was discussed on the OASIS list 2 weeks ago, and at the meeting this week the TC agreed to treat it as an errata, to be updated in the second DITA 1.3 errata.

This pull request makes the relevant changes (removing the default), and updates the only test case where expected output included that default value. With this change, we will be able to implement #1779 (currently blocked by the default value).

### DIFF
--- a/src/main/plugins/org.oasis-open.dita.v1_3/dtd/base/dtd/tblDecl.mod
+++ b/src/main/plugins/org.oasis-open.dita.v1_3/dtd/base/dtd/tblDecl.mod
@@ -322,7 +322,7 @@
                            headers |
                            norowheader |
                            -dita-use-conref-target)
-                                    'headers'
+                                    #IMPLIED
                %tbl.colspec.att;
                %dita.colspec.attributes;"
 >
--- a/src/main/plugins/org.oasis-open.dita.v1_3/rng/base/rng/tblDeclMod.rng
+++ b/src/main/plugins/org.oasis-open.dita.v1_3/rng/base/rng/tblDeclMod.rng
@@ -439,8 +439,7 @@ For <entry>, add:
       </attribute>
     </optional>
     <optional>
-      <attribute name="rowheader" dita:since="1.3"
-        a:defaultValue="headers">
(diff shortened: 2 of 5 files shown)


raducoravu:
I'm not sure I understand the problem. I published a CALS table to XHTML with the DTDs as they look before your change.
The generated XHTML header cells looked like:

               <th class="entry cellrowborder" id="d1687e40">Header text1 </th>

and each cell under this header cell looked like this:

                <td class="entry cellrowborder" headers="d1687e40 "> cell text</td>

referencing the top cell on the column as a header cell.

jelovirt:
@robander Is this ready for merge?

robander:
@raducoravu We have not currently implemented anything for this new attribute (that's open as an issue #1779). I found this bug when starting to implement 1779. Basically -- the grammar files mean that every `colspec` has a default of `rowheader="headers"`. The value `rowheader="headers"` on `<colspec>` means "This column is a header column" [http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part1-base/langRef/attributes/calsTableAttributes.html#oet-atts__rowheader]. So, if you respect that default, every column becomes a "header column", and every cell becomes a header cell. Removing the default allows the attribute to be evaluated as intended, allowing an implementation of 1779.

@jelovirt yes, ready for merge.

raducoravu:
@robander Thanks for the explanation, I did not know how urgent this would be to patch in the DITA OT 2.5.2 bundled with Oxygen 19.1, but as it is not yet used for anything in 2.5.2...

robander:
@raducoravu yep that's correct - not urgent, unless / until you plan to implement additional support for `<colspec rowheader="headers"/>`. I'd hoped to have that in HTML processing for 3.0 but finding this slowed me down and I got distracted with other issues.

PR #2811, opened 2017-10-12
Refactor HTML5 simpletable accessibility for modern approach

### LINKED ISSUE #2448: Bogus row header @id generated for <stentry> elements

robander:

This pull request modifies the way simpletable headers are marked up for accessibility.

The original approach specified `@id` for every header cell, and used `@headers` on body cells to reference each header. This was the recommended approach when DITA-OT 1.0 came out and was used in XHTML.

There are a number of problems with this:

1. It used generated IDs, which are a problem for many systems that use diff to compare output
1. The output is extremely verbose, while many HTML5 developers prefer clean files / minimal noise
1. In our implementation, generated IDs for cells inside `sthead` often override specified IDs, so links to those entries do not resolve
1. Even when specified IDs are preserved today, the actual ID is used instead of `topic__id` - so again, links to those IDs do not resolve
1. In 2.5, when a header row is specified for `<choicetable>`, the same generated ID is used for each cell of the header, so those are broken today (it only works when the header row is generated, or when the header entries specify their own ID)

The updated code here switches to use the `@scope` attribute to mark header cells. It is always used for entries inside `sthead` (with `scope="col"`), and is also used for any designated key column (with `scope="row"`). The code to generate these values is much simpler, which makes it easy to fix items 3/4/5 above. By removing extra noise from the HTML5 output, we also move to the now-recommended approach for data tables. See: https://webaim.org/techniques/tables/data#headers

> The scope attribute tells the browser and screen reader that everything within a column that is associated to the header with scope="col" in that column, and that a cell with scope="row" is a header for all cells in that row.
> All <th> elements should generally always have a scope attribute. While screen readers may correctly guess whether a header is a column header or a row header based on the table layout, assigning a scope makes this unambiguous.
> ...
> Another way to associate data cells and headers is to use the headers and id attributes. This method is NOT generally recommended because scope is usually sufficient for most tables, even if the table is complex with multiple levels of headers.

A few notes / caveats:

- This update is straightforward for simpletables, where entries cannot span and headers are easy to calculate. I'd like to make the same update for complex tables but am not sure I'll have time before 3.0 is frozen. The updates for complex tables are closely related to #1779 so should probably be done together.
- I've only done this for HTML5. Given that the existing support is valid (although outdated), and given the churn this would cause for existing XHTML table overrides, I'd prefer to leave that alone (much higher cost for a lower return).
- The update here covers simpletables and OASIS specializations of simple tables (`<properties>`, `<choicetable>`).
- When integrated, this should close #2448 and #2449 that both suggest removing generated headers in simple tables

Tested using the following files, which cover simpletable (with / without headers and keycol), properties (with/without generated headers and keycol), and choice tables (with / without headers and keycol).
[1779.zip](https://github.com/dita-ot/dita-ot/files/1380262/1779.zip)

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/choicetable.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/choicetable.xsl
@@ -20,33 +20,13 @@ See the accompanying LICENSE file for applicable license.
   ">
     <chhead class="- topic/sthead task/chhead ">
       <choptionhd class="- topic/stentry task/choptionhd ">
-        <xsl:attribute name="id" select="simpletable:generate-headers(., 'option')"/>
         <xsl:sequence select="dita-ot:get-variable(., 'Option')"/>
       </choptionhd>
 
       <chdeschd class="- topic/stentry task/chdeschd ">
-        <xsl:attribute name="id" select="simpletable:generate-headers(., 'desc')"/>
         <xsl:sequence select="dita-ot:get-variable(., 'Description')"/>
       </chdeschd>
     </chhead>
-  </xsl:template>
-
-  <xsl:template mode="headers" match="
-    *[contains(@class,' task/choption ')]
-     [empty(simpletable:get-current-table(.)/*[contains(@class,' task/chhead ')])]
-  ">
-    <xsl:attribute name="headers" select="
-      simpletable:generate-headers(simpletable:get-current-table(.), 'option')
-    "/>
-  </xsl:template>
-
-  <xsl:template mode="headers" match="
-    *[contains(@class,' task/chdesc ')]
-     [empty(simpletable:get-current-table(.)/*[contains(@class,' task/chhead ')])]
-  ">
-    <xsl:attribute name="headers" select="
-      simpletable:generate-headers(simpletable:get-current-table(.), 'desc')
-    "/>
   </xsl:template>
 
 </xsl:stylesheet>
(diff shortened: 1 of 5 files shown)


jelovirt:
@robander Doesn't this make #2449 obsolete.

robander:
Yes - I'll mark it closed.

PR #2825, opened 2017-10-28
Update Gradle wrapper to 4.2.1

infotexture:
This syncs Gradle to the current release v4.2.1 as used in the docs repository.

Update Gradle distribution version & regenerate wrapper to pick up recent changes per
https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:wrapper_generation


infotexture:
This should help to prevent any issues with Gradle versions when https://github.com/dita-ot/docs/pull/178 is merged.

Issue #1271, opened 2012-05-08
Chm TOC shows wrong characters for many languages


lade:
As one example, the letter Ά shows incorrectly in TOC for chm for Greek. By changing html output to 'windows-1253' instead of 'iso-8859-7' in codepages.xml I got it to show correctly.

The compiler of HTML Help Workshop uses different encodings from what are used by DITA-OT currently. The ones that work correctly are:
- windows-1250 for East European languages (instead of iso-8859-2)
- windows-1252 for Western languages (instead of iso-8859-1)
- windows-1253 for Greek (instead of iso-8859-7)
- windows-1254 for Turkish (instead of iso-8859-9)
- windows-1257 for Estonian (instead of iso-8859-1)

In some of these cases the differences are only about some special characters. And windows-1252 is a superset of iso-8859-1 (except for control characters in the range 80-9F). But in some there are more relevant differences, and it would be nice if the correct codepage would be used.

robander:
This has just come up as an urgent issue for one of my customers.

In the past, we had a hacked version of DITA-OT that converted _some_ characters to entities, but not all, followed by a codepage conversion. What made that work was that the entity conversion was language-aware, so for characters like Ý that exist in iso-8859-2, we do not generate `&Yacute;` (keep the character as Ý) when the target codepage is iso-8859-2. The `convert-lang` utility in DITA-OT always does the conversion, so common letters in iso-8859-2 languages are corrupted. Similarly, I've always ignored entities for Greek characters when the input file is Greek.

If nothing else - could we just add a property that allows skipping the `convert-lang` task, so that when all characters are common/expected characters from the language's CHM charset, everything will work? We've got similar properties on a lot of targets, but none available here. @jelovirt I don't know if adding such a property qualifies as a bug fix for a patch, as it doesn't quite _fix_ a bug but would allow us to work around one...

robander:
FWIW our earlier process (run outside of DITA-OT) was actually based on target codepage rather than language. For example, if target codepage is windows-1250 (as with Czech), none of the common Czech characters are converted. In this case if I'm reading right, we convert the HHP file to 1250 but the HHC contents file, HHK index, and HTML files to the HTML codepage. I'm not sure if that's causing additional trouble -- it makes sense that HTML would be converted to the HTML codepage (with charset properly updated), but I would think that the CHM compiler on Windows would want the other project files (not just HHP) to use the Windows codepage. :/

robander:
[...]

Issue #1151, opened 2012-04-26
HHC in HTMLHelp has no encoding set


jelovirt:
Converted from [SourceForge issue 3445877](http://sourceforge.net/support/tracker.php?aid=3445877), submitted by [jelovirt](/jelovirt)

The HHC files generated by HTMLHelp plug-in do not have the encoding defined with:

``` xml
  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
```

before convert-lang is run. This results in wrong encoding to be used in table of contents.

The HHC file is generated with XML output mode, whereas the actual topic files are with HTML output. This will make the XSLT processor to generate the meta element and allow convert-lang to correctly identify the encoding and change it.

robander:
Thinking this may be a duplicate of #1271, which was opened a bit later but has more discussion.

robander:
Fixed with #2852

PR #2852, opened 2017-11-28
Fix corrupted chars in HHC, HHK #1271 #1151 

robander:

We have 2 issues with the project files generated for HTML Help compilation. This fix handles both issues together.

1. The compiler expects the TOC (HHC) and index (HHK) to be in the Windows codepage rather than the ISO codepage. They are very similar but the few differences can cause corrupted characters. We currently use the Windows codepage for the project (HHP) file, but should use for all 3 (I've been doing that for all 3 for many years in another system).
1. Within the project files, HTML entities are not valid. We currently convert any accented or special character that has a matching HTML entity to use that entity, which then will not display in the TOC or Index. Any character that is supported in the target codepage should be allowed to use that codepoint rather than the entity. This has been reported several times in duplicate issues, but most of the discussion is in #1271.

The fix switches HHC and HHK to use the same codepage as HHP. It also sets up an exception list of entities for each codepage encountered. Currently, the logic is "If an entity exists, use it". The new logic is "If an entity exists and is not in the exception list, use it". The exception list is based on the target charset, so that (for example) Ω will be converted to `&Omega;` where the character does not exist but will map to the proper codepoint when the target codepage is Greek.

## How Has This Been Tested?

The new codepage logic (Windows codepage for both HHC and HHK) is the same logic I've been using elsewhere for HTML Help for 10+ years, with all languages covered by DITA-OT. Same for the entity mapping, where I've avoided creating entities whenever possible.

To test the code changes here I've set up English, Czech, and Greek test cases.

- Accented characters available in Latin-1 character sets now work in English. Greek characters and characters like Š still work in the content as entities, and still fail in the TOC as entities.
- Accented characters available in Latin-2 character sets now work in the Windows-1250 TOC and index files. Tested with the full Czech character set: `Aa Áá Bb Cc Čč Dd Ďď Ee Éé Ěě Ff Gg Hh Ii Íí Jj Kk Ll Mm Nn Ňň Oo Óó Pp Qq Rr Řř Ss Šš Tt Ťť Uu Úú Ůů Vv Ww Xx Yy Ýý Zz Žž`
- Greek characters now work in Greek TOC, index, and contents: `ΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩ αβγδεζηθικλμνξοπρςστυφχψω`

Test cases attached.
[chars.zip](https://github.com/dita-ot/dita-ot/files/1511049/chars.zip)

## Type of Changes

Bug fixes to handle 2 overlapping issues.

## Checklist

- I have signed-off my commits per http://www.dita-ot.org/DCO.
- Builds & tests completed successfully (`./gradlew test integrationTest`).

### DIFF
--- a/src/main/java/org/dita/dost/ant/ConvertLang.java
+++ b/src/main/java/org/dita/dost/ant/ConvertLang.java
@@ -23,8 +23,10 @@ import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 import java.util.Map.Entry;
 
 import javax.xml.parsers.DocumentBuilder;
@@ -56,6 +58,13 @@ public final class ConvertLang extends Task {
     private static final String tag1 = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>";
     private static final String tag2 = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>[OPTIONS]";
     private static final String tag3 = "&lt;?xml version=\"1.0\" encoding=\"utf-8\"?&gt;";
+    
+    private static final String CODEPAGE_ISO_8859_1 = "iso-8859-1";
+    private static final String CODEPAGE_ISO_8859_2 = "iso-8859-2";
+    private static final String CODEPAGE_ISO_8859_7 = "iso-8859-7";
+    private static final String CODEPAGE_1250 = "windows-1250";
+    private static final String CODEPAGE_1252 = "windows-1252";
+    private static final String CODEPAGE_1253 = "windows-1253";
 
     private String basedir;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/dita/dost/ant/ConvertLang.java
```diff
@@ -415,6 +432,82 @@ else if(FileUtils.isHHPFile(inputFile.getName())){
...
+        if (exceptionCharset != null && exceptionCharset.equals(charset)) {
+            return;
+        }
+        exceptionCharset = charset;
+        if (!entityExceptionSet.isEmpty()) {
+            entityExceptionSet.clear();
+        }
+        if (charset.equals(CODEPAGE_ISO_8859_2) || charset.equals(CODEPAGE_1250) ||
+                charset.equals(CODEPAGE_ISO_8859_1) || charset.equals(CODEPAGE_1252)) {
+            entityExceptionSet.add(193); entityExceptionSet.add(225);//A-acute
```
jelovirt: IMO it would look better as

```java
entityExceptionSet.add(Arrays.asList(
  193, 225, //A-acute
  194, 226, //A-circumflex
  ...
));
```

src/main/java/org/dita/dost/ant/ConvertLang.java
```diff
@@ -91,6 +104,8 @@ public void execute(){
         createLangMap();
         //initialize entity map
         createEntityMap();
+        //Initialize entitye exceptions
+        entityExceptionSet = new HashSet<>(128);
```
jelovirt: No need to init size.

PR #2871, opened 2018-01-03
Change label of note type 'danger' in German from 'Vorsicht' to 'Gefa…

xephon2:
Fixed German label for `<note type="danger">`. Changed from `VORSICHT` to `GEFAHR` as suggested by **ANSI Z535.4**, see [NORM 5 – Neue ANSI Z535 – vom vernünftigen Umgang mit Sicherheitshinweisen](http://www.schmeling-consultants.de/uploads/media/NORM_5_Schmeling_ANSI_Z535_neu.pdf), page **21**.

### DIFF
--- a/src/main/xsl/common/strings-de-ch.xml
+++ b/src/main/xsl/common/strings-de-ch.xml
@@ -37,7 +37,7 @@ See the accompanying LICENSE file for applicable license.
   <str name="Important">Wichtig</str>
   <str name="Attention">Achtung</str>
   <str name="Caution">ACHTUNG</str>
-  <str name="Danger">VORSICHT</str>
+  <str name="Danger">GEFAHR</str>
   <str name="Remember">Hinweis</str>
   <str name="Restriction">Einschränkung</str>
   <str name="Warning">Warnung</str>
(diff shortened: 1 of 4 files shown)


robander:
Looks good to me, though if I understand this correctly the current translation is considered broken for most cases - so this might be more appropriate for `hotfix/3.0.2` rather than waiting for the next release.

infotexture:
If we do decide to implement this change for 3.0.2, 3738c59 would need to be cherry picked to the hotfix branch, since the commit was made to the `develop` branch on the fork, rather than a dedicated feature branch, so it can't be merged via the web UI even if the base branch is changed.

xephon2:
@infotexture Should I redo the PR on another branch?

infotexture:
@xephon2 No need to repeat the work, the cherry pick operation is trivial if we decide to ship this with 3.0.2.

jelovirt:
@raducoravu Even though this was merged to `develop`, this should also be cherry-picked to 3.0.2.

raducoravu:
@jelovirt I cherry picked it to 3.0.2.

jelovirt:
@raducoravu In the future, it's preferable in cases like this to cherry-pick the commit to hotfix branch and then just wait for that hotfix branch to be merged to `develop`, instead of merging the commit directly to `develop` and duplicating the commit to hotfix branch.

Issue #2872, opened 2018-01-03
Add "Related terms" category for related links to glossentries


raducoravu:
When the relationship table defines links between a topic and others, those links are automatically grouped in the published output at the end of the topic HTML content in various categories: "Related topics", "Related tasks", "Related concepts" and "Related information".
Our particular end user would like to have a particular category for related links to glossentry topic types, category which could be called "Related terms". 

So for example in the translation files like "xsl/common/strings-en-us.xml" we would need a new mapping like:

          <str name="Related glossentrys">Related terms</str>  

which would be used in the XSLTs for HTML and PDF based outputs similar to the ones for concepts, tasks and references for the category labels when a topic has related links to glossary entries.

xephon2:
This requirement should be generalized. `<glossentry>` is just another specialized DITA topic. We need an easy to configure mechanism based on conventions, how users can quickly set this up for specialized topic types.

robander:
It _should_ be possible to do this by putting a column header on the reltable column. But, this doesn't seem to be working for me now. When I put the title in the column using `@navtitle` I get the issue reported in #2432. If I add a dummy `@href` to that reference the build continues but I get an ambiguous rule match for the glossary topic. If I use `<title>` inside the column instead, I also get the ambiguous rule match. In either case, my two glossary entries end up under "Related concepts".

So - that _should_ be the easy configuration mechanism, because it's part of the spec and _should_ work across implementations, it just ... doesn't seem to work now, or else I'm making some obvious error in the sample I quickly created.

All that being said - @raducoravu I wonder, were you asking for DITA-OT to add explicit / universal grouping support for glossary terms along with task/concept/reference, as opposed to "some ability to make new groups"?

robander:
Ahh update on the earlier response -- the `<relcolspec>` approach _is_ working when:

1. I fix the `$href` bug so that you can just specify a title
1. I fix the ambiguous rule message (for HTML5, that issue doesn't exist for XHTML)
1. I add a title to _every_ relcolspec -- if I only add to some of the columns, either the headers are ignored or links are ignored; if I add to all columns, headers are used and links appear properly

raducoravu:
Similar to what's done for tasks/concepts we could have also had out of the box support for naming the category "Related terms" when the relcolspec had a glossentry type:

           <relcolspec type="glossentry"/> 

But I think it's great that the generic way of naming categories now works because from what I understand customers can now name the categories whatever they want without the need of XSLT customizations.

[...]

PR #2873, opened 2018-01-04
Fix link group titles based on reltable columns

### LINKED ISSUE #2432: An empty sequence is not allowed as the value of parameter $href [DOT 2.3.1]
Using one of my client's samples I obtain an error when publishing to XHTML-based outputs, the error is something like this:

```
        maplink:
          [maplink] Error on line 418 of maplinkImpl.xsl:
          [maplink]   XPTY0004: An empty sequence is not allowed as the value of parameter $href
          [maplink]   at xsl:apply-templates (file:/D:/projects/eXml/frameworks/dita/DITA-OT2.x/xsl/preprocess/maplinkImpl.xsl#410)
          [maplink]      processing /map/reltable[1]/relheader[1]/relcolspec[1]/topichead[1]
          [maplink]   at xsl:apply-templates (file:/D:/projects/eXml/frameworks/dita/DITA-OT2.x/xsl/preprocess/maplinkImpl.xsl#335)
```

and it occurs at line:

```
    <xsl:template match="*[contains(@class, ' map/topicref ')]" mode="grab-group-title" as="xs:string?">
      <xsl:variable name="file-origin">
        <xsl:call-template name="get-file-uri">
          <xsl:with-param name="href" select="@href"/>
```

so somehow the topicref does not have an @href attribute set to it and the publishing breaks.
I added an xsl:message and it breaks when processing this topichead:

```
        <topichead xmlns:ditaarch="http://dita.oasis-open.org/architecture/2005/" xmlns:dita-ot="http://dita-ot.sourceforge.net/ns/201007/dita-ot" class="+ map/topicref mapgroup-d/topichead " xtrc="topichead:1;28:28" xtrf="file:/C:/Users/radu_coravu/Desktop/test/abbreviated/Producer_UG.ditamap" toc="no" print="no"><topicmeta class="- map/topicmeta " xtrc="topicmeta:2;29:32" xtrf="file:/C:/Users/radu_coravu/Desktop/test/abbreviated/Producer_UG.ditamap"><navtitle class="- topic/navtitle " xtrc="navtitle:1;30:35" xtrf="file:/C:/Users/radu_coravu/Desktop/test/abbreviated/Producer_UG.ditamap">See also</navtitle></topicmeta></topichead>
```

robander:
## Description

Fixes several issues related to relationship table column headings.

It should be possible to set a heading on a relationship table column, and have that heading used for the group of links generated by cells in that column. For example, it should be able to make a column for glossentry topics, set a title of "Related terms", and have generated links appear in a group titled "Related terms" (instead of with the default "Related concepts").

This is currently broken in several cases, as described below.

## Motivation and Context

1. First commit fixes #2432 -- if a relationship table heading uses a `navtitle` without a topic, the build currently fails with an XSLT error about `$href`. The fix checks to ensure `@href` is specified before setting and using the variable; if `@href` is not present the navigation title is used as expected.
1. Second commit is a simple fix in the same set of code to use the `text-only` mode, so that unwanted items like index terms won't be picked up from titles as they would be today with `xsl:value-of`.
1. Third commit fixes the import of glossary overrides for HTML5. All other topic-type overrides (concept, task, and reference) are part of the shell, as they should be; for some reason glossary is pulled into `topic.xsl`, and it's done with `xsl:include` so that we end up with precedence conflicts for link groups. I noted that error message in #2872.
1. Fourth commit: currently column headings only result in link group headings when _every_ column sets a heading. If only one column sets a heading, either link groups are broken (do not use the heading) or links are missing entirely. Tested with a 2-column relationship table with a concept in one column and two glossentry topics + a header in the second column. Also noticed when testing #2872.
1. Fifth commit: code assumes that if you have `<topicref>` inside of `<relcolspec>`, any file reference is local DITA. If I add a link to `dita-ot.org`, the build fails with the following; fixed so that we only try to read local DITA: `Error: Failed to run pipeline: Failed to read links from C:\DCS\2.5.3\temp\terms.ditamap: A sequence of more than one item is not allowed as the value of variable $file-origin ("file:/C:/DCS/2.5.3/temp/", "http://dita-ot.org")`
1. Sixth commit: even when reading local DITA, the code expects a path constructed as `xs:string` but the function for getting the file path returns two strings, resulting in the same type of failure: `Error: Failed to run pipeline: Failed to read links from C:\DCS\2.5.3\temp\terms.ditamap: A sequence of more than one item is not allowed as the value of variable $file-origin ("file:/C:/DCS/2.5.3/temp/", "reftopic.dita")`

## How Has This Been Tested?

- With a `relcolspec` heading that includes `<topicref navtitle="test"/>` but does not reference a topic (currently broken, now working)
- With a relationship table using a heading for one column but not the other (currently broken, now fixed)
- Tested with headings in each column (currently working for some cases, still working for those cases)
- Tested with headings that use `<title`>, `@navtitle`, `<navtitle>`, no title (non-DITA reference), and no title (DITA reference)

## Type of Changes

Fixes for several closely related issues.

## Checklist

X New test case added to integration test to cover preprocess of reltable headers


robander:
Found two additional bugs, code failed when using actual topic references in `<relcolspec>` -- updated description above to include a description of each condition.

infotexture:
@robander Looks like these changes mean we can close #2872 — or are there other aspects of that issue that remain unresolved?

robander:
@infotexture not exactly -- #2872 is a feature request, asking that we add support for "Related terms" as a _default_ heading, comparable to "Related tasks" or "Related concepts". It's just what keyed me into the fact that the current code was broken for single-column manual headers, resulting in this pull request (you had to customize headers for all or none).

Issue #2862, opened 2017-12-04
xref to web page kills PDF, hangs HTML5 for a while


robander:
## Expected Behavior

One of my customers had this link in a topic -- it's missing `@scope="external"` but is otherwise normal and worked in the past:
```
<p>For more information about each edition, see the 
<xref format="html"
href="http://www-01.ibm.com/common/ssi/ShowDoc.wss?docURL=/common/ssi/rep_ca/0/897/ENUS217-240/index.html"
>IBM 
  Software Announcement</xref>
```

Expected: the link works. Based on the specification, processing default for links beginning with http, https, ftp, and mailto should be treated as external, so my customers shouldn't be forced to add it (though I tend to encourage adding it), and certainly shouldn't get failing builds because of it:
https://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part1-base/langRef/attributes/thescopeattribute.html

## Actual Behavior

With a simple one-topic map, and that link in the topic, HTML5 hangs for an extended period at `topicpull`.

In 2.5.4, the HTML5 build otherwise works locally on my Windows machine, but when I run it on a Linux machine I get this error from the `topic-fragment` step -- I don't know why only there: 
`Error: java.lang.IllegalArgumentException: Only file URI scheme supported: http://www-01.ibm.com/common/ssi/ShowDoc.wss?docURL=/common/ssi/rep_ca/0/897/ENUS217-240/index.html`

I don't have 3.0 on that machine to test HTML5 there. But locally, with PDF and `preprocess2`, that link now kills the build in the merge step:
```
map2pdf2:
[topic-merge] Processing C:\Users\IBM_AD~1\AppData\Local\Temp\temp20171204144909813\e17ddccb7429a132b18bfa84b28d3bed6f844f1a.ditamap
[topic-merge] Processing C:\Users\IBM_AD~1\AppData\Local\Temp\temp20171204144909813\b7e631acad95a349a712d265a6b2a2d6792ac95e.dita
[topic-merge] null
Error: Failed to run pipeline: Failed to process merged topics: Supplied element name is a zero-length string
```

## Possible Solution

Per the specification, treat the 4 known prefixes as "external" when no other scope is specified or inherited (should be relatively straightforward for `xref` as in this case but not sure how this will affect cascading in maps).

[...]

PR #2877, opened 2018-01-11
Support incorrectly defined absolute external links

jelovirt:
When a topic has an external link without external `@scope` attribute, current processing has special handling that results in incorrect output in preprocess2. For example:

```xml
<xref href="http://example.com/topic.html" format="html">topic</xref>
<xref href="http://example.com/topic.dita">topic</xref>
```

This PR changes the preprocess validation filter to fix these in lax processing mode to:

```xml
<xref href="http://example.com/topic.html" format="html" scope="external">topic</xref>
<xref href="http://example.com/topic.dita" scope="external">topic</xref>
```

### DIFF
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -358,6 +358,16 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ075W" type="WARN">
+    <reason>Absolute link '%1' without correct 'scope' attribute.</reason>
+    <response>Using 'scope' attribute value 'external'.</response>
+  </message>
+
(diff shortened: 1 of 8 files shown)


robander:
This would fix all the broken cases I have in #2862 -- I've ever seen a case from my customers with `http` references that were local. And technically the spec says to default to external for those: http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part1-base/langRef/attributes/thescopeattribute.html

@raducoravu do you have a moment to comment on this? I wonder if you've seen local DITA references, without `@scope`, but using `http://`?

raducoravu:
@robander Yes, that's why I implemented such a check in our "Validate and check for completeness" action. I've not seen many cases though.

Issue #2874, opened 2018-01-07
strictTaskbodyConstraintMod.xsd refers to taskMod.xsd unversioned


nfekete:
## Expected Behavior
The reference to `taskMod.xsd` should be versioned as with all other references in the schema files.

## Actual Behavior
It references `taskMod.xsd` without version qualifier and it causes problems when running transformations that are based on versioned schema references with version tied to 1.2
The following command shows pretty well how `strictTaskbodyConstraintMod.xsd` is different from all other files:
```
find src/main/plugins/org.oasis-open.dita.v1_2/ -name *.xsd | xargs grep "redefine schemaLocation"
```

## Possible Solution
change schema reference from
```
<xs:redefine schemaLocation="urn:oasis:names:tc:dita:xsd:taskMod.xsd">
```
to
```
<xs:redefine schemaLocation="urn:oasis:names:tc:dita:xsd:taskMod.xsd:1.2">
```

## Steps to Reproduce
Run a transformation on a map that contains a task with `xsi:noNamespaceSchemaLocation="urn:oasis:names:tc:dita:xsd:task.xsd:1.2"`.

## Copy of the error message, log file or stack trace
The relevant log output line is:
```
src-resolve: Cannot resolve the name 'tasktroubleshooting' to a(n) 'group' component.
```

## Environment

* DITA-OT version: 2.5.4, possibly all others that support both DITA 1.2 and 1.3
* Operating system and version: independent of
* How did you run DITA-OT? `dita` command
* Transformation type: custom on top of PDF2

nfekete:
Also this command:
```
$ find src/main/plugins/org.oasis-open.dita.v1_2/ -name *.xsd | xargs grep "<xs:\(import\|include\|redefine\)" | grep -v "1\.2"
src/main/plugins/org.oasis-open.dita.v1_2/schema/technicalContent/xsd/ditabase.xsd:  <xs:include schemaLocation="urn:oasis:names:tc:dita:xsd:ditabaseMod.xsd"/>
src/main/plugins/org.oasis-open.dita.v1_2/schema/technicalContent/xsd/ditabaseMod.xsd:<xs:import namespace="http://www.w3.org/XML/1998/namespace"  />
src/main/plugins/org.oasis-open.dita.v1_2/schema/technicalContent/xsd/strictTaskbodyConstraintMod.xsd:  <xs:redefine schemaLocation="urn:oasis:names:tc:dita:xsd:taskMod.xsd">
```

[...]

PR #2878, opened 2018-01-12
Fix DITA 1.2 XSDs to use version-specific ID #2874

robander:

The DITA 1.2 Schemas were supposed to use version-specific identifiers in the `@schemaLocation` references. As noted in #2874, the version number is missing in two locations, which means the schemas do not work properly in environments where "latest module" inclusions resolve to DITA 1.3.

OASIS is not going to issue an errata for this, but could update the copy stored in SVN; either way, we should be able to fix the versions in DITA-OT so that toolkit users can use DITA 1.2 as intended.

### DIFF
--- a/src/main/plugins/org.oasis-open.dita.v1_2/schema/technicalContent/xsd/ditabase.xsd
+++ b/src/main/plugins/org.oasis-open.dita.v1_2/schema/technicalContent/xsd/ditabase.xsd
@@ -179,7 +179,7 @@
 
   </xs:redefine>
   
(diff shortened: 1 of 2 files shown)


robander:
@easirois As the one-time maintainer of these (now un-maintained) files, do you want to double check that this looks good? Pretty simple fix here, both of the location values with `:1.2` are in the catalog and used in other files.

drmacro:
Is it too late to make this fix in the 1.3 Errata 2?

drmacro:
Doing this search:

//*/@schemaLocation[ends-with(., '.xsd')]

Across all the files in the 1.3 schema/ files I only find ditabase.xsd as having an unqualified URI reference (that is not an SVG or MathML XSD).

PR #2882, opened 2018-01-27
Use ditafileset for all XSLT processing

jelovirt:
## Description
Use `<ditafileset>` custom [resource collection](https://ant.apache.org/manual/Types/resources.html#collection) in XSLT processing and replace built-in Ant `<xslt>` use with `<pipeline>` XSLT module.

## Motivation and Context
Using a custom task for XSLT processing will allow more control over XSLT, e.g. custom extension and using job configuration to select files.

### DIFF
--- a/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
+++ b/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
@@ -439,6 +439,7 @@ public final class ExtensibleAntInvoker extends Task {
     public static class FileInfoFilterElem extends ConfElem {
         private String format;
         private Boolean hasConref;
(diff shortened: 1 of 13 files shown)


robander:
Is there a reason the code has both `<ditafileset>` lower-case and `<ditaFileset>` mixed-case? I see one switched to lower, but some new ones were added with mixed.

jelovirt:
Backwards compatibility, because we've had both at one point. I'll take a look at harmonizing these.

robander:
OK, makes sense - just seemed odd to be adding a couple of `ditaFileset` while also changing one to `ditafileset` - left me not sure which is preferred. (At least, if I'm reading the commits correctly.)

PR #2891, opened 2018-02-06
Move HTMLHelp source and resources to HTML plugin JAR

jelovirt:
Move HTMLHelp related Java sources and resources to HTMLHelp plugin and bundle into the plugin JAR.


jelovirt:
@robander, I think this needs to be verified on Windows, just to be sure.

Issue #2187, opened 2016-01-08
Navtitle specified in topicref metadata seems to override title specified in topic [DOT 2.x]


raducoravu:
I have a topicref like:

``` xml
<topicref href="my_service.dita">
  <topicmeta>
    <navtitle>SHOULD NOT OVERRIDE</navtitle>
  </topicmeta>
</topicref>
```

and the published XHTML output should ignore the navtitle (because the @locktitle attribute has not been set) and use instead the title from the topic.
But somehow with DITA OT 2.x the title displayed in the table of contents is taken from the topicmeta.
This works properly with DITA OT 1.8.

jelovirt:
Is the issue that `navtitle` from map is used instead of `navtitle` in topic?

raducoravu:
The issue is that the navtitle element from the topicmeta is used instead of the topic title element. Even though the @locktitle is not set (and thus the title is not locked).
I cannot reproduce the issue if instead of topicmeta with navtitle I'm setting the navtitle as an attribute on the topicref.

eerohele:
Introduced in 77d3cd64eccf66868753f4853c7f86117daaa8bd.

raducoravu:
We had one more user reporting this problem. I added an extra vote on the issue.

raducoravu:
Another one of our users encountered this problem. As an extra behavior, when the topic also specifies a navtitle in its titlealts element:

    <titlealts>
        <navtitle>Topic Overview</navtitle>
        <searchtitle>Overview</searchtitle>
    </titlealts>

the table of contents for the XHTML output will concatenate the navtitle specified in the DITA Map with the navtitle set inside the topic.

robander:
Fixed with #2897

PR #2897, opened 2018-02-19
Fix processing for <navtitle> in maps

robander:
## Description

Fixes current processing of `<navtitle>` in maps, which does not respect the `@locktitle` attribute.

The `@navtitle` attribute is correctly implemented in OT:
* If `@locktitle="yes"`, that value stays in the map
* Otherwise we look in the topic for `<navtitle>` and then for `<title>` and use the found value to update the `@navtitle` value in the map

In DITA 1.2 the `@navtitle` attribute was deprecated on topic references in favor of `<navtitle>`. Several years ago our `preprocess` was updated to use that element; if needed, we move the value from the attribute to that element during preprocess. The element is supposed to work the same - it acts as a hint for the author, but should be ignored unless `@locktitle="yes"`: http://docs.oasis-open.org/dita/dita/v1.3/errata01/os/complete/part2-tech-content/langRef/base/navtitle.html#navtitle
> When `<navtitle>` is used in a map, it functions in the same way as the `@navtitle` attribute; both are used to specify a navigation title for the target of the `<topicref>` element. That is, the title itself will only be used as an actual navigation title when the title is locked; the title is locked when the `<topicref>` element sets the attribute `locktitle="yes"`. If the title is not locked, processing systems will typically retrieve the current title from the target topic, looking first for a `<navtitle>` element and second for the general `<title>`.

The problem is that in our `MoveMetaModule` we push `<navtitle>` from the map into topics along with other metadata. (We've never pushed `@navtitle`.) Our code that updates navigation titles in the map comes later in preprocess. It is supposed to replace the `navtitle` in map with the topic title, but when it looks in the title, it _always_ finds the same navigation title that was pushed in earlier. Result: it replaces the navtitle with itself, and the actual topic title is never used. [It's probably good that we push this in when it is locked -- it's like saying "Yes, this really is the navtitle we want for this topic" -- but doing so when not locked is wrong. And I don't see any reason to update to do this for the deprecated `@navtitle`.]

I found a few ways to fix this that all seemed like late-preprocess-hacks (like updating `mappull` to check if the `navtitle` seemed to come from a map, and ignoring it). I'm not sure this fix is ideal, but the general idea is just to avoid pushing navtitles into topics _unless_ they are locked:

- When we find `<navtitle>` in the map, annotate it with `dita-ot:locktitle` that uses either the actual `@locktitle` or the default `no`
- When we write metadata into topics, skip writing navigation titles unless `dita-ot:locktitle="yes"`. (It's possible you might end up with an empty `<titlealts/>` but that's already legal and not a problem.)

As usual I'm not sure if this is the optimal code path, but the logic (don't push unless locked) seemed a lot better than the first couple approaches I took of pushing in all cases and then trying to ignore later.

## Motivation and Context

OASIS has been discussing the plan to remove `@navtitle` for DITA 2.0: https://lists.oasis-open.org/archives/dita/201802/msg00059.html

`@navtitle` is used pretty heavily in some of my real and test maps, usually as a hint that isn't updated as the real title gets updated. I tried migrating some of those to use `<navtitle>` and noticed that this changed the TOC, all the navtitles that were previously ignored now showed up.

@drmacro mentioned that he'd seen this happening with translated docs, where the English navigation titles that should be ignored still showed up in TOC. I haven't seen his samples but I'm _guessing_ the maps used `<navtitle>`, because `@navtitle` has been working properly for years.

## How Has This Been Tested?

Ran and updated the existing test cases for changes.

Added a new integration test that ensures we test the `MoveMeta` module together with `mappull`. The new test attempts to test all combos, using a topic that has `<title>` but no `<navtitle>`:

1. Reference with no navigation title
1. Reference with `@navtitle` and no `@locktitle`
1. Reference with `@navtitle` and `@locktitle="no"`
1. Reference with `@navtitle` and `@locktitle="yes"`
1. Reference with `<navtitle>` and no `@locktitle`
1. Reference with `<navtitle>` and `@locktitle="no"`
1. Reference with `<navtitle>` and `@locktitle="yes"`
1. Reference with both `@navtitle` and `<navtitle>` and `@locktitle="yes"`

Then repeat tests 2 through 8 with topics that have `titlealts/navtitle` already specified.

Can be easily verified by building the new sample `TestingLocktitle.ditamap` to HTML5 and checking the TOC.

## Type of Changes

## Checklist


raducoravu:
Is this a fix which potentially solves also:
https://github.com/dita-ot/dita-ot/issues/2187
?

robander:
@raducoravu looks like it is the same issue, yes. I caught this in my own content and worked on a fix without looking for existing reports. I think this is the first time I've _accidentally_ tried to fix on one from your "list of issues that need to be fixed" :-)

robander:
Tagging with `P1` because this is in Radu's list of high priority items and because it's critical to fix as we get closer to OASIS officially removing `@navtitle`.

robander:
@infotexture I don't think so - I think that one might actually have been resolved in 2.5.1 with #2719 -- that one fixed several older issues so I may have missed one when searching for things it covered. If so, we should go ahead and close it (and possibly update those old release notes for 2.5.1).

Issue #2917, opened 2018-03-16
Image with keyref not included in output


jlowther:
## Expected Behavior
I'm trying to build XHTML output using DITA-OT 3.0.2.

My map contains the following key definition:

     <keydef keys="AddButton" href="images/AddButton.png" format="png"/>

I expect AddButton.png to be copied to the output folder. 

## Actual Behavior
The image is not copied to the output folder and the HTML page displays a broken link. 

## Steps to Reproduce

1. Create a bookmap and insert the keydef in the backmatter element:
<backmatter>
  <keydef keys="AddButton" href="sfos/images/AddButton.png" format="png" />
 </backmatter>
2. Build output using DITA-OT 3.0.2.
3.View output folder. The image file is not present. 

## Copy of the error message, log file or stack trace
No error message present in output log.

## Environment
Windows 10

* DITA-OT version: 3.0.2
* Operating system and version:
Windows 10
* How did you run DITA-OT?
Ant, dita command
* Transformation type:
XHTML

raducoravu:
I can reproduce the problem on my side with DITA OT 3.0.2, I'm attaching a sample project.
[imageKeyref.zip](https://github.com/dita-ot/dita-ot/files/1824254/imageKeyref.zip)

Somehow for direct image references (via @href) the "image.list" file in the temp files folder is properly populated. Also the .job.xml uses "format='image'" when referring to the image. 
For indirect images the "image.list" does not contain the image references.
But I'm not sure where the problem is.

jelovirt:
I think the cause is that `copy-image` task is ran in the middle of `preprocess`, and at that point the image format is just the file extension, `png`. When image metadata processing is ran and we collect all image references, the format is changed to `image`.

The fix to this would be to move the image copy to the end of preprocessing or beginning of HTML tasks. Issue #1988 is basically about implementing this.

I'm not sure what would be the best fix for this in 3.0.x because we can't really move the responsibility of copying to HTML in a bug fix release. @robander how do you feel about just moving `copy-files` to the end of `preprocess` to remedy this?

jelovirt:
Tested this and it's not enough to just move `copy-files` to end of `preprocess`. Copy should take place after `xhtml.image-metadata` and that's part of XHTML tasks. An alternative workaround would be to replace

```xml
<copy todir="${copy-image.todir}" failonerror="false">
  <ditafileset format="image" />
</copy>
```

with

```xml
[...]

PR #2921, opened 2018-03-21
Use configured image extensions as fallback to image format #2917

jelovirt:
Use configured image extensions as fallback to `image` format, because correct `image` format is not populated until image metadata is read.

This is a workaround and should be replaced by moving image copy to transtype specific code.

### DIFF
--- a/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
+++ b/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
@@ -439,7 +439,7 @@ public final class ExtensibleAntInvoker extends Task {
     }
 
     public static class FileInfoFilterElem extends ConfElem {
-        private Set<String> format = Collections.emptySet();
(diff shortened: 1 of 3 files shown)


raducoravu:
Couple of cosmetic remarks:

Maybe instead of making the  "org.dita.dost.util.FileUtils.supportedImageExtensions" a public field we could add a getter method for it FileUtils.getSupportedImageExtensions().
Also the "supportedImageExtensions" is right now marked as deprecated. If we want this fix as a long term solution, we should probably remove the deprecated marker from the field.

In JobSourceSet maybe rename the Set<String> format => to "formats" or "filterFormats" to better indicate there are multiple formats which will be used to filter the job file.

jelovirt:
The codebase tries to avoid getter methods and favoring public final fields.

Since we still consider the image extension list a deprecated feature, and only use it as a workaround in this case, I think we should keep it deprecated.

Issue #2494, opened 2016-09-28
Cannot conref from one submap to another


raducoravu:
I'm attaching a sample:

[topicref-conref-problem.zip](https://github.com/dita-ot/dita-ot/files/497186/topicref-conref-problem.zip)

Basically one submap conrefs to an ID set on a topicref in another submap. The conref is not expanded at all in the output.
In my opinion the problem is that the [conref] stage applies only on the submap which has the actual original conref, when in fact it should apply on all DITA Maps, because all DITA Maps have been expanded by the previous [mapref] stage.

raducoravu:
As a possible quick fix in "build_preprocess.xml" the target named "conref" besides having a fileset like this:

            <ditaFileset conref="true"/>

could have an additional file set with all DITA Maps:

               <ditaFileset format="ditamap"/>

raducoravu:
Or have two conref stages, one for DITA Maps which is applied before mapref, and one for topics which is applied later on...

jelovirt:
The correct fix is to update job configuration after map merge, to cascade conref flag up to the main map from submaps.

robander:
@raducoravu or @jelovirt do you know if this might be fixed with `preprocess2`? If so it might still need to be open against `preprocess` but it would be nice if we've already made some progress.

raducoravu:
I tested but it still does not work with the default DITA OT 3.0.1 processing (which should be from what I know proprocess2).

robander:
Just FYI -- `preprocess2` is now the default for PDF, but not for XHTML / HTML5, so depending on your transform type you can still get either one.

raducoravu:
I did not know that, from what I remember I tried also with PDF and it did not work though.

jelovirt:
IMO the fix that after map merge we check the map files for conref is still the correct one, but the fix is a bit tricky because map merge is implemented with XSLT. Thus we have no simple post-tranformation hook where we can check whether `@conref` attributes have been added. If we change the code to not just use `<pipeline><xslt>`, then that cannot really go into a bug fix release due to the extend of the changes. So this should probably go to 3.1.

robander:
No strong preference - sounds like there are advantages to both. It sounds like the latter approach with dedicated module might be _slightly_ better because if other things like this come up in the future they'll be easier to handle. But yeah that would need to be for 3.1.

jelovirt:
@robander #2875 fixes this. Are you able to review?

robander:
[...]

PR #2928, opened 2018-04-09
New test case with mapref to conref

robander:

## Description

Currently includes only the test case, have not made the fix yet but working on it.

## Motivation and Context

Currently processing with both `preprocess` and `preprocess2` will not update the job configuration file after running the `mapref` resolution.

I have a root map that refers to an intermediate map using `<mapref>`; the intermediate map then uses `@conref` to pull in a branch from a third map. In the job configuration, only the intermediate map is marked with `hasconref="true"`. The root map does not have that marker because it does not have `@conref` at the start of the process.

After the `mapref` resolution runs, the content with `@conref` is pulled into the root map; at that point, the root map _does_ have conref, but it's still missing in the job configuration.

When the `conref` or `map-conref` step runs later, it's currently skipping the root map, so the conref never resolves.

The job configuration should be updated to indicate that the root map has conref when that's the result of `mapref`.

## How Has This Been Tested?

New test case attached as part of the pull request. Easiest to see the problem when building to PDF, where the topics pulled in via `@conref` are simply missing.

## Type of Changes

## Checklist


robander:
@jelovirt noticed this is a duplicate of #2494 which is fixed in 3.1. It used a new Java module to help make the change so unlikely to be easily ported back for a 3.0 hotfix.

Jarno also noticed that the test case I initially pushed up is incorrect (fixing the problem results in new parent/child links that I didn't get into the expected topic output). I'll redo this and change the target branch to `develop` ... would be good to have this case combining 2 major features as part of our integration test.

Issue #2870, opened 2017-12-22
Cannot produce EclipseHelp with DITA OT 3.0.1


raducoravu:
Eclipsehelp "plugin.xml" does not refer to "eclipsehelp.jar". So with DITA OT 3.0.1, publishing to Eclipse Help using the command line:

    dita -i D:\projects\eXml\test\dita\transform\hierarchy.ditamap -f eclipsehelp

generates these problems:

            dita.out.map.eclipse.index:
                 [echo]  args.eclipsehelp.indexsee = false 
             [pipeline] [Fatal Error] :-1:-1: Premature end of file.
             [pipeline] [DOTJ013E][ERROR] Failed to parse the referenced file 'file:/D:/projects/eXml/test/dita/transform/temp/eclipsehelp/oxygen_dita_temp/tasks/flowers.pdf'.Premature end of file.
             [pipeline] java.lang.ClassNotFoundException: org.dita.dost.writer.EclipseIndexWriter
             [pipeline] 	at java.net.URLClassLoader.findClass(Unknown Source)
             [pipeline] 	at java.lang.ClassLoader.loadClass(Unknown Source)
             [pipeline] 	at java.lang.ClassLoader.loadClass(Unknown Source)
             [pipeline] 	at java.lang.Class.forName0(Native Method)
             [pipeline] 	at java.lang.Class.forName(Unknown Source)
             [pipeline] 	at org.dita.dost.index.IndexTermCollection.outputTerms(IndexTermCollection.java:165)
             [pipeline] 	at org.dita.dost.module.IndexTermExtractModule.execute(IndexTermExtractModule.java:77)
             [pipeline] 	at org.dita.dost.ant.ExtensibleAntInvoker.execute(ExtensibleAntInvoker.java:163)
             [pipeline] 	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:293)
             [pipeline] 	at sun.reflect.GeneratedMethodAccessor6.invoke(Unknown Source)
             [pipeline] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
             [pipeline] 	at java.lang.reflect.Method.invoke(Unknown Source)
             [pipeline] 	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:106)
             [pipeline] 	at org.apache.tools.ant.Task.perform(Task.java:348)
[...]

PR #2932, opened 2018-04-12
Add end to end tests for transtype regression test

jelovirt:
## Description
Add a new test to run DITA-OT transtypes to catch regression bugs. The test doesn't compare actual output with expected output, it only runs a transtype.

## Motivation and Context
The goal is to run every transtype during testing phase, even if the output cannot be verified to be correct. This allows us to catch regression bugs that break the whole transtype process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
 - git submodule update --init --recursive
 - "./gradlew"
 script:
-- "./gradlew check integrationTest --info --stacktrace --no-daemon"
+- "./gradlew check integrationTest e2eTest --info --stacktrace --no-daemon"
 after_success:
 - "./gradlew dist --stacktrace --no-daemon -Pcommit=${TRAVIS_COMMIT} -Ptag=${TRAVIS_TAG} -PskipGenerateDocs=${SKIP_GENERATE_DOCS}"
 notifications:
--- a/build.gradle
+++ b/build.gradle
@@ -67,6 +67,7 @@ artifacts {
 
 test {
     exclude '**/IntegrationTest*.class'
+    exclude '**/EndToEndTest*.class'
 }
 
 // Integration test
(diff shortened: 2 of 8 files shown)


jelovirt:
Using those files was a random choice. It might be a good idea to create a simple E2E source set.

robander:
I think if we want to go for simple -- for example, goal is to make sure the code that runs after preprocess actually compiles and finishes -- then I think that's a good test.

If we want a bigger / valid test set that uses lots of DITA features, our own docs might be a good set? But if we compare output, then that's going to need us to update the `exp/` copy much more often.

infotexture:
> If we want a bigger / valid test set that uses lots of DITA features, our own docs might be a good set? 

👍

xephon2:
If you modify a docs topic, you have to modify the compared expected result as well. Could be a show stopper and eat to much docs resources.

jelovirt:
The problem with using latest docs for regression testing is that

* if the docs are broken the test will fail, even though there is no regression bug
* the docs set is larger than needed for testing for regression, thus making the test run longer.

infotexture:
The idea here would be not to use the bleeding edge docs state from `develop`, but the last released version from the `master` branch, which should never be “broken” otherwise they wouldn’t have been released.

infotexture:
As for test time, both HTML & PDF  builds together take about 1m to run. HTML alone is much faster.

robander:
Note on my approval -- I think this is a good first step -- having an as-simple-as-possible test case that runs through each of our transform types as the minimum end-to-end test is a big improvement on what we have and meets the goal of this enhancement.

I still think having a stable version of our docs in there as an additional end-to-end test would be a good idea, because we've seen that it is good at catching failures that don't show up elsewhere, but I'm not sure of the best way to add that in.

thendarion:
will this catch stuff like #2870 in the future? :)

jelovirt:
@thendarion Right now we E2E will not run the tests via `dita` command, so no. Maybe the E2E test harness will be changed to do this.

Issue #2907, opened 2018-03-04
Remove leading spaces in coderefs


xephon2:
I really like the coderef feature to pick certain parts of config files for explanation.

```xml
<codeblock outputclass="language-xml">
  <coderef href="DBService_App.config#token=snippet-START,snippet-END"/>
</codeblock>
```

**Problem**

It would be a great enhancement, if the indentation of these blocks would be fixed. Some codeblocks have very long lines, that look messy in the out. Think of this block:

```xml
            <property name="dataStorageConfiguration">
                <bean class="org.apache.ignite.configuration.DataStorageConfiguration">
                    <property name="defaultDataRegionConfiguration">
                        <bean class="org.apache.ignite.configuration.DataRegionConfiguration">
                            <property name="name"
                                      value="ContactGatewayDefault"/>
                            <property name="initialSize"
                                      value="#{1L * 1024 * 1024 * 1024}"/>
                            <property name="maxSize"
                                      value="#{4L * 1024 * 1024 * 1024}"/>
                            <property name="metricsEnabled"
                                      value = "true"/>
                            <property name="persistenceEnabled"
                                      value="false"/>
                        </bean>
                    </property>
                </bean>
            </property>
```

To make this block look good, you have to manually remove the leading spaces to make it look like this:

```xml
<property name="dataStorageConfiguration">
    <bean class="org.apache.ignite.configuration.DataStorageConfiguration">
        <property name="defaultDataRegionConfiguration">
            <bean class="org.apache.ignite.configuration.DataRegionConfiguration">
                <property name="name"
                          value="ContactGatewayDefault"/>
                <property name="initialSize"
                          value="#{1L * 1024 * 1024 * 1024}"/>
                <property name="maxSize"
                          value="#{4L * 1024 * 1024 * 1024}"/>
                <property name="metricsEnabled"
                          value = "true"/>
                <property name="persistenceEnabled"
                          value="false"/>
            </bean>
        </property>
    </bean>
</property>
```

That is boring and expensive work. It makes the file looking good for the tech writer, but ugly for the dev, because it destroys the indentation of the original file.

**Solution**
All referenced lines of a code block need to be parsed. If all lines have leading spaces, find the line with the fewest spaces, count the spaces and remove the amount of spaces from all referenced lines.

jelovirt:
[...]

PR #2953, opened 2018-05-05
Remove leading spaces in coderefs #2907

jelovirt:
Add filter to remove leading whitespace from `codeblock` elements. Implements #2907.

### DIFF
--- a/src/main/java/org/dita/dost/util/SaxCache.java
+++ b/src/main/java/org/dita/dost/util/SaxCache.java
@@ -1,0 +1,133 @@
+/*
+ * This file is part of the DITA Open Toolkit project.
+ *
+ * Copyright 2018 Jarno Elovirta
+ *
+ * See the accompanying LICENSE file for applicable license.
+ */
+
+package org.dita.dost.util;
+
+import org.xml.sax.Attributes;
+import org.xml.sax.ContentHandler;
+import org.xml.sax.SAXException;
+
+public class SaxCache {
+
+    public interface SaxEvent {
+        void write(ContentHandler handler) throws SAXException;
+    }
+
+    public static class StartPrefixMappingEvent implements SaxEvent {
+        public final String prefix;
+        public final String uri;
+
+        public StartPrefixMappingEvent(String prefix, String uri) {
+            this.prefix = prefix;
+            this.uri = uri;
+        }
+
+        @Override
+        public void write(ContentHandler handler) throws SAXException {
+            handler.startPrefixMapping(prefix, uri);
+        }
+    }
+
+    public static class EndPrefixMappingEvent implements SaxEvent {
+        public final String prefix;
+
+        public EndPrefixMappingEvent(String prefix) {
+            this.prefix = prefix;
+        }
+
+        @Override
+        public void write(ContentHandler handler) throws SAXException {
+            handler.endPrefixMapping(prefix);
+        }
+    }
+
+    public static class StartElementEvent implements SaxEvent {
+        public final String uri;
+        public final String localName;
+        public final String qName;
+        public final Attributes atts;
+
+        public StartElementEvent(String uri, String localName, String qName, Attributes atts) {
+            this.uri = uri;
+            this.localName = localName;
+            this.qName = qName;
+            this.atts = atts;
+        }
+
+        @Override
+        public void write(ContentHandler handler) throws SAXException {
+            handler.startElement(uri, localName, qName, atts);
+        }
+    }
+
+    public static class EndElementEvent implements SaxEvent {
+        public final String uri;
+        public final String localName;
+        public final String qName;
+
+        public EndElementEvent(String uri, String localName, String qName) {
+            this.uri = uri;
+            this.localName = localName;
+            this.qName = qName;
+        }
+
+        @Override
+        public void write(ContentHandler handler) throws SAXException {
+            handler.endElement(uri, localName, qName);
+        }
+    }
+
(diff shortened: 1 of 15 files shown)


jelovirt:
@robander Do you think we should make the outputclass that triggers whitespace normalization configurable via plugin extension point, or just hardcode `normalize-space` for now?

src/main/plugins/org.dita.base/build_preprocess2_template.xml
```diff
@@ -354,6 +354,7 @@ See the accompanying LICENSE file for applicable license.
           <param name="processing-mode" value="${processing-mode}" if:set="processing-mode"/>
         </filter>
         <filter class="org.dita.dost.writer.CoderefResolver" unless:set="preprocess.coderef.skip"/>
+        <filter class="org.dita.dost.writer.NormalizeCodeblock" unless:set="preprocess.codeblock.skip"/>
```
jelovirt: Is this a good guard condition property name? Maybe `preprocess.normalize-codeblock.skip` would be better?

src/main/java/org/dita/dost/util/SaxCache.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    public static class CharactersEvent implements SaxEvent {
+        public final char[] ch;
+        public final int start;
+        public final int length;
+
+        public CharactersEvent(char ch[], int start, int length) {
+            this.ch = ch;
```
jelovirt: Maybe this should make a copy of the char array, to prevent modification and allow GC to collect the array.

src/main/plugins/org.dita.base/build_preprocess2_template.xml
robander: I think `preprocess.normalize-codeblock.skip` is probably better? It feels too long -- but like many of the `skip` conditions, it will probably only be seen by / used by a small number of people, so length is not a big issue.

robander:
Missed that question 6 days ago -- I don't think it's really necessary to make that configurable, seems like if you want to use the feature you can use that token?

Looks good, only added one comment (to answer existing question from @jelovirt ).

Only thing I wonder is if there is any need to add a test case like this, where the sub-element includes newlines; I remember, way back before DITA in our SGML days, we had some parsers that reliably broke down on this - but it may not be an issue anymore. I've usually seen this as a creative way to use add a profiled line into a codeblock (filtered out for some conditions):
```
<codeblock>   test<b>
      stuff here</b>
   and finish</codeblock>
```

src/main/plugins/org.dita.base/build_preprocess2_template.xml
jelovirt: Changed

jelovirt:
Additional test with nested linebreak added

infotexture:
@jelovirt Found an edge case where non-whitespace characters are trimmed from the first line in the codeblock after these changes.

In the “Branch filtering” section of the [DITA features in docs](http://www.dita-ot.org/dev/reference/DITA-features-in-docs.html) topic, there's a code block with hanging indentation, where the first line is indented less than the lines that follow. The source of this code block is a coderef on [Line 55](https://github.com/dita-ot/docs/blob/develop/reference/DITA-features-in-docs.dita#L55) that pulls a range of [several lines from the publishing.ditamap](https://github.com/dita-ot/docs/blob/develop/topics/publishing.ditamap#L7-L17).

```xml
<topicref href="using-dita-command.dita" keys="build-using-dita-command" locktitle="yes">
  <topicmeta>
    <navtitle>Using the <cmdname>dita</cmdname> command</navtitle>
  </topicmeta>
  <ditavalref href="../resources/expert.ditaval">
    <ditavalmeta>
      <dvrResourcePrefix>build-</dvrResourcePrefix>
    </ditavalmeta>
  </ditavalref>
```

After the changes in this PR, the hanging indentation is removed when the leading whitespace is normalized — the first two characters are missing from the first line: `<t`.

```xml
opicref href="using-dita-command.dita" keys="build-using-dita-command" locktitle="yes">
<topicmeta>
  <navtitle>Using the <cmdname>dita</cmdname> command</navtitle>
</topicmeta>
<ditavalref href="../resources/expert.ditaval">
  <ditavalmeta>
    <dvrResourcePrefix>build-</dvrResourcePrefix>
  </ditavalmeta>
</ditavalref>
```

While this particular case is certainly unusual and could be worked around in source by referencing a different range of lines with uniform indentation, the normalization process should probably guard against this sort of thing to ensure that only whitespace characters are deleted — never actual content.

infotexture:
Found another case today where non-whitespace content is removed from the beginning of each line.

In a codeblock that pulls in the following [properties file](https://github.com/dita-ot/docs/blob/develop/samples/properties/pdf.properties) via `<coderef>`,

```properties
output.dir = out/pdf
args.gen.task.lbl = YES
args.rellinks = nofamily
```

the first 6 characters are stripped from each line when `@outputclass="normalize-space"`:

```properties
.dir = out/pdf
en.task.lbl = YES
ellinks = nofamily
```

Of course in this case there's no leading whitespace to adjust, so the problem can be solved by simply removing the `normalize-space` keyword from the outputclass.

But since the primary use case for this option is to normalize the content of external `<coderef>` sources that may change after the `<codeblock>` markup is authored, it might be useful to be able to enable normalization in advance, and trust that it will do no harm.

infotexture:
The remaining character loss issues described above were resolved with 8f16633.

PR #2960, opened 2018-05-14
Add support for SVG domain elements

jelovirt:
Add support for [SVG domain elements](https://docs.oasis-open.org/dita/dita/v1.3/os/part2-tech-content/langRef/containers/SVG-elements.html).

### DIFF
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -83,6 +83,7 @@ public final class KeyrefPaser extends AbstractXMLFilter {
         ki.add(new KeyrefInfo(TOPIC_DATA, ATTRIBUTE_NAME_HREF, false, true));
         ki.add(new KeyrefInfo(TOPIC_DATA_ABOUT, ATTRIBUTE_NAME_HREF, false, true));
         ki.add(new KeyrefInfo(TOPIC_IMAGE, ATTRIBUTE_NAME_HREF, false, true));
+        ki.add(new KeyrefInfo(SVG_D_SVGREF, ATTRIBUTE_NAME_HREF, true, false));
         ki.add(new KeyrefInfo(TOPIC_LINK, ATTRIBUTE_NAME_HREF, false, true));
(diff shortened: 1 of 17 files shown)


src/main/java/org/dita/dost/module/CopyToModule.java
```diff
@@ -251,7 +251,7 @@ private void copyFileWithPIReplaced(final URI src, final URI target, final URI c
         public void startElement(final String uri, final String localName, final String qName, final Attributes atts)
                 throws SAXException {
             Attributes resAtts = atts;
-            if ((TOPIC_XREF.matches(atts) || TOPIC_LINK.matches(atts) || TOPIC_IMAGE.matches(atts))
+            if ((TOPIC_XREF.matches(atts) || TOPIC_LINK.matches(atts) || TOPIC_IMAGE.matches(atts) || SVG_D_SVGREF.matches(atts))
```
robander: I don't think the addition here does anything, because `<svgref>` is specialized from `xref` so should already match the first condition.

jelovirt: 👍

src/main/java/org/dita/dost/writer/KeyrefPaser.java
```diff
@@ -83,6 +83,7 @@
         ki.add(new KeyrefInfo(TOPIC_DATA, ATTRIBUTE_NAME_HREF, false, true));
         ki.add(new KeyrefInfo(TOPIC_DATA_ABOUT, ATTRIBUTE_NAME_HREF, false, true));
         ki.add(new KeyrefInfo(TOPIC_IMAGE, ATTRIBUTE_NAME_HREF, false, true));
+        ki.add(new KeyrefInfo(SVG_D_SVGREF, ATTRIBUTE_NAME_HREF, false, true));
```
robander: Also not sure if this is going to do anything; there is a check for `TOPIC_XREF` a few lines down that should also catch any specialized `<xref>`. But I might be missing something else that is gained by having an entry for `SVG_D_SVGREF`?

jelovirt: This is actually needed, because `xref` can contain mixed content, and `svgref` is empty. Thus they need separate element configurations.

PR #2964, opened 2018-05-17
Support alt text for images in PDF

robander:

## Description

FOP, Antenna House, and XEP all have extensions to support alternate text in images. The alternate text is embedded with the image when formatters are produce Tagged PDF output (producing tagged output may be a configuration option on the formatter).

Today, for all FO output, the alternate text is discarded. This update ensures that the alternate text is passed to the FO file using the appropriate extension for each formatter; when the formatter produces tagged / accessible PDF, the alternate text will be used automatically without the need for additional plugins or customizations.

## Motivation and Context

In chatting with @xephon2 he mentioned the FOP extension for alternate text. This also came up as a question for some of my customers recently.

Even though the tagged PDF output is off by default in FOP, we should include the alternate text so that it is already there when somebody turns on tagging. (In general, we should preserve all possible content, and let the rendering step determine whether it is discarded.)

AH and XEP each have their own extensions to support this text, so I've added the update for all three formatters.

## How Has This Been Tested?

- Added alternate text to an image
- Tested to ensure the expected attribute is present in all three copies of `topic.fo`
- Ensured that when tagging is _off_, the alt text does not show up unexpectedly (tested for FOP and AH; I don't have a working copy of XEP to try out, so I've just tested that the attribute matches the one in an [XEP white paper](http://www.renderx.com/Images/Section_508_Showcase.pdf))

## Type of Changes

Sort of a bug fix (we should not have discarded the text), more of an enhancement (improving image processing by supporting alt-text).

### DIFF
--- a/src/main/plugins/org.dita.pdf2.axf/xsl/fo/topic2fo_shell_axf_template.xsl
+++ b/src/main/plugins/org.dita.pdf2.axf/xsl/fo/topic2fo_shell_axf_template.xsl
@@ -16,6 +16,7 @@ See the accompanying LICENSE file for applicable license.
   <xsl:import href="../../cfg/fo/attrs/index-attr_axf.xsl" />
   <xsl:import href="root-processing_axf.xsl"/>
   <xsl:import href="index_axf.xsl"/>
+  <xsl:import href="topic_axf.xsl"/>
   
   <dita:extension id="dita.xsl.xslfo" behavior="org.dita.dost.platform.ImportXSLAction" xmlns:dita="http://dita-ot.sourceforge.net"/>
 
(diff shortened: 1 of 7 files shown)


infotexture:
@robander I have an XEP license, so I should be able to test this if you can provide your test case(s) & instructions on how to verify (as you did for FOP & AH).

robander:
I think @simalot is testing already, but thanks -- I've already tested that the attribute matches the docs, so I was really just testing to ensure that when output is NOT tagged output the attribute does cause any problems. Basically making sure we do no harm by adding what's documented -- make sure that the alt text doesn't suddenly show up next to the image.

PR #2967, opened 2018-05-21
Use SLF4J with FOP

jelovirt:
FOP logging from PDF2 has used a custom log formatter where each log message takes two lines. This PR adds configuration to use SLF4J for FOP, thus allowing to configure the output format. The log messages are all at INFO level, because they are piped through STDERR.

### DIFF
--- a/src/main/config/logback.xml
+++ b/src/main/config/logback.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+      <pattern>[%level] %msg%n</pattern>
+      <!--pattern>
+        %d{HH:mm:ss.SSS} [%thread] %-5level %logger - %msg%n
+      </pattern-->
+    </encoder>
+  </appender>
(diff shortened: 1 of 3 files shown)


infotexture:
> This PR adds configuration to use SLF4J for FOP, thus allowing to configure the output format.

Is this something that users would want/need to configure (and thus require docs), or simply an internal change that implements a new logging format that would be used without user intervention?

jelovirt:
It's something users could in theory configure, but they don't need to. Unless they want to see FOP debug logging.

infotexture:
@jelovirt If users _did_ want access to FOP debug logging, how would they go about enabling that?

@robander Do you think that use case is common enough that we should explain it in the docs?

jelovirt:
Modify `config/logback.xml` or add your own `logback.xml` into classpath with higher priority. [Logback configuration docs](https://logback.qos.ch/manual/configuration.html)

robander:
@infotexture missed that question somehow. If I understand correctly, this means all FOP messages are info level and would be hidden from the console by default. For most cases I like that (there is generally a lot of noise that's not useful for our builds) but sometimes real issues pop out, so it might be good to highlight either how to get them out or highlight that if FOP crashes you might be able to get more information by turning on debugging?

PR #2970, opened 2018-05-22
Inline toc mode from map to HTML5 stylesheet into cover stylesheet

jelovirt:
This fixes the double import of map2html5Impl.xsl stylesheet.

### DIFF
--- a/src/main/java/org/dita/dost/module/XsltModule.java
+++ b/src/main/java/org/dita/dost/module/XsltModule.java
@@ -30,6 +30,7 @@ import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


jelovirt:
@infotexture I think I've fixed this now. Are you able to test?

This really needs integration tests.

infotexture:
> This really needs integration tests.

Yes, this is why I firmly believe that a snapshot of the _last released version_ of our own docs should be used for integration testing as suggested in [various](https://github.com/dita-ot/dita-ot/pull/2932#issuecomment-380858922) [different](https://github.com/dita-ot/dita-ot/pull/2932#issuecomment-380882053) [comments](https://github.com/dita-ot/dita-ot/pull/2932#issuecomment-390408508) in #2932.

That would help to “fail faster” so that this sort of regression is recognized earlier in the development process.

PR #2975, opened 2018-05-26
Show whitespace characters in codeblock in PDF

jelovirt:
Add support for showing invisible whitespace characters and defining tab character width in `codeblock` elements in PDF. Add outputclass based trigger for line numbering.

test

New variables and template modes:

* `codeblock.tab-width` variable to define width of tab characters. Defaults to `4`.
* `codeblock.show-whitespace` template mode to define whether whitespace characters are made visible. Default is to check `show-whitespace` in `@outputclass`.
* `codeblock.whitespace-character.space` template mode to output a visible space character. Defaults to `·`.
* `codeblock.whitespace-character.tab` template mode to output a visible tab character. Defaults to `→   `.

New attribute sets:
* `codeblock.whitespace` to define visible whitespace styling.

Changed defaults:
* `codeblock.generate-line-number` template mode default has been changed to check `show-line-numbers` in `@outputclass`.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/pr-domain-attr.xsl
+++ b/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/pr-domain-attr.xsl
@@ -64,6 +64,10 @@ See the accompanying LICENSE file for applicable license.
       <xsl:attribute name="color">gray</xsl:attribute>
       <xsl:attribute name="padding-end">1em</xsl:attribute>
     </xsl:attribute-set>
+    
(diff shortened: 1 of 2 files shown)


infotexture:
👍 This looks like a very handy feature to have.

I see the current implementation is for PDF output only. Would it make sense to implement these features for HTML output as well? 

It's certainly more vital in print contexts, where users cannot simply copy/paste to get the exact indentation, etc. — but it in some cases it might be handy to have visible whitespace in HTML too.

Issue #2974, opened 2018-05-25
Update to Apache FOP 2.3


raducoravu:
Apache FOP 2.3 just got released:

https://xmlgraphics.apache.org/fop/download.html

The XML graphics commons library on which FOP depends also needs to be updated to the latest version.
And maybe also the fontbox library.

infotexture:
:+1: Might be a good opportunity to include this in the next release. 

Would you be able to submit a PR with the necessary libraries and a summary of the relevant changes in the new versions?

raducoravu:
List of Apache FOP 2.3 changes is here:

https://xmlgraphics.apache.org/fop/changes.html

I'm not sure what updating to it implies, probably some changes to the Gradle config file, I'd rather not do that myself.

infotexture:
Implemented in #2976 for 3.1 with 39f1ac4.

PR #2976, opened 2018-05-26
Update to FOP 2.3

jelovirt:
Update to PDF plugin to FOP 2.3. Implements #2974.


infotexture:
Tested with latest source files from dita-ot/docs@65d3d99 with FOP version (2.2) from last OT release, and new FOP version 2.3.

| FOP | Build time |
| ------ | :----- |
| 2.2   | 46s   |
| 2.3 | 56s _(first build)_ |
| 2.3 | 52s _(second build)_ |
| 2.3 | 49s _(third build)_ |

The new version seems marginally slower at first, but subsequent build times appear to converge, so it's probably not a show-stopper.

The resulting output looks nearly identical, though there are minor changes in pagination (probably due to changes in `keep-with-next` properties), for example on Page 31 (first page of Chapter 7) where the last list item moves to the next page along with the parameter list it contains.

raducoravu:
👍 The first time a new FOP runs it will create a cache containing descriptors for all fonts installed in the system. On subsequent runs it will reuse that cache.

PR #2978, opened 2018-05-30
Support frame attribute on choice table

robander:
## Description

1. Moves logic for `simpletable/@frame` into mode templates to make the logic reusable
1. Reuse the logic from `choicetable`

## Motivation and Context

One of my customers noticed that `@frame` is ignored on `<choicetable>` -- all choice tables in PDF come out with no frames, no row separators, and no column separators. When I checked I noticed that the logic for supporting these on `<simpletable>` was overly verbose and not reusable without copying.

This pull request 1) puts the frame logic into mode templates that can be reused, and 2) calls those mode templates from the `<choicetable>` elements. The logic is set up so that it also works when called from the context of the actual `<choicetable>` context for generated headers.

Not currently adding the same support to `<properties>` because (unlike choice tables) it has columns that can disappear; with choice tables we know what the final column is.

## How Has This Been Tested?

Created a `<choicetable>` with each frame value, with and without default headers. Also created a `<simpletable>` with each frame value. Verified that the simple table appears the same before and after the fix, and verify that the frame values operate the same in equivalent choice tables.

Sample files:
[choicesample.zip](https://github.com/dita-ot/dita-ot/files/2052213/choicesample.zip)

## Type of Changes

Minor enhancement to simple tables to make logic reusable; bug fix for choice tables.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
@@ -929,37 +929,20 @@ See the accompanying LICENSE file for applicable license.
           <xsl:for-each select="$current">
             <fo:table-cell xsl:use-attribute-sets="strow.stentry">
                 <xsl:call-template name="commonattributes"/>
-                <xsl:variable name="frame" as="xs:string">
-                    <xsl:choose>
-                        <xsl:when test="../@frame">
-                            <xsl:value-of select="../@frame"/>
-                        </xsl:when>
-                        <xsl:otherwise>
-                            <xsl:value-of select="$table.frame-default"/>
-                        </xsl:otherwise>
-                    </xsl:choose>
-                </xsl:variable>
+                <xsl:variable name="frame" as="xs:string" select="if (../@frame)
+                    then (../@frame)
+                    else ($table.frame-default)"/>
                 <xsl:if test="following-sibling::*[contains(@class, ' topic/strow ')]">
-                    <xsl:call-template name="generateSimpleTableHorizontalBorders">
-                        <xsl:with-param name="frame" select="$frame"/>
-                    </xsl:call-template>
+                    <xsl:apply-templates select="." mode="simpletableHorizontalBorders">
+                        <xsl:with-param name="frame" select="$frame" as="xs:string"/>
+                    </xsl:apply-templates>
                 </xsl:if>
-                <xsl:if test="$frame = 'all' or $frame = 'topbot' or $frame = 'top'">
-                    <xsl:call-template name="processAttrSetReflection">
-                        <xsl:with-param name="attrSet" select="'__tableframe__top'"/>
-                        <xsl:with-param name="path" select="$tableAttrs"/>
-                    </xsl:call-template>
-                </xsl:if>
-                <xsl:if test="($frame = 'all') or ($frame = 'topbot') or ($frame = 'sides')">
-                    <xsl:call-template name="processAttrSetReflection">
-                        <xsl:with-param name="attrSet" select="'__tableframe__left'"/>
-                        <xsl:with-param name="path" select="$tableAttrs"/>
-                    </xsl:call-template>
-                    <xsl:call-template name="processAttrSetReflection">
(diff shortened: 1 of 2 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -929,37 +929,20 @@ See the accompanying LICENSE file for applicable license.
...
-                    <xsl:choose>
-                        <xsl:when test="../@frame">
-                            <xsl:value-of select="../@frame"/>
-                        </xsl:when>
-                        <xsl:otherwise>
-                            <xsl:value-of select="$table.frame-default"/>
-                        </xsl:otherwise>
-                    </xsl:choose>
-                </xsl:variable>
+                <xsl:variable name="frame" as="xs:string" select="if (../@frame)
```
jelovirt: This could be even more terse as `(../@frame, $table.frame-default)[1]` but I don't know if we want to start using a pattern like that. In e.g. JS you can do `a || b` for the same thing and IMO it's more readable.

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -1110,6 +1074,52 @@ See the accompanying LICENSE file for applicable license.
...
+        <xsl:call-template name="generateSimpleTableHorizontalBorders">
+            <xsl:with-param name="frame" select="$frame"/>
+        </xsl:call-template>
+    </xsl:template>
+    <xsl:template match="*" mode="simpletableTopBorder">
+        <xsl:param name="frame" select="if (ancestor-or-self::*[contains(@class, ' topic/simpletable ')][1]/@frame)
+            then (ancestor-or-self::*[contains(@class, ' topic/simpletable ')][1]/@frame)
+            else ($table.frame-default)"
+            as="xs:string"/>
+        <xsl:if test="$frame = 'all' or $frame = 'topbot' or $frame = 'top'">
```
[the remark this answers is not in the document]
robander: :+1: didn't even think about refactoring that one - it was just copy/paste of the existing logic.

robander:
Updated / simplified all of the `$frame` declarations and references based on the review comments (including those for `<table>`, which didn't have any other logic changes for this pull request).

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -1110,10 +1049,49 @@ See the accompanying LICENSE file for applicable license.
...
+            <xsl:call-template name="processAttrSetReflection">
+                <xsl:with-param name="attrSet" select="'__tableframe__top'"/>
+                <xsl:with-param name="path" select="$tableAttrs"/>
+            </xsl:call-template>
+        </xsl:if>
+    </xsl:template>
+    <xsl:template match="*" mode="simpletableSideBorders">
+        <xsl:param name="frame" select="(ancestor-or-self::*[contains(@class, ' topic/simpletable ')][1]/@frame, $table.frame-default)[1]"
+            as="xs:string"/>
+        <xsl:if test="($frame = 'all') or ($frame = 'topbot') or ($frame = 'sides')">
```
jelovirt: This too could be written as `$frame = ('all', 'topbot', 'sides')`

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -1110,10 +1049,49 @@ See the accompanying LICENSE file for applicable license.
...
+        <xsl:call-template name="generateSimpleTableVerticalBorders">
+            <xsl:with-param name="frame" select="$frame"/>
+        </xsl:call-template>
+    </xsl:template>
+
     <xsl:template name="generateSimpleTableHorizontalBorders">
         <xsl:param name="frame" as="xs:string?"/>
         <xsl:choose>
-            <xsl:when test="($frame = 'all') or ($frame = 'topbot') or ($frame = 'sides') or not($frame)">
+            <xsl:when test="$frame = ('all', 'topbot', 'sides') or not($frame)">
```
jelovirt: Instead of `not($frame)` we should use `empty($frame)`

PR #2988, opened 2018-06-03
Fix distribution docs build

jelovirt:
The docs build no longer uses Ant to build docs. The new Gradle build has not been written to be extensible and to be ran with an external docs source directory. This is a temporary fix around the issue to enable distribution package build with docs.

This should be reverted when the docs Gradle build has been modified to allow creating docs output from an external docs source directory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,8 +5,15 @@
  *
  * See the accompanying LICENSE file for applicable license.
  */
+plugins {
+    id 'com.github.eerohele.dita-ot-gradle' version '0.5.0'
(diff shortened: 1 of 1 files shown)


infotexture:
The docs build can be run from directories other than `docsrc` by passing the path to the OT installation as a parameter.

For example: `./gradlew -PditaHome=../dita-ot/src/main`.

I use this approach regularly to build from a standalone clone of the docs repo, so I assume a similar tactic could be used to run the docs build from the dist temp folder.

PR #3011, opened 2018-07-04
Deprecate and remove Ant properties

jelovirt:
## Description
Remove `dost.jar.path` property and deprecate `dost.class.path` properties.

## Motivation and Context
The currently correct way to define the classpath is outside the Ant process, thus `dost.class.path` is obsolete. The `dost.jar.path` is a previously deprecated alias for `dost.class.path` and can be removed after being deprecated since 2.1.

### DIFF
--- a/src/main/build_template.xml
+++ b/src/main/build_template.xml
@@ -23,6 +23,7 @@ See the accompanying LICENSE file for applicable license.
     </not>
   </condition>
     
+  <!-- Deprecated since 3.1 -->
   <path id="dost.class.path">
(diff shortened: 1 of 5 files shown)


jelovirt:
I'm having trouble with the classpath configuration in end to end tests. The modifications work when running DITA-OT in a normal environment, but in end to end tests Ant is unable to find `Fop` class.

jelovirt:
Had to tweak test classpath and remove Xalan as a transitive dependency.

Issue #2815, opened 2017-10-17
Support relative paths for HTML headers/footers


infotexture:
Currently several HTML parameters require absolute path values, which prevents portable cross-platform builds with the `dita` command:

* `args.ftr`
* `args.hdf`
* `args.hdr`

If absolute paths for these parameters are defined on one system, the same command/value may not work on other operating systems or machines where paths differ.

Allowing the `dita` command to interpret relative paths in these cases would reduce the need for Ant or other build scripts to set paths with variables that resolve to the correct local path when the build runs. 

These parameter values could then be passed in `.properties` files that work on multiple systems, which is apparently not possible at the moment per [comments](https://github.com/dita-ot/docs/pull/163#issuecomment-337163323) in https://github.com/dita-ot/docs/pull/163:

> WRT the absolute paths and properties file: cannot be fixed so that it will work every[w]here and will not break OT itself.

Various toolkit parameters currently use different approaches to path resolution. Perhaps the resolution mechanism for the HTML header/footer files could be aligned with one of these other approaches:

* `args.filter`

	> Relative paths are resolved against the DITA-OT base directory (for backwards compatibility) and internally converted to absolute paths.

* `args.input`

	> The path can be absolute, relative to `args.input.dir`, or relative to the directory where your project's Ant build script resides if `args.input.dir` is not defined.

jelovirt:
If those properties are made to support relative paths, it will only be at the outer API, i.e. in `dita` command. They will need to be resolved to absolute paths before the files are read by XSLT.

shaneataylor:
I think it's reasonable only to support this enhancement for the dita command and not Ant.

I'd also like to suggest that relative path resolution follow a common set of rules that minimizes the likelihood of unresolved paths by accepting the first of the following potential resolutions for which the target is found:

1. Relative to the properties file (if one is specified)
2. Relative to `input.dir` (if specified)
3. Relative to the `args.input` path (maybe set the default value of `input.dir` based on `args.input`?)
4. Relative to the transtype's plugin directory (not unusual to set headers and footers as part of an HTML customization)
5. Relative to DITA-OT home

I realize this adds some complexity, but it also (I think) addresses the most common assumptions users have about how such relative paths _should_ be resolved.

jelovirt:
The only correct base path is the properties file when the properties are set in a properties file, and the current directory when set using `-D`.

infotexture:
[...]

Issue #3016, opened 2018-07-11
Must set absolute value for args.cssroot to get CSS included in html5 and xhtml


drmacro:
## Expected Behavior

specifying only `args.css` with an absolute path and `no` value for `args.cssroot` should result in the CSS being included in HTML output when `args.copycss` is `yes`

## Actual Behavior

Final value of `args.css.file` is just the CSS filename so copying of CSS fails.

## Possible Solution

Not sure how best to fix this since it depends on whether or not `args.css` is or is not absolute,
but the bug is

```xml
<basename property="args.css.file.temp" file="${args.css}"/>
```
in `build_dita2html5_template.xml` (and in `org.dita.xhtml/build_general_template.xml`).

As this assumes that `args.cssroot` is also set.

## Steps to Reproduce

1. Set `args.css` to the absolute path of the CSS file
2. Do not set `args.cssroot` to a value
3. Set `args.copycss` to `true`
4. Generate HTML5
5. Observe that CSS was not copied and not referenced in generated HTML

## Environment

* DITA-OT version: 2.5.4
* Operating system and version: macos
* How did you run DITA-OT? dita command
* Transformation type: html5

robander:
Pretty sure this goes back to the original mess of options built up around CSS parameters, which goes back to pre-DITA-OT beta days, which goes back to tools that existed before DITA-OT. When something didn't work, typically another parameter was added to handle the problem rather than fixing the problem (because fixing the problem also meant breaking existing builds). The result is kind of a mess.

In the early days `args.css` could *only* be a file name, but I see that the documentation now allows either the file name or an absolute path, so I assume the absolute path worked at some point. It's not something I've used recently so I'm not sure what might have changed, or when...

drmacro:
I suspect that most users either just used args.cssroot from the start or they had multiple CSS files and so had to use it.

In my case the CSS is set in the plugin so there didn't seem to be a reason to use two parameters where one should work.

I can try to puzzle out the best solution at some point but it's not immediately obvious from the Ant code what that is.

larrykollar:
Indeed, `args.cssroot` requires a full path to work. But [the documentation](https://www.dita-ot.org/3.1/parameters/parameters-html5.html) says the cssroot is relative to the location of the input map. So one would *expect*, if the map and CSS files are in the same directory, that `args.cssroot` should default to "."

[...]

PR #3020, opened 2018-07-17
Refactor HTML5 CSS configuration

jelovirt:
Add tests for HTML5 CSS configuration and refactor code. The PR should not change the expected output.


robander:
Do the refactored checks for absolute paths also address the bug reported in #3016?

infotexture:
Could a similar approach be applied to header/footer files and other cases where various toolkit parameters currently use different approaches to path resolution? 

Aligning these behaviors would allow us to resolve #2815 and perhaps a few related issues.

jelovirt:
@robander No, this is just refactoring the code.

@infotexture Yes, but not in this PR.

PR #3024, opened 2018-07-19
Fix configuration file generation and reading

jelovirt:
Various configuration files are not read from correct locations, thus not all configuration data is available in all places. Fix the issue by removing `lib/configuration.properties` and retain only `config/configuration.properties`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -202,8 +202,6 @@ task copyDistTemp(type: Copy, dependsOn: initDist) {
         exclude "docsrc/temp"
         include "docsrc/**"
         include "dtd/**"
-        //include "lib/**/*.properties"
-        exclude "lib/configuration.properties"
         include "plugins/org.dita.base/**"
         include "plugins/org.dita.eclipsehelp/**"
         include "plugins/org.dita.html5/**"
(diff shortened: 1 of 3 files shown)


raducoravu:
@jelovirt One question, the now removed configuration.properties had a dita ot version key inside. Was this key moved to the other one? This information was quite useful even for plugins, for example a plugin could load the properties file in its build.xml and depending on the DITA OT version integrate in one way or another.

jelovirt:
@raducoravu Version info was already duplicated in [`application.properties`](https://github.com/dita-ot/dita-ot/blob/develop/src/main/resources/application.properties) and that's the correct location for it. `application.properties` is added to `dost.jar` and cannot be modified (easily), thus that's the configuration file for configuration items that should not be modified by users. `configuration.properties` on the other hand is for configuration items that are considered to be under user control.

raducoravu:
Makes sense, I'll look to see how can I read a properties file from inside a JAR file directly from the build.xml. Probably it's feasible using the jar:file protocol.

jelovirt:
@raducoravu See [build_init.xml](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/build_init.xml#L13) for example of reading resources from classpath in Ant.

raducoravu:
Oh, this one, nice:

    <loadproperties>
       <javaresource name="application.properties"/>
     </loadproperties>

Issue #1691, opened 2014-06-12
Preprocessing conref step does not fix up external links


vadi2:
https://dita-ot.github.io/1.8/dev_ref/preprocess-conref.html describes how links internal to the conreffed content are randomised to make them unique within the topic they get included in, and how internal links within the conreffed content are updated to match the new targets.

That works fine, but this breaks when content external to the conreffed content links inside the conreffed content. The external links are not fixed up by the process and they are thus broken.

PR #3027, opened 2018-07-27
Fix #./element style XREF that is pulled with conref

robander:
## Description

Fixes a bug in processing of `href="#./fig"` style "same-topic" references that are pulled with conref.

The specification describes how to process `<xref>` elements within content that is referenced with `@conref`:

> When the address is a same-topic fragment identifier, processors _MUST_ resolve it relative to the location of the content reference (referencing context). 

See the [spec topic on xref with conref]((https://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part1-base/archSpec/base/handling-xref-and-conref-within-topics.html). There is an example of this exact case (paragraph `C` referencing `p3`) showing the expected result.

Currently, when _any_ local cross reference begins with `#` and is pulled into another file via `@conref`, it is treated the same way: the file name of the original file is added so that the resulting link always goes to the original file. This is correct for links that begin with a topic ID such as `#topic`, but the `#.` syntax is explicitly intended as a way around this behavior, allowing pulled cross references to link to an ID within the _referencing_ file. The result today is always broken.

## Motivation and Context

Reported as a bug by one of my internal customers, who was pulling a section with cross references into another topic as part of a PDF publishing process. The resulting PDF ends up with broken links (the target of the `@conref` is not part of the PDF; even when part of the PDF, the link is still broken, because the modified href value `filename.dita#./d1e123` is invalid). 

This does *not* handle the case where you might have a section that includes both `id="test"` and `href="#./test"`. This cannot be handled with the current code, which always modifies IDs to generated values to ensure the resulting topic is correct. I think any fix for that issue likely requires larger design changes inappropriate for a hotfix.

## How Has This Been Tested?

Tested with a document based on the source I originally got with the problem.

Also updated an existing Integration test case where `@conref` pulls in a section full of `#topic` style cross references, so that it also tests the `#.` condition.

## Type of Changes

## Checklist

### DIFF
--- a/src/main/xsl/preprocess/conrefImpl.xsl
+++ b/src/main/xsl/preprocess/conrefImpl.xsl
@@ -539,6 +539,10 @@ See the accompanying LICENSE file for applicable license.
         <xsl:when test="starts-with(., 'http://') or starts-with(., 'https://') or starts-with(., 'ftp://')">
           <xsl:value-of select="."/>
         </xsl:when>
+        
+        <xsl:when test=". = '#.' or starts-with(., '#./')">
+          <xsl:value-of select="."/>
+        </xsl:when>
 
         <xsl:when test="starts-with(., '#')">
           <xsl:choose>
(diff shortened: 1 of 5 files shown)


robander:
RE: the "not handled" portion in the original pull request:
> This does _not_ handle the case where you might have a section that includes both id="test" and href="#./test". This cannot be handled with the current code, which always modifies IDs to generated values to ensure the resulting topic is correct. I think any fix for that issue likely requires larger design changes inappropriate for a hotfix.

I've just noticed that larger issue is open as its own item, #1691

PR #3028, opened 2018-07-29
Support for plugin registry

jelovirt:
## Description
Add support for installing plugins based on plugin registry information.

## Motivation and Context

Current plugin installation requires either a local plugin ZIP file or URI of a remote plugin ZIP file. Using a registry to store the distribution package URI allows installing plugins with just plugin ID, e.g.

```sh
dita --install org.lwdita
```

or

```sh
dita --install com.elovirta.ooxml@1.3
```

## Design

The plugin registry is a simple URL accessible set of JSON files, where the file name follows the convention `<plugin-name>.json`. The repository file structure:

```json
[
  {
    "name": "<name of the plugin>",
    "version": "<plugin version>",
    "url": "<download URL>",
    "cksum": "<optional SHA-256 checksum of the download resource>",
    "deps": [
      {
        "name": "<name of the dependency plugin>",
        "req": "<optional plugin dependency version>"
      }
    ]
  }
]
```

When a plugin is installed using just the plugin name, the integrator process will fetch plugin definitions from all configured repositories and will install the first one that matches the platform and desired version requirements. If no plugin is found with matching name and version requirements, installation will fail. Dependencies of plugins that are not yet installed will be recursively installed

The argument to `--install` will be processed as follows in order, stopping on first match and then trying to install the plugin from that, in pseudo-code:

```pseudo
if Files.exists($arg):
  installLocal($arg)
else if URI.isAbsolute($arg):
  installDownload($arg)
else:
  foreach repository in repositories:
    $plugins = downloadRepository($repository)
    if $plugins.contains($arg):
      $plugin = $plugins.get($arg)
  if $plugin:
    installDownload($plugin.url)
```

The plugin repository URL could be just the plain GitHub URL, but it would likely make more sense to use a redirect from `https://repository.dita-ot.org/` to `https://raw.githubusercontent.com/dita-ot/dita-ot-index/master/`.


src/main/config/configuration.properties
```diff
@@ -14,6 +14,7 @@ cli.color = true
 plugindirs = plugins;demo
 plugin.ignores =
 plugin.order = org.dita.base org.oasis-open.dita.v1_3 org.oasis-open.dita.v1_2
+registry = https://raw.githubusercontent.com/jelovirt/dita-ot-index/master/
```
jelovirt: The jelovirt/dita-ot-index repository needs to be moved to dita-ot ownership and this changed accordingly before merge.

jelovirt:
One of the design decisions is that do we implement automatic installation of dependent plugins in this PR, or later on as an enhancement. E.g. PDF2 FOP plugin depends on PDF2 plugins, so if the user installs `org.dita.pdf2.fop` plugin, will we implement dependency management to a degree that integrator will also fetch `org.dita.pdf2` plugin and install both at the same time.

Dependency management and resolution will be challenging to implement, thus I would prefer to implement that in a separate PR.

robander:
> Dependency management and resolution will be challenging to implement, thus I would prefer to implement that in a separate PR.

That sounds right to me. Thinking about minimum implementation to get feedback, it seems right to add support for the registry, change as needed, and dependency management can come later. Best result now would probably be an error message "This cannot be installed without X, Y, and Z" -- can install those independently.

Thinking about my own (many) plugins -- over time I've generally stopped making too many dependencies because most of them were unnecessary and just ended up annoying me when I got "failed to install" messages.

src/main/resources/plugin.rnc
```diff
@@ -59,8 +60,10 @@ extension-point =
 require =
   element require {
     attribute plugin { xsd:NCName },
+    attribute version { version },
```
robander: Does this mean `@version` will now be required on the `<require>` element? Might be better to make it optional (default to `0.0.0` which is the default for plugins without a version)... otherwise we'll make a lot of existing plugins invalid.

robander:
I'm really liking this feature. Scanned through the code and made a simple comment, I don't feel qualified to make a full review but looks good as far as I can tell. Not sure how close it is to complete so I haven't marked it approve or added an official review.

infotexture:
👍 Also very glad to see this taking shape. 🎉 

_A few questions come to mind…_

1. **Will there be a way to configure alternate repository locations that the registry should parse?** 

    I imagine this is conceived to start with a set of "official" plug-ins stored in a pre-configured repository, but it would be most useful if users could add other locations from which to retrieve plug-ins, such as [dita-community-ot-plugins](https://github.com/dita-community/dita-community-ot-plugins), a company plug-in repository, or individual developers' open-source plug-in repos on GitHub or elsewhere.

2. **How will collisions be handled if there are multiple plug-ins of the same name/ID in different configured repositories?**

jelovirt:
Yes, just add more URLs to the repository key in configuration.properties.

First one in order will win.

jelovirt:
@robander fixed the optional version attribute.

PR #3029, opened 2018-07-30
Avoid failure for cite in preprocessed link node

robander:

## Description

Avoids XSLT failure with PDF for inline markup inside of `<linktext>` elements.

## Motivation and Context

Link markup is initially preprocessed for PDF to place it into a variable. The variable is then processed to create the correct page layout.

When a `<linktext>` element contains inline markup like `<cite>` (phrase-like elements that can turn into links using a key), the phrase-like markup is tested to resolve that link. This uses a key to check if the target is part of the PDF. When the element is reprocessed as part of the variable, the context is no longer part of the original document, resulting in an XSLT failure:
```
 [pipeline] Failed to transform document: In the key() function, the node supplied
 in the third argument (or the context node if absent) must be in a tree whose
 root is a document node
```

The fix here ensures that if such inline markup is processed outside of the document context, the build will not fail, and the phrase contents will be used.

## How Has This Been Tested?

Create a topic that includes the following markup:
```
<related-links>
  <link href="http://example.com" scope="external" format="html">
    <linktext><ph><cite keyref="test">World without Mind</cite> by Franklin Foer</ph></linktext>
    <desc>Testing a link to a book without an href</desc>
  </link>
</related-links>
```

Tested adding that topic to one map that does and one map that does _not_ declare `keys="test"` -- the failure is now avoided in each case, and the specified link text is used.

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
@@ -210,7 +210,8 @@ See the accompanying LICENSE file for applicable license.
   <xsl:template match="*[contains(@class,' topic/term ')]" name="topic.term">
     <xsl:param name="keys" select="@keyref" as="attribute()?"/>
     <xsl:param name="contents" as="node()*">
-      <xsl:variable name="target" select="key('id', substring(@href, 2))"/>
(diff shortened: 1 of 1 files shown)


robander:
Needed the same update for `<term>` which has its own complicated test / logic for inline linking, but also includes the same `key('id')` syntax that fails when not in a document node.

Issue #1132, opened 2012-04-25
Support generating text only reference in reltable


jelovirt:
Converted from [SourceForge issue 1989420](http://sourceforge.net/support/tracker.php?aid=1989420), submitted by amberink

Requirement: be able to use relationship table for all references, including text only references to other deliverables.

Enhancement: support processing `<topicref>` with no href specified and generate text in <linktext> element.

Example DITA markup:

``` xml
<topicref>
  <topicmeta>
    <linktext>For more information about twinkies, see "All about Twinkies".</linktext>
  </topicmeta>
</topicref>
```

in a relationship table.

```
Output: For more information about twinkies, see "All about Twinkies".
```

robander:
For some reason decided to take a look at this one and found 3 different issues all with a simple test case:

- Original issue of course is still there: a `linktext`-only entry in the reltable doesn't get added to the topic
- If a text-only link appears inside of a topic, the `linktext` element is dropped by the `topicpull` step: `<link><linktext>something</linktext></link>` is converted to `<link></link>` (after error message `DOTX028E` that it is missing href/keyref)
- If such an empty link is coded in a topic (or created by that `topicpull` bug), FOP crashes with the following message:
```
      [fop] [ERROR] Anttask - Error rendering fo file: C:\DITA-OT\TEST\
1132textReltable\temp\topic.fo <net.sf.saxon.trans.XPathException: org.apache.fop.f
o.ValidationException: Missing attribute on fo:basic-link: Either external-desti
nation or internal-destination must be specified. (See position 3:405)>net.sf.sa
xon.trans.XPathException: org.apache.fop.fo.ValidationException: Missing attribu
te on fo:basic-link: Either external-destination or internal-destination must be
 specified. (See position 3:405)
```
[1132.zip](https://github.com/dita-ot/dita-ot/files/2117570/1132.zip)

PR #3030, opened 2018-07-30
Allow text-only links to appear with related-links

robander:
## Description

Includes several related fixes / enhancements in order to implement #1132:

- Bug fix: currently `topicpull` removes any specified `<linktext>` or `<desc>` content when there is no link target, resulting in a truly empty link. This should be deferred to rendering to determine whether / how to render links without `@href`.
- Bug fix / enhancement: if a link with link text but no `@href` does make its way to PDF processing today, it results in a build failure. The build failure has been removed and the text is rendered with other links.
- Enhancement: If text-only links appear in a reltable (`<topicref> that does not have a link target, but _does_ include `<linktext>`), those links are generated, and the final rendering step is allowed to render or hide them as needed.
- Bug fix: when `<xref>` does not specify any `@href` value at all, remove the error message that complains about an _empty_ `@href` value
- Enhancement: When `<link>` does not specify `@href`, avoid generating a build error in `topicpull`, and leave this to rendering steps or customizations to determine if it is an error.

## Motivation and Context

Reported / requested long ago with #1132, request is still valid and outstanding.

## How Has This Been Tested?

Test files attached in the original issue to reproduce all issues. 

Also encountered #3029 while testing this issue, but that is a more general issue that should be fixed ASAP in a hotfix rather than bundled with this enhancement.

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/links.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/links.xsl
@@ -489,7 +489,7 @@ See the accompanying LICENSE file for applicable license.
       </xsl:choose>
     </xsl:template>
 
-  <xsl:template match="*[contains(@class,' topic/link ')]" mode="processLink">
+  <xsl:template match="*[contains(@class,' topic/link ')][not(empty(@href) or @href='')]" mode="processLink">
     <xsl:variable name="destination" select="opentopic-func:getDestinationId(@href)"/>
     <xsl:variable name="element" select="key('key_anchor',$destination, $root)[1]"/>
 
@@ -544,6 +544,21 @@ See the accompanying LICENSE file for applicable license.
       <xsl:with-param name="linkScope" select="$linkScope"/>
     </xsl:call-template>
     </fo:block>
+  </xsl:template>
+
+  <xsl:template match="*[contains(@class,' topic/link ')][empty(@href) or @href='']" mode="processLink">   
+    <xsl:if test="*[contains(@class, ' topic/linktext ')]">
+      <fo:block xsl:use-attribute-sets="link">
+        <fo:inline>
+          <xsl:apply-templates select="*[contains(@class, ' topic/linktext ')]"/>
+        </fo:inline>
+        <xsl:if test="*[contains(@class, ' topic/desc ')]">
+          <fo:block xsl:use-attribute-sets="link__shortdesc">
+            <xsl:apply-templates select="*[contains(@class, ' topic/desc ')]"/>
+          </fo:block>
+        </xsl:if>
+      </fo:block>
+    </xsl:if>
   </xsl:template>
 
     <xsl:template name="buildBasicLinkDestination">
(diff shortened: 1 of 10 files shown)


src/main/xsl/preprocess/maplinkImpl.xsl
```diff
@@ -340,7 +340,7 @@ See the accompanying LICENSE file for applicable license.
     </xsl:variable>
     
     <xsl:for-each select="ancestor::*[contains(@class, ' map/relrow ')]/*[contains(@class, ' map/relcell ')][position()!=$position]">
-      <xsl:if test="descendant::*[contains(@class, ' map/topicref ')][@href and not(@href = '')][not(@linking = ('none', 'sourceonly'))]">
+      <xsl:if test="descendant::*[contains(@class, ' map/topicref ')][(@href and not(@href = '')) or */*[contains(@class,' map/linktext ') or contains(@class,' topic/linktext ')]][not(@linking = ('none', 'sourceonly'))]">
```
jelovirt: The added set of predicate conditions are repeated in multiple placed in this stylesheet. IMO this should be extracted into a function with a descriptive name to make the code easier to read and understand.

robander: Made the `[(href and not href='') or linktext]` condition into a function.

Still need to add an integration test for this to ensure the links are updated properly by `maplink` + not dropped by `topicpull`.

src/main/plugins/org.dita.pdf2/xsl/fo/links.xsl
```diff
@@ -489,7 +489,7 @@ See the accompanying LICENSE file for applicable license.
       </xsl:choose>
     </xsl:template>
 
-  <xsl:template match="*[contains(@class,' topic/link ')]" mode="processLink">
+  <xsl:template match="*[contains(@class,' topic/link ')][not(empty(@href) or @href='')]" mode="processLink">
```
jelovirt: I would like us to at some point consider adding empty `@href` removal to preprocessing. Then we would not have to have `@href = ''` tests in transtype specific code. Or is there any reason why to use an empty `@href` attribute? `URI.create("file:/base/topic.xml").resolve("")` will resolve to `file:/base/`, but I don't see any reason to allow users to link to _directories_. If we remove empty `@href`, you could still link to those with `href="./"`.

robander: :+1: for removing empty `@href` (and empty `@conref`) early in initial processing. It's caused problems several times over the year when other steps don't test for it, and try to open the directory -- exactly as you've described. For an author running a build it's extremely confusing and hard to track down, because it usually appears as "Cannot open `/base/`" in the error. So yeah -- leaving it in means a lot of messy checks in the code, and means if you forget the check, things go wrong.

The only reason I can think of for having it is if you have some element with required `@href` attribute, but you don't actually want to specify one -- maybe because you use `@conref` or `@keyref`. Technically that's what the `-dita-use-conref-target` token is for but most people don't know about that. Even for that case though -- the empty `@href` would only be needed for initial parsing / validation, and generic DITA processing code that follows won't look for it.

Issue #2481, opened 2016-09-16
Handling of <q> element in output


raducoravu:
In the XHTML output by default text inside the "q" element is surrounded in dumb double quotes.
But in the PDF output the same text is surrounded in smart double quotes.
Maybe we should have consistency in these outputs.

shaneataylor:
There exist `OpenQuote` and `CloseQuote` strings (`xsl/common/strings-en-us.xml`), which I assume are being used for XHTML and to facilitate both internationalization and house style preferences for "straight" or "curly" quotes. Are you saying we should simply change the default values for some languages to the curly quote symbols?

shaneataylor:
...and that assumption seems to have been incorrect. 😦

jelovirt:
This should be fixed in SASS.

raducoravu:
Right now as @shaneataylor says the "strings-en-us.xml" generates actual characters for those elements. So @jelovirt, are you suggesting giving up the  "strings-en-us.xml" approach for this case and style the spans via the CSS?

robander:
Long ago when these quote strings were added + used in XHTML, the browsers did not have reliable formatting for `<q>` (some showed quotes, some did not, and I think there were also issues with localization). 

I tested this a couple of months ago when somebody's validation tool complained about the quotes, and all of the browsers I tried with correctly displayed quotes for something like `<q>test</q>` in the HTML. I think it's time to stop pulling in the string variables for X/HTML and instead just use the native `<q>` element.

robander:
Fixed with https://github.com/robander/dita-ot/commit/d5db3f2a2697f2bebce5f6a92527c9678480cd62

PR #3032, opened 2018-07-31
Improve handling of <q> in html5 / xhtml

robander:
## Description

When processing the DITA `<q>` element for quotes, we currently generate `<span>` elements and quote characters based on variables. This request switches to using the native XHTML / HTML5 `<q>` element.

When originally implemented 15+ years ago, browser support for `<q>` was somewhat inconsistent. At this point, any recent browser does a good job of rendering the element with quotes, and if alternate quotes are desired they can be handled with CSS.

## Motivation and Context

Originally reported here as #2481.

Not long ago, I also got a report about this in other channels, because an accessibility checker was flagging `<span>"quote"</span>` for improper markup (as with any semantic markup, using it accurately to say the quote is a quote results in better automation from all sorts of tools).

## How Has This Been Tested?

Tested using the `<q>` element on my machine with Firefox, Chrome, and IE. All three automatically displayed quotes, though the actual quote characters were inconsistent. (Firefox displayed open/close quotes, Chrome displayed simple " quotes, and IE displayed locale-specific quotes).

## Type of Changes

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl
+++ b/src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl
@@ -1151,16 +1151,10 @@ See the accompanying LICENSE file for applicable license.
 
 <!-- quotes - only do 1 level, no flip-flopping -->
 <xsl:template match="*[contains(@class, ' topic/q ')]" name="topic.q">
-  <span class="q">
+  <q>
     <xsl:call-template name="commonattributes"/>
-    <xsl:call-template name="getVariable">
-      <xsl:with-param name="id" select="'OpenQuote'"/>
-    </xsl:call-template>
     <xsl:apply-templates/>
-    <xsl:call-template name="getVariable">
-      <xsl:with-param name="id" select="'CloseQuote'"/>
-    </xsl:call-template>
-  </span>
+  </q>
 </xsl:template>
 
 <xsl:template match="*[contains(@class, ' topic/term ')]" mode="output-term">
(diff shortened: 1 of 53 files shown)


src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl
```diff
@@ -1151,16 +1151,10 @@ See the accompanying LICENSE file for applicable license.
 
 <!-- quotes - only do 1 level, no flip-flopping -->
 <xsl:template match="*[contains(@class, ' topic/q ')]" name="topic.q">
-  <span class="q">
+  <q>
     <xsl:call-template name="commonattributes"/>
-    <xsl:call-template name="getVariable">
```
jelovirt: See e.g. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/q

> **Usage note**: Most modern browsers will automatically add quotation marks around text inside a `<q>` element. A style rule may be needed to add quotation marks in older browsers.

Thus I would prefer us **not** generate quotation marks into HTML output and handle the generation in CSS. If users want hard coded quotation marks, they can override this XSLT.

robander: I'm confused ... this pull request gets rid of the variable use / gets rid of the hard coded quotation marks, and instead uses the native HTML `<q>` element. The browser's default CSS is now used to generate the quotation marks, and custom CSS can be used to change / remove the characters (or they can use XSLT as you suggested).

It looks like this is a comment disliking the code we should remove, rather than a comment on the new code?

jelovirt: Sorry, my fried brain mixed red and green. Yes, this is the correct approach.

PR #3033, opened 2018-08-01
Fix NPE when conref range used on topics

### LINKED ISSUE #2222: NPE in TopicFragmentFilter module [DITA OT 2.x]
DITA Map:

``` xml
<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "map.dtd">
<map>
 <title>DITA Topic Map</title>
 <keydef href="zuzu.dita" keys="zuzu"/>
 <topicref href="test.dita"/>
</map>
```

zuzu.dita:

``` xml
<!DOCTYPE dita PUBLIC "-//OASIS//DTD DITA Composite//EN" "ditabase.dtd">
<dita>
  <topic id="a">
    <title>A</title>
  </topic>
  <topic id="Z">
    <title>A</title>
  </topic>
</dita>
```

test.dita:

``` xml
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
<topic id="topic_fyk_zzc_x5">
  <title>AAA</title>
  <body/>
  <topic id="id_kq2_11d_x5" conkeyref="zuzu" conrefend="default.dita#default/Z">
    <title/>
  </topic>
</topic>
```

NPE when publishing:

```
BUILD FAILED
D:\projects\eXml\frameworks\dita\DITA-OT2.x\build.xml:41: The following error occurred while executing this line:
D:\projects\eXml\frameworks\dita\DITA-OT2.x\plugins\org.dita.base\build_preprocess.xml:226: java.lang.NullPointerException
    at java.util.ArrayDeque.addFirst(Unknown Source)
    at org.dita.dost.writer.TopicFragmentFilter.startElement(TopicFragmentFilter.java:48)
    at org.apache.xerces.parsers.AbstractSAXParser.startElement(Unknown Source)
    at org.ditang.relaxng.defaults.RelaxNGDefaultsComponent.startElement(RelaxNGDefaultsComponent.java:200)
    at org.apache.xerces.impl.XMLNSDocumentScannerImpl.scanStartElement(XMLNSDocumentScannerImpl.java:268)
    at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl$FragmentContentDispatcher.dispatch(XMLDocumentFragmentScannerImpl.java:1655)
    at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanDocument(XMLDocumentFragmentScannerImpl.java:325)
```

robander:
## Description

When conref range is used on `topic` elements, the generated result from the `conref` step has an ID on the first topic (as specified on the source element that uses `conref` or `conkeyref`), but IDs are explicitly dropped from all following topics in the range. Topics are required by DITA to have an ID, and many steps assume it is present (a source document will not parse when `topic` is missing `@id`). So, any step after `conref` that assumes topic IDs are present will fail.

The fix checks to see if the element in the range is a topic; if so, a generated ID is added to ensure the topic is valid.

## Motivation and Context

Fixes #2222 

## How Has This Been Tested?

Tested with the original source doc. Added a variant of the source doc into our integration test. For each test case, the build failed without the fix, and works with the fix.

## Type of Changes

## Checklist

Our integration test does not appear to have any tests for `conrefend` (even if they did, they would only have caught this if used on `<topic>`). So, the test case I've added uses `conrefend` on normal body content as well as on a topic.


src/main/xsl/preprocess/conrefImpl.xsl
```diff
@@ -462,22 +462,25 @@ See the accompanying LICENSE file for applicable license.
...
-                          </xsl:when>
-                          <xsl:otherwise>
-                            <xsl:copy/>
-                          </xsl:otherwise>
-                        </xsl:choose>
-                      </xsl:if>
+                      <xsl:choose>
+                        <xsl:when test="name() = 'id'">
+                          <xsl:if test="contains(../@class,' topic/topic ')">
+                            <xsl:attribute name="id" select="generate-id(.)"/>
```
jelovirt: This generates the new `@id` from `generate-id(@id)`. Should it instead be from the element, so `generate-id(..)` in this particular case? My preference is to use the element for source of identity, not attributes.

robander: Yep, not intentional, and more obvious / easily fixed with the refactoring suggested above.

Issue #1960, opened 2015-07-03
Need dita action to list IDs of installed plugins


drmacro:
The dita -uninstall action requires knowledge of the plugin ID but there is no way to get the plugin IDs of the installed plugins from the dita command--you have to look in the plugin.xml files for each plugin.

Would like an action that lists the plugin IDs and plugin versions

This would also allow for scripts that can check dependencies and current installation status in support of coordinated installation of multiple plugins.

jelovirt:
Argument like -plugins would be suitable for this. In tools like Homebrew this would be -list, in OT it's not the best command word. Similarly -transtypes could list all transtypes.

drmacro:
I would find -plugins and -transtypes intuitive. I guess the danger is if there are more things that would be usefully listed, a -list {type} command would be more general, but for now there is really only plugins and transtypes.

blefort-uo:
+ 1

jelovirt:
@robander this could be fixed as part of #3028.

robander:
Yep - just noticed it while thinking about how easy this might be to implement, and put in the comment so it wouldn't get lost.

robander:
Implemented for 3.2 with #3037

PR #3037, opened 2018-08-03
Add --plugins argument to dita command line #1960

robander:

## Description

Adds a new command line arguments `--plugins` and `--transtypes` to the `dita` command line. When used, they provide a list of installed plug-in IDs and transtype names, one per line.

Pretty sure my implementation could be improved, but I won't know until I open the pull request and wait for the comments...

## Motivation and Context

Requested in #1960 along with follow-up discussion.

There is also a request for an option that would display a list of transform types, but that is not (yet) part of this request - waiting to see how the `--plugins` implementation goes over.

## How Has This Been Tested?

Tested by running the command line `bin/dita --plugins` and `bin/dita --transtypes`, verifying that each installed plugin is listed.

## Type of Changes

## Checklist

Have not updated unit tests - I don't see unit tests for the other command line parameters so not sure where this one should go.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -64,6 +64,7 @@ import org.xml.sax.SAXException;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 
+import static org.dita.dost.util.Configuration.transtypes;
 import static org.dita.dost.util.Constants.ANT_TEMP_DIR;
 import static org.dita.dost.util.Constants.PLUGIN_CONF;
 import static org.dita.dost.util.XMLUtils.getChildElements;
@@ -527,6 +528,7 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
         boolean justPrintVersion = false;
         boolean justPrintDiagnostics = false;
         boolean justPrintPlugins = false;
+        boolean justPrintTranstypes = false;
         useColor = getUseColor();
 
         final Deque<String> args = new ArrayDeque<>(Arrays.asList(arguments));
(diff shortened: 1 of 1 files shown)


src/main/java/org/dita/dost/invoker/Main.java
```diff
@@ -767,6 +788,19 @@ private void handleArgUninstall(final Deque<String> args) {
         }
         uninstallId = value;
     }
+    
+    /** Handle the --plugins argument */
+    private void printPlugins() {
+        final List<String> installedPlugins = getInstalledPlugins();
+        if (installedPlugins != null) {
+            System.out.println("The following DITA-OT plugins are installed:");
```
jelovirt: I'm not sure if I want to have this line output and each ID prefixed with spaces. My preference would be to just have one ID per line, without any prefixes. That way the output could be piped to other processes without parsing out the pretty printing.

robander: I don't have a strong feeling on it, either is fine. 

When I got the first one working it didn't have spaces, and that seemed to look a little odd in my DOS shell. I put in the 3 spaces because they matched the `--help` output. But ... yeah it's easy to convince me to remove them here.

jelovirt: The question is, for what is the output used for. In the original request

> This would also allow for scripts that can check dependencies and current installation status in support of coordinated installation of multiple plugins.

jelovirt:
Added support for listing trantypes from CLI, IMO that's in the same class or requirement as listing plugins. @robander are you ok with this addition?

robander:
> Added support for listing trantypes from CLI, IMO that's in the same class or requirement as listing plugins. @robander are you ok with this addition?

:+1: yes, was hoping we'd get that included as well.

src/main/java/org/dita/dost/invoker/Main.java
jelovirt: @robander should we then change the output be plumbing output, i.e. only output the plug-in IDs or trantype names, without prefix?

robander: I think that's probably a good idea -- just generate the list, one per line, no extra spacing.

PR #3039, opened 2018-08-06
Implement defaults output for hazard domain in PDF

jelovirt:
Implement a generic hazard statement styling for PDF2.

screen shot 2018-08-10 at 18 33 01

## References

* https://en.wikipedia.org/wiki/ISO_3864
* https://en.wikipedia.org/wiki/ANSI_Z535

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/hazard-d.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/hazard-d.xsl
@@ -6,8 +6,9 @@ See the accompanying license.txt file for applicable licenses.
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
                 xmlns:fo="http://www.w3.org/1999/XSL/Format"
                 xmlns:xs="http://www.w3.org/2001/XMLSchema"
+                xmlns:dita-ot="http://dita-ot.sourceforge.net/ns/201007/dita-ot"
                 version="2.0"
-                exclude-result-prefixes="xs">
+                exclude-result-prefixes="xs dita-ot">
 
   <xsl:template match="*[contains(@class, ' hazard-d/hazardstatement ')]">
     <xsl:variable name="number-cells" as="xs:integer" select="2"/>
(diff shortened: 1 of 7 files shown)


jelovirt:
In order to make all custom symbol images the same size, we would need to refactor the image templates. Or rather, I could just copy image templates and change how the size is calculated, or then we change the normal image templates to make it easier to have custom size configuration.

jelovirt:
I haven't chosen whether to implement ANSI or ISO spec for this. I have access to the ANSI spec, but not ISO. Preferably it should be user definable via property.

I don't have HTML5 implementation planned right now.

src/main/plugins/org.dita.pdf2/xsl/fo/hazard-d.xsl
```diff
@@ -0,0 +1,123 @@
...
+                  <w xsl:use-attribute-sets="hazardstatement.title.caution"/>
+                </xsl:when>
+                <xsl:otherwise>
+                  <w xsl:use-attribute-sets="hazardstatement.title.notice"/>
+                </xsl:otherwise>
+              </xsl:choose>
+            </xsl:variable>
+            <xsl:sequence select="$atts/@*"/>
+            <fo:block>
+              <xsl:if test="@type = ('danger', 'warning', 'caution')">
```
robander: Because it defaults to Caution below - we also check for `empty(@type)`. Current result seems a little odd when not setting `@type` gets me `CAUTION` (no image) but setting `type="caution"` gets me the image + `CAUTION`.

src/main/plugins/org.dita.pdf2/xsl/fo/hazard-d.xsl
```diff
@@ -0,0 +1,123 @@
...
+          <fo:table-cell xsl:use-attribute-sets="hazardstatement.title">
+            <xsl:variable name="atts" as="element()">
+              <xsl:choose>
+                <xsl:when test="@type = 'danger'">
+                  <w xsl:use-attribute-sets="hazardstatement.title.danger"/>
+                </xsl:when>
+                <xsl:when test="@type = 'warning'">
+                  <w xsl:use-attribute-sets="hazardstatement.title.warning"/>
+                </xsl:when>
+                <xsl:when test="@type = 'caution'">
```
robander: If `caution` is the default for generated text, then I think it should also be the test here for if `@type` is missing, so the CAUTION output matches.

jelovirt:
@robander changed the inconsistent caution output when `@type` was not defined.

Issue #2940, opened 2018-04-24
Index collection fails if title is wrapped in ph indexterm


Enigma644:
## Expected Behavior

Trying to do HTML Help output to CHM file.

## Actual Behavior

No CHM produced, NullPointerException in log.

## Possible Solution

Seems to be something to do with indexterm being in the title and also being in the body.

## Steps to Reproduce

[testFiles.zip](https://github.com/dita-ot/dita-ot/files/1943570/testFiles.zip)
1. Unzip.
2. Correct paths in bat for your setup.
3. Run makeCHM.bat

## Copy of the error message, log file or stack trace

 [pipeline] file:/C:/Users/James/Desktop/swap/DOTK/input/TestTopic.dita:3:146: [DOTX037W][WARN]: No title found for this document; using "***" in XHTML title bar. 
Error: The following error occurred while executing this line:
C:\Stuff\DITA-OT-3.0.3\plugins\org.dita.htmlhelp\build_dita2htmlhelp.xml:90: java.lang.NullPointerException

## Environment

* DITA-OT version:3.0.3
* Operating system and version: Winx64
* How did you run DITA-OT?
Call "C:\Stuff\DITA-OT-3.0.3\bin\dita.bat" /transtype:htmlhelp -i="C:\Users\James\Desktop\swap\DOTK\input\map.ditamap" -o="C:\Users\James\Desktop\swap\DOTK\output" -l="C:\Users\James\Desktop\swap\DOTK\log.log"
* Transformation type:
htmlhelp

Enigma644:
Yes, the whole title. It was a convenient way to index the topic as I'm also doing args.indexshow=yes

Enigma644:
Also worth noting is that if a section title is fully wrapped in <ph><indexterm>, then it is missing from the output and you just get a  comment.

As a workaround, I'm currently putting a single space after the indexterm, as this makes it work.

drmacro:
I would suggest that using "indexshow" to get around the general need to have index term markup in addition to the text that warrants the index entry is bad practice.

It will only produce appropriate results when all the index entries are identical to the base text, which cannot always be the case generally.

If you're indexing in the topic title the better markup would be to put the index entry in the topic prolog.

Enigma644:
All my index terms are in the base text and adding markup there is a far better user experience that having a mess of terms in the prolog. Also, marking up inline means that index terms are already there when sections are re-used between documents.

robander:
[...]

PR #3043, opened 2018-08-09
Fix NPE from index process when topic has empty title

robander:
## Description

If a topic:
- Has index terms, and
- Has an empty title, or title with only metadata, such as `<title><indexterm>test</indexterm></title>`, and
- Is part of a map build to HTML Help or Eclipse help (or, probably, Javahelp using the older plugin)

Then the build will crash with a null pointer error. This is caused by a `Map` that stores topic IDs with their title for generating index output; a title with zero content results in a `null` title, which fails in the various steps that try to generate the title later.

Fixed by replacing the null title with `***`, the same value already used (with a warning) for empty index terms and for the HTML `<title>` output with the same empty condition.

## Motivation and Context

Fixes the NPE reported in #2940

## How Has This Been Tested?

Fixes the NPE in the original test files. Also added the empty-title condition to the `IndexTermReader` unit test; as expected, that resulted in the NPE, but passed again with this fix.

## Type of Changes

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/reader/IndexTermReader.java
+++ b/src/main/java/org/dita/dost/reader/IndexTermReader.java
@@ -237,6 +237,9 @@ public final class IndexTermReader extends AbstractXMLReader {
             inTitleElement = false;
             if (!topicIdStack.empty() && !titleMap.containsKey(topicIdStack.peek())) {
                 //If this is the first topic title
+                if (title == null) {  //Guard against zero-content titles
+                    title = "***";
+                }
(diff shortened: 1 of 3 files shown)


jelovirt:
Should this go into 3.1.2 because it fixes a NPE?

robander:
> Should this go into 3.1.2 because it fixes a NPE?

Sure ... only reason I didn't put it there was because (in my head) we were already trying to close out 3.1.2. I can rebase on the hotfix this morning.

PR #3050, opened 2018-08-21
Fix pdf2 import

xephon2:
## Description
Imports of `topic-attr.xsl` and `concept-attr.xsl` moved to the shell.

## Motivation and Context
Fixes a pretty old FIXME. Now it is easier to completely replace `topic-attr.xsl` and `concept-attr.xsl` completely by stripping those files from the shell. Before this fix it was required to also replace `commons-attr.xsl` to get rid of the imports.

## How Has This Been Tested?
Local PDF2 publication

## Type of Changes
Bug fix

## Checklist


jelovirt:
Refactoring like this is a good direction because it makes the import tree more flat. 👍

PR #3058, opened 2018-08-27
Add Homebrew installation details to ReadMe

infotexture:
This PR builds on the changes proposed by @SMillerDev in #3055 to provide additional details on downloading and installing the toolkit via [Homebrew](https://brew.sh).

_(There are a few things we could tweak in the formula, but I’ll create a [separate issue](https://github.com/dita-ot/dita-ot/issues/3059) for that.)_

To simplify the instructions for new users, this version also wraps developer details in a disclosure section that can be clicked to show info on building from source and compiling the distribution package.

Since the diff is a bit busy, see the [rendered results in my fork](https://github.com/infotexture/dita-ot/blob/feature/update-readme-homebrew/README.markdown).

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -5,17 +5,18 @@ The _DITA Open Toolkit_, or _DITA-OT_ for short, is an open-source publishing en
 Visit the project website at [dita-ot.org][site] for documentation, information about releases, and [download packages][dist].  
 For information on additional DITA and DITA-OT resources, see [SUPPORT].
 
-- [Prerequisites](#prerequisites)
+- [Prerequisites: Java 8](#prerequisites-java-8)
 - [Installing](#installing)
 - [Building output](#building-output)
 - [For developers](#for-developers)
 - [License](#license)
 
-## Prerequisites
+## Prerequisites: Java 8
 
-To build and use DITA-OT, you’ll need:
+- To _build_ DITA-OT, you’ll need Java Development Kit (JDK), version 8 or newer
+- To _run_ DITA-OT, the Java Runtime Environment (JRE) is sufficient
 
-- Java Development Kit 8 or newer
+You can download the Oracle JRE or JDK from [oracle.com/technetwork/java][java].
 
 ## Installing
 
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -1,68 +1,104 @@
...
 
-1. Clone the DITA-OT Git repository:
+1.  Download the distribution package from [dita-ot.org/download][4].
+2.  Extract the contents of the package to the directory where you want to install the DITA-OT.
 
-        git clone git://github.com/dita-ot/dita-ot.git
+### Installing on macOS via Homebrew
 
-2. Move to the DITA-OT directory:
+On macOS, you can also install DITA-OT using the [Homebrew][6] package manager:
```
eerohele: Just a tip (that you maybe already know, but just in case):

To avoid fiddling with link reference numbers, you can just leave them out and create the links like this:

```markdown
…using the [Homebrew] package manager

[Homebrew]: https://brew..sh
```

infotexture: 👍  I'm aware of that option, but I generally prefer the readability of numbered link references,  especially in cases where the same target is used for different display text, as in the download links here:  `[download packages][4]` & `[dita-ot.org/download][4]`.

I have a script that de-duplicates & re-numbers them automatically, so easy to maintain too.

README.markdown
```diff
@@ -1,68 +1,104 @@
-# DITA Open Toolkit [![Build Status][1]](http://travis-ci.org/dita-ot/dita-ot) [![Slack][7]](http://slack.dita-ot.org/)
```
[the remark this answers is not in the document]
infotexture: 👍 Not necessarily if there's a clear preference for another style. 

_(So far, the Markdown files in our various repos use a mix of inline and reference links, some of which are numbered, others use words for the link identifiers, and others use implicit link names.)_

Since named link identifiers mimic DITA keys, and can be called with the [shortcut reference link syntax](https://spec.commonmark.org/0.18/#shortcut-reference-link) @eerohele proposed, I'm fine standardizing on those if that's what we want.

README.markdown
```diff
@@ -1,68 +1,104 @@
...
+- [Building output](#building-output)
+- [For developers](#for-developers)
+- [License](#license)
 
 ## Prerequisites
 
 To build and use DITA-OT, you’ll need:
 
-* Java Development Kit 8 or newer
+- Java Development Kit 8 or newer
```
jelovirt: This could be fixed in this PR. So to build DITA-OT you need JDK (Java Development Kit), but to use it, you can just use JRE (Java Runtime Environment.)

infotexture: Sure, I can clarify that while we're at it.

PR #3066, opened 2018-09-07
Fix PDF error when indexterm is in draft section + draft is off

robander:

## Description

The `pdf2` index pre-processing code adds indexing anchors in the content + references in the index for all terms. Later steps convert this to the proper XSL-FO markup.

The index pre-process does not check the `args.draft` parameter, which means that index entries inside of draft comments always generate anchors (inside of `draft-comment` or `required-cleanup`), and always generate references. When the processed file is converted to XSL-FO and `args.draft=no`, the anchors do not end up in the FO but the references do, resulting in errors like:
`      [fop] WARNING: Page 5: Unresolved ID reference "index in draft comment:_unique_d83e47" found.`
or
```
      [ahf] Error Code  : 10777 (2A19)
      [ahf] Unresolved index-key value: "index in draft comment:".
```

The fix checks the value for `args.draft`. If it is `yes`, these index entries are processed as today in PDF (they are included and links work). If it is `no` the index entries are not processed, no anchors and no references are generated.

Originally I looked for a quicker fix in the XSL-FO, but if we drop entries at that point we can still be left with empty group headings (for example, if all entries beginning with `z` are in draft sections, we would have an empty `Z` heading in the index). So, this really has to be handled in the index preprocessor that creates the groups.

## Motivation and Context

Came up in a customer document where index entries were moved into a `<required-cleanup>` element, and the build started throwing out unresolved ID errors.

## How Has This Been Tested?

Test bookmap included with index entries in a paragraph, in `draft-comment`, and in `required-cleanup`.
[3066.zip](https://github.com/dita-ot/dita-ot/files/2362141/3066.zip)

When `args.draft=yes`, all entries appear, and the draft content is indexed.

When `args.draft=no`, entries from draft content do not appear in the index. Also added entries beginning with `B` that *only* appear in draft content, and in this case the **B** heading is also removed.

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexPreprocessor.java
+++ b/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexPreprocessor.java
@@ -73,7 +73,7 @@ public final class IndexPreprocessor {
         this.namespace_url = theNamespace_url;
         this.excludedDraftSection.clear();
         this.excludedDraftSection.add(false);
-        includeDraft = draftParameter.equals(ARGS_DRAFT_YES) ? true : false;
+        includeDraft = draftParameter.equals(ARGS_DRAFT_YES);
         indexDitaProcessor = new IndexDitaProcessor();
         indexGroupProcessor = new IndexGroupProcessor();
     }
--- a/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexPreprocessor.java
+++ b/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexPreprocessor.java
@@ -54,8 +54,10 @@ public final class IndexPreprocessor {
     
     private final String prefix;
     private final String namespace_url;
+    private final Stack<Boolean> excludedDraftSection = new Stack<>();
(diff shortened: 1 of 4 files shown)


src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexPreprocessor.java
```diff
@@ -66,9 +68,12 @@
      * @param prefix index prefix
      * @param theNamespace_url index element namespace URI
      */
-    public IndexPreprocessor(final String prefix, final String theNamespace_url) {
+    public IndexPreprocessor(final String prefix, final String theNamespace_url, final String draftParameter) {
         this.prefix = prefix;
         this.namespace_url = theNamespace_url;
+        this.excludedDraftSection.clear();
+        this.excludedDraftSection.add(false);
+        includeDraft = draftParameter.equals(ARGS_DRAFT_YES) ? true : false;
```
jelovirt: Simplify this to just

```java
includeDraft = draftParameter.equals(ARGS_DRAFT_YES);
```

robander:
Not sure what happened - I had made this change locally and thought I'd pushed right after the review, but looks like I had a typo in branch names `draftindex` vs `draftIndex`. Not sure how I managed that :roll_eyes: Anyway updated now with the code review suggestion.

Issue #2110, opened 2015-10-23
Validation errors/warning in XSL-FO generated using DITA OT


raducoravu:
I generate PDF from a sample DITA Map using DITA OT 2.x (but probably it's the same issue with older versions). The PDF is generated succesfully. But if you open the XSL-FO file, it has lots of problems reported in it. And they seem to be quite easily fixable.

Like:

```
 <fo:wrapper id="tasks:_unique_d87e216"/>

 Description: value of attribute "id" is invalid; must be an XML name without colons
```

We should try to avoid ":" when constructing such ID attributes.

```
 <fo:basic-link color="blue" font-style="italic"
                                              internal-destination="flowers:autumn:chrysanthemums:_unique_d87e988"

value of attribute "internal-destination" is invalid; must be an XML name without colons
```

Same thing again, colons in ID references.

```
  <fo:page-number-citation
                                              ref-id="flowers:autumn:chrysanthemums:_unique_d87e988"
                                              />
```

Same thing again, colons in ID references.

```
    <fo:list-item-label end-indent="label-end()"
                                keep-together.within-line="always"
                                keep-with-next.within-line="always">

 Error: attribute "keep-with-next.within-line" not allowed here
```

The specs seems not to allow that attribute there.

jelovirt:
That looks like indexing code. Instead of colon we should use e.g. code point from Unicode private use area, that way we're pretty sure delimiter will not clash with index key value.

raducoravu:
Sure, as long as it remains a valid ID value. Or use instead of it some pattern like "-.-" or "-_-"

jelovirt:
Since in theory it's possible that an index term may use "-.-" in the key, IMO it's better to just use [a code point from PUA](https://en.wikipedia.org/wiki/Private_Use_Areas), e.g. simply `&#xE000;`

robander:
Ran across this when trying to debug a crash during my PDF build -- the PDF is 1500+ pages with 25+ pages of index, and the log is so full of errors about `:` that I can't find the actual errors. I tried replacing every one of the `:` characters in an index with `&#xF000;` as suggested and AH didn't like that character either --
`AHFCmd :WARNING: Invalid property value: id="publications_unique_d117e24383".`

robander:
[...]

PR #3073, opened 2018-09-14
Fix invalid IDs in FO index links

robander:
## Description

(Builds on #3066 so will need to be rebased if that one does not go in or requires further changes.)

Fix updates index processing in XSL-FO to ensure that index IDs are valid.

Currently index IDs use the literal term text plus the text of any ancestor terms, separated by a colon. These IDs are added during the Java preprocessor, and later copied as-is into XSL-FO attributes like `@id` / `@ref-id` / others. The colon is technically invalid in IDs, as reported in #2110, so some processors report this as an error. In addition, because the actual index text is part of the ID, any term that contains an invalid ID character (such as a space, comma, or most special characters) is reported as invalid.

The colon separator within IDs is used extensively in in the indexing logic to manage index terms, in particular to handle see, see-also, and index start/end ranges. This makes it difficult to fix the bug without modifying the whole design. Replacing every special character with a valid character or pattern is another option, but would result in more conflicts (replacing every special character with `_` would mean `test/key`, `test_key`, and `test`>`key` conflict).

The fix here leaves all of the existing linking logic alone. Instead the Java preprocessor adds a new attribute `indexid` that contains a hashed version of the current string; the characters in the hash are guaranteed to be valid as an XSL-FO ID. The logic for connecting terms in our XSLT is the same, still based on terms and `:` separators, but any place that previously generated an actual ID now uses the hash. 

This update also fixes a previously unreported bug with index ranges. The range used literal values from `indexterm/@start` in the FO `fo:index-range-begin/@id` attribute. When the start/end value contained spaces, Antenna House reported an error and no links appeared for the term. With the fix, the range appears and both start/end links work correctly.

## Motivation and Context

Motivated by #2110 which came up with a large (1000+ term) index, where most terms contained special characters; the formatting log was so full of index warnings that I could not find the meaningful warnings.

## How Has This Been Tested?

Tested with the sample set below, which extends the set from #3066 to add see, see-also, and index ranges. All links for all terms format properly with the fix, and the FO does not contain any invalid attributes.
[2110.zip](https://github.com/dita-ot/dita-ot/files/2383806/2110.zip)

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexPreprocessor.java
+++ b/src/main/plugins/org.dita.pdf2/src/com/idiominc/ws/opentopic/fo/index2/IndexPreprocessor.java
@@ -61,6 +61,7 @@ public final class IndexPreprocessor {
     private DITAOTLogger logger;
     private static final String elIndexRangeStartName = "start";
     private static final String elIndexRangeEndName = "end";
+    private static final String hashPrefix = "indexid";
 
     /**
      * Create new index preprocessor.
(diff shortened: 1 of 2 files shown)


robander:
Just noting -- I fixed the commit history to untangle from another index update in 3.1.3, the single commit now is the same one made earlier but without other obsolete commits in the way.

PR #3077, opened 2018-09-21
Fix Java failure when chunking is combined with uplevels

robander:
## Description

Fixes a bug with maps that use `by-topic` chunking but also have content outside of the map directory.

The update ensures that when a file is split into multiple chunks, the new chunked topics are created in the same directory as the original topic. Currently, this works correctly when the map is in the root of the temp directory, but fails when the map is _not_ in the root of the temp directory due to `uplevels` processing.

The fix changes how the temporary file name for the new chunk is created. Currently, the build properly determines what the result file would be in the original source context. However, when it creates the file name used to _generate_ the fragment, it includes the full path from the root of temp, and then appends that to the path of the current topic.

For example, assume I have this construct in the temp dir:
```
common/reuse.dita
main/input.ditamap
main/topic.dita
```

When `topic.dita` is split into multiple chunks, file names are created based on the IDs of each topic. For any but the first-in-file, those file names result in a relative path from temp such as `main/id.dita`. This is then evaluated _against_ the topic document itself, resulting in an attempt to open `main/main/id.dita`. This fails, which means variables are not initialized, which results in a build failure later in the chunking step: 
`Error: java.util.NoSuchElementException`

The update ensures we use only the base file name of the new chunk, relative to the document that is splitting.

## Motivation and Context

Encountered in customer document that uses the same organization as the new sample integration test doc.

## How Has This Been Tested?

Passes unit / integration tests; new integration test added that is fixed with the update.

## Type of Changes

## Checklist


src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
```diff
@@ -242,12 +242,12 @@ FileInfo generateFileInfo(final URI output) {
     URI generateOutputFilename(final String id) {
         final FileInfo cfi = job.getFileInfo(currentFile);
         URI result = cfi.result.resolve(id + FILE_EXTENSION_DITA);
-        URI temp = tempFileNameScheme.generateTempFileName(result);
```
jelovirt: I think this will break things. The `generateTempFileName` method will turn a source file name into a temporary file name, e.g. a hash of the absolute URI. Just taking the source file name and using that is probably incorrect.

robander: I thought so initially ... but in the course of changing this line, I added 30 or 40 debug lines, and that doesn't seem to be the case with the current code (you're commenting on the old code but I assume you're questioning the change away from this). Details:

- `currentFile` is the map
- `URI result` creates a URI using the ID of the new chunked topic; it creates the file name as a "source file", in the same source directory as `currentFile'. So you get a value like: `/path/to/map/topicid.dita`
- `generateTempFilename` returns the portion of that path that would exist inside `temp` -- the portion relative to the common directory accounting for uplevels. If there are no uplevels, it just returns `topicid.dita`. If uplevels goes up one directory, it returns `map/topicid.dita`. 
- That portion of the path is then evaluated, and the file generated in temp, _relative to the chunked document_. So with uplevels, when chunking `temp/map/doc.dita`, we try to chunk the new topic as `temp/map/map/topicid.dita` (This fails further down, meaning something isn't added to a stack, which even later results in an NPE.)
- The method name is confusing, but ... `generateTempFileName` isn't generating a hash, it's taking the already-determined file name in a source directory, and returning the portion of path which fits in `temp`.

So there is confusion between `generateTempFilename(result)` -- which returns part of the URI without modifying it -- and `generateFilename()`, which actually generates a name.

If you can ignore the method names, what's happening today is:

- Generate a file name with the topic ID + `.dita`, as if it was a source file
- Take from that the portion that fits within temp -- _this bit is broken and later ends up with an NPE_
- If we couldn't get a file name, OR if that file name already exists within temp, go into the exception code on line 246 to find a new file name
- Exception code calls `generateFilename()` to get a hash. Today, that code also has the same problem as above (hash is created as a source file, we take the part that fits within `temp`, and then evaluate that relative to the chunked doc). So I've made the same fix within the exception code -- the hashed name should be relative to the chunked doc, regardless of uplevels.

robander:
Reworked after discussion with Jarno about how the file name generation was working.

Current version is changed to:

- Generate the file name relative to the file that's being chunked (rather than relative to the map). This means that when file/directory names are relevant, splitting up a document results in new chunks within the same directory
- The temporary chunk name is created relative to the temp directory, as before. To fix the original bug that prompted this, that value is now made absolute by evaluating it relative to the same temp directory, rather than relative to the chunked document.

src/test/resources/chunk_uplevel/src/main/composite.dita
```diff
@@ -0,0 +1,37 @@
...
+<concept id="chunkmetoo">
+<title>Second topic in main dir!</title>
+<conbody>
+<p>thing</p>
+</conbody>
+</concept>
+</topic>
+</dita>
+
+<!--
```
jelovirt: Maybe remove this comment because the log shown here is not what the final modifications will output.

robander: Shouldn't have been in there either way, leftover from my debugging of the test file itself. Thanks...

PR #3079, opened 2018-09-25
Optimize topicpull processing by skipping resource-only

robander:
## Description

Updates `topicpull` to skip processing `resource-only` topics / topics that are treated as `resource-only`.

None of the default processing in DITA-OT needs to run `topicpull` on files that are explicitly resource-only. The topics will not be published; `topicpull` comes last in the preprocessing pipelines, so for normal cases, we are done with those topics at the end of `preprocess`. Skipping `topicpull` on those topics optimizes processing for builds with a large number of `resource-only` topics, and also avoids errors for the condition described below.

## Motivation and Context

For all of our builds we set `onlytopic.in.map=true` by default. This parameter has a few problems around the edge cases, but many of the problems have been fixed in the last couple releases. One thing it does well is handle the condition where `a.dita` has a link to `b.dita` (where B is not in the map), and `b.dita` has unrelated cross references to other topics; the early stages of processing no longer try to follow those cross references, and the link targets are not copied to `temp` even when present.

However, the `resource-only` topics (with all other topics) are tossed into `topicpull`. When the `resource-only` file `b.dita` runs, it attempts to pull link text for cross references that will never be used. When those files were not pulled into `temp` to begin with, this results in Java errors and `topicpull` errors. For example:
```
[topicpull] I/O error reported by XML parser processing file:/C:/DITA-OT/TESTGITHUB/topicpull/temp/linkFromResourceOnly-NotNeeded.dita: C:\DITA-OT\TESTGITHUB\topicpull\temp31\linkFromResourceOnly-NotNeeded.dita (The system cannot find the file specified.)
[topicpull] file:/C:/DITA-OT/TESTGITHUB/topicpull/resourceonly.dita:6:77: [DOTX056W][WARN]: The file 'linkFromResourceOnly-NotNeeded.dita' is not available to resolve link information.
```

- When `onlytopic.in.map=true`, files that are only link targets (not in the map) are treated as `resource-only` along with any topics that the map explicitly marks `resource-only`. In this case, there is no need to run `topicpull` on any of those files. Using the sample above, there is a potential edge case where `b.dita` has a cross reference within the title, so that link text must be resolved in order to establish link text for links to `b.dita`. This is already broken today, and should probably be handled by adding that link target to the map as `resource-only`.
- Otherwise, in the scenario above, both `a.dita` and `b.dita` will generate output; neither is marked as `resource-only`, so all links are followed, and `topicpull` runs on each. So for that condition, there is no change to today's behavior.

## How Has This Been Tested?

Created sample files to produce the error condition; this is fixed with the update.

Two sample sets in our integrationTest were affected by this update. In each case, we were doing extra (unnecessary) processing on topics that were link or conref targets; those topics were not rendered. So, it is safe to update those sample sets so that the `resource-only` topics are left with unresolved cross references.

## Type of Changes

Blurs the line between bug fix (it removes an invalid error we have today) and breaking change (feature that optimizes processing while producing different results within some temporary artifacts).

## Checklist


jelovirt:
I'm a bit worried this might break something we haven't thought of, but this is easy to revert.

PR #3090, opened 2018-10-06
Extract attributes from structural HTML5 elements into attribute sets

jelovirt:
Extract ARIA attributes from structural HTML5 elements into attribute sets to enable easier override and extension with e.g. custom `class` attributes.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/nav.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/nav.xsl
@@ -24,9 +24,13 @@ See the accompanying LICENSE file for applicable license.
     <xsl:apply-templates select="document($input.map.url)" mode="normalize-map"/>
   </xsl:variable>
 
+  <xsl:attribute-set name="toc">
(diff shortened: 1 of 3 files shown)


infotexture:
> enable easier override and extension with e.g. custom `class` attributes.

Do I understand correctly that this would be achieved by overriding the named attribute sets in a custom plug-in to add whatever other attributes are required?

So for output with an ID and custom classes such as: 

```html
<main id="content" class="col-md-9 container" role="main">
  ⋮
</main>
```

a custom plug-in would need to re-define the "main" attribute set as:

```xml
<xsl:attribute-set name="main">
  <xsl:attribute name="id">content</xsl:attribute>
  <xsl:attribute name="class">col-md-9 container</xsl:attribute>
  <xsl:attribute name="role">main</xsl:attribute>
</xsl:attribute-set>
```

❔

jelovirt:
@infotexture Yes, exactly. However in some cases the output `class` attribute is generate using `@outputclass` value, so this may not always apply. But generally speaking this is the goal.

PR #3091, opened 2018-10-08
PDF2: Force footnotes to 'normal' when anchor is bold

lief-erickson:
PDF2: Force footnotes to 'normal' in situations were the footnote was added to a title, such as a fig or table title, or some other style that is bolded. The footnote body was inheriting the bold property.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/topic-attr.xsl
+++ b/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/topic-attr.xsl
@@ -455,6 +455,8 @@ See the accompanying LICENSE file for applicable license.
         <xsl:attribute name="line-height">1.2</xsl:attribute>
         <xsl:attribute name="start-indent">0pt</xsl:attribute>
         <xsl:attribute name="font-weight">normal</xsl:attribute>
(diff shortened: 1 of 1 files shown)


lief-erickson:
@robander, do you want me to add `<xsl:attribute name="font-style">normal</xsl:attribute>` to my PR?

robander:
@lief-erickson Sure, that would be great. I assume you've got a test case with bold, so should double check one with italic to make sure that fixes it. And of course while thinking of it ... may want to add footnotes in `<u>` and `<overline>`, see if they have the issue + see if the following fixes it?
`<xsl:attribute name="text-decoration">no-underline no-overline</xsl:attribute>`

lief-erickson:
@robander, to test further I put footnotes in the following:

- `<i>`: Result = italicized footnote
- `<b>`: Resulte = bolded footnote
- `<title>` (both `<fig>` and `<table>`: Result = bolded footnote
- `<thead>`: Result = bolded footnote
- `<u>`: Result = underlined footnote
- `<overline>`: Result = overlined footnote

Some examples required wrapping the `<fn>` in a `<ph>` to make the DITA valid, since `<fn>` cannot be a direct child of most of these elements.

PR #3103, opened 2018-10-14
HTML5: Made Dublin Core property names lower-case

StefanGentz:
## Expected Behavior

[The Guidelines for implementing Dublin Core in XML](http://dublincore.org/documents/dc-xml-guidelines/), Section 4.2 Implementation guidelines says:

> “Recommendation 4. The property names for the 15 DC elements should be all lower-case.”
[Update:] The prefic DC needs to be all upper-case, the property name all-lowercase.
```xml
<meta name="DC.property1" content="foo" />
<meta name="DC.property2" content="foo" />
```

## Actual Behavior
Dublin Core Properties are currently output in upper-case and/or a mixture of them
```xml
<meta name="DC.property1" content="foo" />
<meta name="DC.Property2" content="foo" />
```

## Possible Solution
I have changed the property names as per Dublin Core recommendation and made them all lower-case.

Signed-off-by: Stefan Gentz <gentz@adobe.com>

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/get-meta.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/get-meta.xsl
@@ -159,7 +159,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:apply-templates select="*[1]" mode="gen-type-metadata"/>
   </xsl:template>
   <xsl:template match="*" mode="gen-type-metadata">
-    <meta name="dc.type" content="{name(.)}"/>
+    <meta name="DC.type" content="{name(.)}"/>
   </xsl:template>
   
   <xsl:template match="*[contains(@class,' topic/abstract ')]" mode="gen-shortdesc-metadata">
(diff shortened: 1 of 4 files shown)


infotexture:
@StefanGentz 👍 Thanks for this suggestion.

Can you update your commit and [sign your work](http://www.dita-ot.org/DCO) to give us permission to use  your contribution?

To sign off your last commit without changing the commit message or associated files, use:

```
git commit -s --amend --no-edit
```

Then force-push to update the branch on your fork with:

```
git push --force-with-lease
```

robander:
This change looks right, but also breaks several of our automated tests -- the generated HTML5 no longer matches the expected result, because the expected result still uses the old values like `DC.Type`. The pull request should also modify the expected result files, like the one here: https://github.com/dita-ot/dita-ot/blob/develop/src/test/resources/html5_css/exp/html5/topic.html#L8

There are only a few here, but the related pull request for XHTML is going to hit a whole lot of them. If you'd like, I can fix these with a search/replace update and push them in here so that you don't need to worry about it this time.

Issue #2019, opened 2015-08-27
Support RELAXNG parsing and validation


drmacro:
Integrate George's RNG DTD compatibility libraries into the OT's DITA parsing code.

nakohdo:
+1 from me

JFTR linking this to the discussion in the DITA-OT Users group
["Using DITA-OT with RNG files?"](https://groups.google.com/forum/#!topic/dita-ot-users/0XZAloq2cOY)

jelovirt:
If possible, we should handle this as plugin to the parsing infrastructure. Can you provide the code that's to be integrated and info how forks of OT currently handle this?

drmacro:
Yes, we have a plugin that does this (George set it up originally as part of his ditarng project). I'll take the task of reworking it as something that can be part of the OT plugin set and submit a pull request.

queshaw:
I have a fork of DITA-NG that adds schema validation support, if that interests anyone. I made a fork here:

https://github.com/queshaw/dita-ng/tree/validation

Essentially, if a properties file named **dita-ng.properties** contains **validate=yes**, after the document is augmented with default values (class attribute etc.), the parse tree is validated using the validation part of Jing (org.thaiopensource.relaxng.sax.PatternValidator).

I also added commands for invoking DITA-OT with the classpath set. They are wrappers around the  dita command. You can install the plugin using the command, e.g.:

bin/dita-ng -install somewhere/dita-ng20150914.zip -v
bin/dita-ng -uninstall org.dita-ng.doctypes -v

And, I added some unit tests.

drmacro:
This seems like appropriate default behavior. I would think most, if not all, users would expect this level of validation.

queshaw:
I've made more changes that I hope might serve as at least exercising dita-ng.

In my fork, I split the plugin into two plugins, one is a library plugin, that adds dita-ng and jing to the classpath, the other is a catalog plugin.

I also changed the flowers demo files to use the OASIS urns and added the dita spec from CS01 part3 all inclusive. And I got rid of the dita-ng commands since you don't need to adjust the classpath.

So, for testing the integration you can process the DITA spec. I hope there isn't a problem with including the spec.

For the build, there is separate dist target to do the conversion and dist thing that was there and then there is dist.minimal which creates the two separate plugins and doesn't run any conversion.

queshaw:
Oh, and I made validation the default.

jelovirt:
[...]

PR #3106, opened 2018-10-16
Add support for using RNG schema in input files

jelovirt:
Copy Xerces parser configuration from https://github.com/georgebina/dita-ng and use forked and patched Jing.

Implements #2019.


jelovirt:
@raducoravu you should be able to add a review and approval to this PR.

jelovirt:
I don't know if I should also format the Java code to match our coding conventions. It makes it more difficult to get upstream changes to files, but given that the files haven't changed in upstream for 6 years, maybe it's not a problem.

Issue #3109, opened 2018-10-18
Provide CLI feedback for plug-in installation


infotexture:
## Description

The `dita --install` option currently provides no confirmation when plug-ins are successfully installed (or removed). 

This is consistent with the typical "no news is good news" approach for basic UNIX utilities, and tolerable enough when installing or removing a single plug-in.

However, the current behavior is less helpful in situations when the `plugins` directory is manually modified to add/remove multiple plug-ins at once, ~or install multiple plug-ins from a single .zip file~.

## Possible Solution

Since DITA-OT maintains a list of currently installed plug-ins and updates this configuration during the integration process, it would be helpful to compare the before/after states and print a summary to the console upon completion, for example:

```console
$ dita --install
+ Installed com.example.plugin1
- Removed com.example.plugin2
```

It might also be useful to conclude the integration process by printing the current list of available plug-ins to the console via the `dita --plugins` option. 

Since the plug-in registry changes for #3028 also permit the installation of plug-ins by alias, this would also serve to confirm the user's intent by displaying the plug-in ID that was actually installed via the alias.

This would be more consistent with the feedback provided by other CLI-based package managers.

_(See related enhancement #3070)._

jelovirt:
>  install multiple plug-ins from a single .zip file

Does that work? 🤔

infotexture:
My recollection was that multiple plug-ins could be installed simultaneously by passing a .zip file that contained several plug-ins, each in top-level folders of the archive.

But I tested now, and only 1 of the 4 plug-ins in the archive was installed _(first in alphabetical order)_.

@robander Did I dream this, or did it used to work that way? 

Having to unpack the archive to the `plugins` folder first seems like an unnecessary step.

robander:
I think I've asked about this in the past but it was never supported - like @jelovirt I was a bit surprised by that in the initial report. It came up for me when first working with some of the D4P and SVG Diagram support, each of which required a bundle of plugins to work. I think I asked about that at the time and it was clearly out of scope for the initial `dita --install` support, but I hadn't thought about it since.

infotexture:
Resolved in #3110 & a17143c.

PR #3110, opened 2018-10-18
Log added and removed plugins during integration #3109

jelovirt:
Log IDs of added and removed plugins during integration at info logging level.

### DIFF
--- a/src/main/java/org/dita/dost/platform/Integrator.java
+++ b/src/main/java/org/dita/dost/platform/Integrator.java
@@ -156,6 +156,8 @@ public final class Integrator {
      * Execute point of Integrator.
      */
     public void execute() throws Exception {
+        final Set<String> pluginList = getPluginIds(readPlugins());
+
         // Read the properties file, if it exists.
         properties = new Properties();
         if (propertiesFile != null) {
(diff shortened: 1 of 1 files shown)


jelovirt:
@infotexture are you using verbose logging in CLI?

infotexture:
No, just the defaults.

jelovirt:
CLI will not log anything when there are no errors in default logging level, so this is working as expected.

infotexture:
OK, I see the messages if I run `dita --install --verbose` instead.

I think it would be more helpful to _always_ see these messages in this case, but I'll defer to majority opinion if I'm the outlier here.

jelovirt:
Since the CLI uses the UNIX convention to not log anything when no errors occur, changing this for install only doesn't make sense to me.

infotexture:
I understand the rationale, it just seems odd compared to the feedback provided by other package managers.

With Homebrew, I get this:

```console
$ brew uninstall svgo
Uninstalling /usr/local/Cellar/svgo/1.0.5... (875 files, 4.6MB)
```

With this approach for `dita --install`, I get either nothing, or much more than I needed:

```console
$ dita --uninstall com.example.something --verbose
Searching for integrator.xml ...

uninstall:

uninstall.delete:
   [delete] Deleting directory /path/to/dita-ot/plugins/com.example.something

integrate:
[integrate] Removed com.example.something
      [jar] Building jar: /path/to/dita-ot/lib/dost-configuration.jar
```

The extraneous info here obscures the main result: `Removed com.example.something`.

— but I'll take what I can get…

jelovirt:
The verbose logging for install is due to the actual process being implemented with Ant. One option to fix this would be to use a custom Ant log listener.

jelovirt:
Tested what the output would be if the logging level was moved to warning. The output would be

```bash
$ dita --install org.dita.troff     
[integrate] Added org.dita.troff
```

Which is not what we want either. We need to tweak the logger that's used with the CLI call to Ant.

PR #3111, opened 2018-10-18
Rev marking should cover whole part, topic

robander:

This update changes when revisions (stored in `dita-ot/startprop`) are processed for full topics. It moves processing to the start of topic titles, for the following benefits:

- When revision change bars are specified for an entire topic, the change bar now extends from the title of the topic down to the end of the body. Previously it only extended along the body, which some of my customers reported to me as a defect.
- When revision images are specified for start-change, the image now appears at the start of the title, which helps clarify that the full topic is revised; previously it appeared on its own between the title and the body.
- Previously change bars were missing for parts, chapters, and a few other bookmap-specific topics when the entire topic was revised, due to the way that those topics are processed to set up special flows. This update ensures that change bars now appear alongside parts / chapters / etc when needed.

This is a follow-on to #3100 which added revision bar support for entries in the TOC (I mentioned some of these outstanding issues in that update). It uses the same source maps + DITAVAL to test basic maps and bookmaps with change bars + image flagged + color revisions. https://github.com/dita-ot/dita-ot/files/2470022/revbar.zip

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -123,9 +123,8 @@ See the accompanying LICENSE file for applicable license.
         <xsl:apply-templates select="." mode="customTopicMarker"/>
         <xsl:apply-templates select="*[contains(@class, ' topic/title ')]"/>
         <xsl:apply-templates select="*[contains(@class, ' topic/prolog ')]"/>
-        <xsl:apply-templates select="*[not(contains(@class, ' topic/title ')) and
-                                       not(contains(@class, ' topic/prolog ')) and
-                                       not(contains(@class, ' topic/topic '))]"/>
+          <xsl:apply-templates select="* except(*[contains(@class, ' topic/title ') or contains(@class,' ditaot-d/ditaval-startprop ') or
+              contains(@class, ' topic/prolog ') or contains(@class, ' topic/topic ')])"/>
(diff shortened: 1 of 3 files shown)


infotexture:
@robander Would you be able to review the information in the docs topic on [Generating revision bars](http://www.dita-ot.org/dev/topics/pdf2-creating-change-bars.html) and provide input on any changes required to reflect the new processing in #3100 and this PR?

robander:
@infotexture These two PRs don't change anything about how the change bars work, just where they appear when specified. So I was all set to say "that looks mostly right" but now I think this must have changed quite a bit over the last few years.

- The current code is just checking `@changebar` for known values that map to the XSL-FO style, and otherwise taking `groove` as the default style (so if you specify `changebar="QATesting"` you'll get the `groove` style. https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2/xsl/fo/flagging-from-preprocess.xsl#L138
- It's taking color from `@color` so that you don't have to specify that in the `changebar` attribute (in fact doing so will get you the groove style): https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2/xsl/fo/flagging-from-preprocess.xsl#L144
- Doesn't look like there is any way to set the offset (defaults to `2mm2), placement, or width

I think this must have been part of the change to work with preprocess-based flagging, moving away from the suite-sol flagging that operated inconsistently in earlier releases, but I'm not certain.

PR #3115, opened 2018-10-19
Div should inherit common block processing in PDF

robander:

Noticed that the PDF format for `<div>` does nothing more than add a newline; seems that it should be rendered like other blocks by default, separated by a bit of space from the previous block (and pick up any standard attribute overrides from the `common.block` set).

Tested with a topic that contains:
```
<body>
  <p>Topic paragraph</p>
  <p>Topic paragraph</p>
  <p>Topic paragraph</p>
  <div>hey this is div</div>
  <div>hey this is div
    <div>hey this is NESTED div</div>
    <div>hey this is NESTED div</div></div>
  <div>hey this is div</div>
  <p>Topic paragraph</p>
  <p>Topic paragraph</p>
  <p>Topic paragraph</p>
  <div>hey this is div</div>
  <div>hey this is div</div>
</body>
```

Without the update, there is no way to distinguish between the divisions other than the newline (which is meaningless if the lines are very long). After the update the `<div>` containment is clear.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/topic-attr.xsl
+++ b/src/main/plugins/org.dita.pdf2/cfg/fo/attrs/topic-attr.xsl
@@ -213,7 +213,7 @@ See the accompanying LICENSE file for applicable license.
         <xsl:attribute name="start-indent">72pt</xsl:attribute>
     </xsl:attribute-set>
 
-    <xsl:attribute-set name="div">
+    <xsl:attribute-set name="div" use-attribute-sets="common.block">
     </xsl:attribute-set>
 
     <xsl:attribute-set name="p" use-attribute-sets="common.block">


drmacro:
Div that is a peer to other blocks should behave like a block but div within blocks should use fo:inline and should not generate a line break the in the result.

robander:
> Div that is a peer to other blocks should behave like a block but div within blocks should use fo:inline and should not generate a line break the in the result.

I can easily see that for specializations of `<div>`, but I think that would be a very strange default. It's always mapped directly to `<div>` in all cases for our XHTML / HTML5 output, and I think anything otherwise would surprise people based on the name and the commonality with the matching HTML `<div>` element. PDF should match that as well - it's always put out a newline for `<div>` regardless of context, it's just that right now that is inconsistent with other blocks.

FWIW, the spec topic on `<div>` doesn't mention formatting, and the non-normative spec appendix with default presentation / translation rules just describes it as having block formatting: http://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part1-base/non-normative/elementsMerged.html#elements

Issue #1381, opened 2012-09-06
Retain conref and keyref information after preprocessing


jelovirt:
Add namespaced attributes to retain conref and keyref source information. This will allow advanced users to create e.g. review PDFs which show the source of each block.

The feature should be controlled by dynamic configuration; by default the feature should be disabled.

ToshihikoMakita:
I **DO** agree Jarno's opinion. When DITA users uses conref/conkeyref more frequently, they needs the original source information for reviewing process. I've requested this capability from user and modified to retain them by modifying DitaWriter.java in DITA-OT 1.7.5. If this capability is implemented, it is strongly appreciated. Are there any possibilities to implement this?

robander:
We already retain `@keyref` today in the temp file (it's still there after the attribute is resolved). We do not retain `@conref` or `@conkeyref`.

PR #3119, opened 2018-10-20
Handle namespace atts in conref push module

robander:
This came up when I was playing with a fix for #1381.

I added a namespaced attribute `@dita-ot:orig-conkeyref` onto an element for debug purposes. The namespace was defined on the same element. When the conref push filter tried to write that element, it threw an exception because it wrote out `@dita-ot:orig-conkeyref` but did not preserve the definition of `xmlns:dita-ot`:
```
conrefpush:
[conref-push] Reading  C:\DCS\test\temp\topic1.dita
[conref-push] The namespace URI "http://dita-ot.sourceforge.net/ns/201007/dita-ot" has not been bound to a prefix.
[conref-push] javax.xml.stream.XMLStreamException: The namespace URI "http://dita-ot.sourceforge.net/ns/201007/dita-ot" has not been bound to a prefix.
[conref-push]   at org.dita.dost.reader.ConrefPushReader.putElement(ConrefPushReader.java:304)
[conref-push]   at org.dita.dost.reader.ConrefPushReader.startElement(ConrefPushReader.java:187)
[conref-push]   at org.apache.xerces.parsers.AbstractSAXParser.startElement(Unknown Source)
[conref-push]   at org.ditang.relaxng.defaults.RelaxNGDefaultsComponent.startElement(RelaxNGDefaultsComponent.java:210)
[conref-push]   at org.apache.xerces.impl.XMLNSDocumentScannerImpl.scanStartElement(Unknown Source)
[conref-push]   at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl$FragmentContentDispatcher.dispatch(Unknown Source)
...
```

When running the integration tests with an update that preserved `@conref` with `@dita-ot:orig-conref`, a large number of tests failed during `conref-push` with that same exception.

I updated the JUnit for `conref-push` and got an exception when it tried to work with the namespaced attribute. The fix here ensures that when writing out any attribute with a namespace, that namespace is also declared. It also ensures that we only define the namespace once for any given element.

### DIFF
--- a/src/main/java/org/dita/dost/reader/ConrefPushReader.java
+++ b/src/main/java/org/dita/dost/reader/ConrefPushReader.java
@@ -14,8 +14,10 @@ import static org.dita.dost.util.URLUtils.*;
 import java.io.File;
 import java.net.URI;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.Map;
+import java.util.Set;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.stream.FactoryConfigurationError;
@@ -253,6 +255,7 @@ public final class ConrefPushReader extends AbstractXMLReader {
         //conref information like @conref @conaction in current element
         //when copying it to pushcontent. True means remove and false means
         //not remove.
+        final Set<String> namespaces = new HashSet<>(32);
         try {
             pushcontentWriter.writeStartElement(elemName);
             for (int index = 0; index < atts.getLength(); index++) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/dita/dost/reader/ConrefPushReader.java
```diff
@@ -253,6 +255,7 @@ private void putElement(final String elemName, final Attributes atts, final bool
         //conref information like @conref @conaction in current element
         //when copying it to pushcontent. True means remove and false means
         //not remove.
+        final Set<String> namespaces = new HashSet<>(32);
```
jelovirt: No need to set the default size of the set because Java in most cases does the correct thing and here the explicit set size makes the user assume 32 actually means something. Using an explicit initial size in the OT code base should be removed.

src/main/java/org/dita/dost/reader/ConrefPushReader.java
```diff
@@ -266,9 +269,14 @@ private void putElement(final String elemName, final Attributes atts, final bool
                     }
                     final int offset = atts.getQName(index).indexOf(":");
                     final String prefix = offset != -1 ? atts.getQName(index).substring(0, offset) : "";
+                    if (!namespaces.contains(prefix)) {
+                        namespaces.add(prefix);
+                        if (!(prefix.equals(""))) {
```
jelovirt:
```suggestion
                        if (!prefix.isEmpty()) {
```

PR #3121, opened 2018-10-20
Add license file generation to dist

### LINKED ISSUE #3105: Review third party dependencies
There is a "NOTICES.txt" which does not seem to enumerate all JAR libraries found in the DITA OT "lib" folder, libraries like "checker-qual", "animal-sniffer", "error_prone_annotations". Maybe some of these libraries are not necessary to be present in the installation kit, maybe they are used only when developing the code.
For example the "checker-qual" has dual licensing and from what I talked with our legal, the NOTICES could state what license is used for it (MIT instead of GPL).

This list in the docs should probably also be updated:
http://www.dita-ot.org/dev/reference/third-party-software.html

jelovirt:
Generate `.LICENSE` file for every third-party Jar file we ship in our distribution.

Fixes #3105


jelovirt:
This unfortunately doesn't cover all cases and we likely have to also add `.NOTICES` file in some cases. For example Saxon contains code that is not in a separate JAR file and for those case we would need to ship a notices file that concatenates these notices.

jelovirt:
The same code could also be modified to generate a single NOTICES that just contains licenses for all third-party code. That actually might be better than the current code.

jelovirt:
@raducoravu Are you able to test the current code? It doesn't copy licenses into the `lib` anymore, it just generates `NOTICES.txt`. One thing that would now be needed would be to go through each third-party library and make sure we also have their NOTICES file included. For Saxon I already did this, so it's license and four additional notices. I think e.g. xml-apis has the same thing, we need to include SAX and DOM licenses.

gradle/dist.gradle
```diff
@@ -0,0 +1,89 @@
+import java.nio.file.Paths
+
+import static groovy.io.FileType.FILES
+
+def licenses = [
+        [name: 'avalon-framework', title: 'Avalon Framework', license: ['apache2.txt']],
```
jelovirt: * `name`: Maven name or common prefix for this library. It's used to detect we have a license definition for each JAR we bundle.
* `title`: Name of the library
* `license`: Array of license and notices files.

gradle/dist.gradle
```diff
@@ -0,0 +1,89 @@
...
+        [name: 'Saxon-HE', title: 'Saxon-HE', license: ['mpl2.txt', 'CERN.txt', 'JAMESCLARK.txt', 'THAI.txt', 'UNICODE.txt']],
+        [name: 'icu4j', title: 'ICU4J', license: ['icu.txt']],
+        [name: 'ant', title: 'Ant', license: ['apache2.txt']],
+        [name: 'guava', title: 'Guava', license: ['apache2.txt']],
+        [name: 'slf4j', title: 'SLF4J', license: ['mit-slf4j.txt']],
+        [name: 'jcl-over-slf4j', title: 'SLF4J', license: ['mit-slf4j.txt']],
+        [name: 'logback', title: 'Logback', license: ['epl-1.0.txt']],
+        [name: 'jackson', title: 'Jackson', license: ['apache2.txt']],
+        [name: 'jing', title: 'Jing', license: ['bsd-jing.txt']],
+        [name: 'gradle-wrapper'],
```
jelovirt: These license entries without `licence` are markers for our own JAR files, so we know to skip them.

gradle/dist.gradle
```diff
@@ -0,0 +1,89 @@
...
+        [name: 'fo'],
+        [name: 'eclipsehelp'],
+        [name: 'eclipsehelp'],
+        [name: 'htmlhelp'],
+        [name: 'htmlhelp'],
+        [name: 'xep'],
+        [name: 'axf'],
+        [name: 'dost'],
+        [name: 'error_prone_annotations'],
+        [name: null, title: 'DITA 1.1 DTDs and Schemas', license: ['oasis-dita11.txt']],
```
jelovirt: In some cases the third-party library is not a JAR file.

gradle/dist.gradle
```diff
@@ -0,0 +1,89 @@
...
+            def key = licenses.find { it.name != null && name.startsWith(it.name) }
+            if (key == null) {
+                throw new GradleScriptException("License for ${name} not found", new RuntimeException());
+            }
+        }
+
+        file("${distTempDir}/NOTICES.txt").withWriter('UTF-8') { out ->
+            out.writeLine("""DITA Open Toolkit: Third Party Notices
+
+This Notices file contains certain notices and important information that the
```
jelovirt: This header comes from the old NOTICES.txt, probably originally from IBM. Ping @robander

gradle/dist.gradle
```diff
@@ -0,0 +1,89 @@
...
+        [name: 'axf'],
+        [name: 'dost'],
+        [name: 'error_prone_annotations'],
+        [name: null, title: 'DITA 1.1 DTDs and Schemas', license: ['oasis-dita11.txt']],
+        [name: null, title: 'DITA 1.2 DTDs and Schemas', license: ['oasis-dita12.txt']]
+]
+
+task generateNotices() {
+    doLast {
+        file(distTempDir).traverse(type: FILES, nameFilter: ~/.*\.jar$/) {
```
jelovirt: A simple verification that each JAR we ship is covered by the NOTICES file.

raducoravu:
@jelovirt I checked out your "feature/notices" branch, how do I generate the "NOTICES.txt"? I invoked the "gradlew" and built the project but the notices file did not appear.

gradle/dist.gradle
robander: Probably correct, though that would have gone back to release DITA-OT 1.0 and I didn't get involved in this part of the project until right _after_ that, so I can't confirm (Original project lead Don and any lawyers involved at that time have all retired by now). Were you hoping to remove the text or just looking for the source?

gradle/dist.gradle
```diff
@@ -1,53 +1,87 @@
...
+            }
+        }
+
+        file("${distTempDir}/NOTICES.txt").withWriter('UTF-8') { out ->
+            out.writeLine("""DITA Open Toolkit: Third Party Notices
+
+This Notices file contains certain notices and important information that the
+DITA Open Toolkit Project licensors and contributors (the “Project”) are
+required to provide to you with respect to certain third party components
+included in the DITA Open Toolkit (the “Program”).
```
infotexture: Drop the article per project preferences:

> included in ~the~ DITA Open Toolkit (the “Program”)

jelovirt: > @jelovirt I checked out your "feature/notices" branch, how do I generate the "NOTICES.txt"? I invoked the "gradlew" and built the project but the notices file did not appear.

You should run `./gradlew dist`

jelovirt:
> @jelovirt I checked out your "feature/notices" branch, how do I generate the "NOTICES.txt"? I invoked the "gradlew" and built the project but the notices file did not appear.

You should run ./gradlew dist

raducoravu:
With the latest code on your notices branch, running the dist target fails for me with:

> Task :generateNotices FAILED

>FAILURE: Build failed with an exception.

>* Where:
>Script 'C:\Users\radu_coravu\Documents\dita-ot\gradle\dist.gradle' line: 42

>* What went wrong:
>License for commons-logging-1.0.4.jar not found
> java.lang.RuntimeException (no error message)

jelovirt:
@raducoravu did you run it from a fresh clone that has no old untracked files or additional plugins?

jelovirt:
The code is ready, but XDITA DTDs plugin is blocking this. We cannot use their [LICENSE.md](https://github.com/oasis-tcs/dita-lwdita/blob/master/LICENSE.md) because it just links to other files and doesn't state the conditions. If they cannot provide a single license file, IMO we cannot bundle their plugin in 3.2 release.

raducoravu:
@jelovirt My checkout branch is clean. Now, even if I manually remove the "build" folder before running "gradlew dist" I get at some point:

    install:
         [get] Getting: https://github.com/dita-ot/org.dita.troff/archive/3.1.1.zip
         [get] To: C:\Users\RADU_C~1\AppData\Local\Temp\4407065594203523696\plugin.zip
         [get] https://github.com/dita-ot/org.dita.troff/archive/3.1.1.zip moved to     https://codeload.github.com/dita-ot/org.dita.troff/zip/3.1.1
        [unzip] Expanding: C:\Users\RADU_C~1\AppData\Local\Temp\4407065594203523696\plugin.zip into C:\Users\RADU_C~1\AppData\Local\Temp\4407065594203523696\plugin
     BUILD FAILED
      C:\Users\radu_coravu\Documents\dita-ot\build\tmp\dist\integrator.xml:70:       java.lang.IllegalStateException: Plug-in org.dita.troff already installed: C:\Users\radu_coravu\Documents\dita-ot\build\tmp\dist\plugins\org.dita.troff

jelovirt:
@raducoravu 

> Plug-in org.dita.troff already installed

It's not enough that the `build` directory is clean. The whole `src` directory needs to be clean too. You seem to have at least troff plugin already in the plugins directory.

raducoravu:
@jelovirt I will check again, if we have any feedback on the notices, we'll add a new issue for it. I would expect that the "dist" target cleans up all existing plugins/folders before trying to do the same thing again.

raducoravu:
@jelovirt tried, again, manually deleted the folder "dita-ot\src\main\plugins\org.dita.troff" folder because somehow it was there although it was not showing in the Git client, run again the "gradlew dist", failed here:

       > Task :generateNotices FAILED
          FAILURE: Build failed with an exception.
         * Where:
           Script 'C:\Users\radu_coravu\Documents\dita-ot\gradle\dist.gradle' line: 51
         * What went wrong:
             License for commons-logging-1.0.4.jar not found
             > java.lang.RuntimeException (no error message)

I give up. Will wait for the final kits, in my opinion the build scripts still need some work in order to provide a consistent build experience.

robander:
@raducoravu I got the same thing when I tried over the weekend, made a new clone, and it went away. I think it came from having JAR files still sitting in the `lib/` dir from earlier "releases" -- appeared the code was looking at JAR files, and then looking for matching licenses, so the older JAR files caused problems. I've had similar issues in the past (some releases ago) where my DITA-OT `dist` build broke because it still had old libraries, so this might be just a new version of a long-standing issue ... our `gradle` build pulls in all the required JAR files for whatever branch you're on, but doesn't clear out old libraries.

If that's right (maybe @jelovirt could confirm) ... then deleting your clone and starting over now should make it work. Longer term, the issue might be that when you get a new branch, it removes the license for obsolete JAR files, but when they still exist in your file system the license code might give the "missing license" failure.

infotexture:
You don't necessarily need to delete your clone or create a new separate clone to revert to a fresh state.

Recent versions of Git (≥2.5) include a dedicated `worktree` command that can be used to check out  a pristine copy of any branch in a separate folder: https://git-scm.com/docs/git-worktree.

From your main clone, run something like this to create a new worktree alongside your clone:

    git worktree add -b test-dist-build ../dita-ot.dist.worktree release/3.2

Then switch to the new worktree (`dita-ot.dist.worktree` in the example above), initialize submodules, and run both Gradle builds:

    git submodule update --init --recursive
    ./gradlew
    ./gradlew dist

When you're done testing, you can safely discard the disposable worktree, passing the `--force` option (required for submodules):

    git worktree remove ../dita-ot.dist.worktree/ --force

I've used this approach recently to test distribution builds, and it produces reliable results.

raducoravu:
@robander Freshly cloned the dita-ot project to a new folder, run "gradlew dist" and it failed, then run "gradlew" and then "gradlew dist" and it worked.

raducoravu:
@jelovirt Stefan our CEO says he likes the new NOTICES file.

Issue #3123, opened 2018-10-22
HTML5 output failing in dev OTs


markgif:
## Expected Behavior
HTML5 output of docsrc/sample content should build successfully

## Actual Behavior
Build fails with error like this:

`C:\Users\Mark\dita-ots\dita-ot-3.1.3+93cf848\plugins\org.dita.base\build_preprocess.xml:62: Failed to run pipeline: [DOTA069F][FATAL] Input file 'file:/C:/Users/Mark/dita-ots/dita-ot-3.1.3+93cf848/docsrc/samples/hierarchy.ditamap' cannot be located or read. Ensure that file was specified properly and that you have permission to access it.`

The same DITA files build successfully with released 3.1.3 OT.

## Possible Solution
Not sure but it seems like the kind of error I get when the right DTDs are not available.

## Steps to Reproduce

1. Use dev OT dita-ot-3.1.3+2be7362 or dita-ot-3.1.3+93cf848.
2. Run HTML output on docsrc/samples/hierarchy.ditamap

## Copy of the error message, log file or stack trace
```
gen-list:
 [gen-list] Using Xerces grammar pool for DTD and schema caching.
 [gen-list] Processing file:/C:/Users/Mark/dita-ots/dita-ot-3.1.3+93cf848/docsrc/samples/hierarchy.ditamap

BUILD FAILED
C:\Users\Mark\dita-ots\dita-ot-3.1.3+93cf848\build.xml:46: The following error occurred while executing this line:
C:\Users\Mark\dita-ots\dita-ot-3.1.3+93cf848\plugins\org.dita.base\build_preprocess.xml:62: Failed to run pipeline: [DOTA069F][FATAL] Input file 'file:/C:/Users/Mark/dita-ots/dita-ot-3.1.3+93cf848/docsrc/samples/hierarchy.ditamap' cannot be located or read. Ensure that file was specified properly and that you have permission to access it.

Total time: 7 seconds

The process finished with exit code: 1

```

## Environment

* DITA-OT version: dita-ot-3.1.3+93cf848
* Operating system and version: Windos 7, macOS
* How did you run DITA-OT? Oxygen 20.1
* Transformation type: HTML5

robander:
I think this is a duplicate of #2414 -- if you remove the `+` sign in the directory name and try again, does the map build properly?

markgif:
[...]

Issue #2475, opened 2016-09-11
Wrong error message in: Error: Failed to run pipeline: [DOTA069F][FATAL] Input file


philiprbrenan:
In the event that the referenced  DTD file is incorrect the error message reports the source file as non existent.  

In:

![1](https://cloud.githubusercontent.com/assets/7548378/18418662/7848b73a-7842-11e6-9a04-77dc49e4edf2.png)

I was able to convert a file successfully.  

In:

![2](https://cloud.githubusercontent.com/assets/7548378/18418663/784c645c-7842-11e6-9042-32183333c278.png)

I changed the DTD file name from /home to /HOME to make it invalid, but the message that comes back is:

Error: Failed to run pipeline: [DOTA069F][FATAL] **Input file 'file:/home/phil/dita/dita-ot/src/main/docsrc/samples/test.xml'** cannot be located or read.

despite:

```
ls -la /home/phil/dita/dita-ot/src/main/docsrc/samples/test.xml
-rw-rw-r-- 1 phil phil 387 Sep 11 17:02 /home/phil/dita/dita-ot/src/main/docsrc/samples/test.xml
```

This is highly misleading and confusing!

The erroneous message arises because in the Java source, a single try statement wraps far too many statements and then attempts to decode errors via an enormous set of catch blocks.  A better solution would be to use a number of try statements each more local to the statements to be executed to obtain a more granular result.

This error is extremely damaging to the credibility of the DITA OT because new users attempting to use it for the first time become highly frustrated with the time they must waste tracking down this error.  Indeed, it will beyond the capability of most new users to resolve this error  expeditiously because the error message is so misleading.

A better error message such as: "DTD file ... not found" would allow even a novice to resolve this error quickly and in doing so gain confidence in their ability to use this toolkit successfully.

I therefore implore you to fix this error promptly in order to preserve the honor and respect that the Open Source Software process so rightly deserves.

eerohele:
Duplicate of #2461.

I do agree that this should be fixed, though -- I've been confused by the same issue myself in the past.

Issue #2461, opened 2016-08-26
Not helpful exception reported when DTD of root map cannot be found


raducoravu:
Let's say I'm publishing a DITA Map which has an unknown specialization (for example you can modify the DOCTYPE to point to some non existing public and system ID).
The publishing breaks with this error message:

       D:\projects\eXml\frameworks\dita\DITA-OT2.x\plugins\org.dita.base\build_preprocess.xml:42: Failed to run pipeline: [DOTA069F][FATAL] Input file 'file:/D:/projects/eXml/samples/dita/flowers/flowers.ditamap' cannot be located or read. Ensure that file was specified properly and that you have permission to access it.

which is not helpful at all. 
The initial FileNotFoundException received in the code is this one:

       Caused by: java.io.FileNotFoundException: D:\projects\eXml\samples\dita\flowers\map.d2td (The system cannot find the file specified)

but when the method "org.dita.dost.module.GenMapAndTopicListModule.processFile(Reference)" catches it, it probably assumes that the exception refers to the published file and ignores its message.

raducoravu:
👍  one more report:

https://www.oxygenxml.com/forum/viewtopic.php?f=20&t=15028&p=44422#p44422

robander:
I've gotten this complaint more often recently as well. Also, probably somewhere in the same series of thrown / caught exceptions, we recently had a broken DTD module where Java reported the broken DTD file but gave no indication which file referenced that DTD (out of many hundreds). Not sure if that's the same section of code or not, but it would be nice if -- when we can't read a file due to missing/broken DTD -- we reported the name of the file.

PR #3124, opened 2018-10-23
New message when grammar file not found #2461

robander:

Implements #2461 (also reported as #2475 and reported in Slack today in relation to #3123) by adding new message to distinguish between "file not found" conditions. Currently, if any file cannot be found, we report the file we're trying to read (usually a DITA map or topic) and state that it cannot be found or loaded. If the DTD/Schema is missing, the `FileNotFoundException` actually applies to that grammar file, but we only report the source file. If the file exists but we receive the `FileNotFound` condition, then most likely it's a grammar file that was not found, so we should modify the message accordingly.

Old behavior:

- If the file cannot be parsed (because the file contents are not valid against the grammar, or because the grammar file itself cannot be parsed), there is a `SaxParseException` that reports the correct file (source or grammar) along with the parsing error
- Otherwise if any file could not be found, catch `FileNotFoundException` and report the _source_ file that could not be used. If it's the input file end with fatal "Input file could not be located or read, ensure it exists and you have permission", otherwise use an Error "File could not be located or read"

New behavior:
- If the file cannot be parsed (because the file contents are not valid against the grammar, or because the grammar file itself cannot be parsed), there is a `SaxParseException` that reports the correct file (source or grammar) along with the parsing error
- If we get the `FileNotFoundException` for the input file, but the input file really does exist, use a new fatal message "Input file XYZ could not be loaded" and point to the grammar file as the response; after that message include the actual exception text that states what file could not be found, such as `map.dtd`
- If we get the `FileNotFoundException` for any other file that really does exist, use a new error "File XYZ cannot be loaded" and point to the grammar file as the cause; after that message include the actual exception text that states what file could not be found, such as `my-topic.dtd`
- Otherwise treat the `FileNotFoundException` as we do today

### DIFF
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -381,6 +381,16 @@ See the accompanying LICENSE file for applicable license.
     <reason>Invalid action attribute '%1' on DITAVAL property.</reason>
     <response></response>
   </message>
+  
+  <message id="DOTJ078F" type="FATAL">
+    <reason>Input file '%1' could not be loaded.</reason>
+    <response>Ensure that grammar files for this document type are referenced and installed properly.</response>
+  </message>
+  
+  <message id="DOTJ079E" type="ERROR">
+    <reason>File '%1' could not be loaded.</reason>
+    <response>Ensure that grammar files for this document type are referenced and installed properly.</response>
+  </message>
 
   <!-- End of Java Messages -->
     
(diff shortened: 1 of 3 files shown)


raducoravu:
👍 We periodically have end users who forget to properly integrate their specialization plugins and encounter this problem.

Issue #2523, opened 2016-11-18
[DOTJ047I] error for keys defined in local scopes


ToshihikoMakita:
DITA-OT 2.4 report this error for keys defined in local scopes. Here is sample bookmap authoring.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE bookmap PUBLIC "-//OASIS//DTD DITA BookMap//EN" "bookmap.dtd">
<bookmap xml:lang="en-US">
    <booktitle>
        <mainbooktitle>Online Help</mainbooktitle>
    </booktitle>
    <frontmatter>
        <mapref href="mKeyDef.ditamap"/>
    </frontmatter>
    <chapter href="topics/cUsingOnlineHelp.dita" navtitle="How to use online-help"/>
    <chapter navtitle="Prepareing to use the software" href="topics/cPreparingToUseSoftware.dita">
        <topicgroup keyscope="SINGLE-PRINTER">
            <keydef keys="REFERENCE-TOPIC" href="topics/tWhenConnectingASinglePrinter.dita"/>
            <keydef keys="CONNECTING_PRINTERS" href="collection-topics/tWhenConnectingPrintersTemplate.dita"/>
            <topicref navtitle="Connecting a single printer" keyref="REFERENCE-TOPIC" format="dita"/>
        </topicgroup>
        <topicgroup keyscope="MULTIPLE-PRINTER">
            <keydef keys="REFERENCE-TOPIC" href="topics/tWhenConnectingMultiplePrinters.dita"/>
            <keydef keys="CONNECTING_PRINTERS" href="collection-topics/tWhenConnectingPrintersTemplate.dita"/>
            <topicref navtitle="Connecting multiple printer"  keyref="REFERENCE-TOPIC" format="dita"/>
        </topicgroup>
    </chapter>
</bookmap>
```

The outputted error message:

```
[DOTJ045I] The key "REFERENCE-TOPIC" is defined more than once in the same map file. The reference href="topics/tWhenConnectingMultiplePrinters.dita" is ignored.
[DOTJ045I] The key "CONNECTING_PRINTERS" is defined more than once in the same map file. The reference href="collection-topics/tWhenConnectingPrintersTemplate.dita" is ignored.
[DOTJ047I] Unable to find key definition for key reference "REFERENCE-TOPIC" in root scope. The href attribute may be used as fallback if it exists
[...]

PR #3141, opened 2018-11-05
Feature/2523 keyscopes in flattened map not handled

simtind:
_Migrated to #3194 to allow commiters to modify code._

Fixes test cases reported in issue #2523.
Two issues were identified:

1. When collecting resolve tasks in KeyRefModule.WalkMap, files referenced through mapref but flattened in map-ref stage are not picked up. They are instead added in collectProcessingTopics. This means they are parsed with root scope instead of their correct scope.
2. When transforming a flattened map key using the KeyRefPaser module it's assumed that one file only has a single key scope. Scopes added outside topicrefs or maprefs are then ignored.

I refactored the elemName deque to keyRefElemNames to make the distinction between the two deques elemNames and keyRefElemNames more clear.


simtind:
No problem, I think this should cover it?

jelovirt:
@simtind This PR was merged in #3194 after some refactoring. Thank you very much for the contribution.

PR #3148, opened 2018-11-12
Fix SemVer bugs

jelovirt:
Add support for pre-release version and build metadata to SemVer and SemVerMatch.


robander:
Code looks good, but had the same question as @infotexture on our development builds. Is the use of `+` in our builds a standard use as in the code test `1.0.0-beta+exp.sha.5114f85` ?

Issue #2495, opened 2016-09-28
Problems when ditavalref points to ditavals outside of main DITA Map folder


raducoravu:
I'm attaching a sample project:

[ditavalref-to-external-ditavals.zip](https://github.com/dita-ot/dita-ot/files/497242/ditavalref-to-external-ditavals.zip)

The DITAVAL files are outside of the folder where the published DITA Map is located. Because of this, in the topic "dita_contextvars\topic.dita" the conkeyref is not expanded anymore.
Also no errors are reported.
If the DITAVAL files are copied next to the DITA Map and links to them are updated, the title of the "dita_contextvars\topic.dita" is correctly computed.

raducoravu:
Reported also more recently on the DITA Users List:
https://groups.yahoo.com/neo/groups/dita-users/conversations/topics/43844

robander:
Added to 3.2.1 because it looked like this would be resolved by #3155, but reading more closely I'm not sure -- need to test to be sure.

robander:
It looks like 3155 only partially fixes this.

The conkeyref attributes are now resolved in the same way, whether the DITAVAL docs are in the same directory or one level up. But they're not filtered in the first branch (each key is resolved as "product-1 product-2" or "handler-1 handler2"). The second branch is filtered. So, there is still an issue here.

jelovirt:
@robander Do you want to tackle this in 3.2.1, or fix in a later release?

robander:
@jelovirt I think this is for later - it's likely an unrelated part of the code so the recent fix won't help much with diagnosing.

PR #3155, opened 2018-11-20
Fix branch filter with uplevels

robander:

Fixes a bug with `<ditavalref>` in `preprocess` when there are references outside of the map directory. The `preprocess2` path currently works and is not affected by this change. The bug only impacts branches that use `dvrResourcePrefix` or `dvrResourceSuffix`; when file names are generated by adding `-1` or `-2` to each copy, processing works correctly.

For example, I have the following map:
```
<topicref href="parent.dita">
 <ditavalref href="advanced.ditaval">
  <ditavalmeta><dvrResourceSuffix name="dvrResourceSuffix">-advanced</dvrResourceSuffix></ditavalmeta>
 </ditavalref>
 <ditavalref href="novice.ditaval">
  <ditavalmeta><dvrResourceSuffix>-novice</dvrResourceSuffix></ditavalmeta>
 </ditavalref>
 <topicref href="kiddo.dita"/>
 <topicref href="../peer/peerkiddo.dita"/>
</topicref>
```

In the temp dir I end up with `temp/main/` and `temp/peer` directories, with the map in the main directory.

When the prefix or suffix are specified, we construct a new `FileInfo` object for the new copy, using the file name. The file info today is correct, we construct full and relative paths for filtered copies of `kiddo.dita` and `../peer/peerkiddo.dita` that are equivalent to the originals. These values are set in `.job.xml` in the root temp directory, and relative values are based on that directory. 

The problem is that the code currently writes exactly the same relative value into the fitlered map, but the map is not in the root temp directory. So, the map in directory `main/` with `href="kiddo.dita"` ends up with `href="main/kiddo-advanced.dita"` and `href="main/kiddo-novice.dita"`. The rest of processing looks for files in both locations, resulting in lots of errors and bad output.

You can reproduce with the sample attached. Our DITA-OT doc uses branch filtering; I was also able to reproduce by placing this map in `doscrc/subdir` and building to HTML5:
```
<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "ibm-map.dtd">
<map xml:lang="en-us"><title>user guide in subdir</title>
<mapref href="../userguide.ditamap"/>
</map>
```

Small sample: 
[uplevels.zip](https://github.com/dita-ot/dita-ot/files/2604284/uplevels.zip)

**About the fix**

Everything that goes into `FileInfo` objects for the new files is correct, and left unmodified. The only change is to the value written out for the new copy inside of the map.

When there is no prefix or suffix, we set up the `FileInfo` object the same, but take the `href` value and add a generated suffix when creating the new name; so, the path on `href` is left exactly the same in the map.

So, to match that, the current processing takes the path from `href` (or `copyTo`), and combines that with the generated file name that includes a prefix/suffix. For `preprocess2` nothing changes because there is never a path. For `preprocess`, this ensures that the new filtered copy of a topic is always created + referenced in the same directory as the original topic.

I still need to add some good tests for this but wanted to get feedback on the code change before creating a new unit or integration test.


raducoravu:
Is this a fix for: https://github.com/dita-ot/dita-ot/issues/2495   ?

robander:
@raducoravu ah, it might be - I was looking for related issues with topics or maps outside of the map directory, but it would likely have the same issue with `ditaval` files outside of the map directory.

src/main/java/org/dita/dost/module/BranchFilterModule.java
```diff
@@ -589,7 +588,7 @@ private void processAttributes(final Element elem, final Branch filter) {
                 final FileInfo dstFileInfo = dstBuilder
                         .build();
 
-                elem.setAttribute(BRANCH_COPY_TO, dstTemp.toString());
+                elem.setAttribute(BRANCH_COPY_TO, dstPathFromMap + FilenameUtils.getName(dstTemp.toString()));
```
jelovirt: Not sure about this path and name tweaking, but if it works. Could you add tests for this? Then it would be easier to rewrite this later on if needed.

robander: :+1: I'll turn the test case I was using into an integration test.

robander: Rather, unit test is preferred. Added that, now cleaning up.

PR #3157, opened 2018-11-22
Move base plugin files to plugin directory

jelovirt:
Move XSLT and other files from the root directory to `org.dita.base` directory.

Add `plugin` URI scheme to allow referencing files in the installation directory. [Eclipse's `plugin` URI scheme](https://help.eclipse.org/neon/topic/org.eclipse.platform.doc.isv/reference/misc/platform-scheme-uri.html?cp=2_1_5_14) has been used as a reference for the design. While `plugin` URIs can be used to address plugin files, the old `plugin` URIs should be used to reference plugin files.


jelovirt:
@robander Now the directory structure is approximately (omitting some files that do not play any role in this change):

```
bin/
  ...
build.xml
config/
  CatalogManager.properties
  configuration.properties
  log4j.properties
  logback.xml
  messages_en_US.properties
  org.dita.dost.platform/
    plugin.properties
  plugins.xml
doc/
  ...
docsrc/
  ...
integrator.xml
lib/
  ...
plugins/
  org.dita.base/
    build_init.xml
    build_preprocess.xml
    build_preprocess2.xml
    catalog-dita.xml
    config/
      messages.xml
    plugin.xml
    xsl/
      ...
  ...
...
```
Are you able to test this with your plugins?

robander:
@jelovirt yes - can test in the next couple of days.

robander:
Tested our plugins. Some require migration unrelated to this PR (PDF still needs minor changes to be up to date with 3.2), but no issues related to file imports. Our styled HTML / plain text output both look good.

Also confirmed that error messages from our plugins are showing up in the `messages.xml` file in the new spot (as expected, but seemed worth checking).

I'm sure there are other things I could look at but the basic test looks good for core DITA-OT transform types, as well as for some of our more complex styled plugins.

src/main/build.xml
```diff
@@ -0,0 +1,23 @@
...
+Copyright 2006 IBM Corporation
+
+See the accompanying LICENSE file for applicable license.
+-->
+<project name="DOST" default="init">
+
+  <!-- Initialize DITA-OT base directory -->
+  <dirname property="ant.file.DOST.dir" file="${ant.file.DOST}" />
+  <property name="dita.dir" location="${ant.file.DOST.dir}"/>
+  <!-- XXX Is this ever used? -->
```
robander: I see `dita.dir` used in tons of plugins and our generated `build.xml` uses it quite a bit ... but I'm guessing this is a question about the second definition? I assume that second definition wouldn't be used...

robander:
Issues on my machine prevented me from building a real copy of the `dist` to test, so I built without docs -- should double check that when the full `dist` build is created, we end up with a valid map from `docscr/userguide.ditamap`. The generated messages topic in particular probably needs a change related to this.

jelovirt:
Moved `messages_template.xml` to root `config` directory. I'm not sure which files should be in the root `config` instead of base plugin's `config` directory. We may need to move them about some more later on.

@infotexture are you able to test this too? I verified that the dist build with docs works and that building the docs with `dita` command from the distribution package works too. Do you have additional ways of building that might be affected by this change?

infotexture:
✅ Successfully tested distribution build from a clean worktree:

```console
git worktree add -b test/base_plugin_dist ../dita-ot.dist.worktree feature/base_plugin

cd ../dita-ot.dist.worktree/

git submodule update --init --recursive
./gradlew
./gradlew dist
```

✅ Verified that docs site builds still work.

PR #3159, opened 2018-11-26
Use absolute URIs in XSLT imports and includes

### LINKED ISSUE #3156: Problems with PDF customizations being taken into account when using Saxon 9.8.0.12
Two of our clients complained that when using the stock DITA OT 3.2 sometimes their XSLT customizations are no longer taken into account, latest report is here:

https://www.oxygenxml.com/forum/topic17977.html

Basically if their "custom.xsl" imports a "front-matter.xsl" placed relative to it, Saxon somehow skips over this import and uses the "front-matter.xsl" specified in the base PDF folder.

I have a plugin with which I can reproduce the problem on my side, with Saxon 9.1 the problem with the XSLT imports precedence cannot be reproduced but unfortunately I'm having trouble putting together a small XML + XSLT package to open an issue on the Saxonica list.
My first idea would be to update to Saxon 9.9 and see if the problem is fixed, unfortunately the DITA OT 3.2 Java code needs to be updated in order to work with Saxon 9.9:

      Error: java.lang.NoSuchMethodError: net.sf.saxon.Controller.setMessageEmitter(Lnet/sf/saxon/event/Receiver;)V

jelovirt:
Use absolute `plugin` URIs in XSLT imports and includes, even in cases where a relative URI inside a plugin could be used.

Possibly fixes #3156

### DIFF
--- a/src/main/plugins/org.dita.eclipsehelp/xsl/map2eclipse_template.xsl
+++ b/src/main/plugins/org.dita.eclipsehelp/xsl/map2eclipse_template.xsl
@@ -11,7 +11,7 @@ See the accompanying LICENSE file for applicable license.
                 xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
 
 <!-- Import the main ditamap to Eclipse TOC Contents conversion -->
-<xsl:import href="map2eclipse/map2eclipseImpl.xsl"/>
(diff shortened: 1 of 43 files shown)


jelovirt:
@robander IMO we don't need to verify if this fixes the Saxon bug. I think this is a change we want, regardless of Saxon.

robander:
@jelovirt agreed -- it's a good change either way.

Issue #2717, opened 2017-06-08
Support table cell entry @rotate attribute


raducoravu:
Based on this [DITA Users List feedback](https://groups.yahoo.com/neo/groups/dita-users/conversations/messages/42339):

>  I have tried text rotation in cells using <entry rotate=”1”> No rotation occurred. Does anyone have any suggestions as to why this might not be working?

As a possible solution in the XSLT stylesheet:

          DITA-OT\plugins\org.dita.pdf2\xsl\fo\tables.xsl

the template with mode "processTableEntry" could be modified as such:

            
                <xsl:template match="*" mode="processTableEntry">
                    <xsl:call-template name="commonattributes"/>
                    <xsl:call-template name="applySpansAttrs"/>
                    <xsl:call-template name="applyAlignAttrs"/>
                    <xsl:call-template name="generateTableEntryBorder"/>
                    <fo:block-container reference-orientation="{if (@rotate eq '1') then 90 else 0}">                
                        <fo:block xsl:use-attribute-sets="tbody.row.entry__content">
                            <xsl:apply-templates select="." mode="ancestor-start-flag"/>
                            <xsl:call-template name="processEntryContent"/>
                            <xsl:apply-templates select="." mode="ancestor-end-flag"/>
                        </fo:block>
                    </fo:block-container>
                </xsl:template>

Only Antenna House seems to have decent support for this from what I tested. Both XEP and Apache FOP do not properly render the table.

jelovirt:
We should _not_ generate `fo:block-container` elements unless they are absolutely needed. They will mess up `@start-indent` and other other similar properties. Thus, if something like this is added, we need to only generate the container if it's rotated.

raducoravu:
👍

ToshihikoMakita:
>  They will mess up @start-indent and other other similar properties.

What the meaning of the phase "mess up"?
Even if `fo:block-container` is located inside the `fo:table-cell`, the `@start-indent` is inherited from `fo:table-caption` to inside the `fo:block-container`. This inheritance should be the feature of XSL-FO specification.

So there should be no influence to generate  redundant `fo:block-container` inside `fo:table-cell`.

kirkilj:
How is this issue related to #1778?

robander:
@kirkilj this appears to be a duplicate of 1778 - I had not noticed the extra copy, or I would have closed one of them a while back. Thanks for helping to clean up the list of issues.

[...]

PR #3161, opened 2018-11-28
Support rotate attribute on table entries #2717

robander:
Implements #2717 -- adds PDF support for `entry/@rotate` (added in DITA 1.3) using the container method suggested by @raducoravu in that issue.

I've tested the result with Antenna House and it works well. FOP gave bad results -- the text was rotated, but the cell kept its original dimensions, so that the rotated text writes over surrounding cells. I've updated the FOP table override to skip adding the container. I'll have the XEP test completed soon; if it works or if the rotation is ignored, I expect to keep the code as-is; if the output is broken as in FOP, we should update the `org.dita.pdf2.xep` plugin to skip rotation here.

As part of this, I've moved where `<row>` based flagging is processed within the entry and put it into the `processEntryContent` named template. This seems like an acceptable spot because any row flag has to be processed as entry content; it also simplifies the code, allowing us to call the `ancestor-start-flag` and `ancestor-end-flag` modes once each rather than 4 times each. Provided we keep this change, any DITA-OT 3.2 migration instructions may want to mention the move in case people have custom processing for `entry` elements.

### DIFF
--- a/src/main/plugins/org.dita.pdf2.xep/xsl/fo/tables_xep.xsl
+++ b/src/main/plugins/org.dita.pdf2.xep/xsl/fo/tables_xep.xsl
@@ -2,7 +2,7 @@
 <!--
 This file is part of the DITA Open Toolkit project.
 
-Copyright 2016 IBM Corporation
+Copyright 2019 IBM Corporation
 
 See the accompanying LICENSE file for applicable license.
 -->
--- a/src/main/plugins/org.dita.pdf2.xep/xsl/fo/tables_xep.xsl
+++ b/src/main/plugins/org.dita.pdf2.xep/xsl/fo/tables_xep.xsl
@@ -13,27 +13,25 @@ See the accompanying LICENSE file for applicable license.
   version="2.0"
   exclude-result-prefixes="xs dita-ot">
   
-  <xsl:template match="*[contains(@class, ' topic/thead ')]/*[contains(@class, ' topic/row ')]/*[contains(@class, ' topic/entry ')]">
-    <xsl:apply-templates select="." mode="validate-entry-position"/>
-    <fo:table-cell xsl:use-attribute-sets="thead.row.entry">
-      <xsl:call-template name="commonattributes"/>
-      <xsl:call-template name="applySpansAttrs"/>
-      <xsl:call-template name="applyAlignAttrs"/>
-      <xsl:call-template name="generateTableEntryBorder"/>
-      <fo:block xsl:use-attribute-sets="thead.row.entry__content">
-        <xsl:call-template name="processEntryContent"/>
-      </fo:block>
-    </fo:table-cell>
+  <!-- By default in XEP, rotated table entries will extend to the end of the page, unless a height is provided for the container.
+       To enable rotation, explicitly set the height (and optionally width) as follows:
+       1) Uncomment the fo:block-container
+       2) Adjust the height and width values to either
+       2a) An appropriate default that is acceptable for all of your rotated cells, or
+       2b) A specific or calculated value based on the cell content --> 
+  <xsl:template match="*[contains(@class, ' topic/thead ')]/*[contains(@class, ' topic/row ')]/*[contains(@class, ' topic/entry ')]" mode="rotateTableEntryContent">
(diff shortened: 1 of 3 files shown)


robander:
@raducoravu When I just try adding an extra container (outer one with `reference-orientation="90"` and inner with `reference-orientation="0"`), I get the same results of text writing over other cells. I also tried adding the extra `<fo:block>` outside of the containers, as in your example:
```
<fo:table-cell text-align="from-table-column()">
 <fo:block>
  <fo:block-container reference-orientation="90">
   <fo:block-container reference-orientation="0">
    <fo:block end-indent="3pt" space-after="3pt" space-after.conditionality="retain" space-before="3pt" space-before.conditionality="retain" start-indent="3pt">testing rotate</fo:block>
   </fo:block-container>
  </fo:block-container>
 </fo:block>
</fo:table-cell>
```

The containers in your example explicitly set the height / width of the box within the table cell -- it appears that's what makes this work, rather than the extra `fo:block-container`. When I add `height="100px" width="150px"` to each container, the cell expands to that size, and the text fits / rotated output looks good. But, I don't think this is something we can automatically calculate or generate without knowing a lot about the table and cell.

robander:
Test with XEP correctly rotated the text and it fit within the cell, but even with a small amount of text, the rotation caused the cell to run from the original starting point all the way to the bottom of the page. So, I've created an override to skip the rotating container there as well. I still need to update the XEP shell to include that override, but for now I'm holding off until 3.2.1 is merged to avoid a merge conflict with https://github.com/dita-ot/dita-ot/pull/3159/files#diff-b2fe3bc975f3eb841843b5bb49084abcR14

kbrown01:
I love spell correct! Progression dimension.

 

Trying again.

raducoravu:
@robander Looked more at our CSS example of how to rotate table cells:

https://www.oxygenxml.com/doc/versions/20.1/ug-editor/topics/dcpp_tables.html

and it says something like:

      The width is required - it will become the height of the cell. If this is not specified, the vertical rotated text will bleed out of the cell. 

    *[outputclass ~= "rotated"] > * {
         .....
          width:150px; 
          ...
       }

so you're right, in order to make this work at least with Apache FOP you need to specify the height of the rotated cell, maybe instead of disabling the rotation support with FOP you could just specify in the XSLTs a default height for the rotated cells and let people override that specified cell height using their own PDF plugins.

robander:
@raducoravu we could do that but it makes me nervous - I don't think we have any other cases where the PDF output is done that way (we pick a default size for your text -- it only looks "right" if you have that amount of text, otherwise you have to customize). The customization route would also mean that an author has to make all their rotated cells the same size, unless they get really clever.

A couple of other ideas might be

- Set a default of 150, but also add support for a custom `@outputclass` token that can be used to set the cell size. This way someone can fix each cell using the correct size, and can also fix the output without requiring a full customization.
- Leave instructions in the XSLT override on how to turn on rotated table entries.

@jelovirt or @infotexture , any thoughts? I really don't like the idea of turning on a feature that might often result in bad output without custom XSL code.

raducoravu:
@robander I like both of your suggestions, actually both could be implemented... maybe even issue a warning in the XSLT that a cell was found which needs to be rotated but it does not have the width set on it.

robander:
Updated based on discussion at yesterday's DITA-OT contributor call.

There was opposition to adding custom tokens into `<entry>` attributes in order to support formatters; instead, I've simplified the code to make it easier to customize output for other formatters to set default or custom height values. This also simplifies the code and reduces what we override for FOP/XEP, because the updated mode only reproduces a 3 lines of code for each instance (minus the dropped container). Code is also now commented to indicate how to turn on custom rotation.

src/main/plugins/org.dita.pdf2.xep/xsl/fo/tables_xep.xsl
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+This file is part of the DITA Open Toolkit project.
+
+Copyright 2016 IBM Corporation
```
jelovirt:
```suggestion
Copyright 2019 IBM Corporation
```

PR #3163, opened 2018-12-01
Use plugin URI scheme in XSLT template imports

jelovirt:
Use `plugin` URI scheme on plugin configured XSLT imports in template files. Old code generated a relative link, but now we want to generate a `plugin` URI where possible.

### DIFF
--- a/src/main/java/org/dita/dost/platform/Integrator.java
+++ b/src/main/java/org/dita/dost/platform/Integrator.java
@@ -100,7 +100,7 @@ public final class Integrator {
 
     /** Plugin table which contains detected plugins. */
     private final Map<String, Features> pluginTable;
-    private final Set<String> templateSet = new HashSet<>(16);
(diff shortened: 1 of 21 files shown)


src/main/java/org/dita/dost/platform/Integrator.java
```diff
@@ -207,7 +208,8 @@ public void execute() throws Exception {
         for (final String tmpl : properties.getProperty(CONF_TEMPLATES, "").split(PARAM_VALUE_SEPARATOR)) {
             final String t = tmpl.trim();
             if (t.length() != 0) {
-                templateSet.add(t);
+                logger.warn(MessageUtils.getMessage("DOTJ080W", "templates", "template").toString());
```
robander: Is it OK to deprecate a common syntax in a hotfix, or are we supposed to leave that to the release boundary? (Seems a bit off with how I think about our rules / SemVer but I'm not certain)

PR #3169, opened 2018-12-10
Configure Probot Stale & No Response

infotexture:
This PR prepares the initial configurations for the Probot [Stale](https://probot.github.io/apps/stale/) & [No Response](https://probot.github.io/apps/no-response/) bots that will close inactive issues after pre-configured thresholds as [discussed at DITA-OT Day](https://www.oxygenxml.com/events/2018/dita-ot_day.html#who_are_you_calling).

- The **Stale** bot is currently configured to mark as `stale` any issue that has not been updated in the past 2 years, and close it after a 2-week grace period that permits users to update the issue to keep it alive.
- The **No Response** bot can be used during the issue triage process to flag incomplete issues with the `need-more-info` label. If no updates are received from the issue author within 30 days, the issue will be automatically closed.

_Review feedback welcome…_

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,16 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 14
+
+# Label requiring a response
+responseRequiredLabel: more-information-needed
+
+# Comment to post when closing an Issue for lack of response. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed because there has been no response
+  to our request for more information from the original author. With only the
+  information that is currently in the issue, we don't have enough information
+  to take action. Please reach out if you have or find the answers we need so
+  that we can investigate further.
+  
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,16 +1,14 @@
 # Configuration for probot-no-response - https://github.com/probot/no-response
 
 # Number of days of inactivity before an Issue is closed for lack of response
-daysUntilClose: 14
+daysUntilClose: 30
 
 # Label requiring a response
-responseRequiredLabel: more-information-needed
+responseRequiredLabel: need-more-info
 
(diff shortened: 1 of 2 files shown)


robander:
If anyone following the issue wants background on why this bot is being used, a recording of the DITA-OT Day discussion Roger mentioned can be found here: https://www.oxygenxml.com/events/2018/dita-ot_day.html#who_are_you_calling

.github/no-response.yml
```diff
@@ -0,0 +1,14 @@
...
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 30
+
+# Label requiring a response
+responseRequiredLabel: need-more-info
+
+# Comment to post when closing an Issue for lack of response. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed because there has been no response
+  to our request for more information from the original author. With only the
```
jelovirt: Could this say "original author or interested parties"?

.github/stale.yml
```diff
@@ -0,0 +1,55 @@
...
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+  - "[Status] Maybe Later"
+  - "priority/high"
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
```
jelovirt: IMO this should be `true`.

robander: Agreed, though it probably shouldn't make a difference in our repo - if I understand properly, adding something to a project would count as activity, so for this to come up with our other settings, the issue would need to be added to a project that's still open 2 years later. Shifting the unfixed issue from a project that's being closed into a project for the next release should count as activity.

.github/no-response.yml
infotexture: Not in this case: the _No Response_ bot is used to require answers from the original author when issue is incomplete, whereas the _Stale_ bot accepts updates from anyone to keep the issue alive.

PR #3179, opened 2019-01-08
Copy topicmeta when keyref used on topicref

### LINKED ISSUE #1439: xref via keyref to external HTML page doesn't work
Hi,

There was a similar issue reported in the past, but the case is a bit different  https://github.com/dita-ot/dita-ot/issues/1174.

DITA-OT version tested (direct from download): 1.6.3, 1.7.1  both have the issue.

In my understanding, either navtitle element or linktext are not honored in the map resolution (Not sure I use the right vocabulary - sorry - I mean when the map is built in the preprocess stage), when a keyref reference an external html page.
The link text is honored in a context of a topic resolution.
## Code

Here is a map using linktext and navtitle

``` xml
<map>
    <title>Keydef on exernal content</title>
    <keydef keys="ncbi" 
                 href="http://www.ncbi.nlm.nih.gov/" 
                 scope="external" 
                 format="html">
        <topicmeta>
            <navtitle>NCBI web site</navtitle>
            <linktext>NCBI web site</linktext>     
        </topicmeta>
    </keydef>
    <topicref href="concept.dita" type="concept"/>
    <topicref keyref="ncbi"/>
</map>
```

After preprocess, the @href value seems to replace the element navtitle and link text

``` xml
<!-- (...) -->
<topicref class="- map/topicref " 
              format="html" 
              href="http://www.ncbi.nlm.nih.gov/"
              keyref="ncbi" 
              scope="external"
              xtrc="topicref:2;12:30" 
              xtrf="/Users/blefort/DITA-OT1.7.1/bugs/DITA-OT/keyref-html/main.ditamap">
        <topicmeta class="- map/topicmeta ">
            <navtitle class="- topic/navtitle ">http://www.ncbi.nlm.nih.gov/</navtitle>
            <?ditaot gentext?>
            <linktext class="- map/linktext ">http://www.ncbi.nlm.nih.gov/</linktext>
        </topicmeta>
    </topicref>
```
## Result

In the navigation, we will see http://www.ncbi.nlm.nih.gov/
In the concept: NCBI web site
## Samples

https://dl.dropbox.com/u/14980268/issues/keyref-html.zip

```
```

robander:
## Description

Today, when I define a key with `<topicmeta>` (including a title, link text, or anything else), and then I refer to that key from a `topicref` (`<topicref keyref="sample"/>`), none of the topicmeta is preserved on the reference -- only attributes are copied. In the Keyref writer, this is because it explicitly checks for `topicref` elements, and does not write out any sub-elements when the element ends. For other elements (such as `link`, `keyword`, etc) it checks for empty content, and then writes out any required text.

This update removes that "if topicref" test that skips all sub-element, processing. Then, if the element is `topicref`, *and* if there is no sub-element, *and* there is a `topicmeta` element with the key definition, it copies that `topicmeta` into the keyref element. This ensures we preserve link text and any other metadata.

There is also an additional tweak to the `domToSax` method. Currently, in all cases, when writing out `<linktext>` from the key definition, the `@class` value is changed from `map/linktext` to `topic/linktext`. This makes sense for all previously handled cases but breaks map processing; when the link text is written out inside of `topicmeta` the original class must be preserved. I've updated the method so that by default it keeps the current behavior (change classes), but when calling from the topicref context, we can turn off that swap and keep the original classes.

This is not a complete fix: it does not add `topicmeta` if the element with `@keyref` has other `topicref` children, AND it does not add or merge metadata if the element with `@keyref` has _different_ metadata. Both of those should be possible, but with the streaming nature of this class I think it would require a *lot* more tests in `startElement` and/or `endElement` to ensure elements are added/merged in the right order. I didn't need any of that so focused here on the simple case.

## Motivation and Context

I have external links defined using keys, for example:
```
<keydef keys="ditaot" href="https://www.dita-ot.org" scope="external" format="html">
  <topicmeta>
    <linktext>DITA Open Toolkit</linktext>
  </topicmeta>
</keydef>
```

I refer to that key in a relationship table:
`<relcell><topicref keyref="ditaot"/></relcell>`

When `@keyref` is resolved today, *only the attributes are preserved* -- so I get the correct link, but everything in `topicmeta` is discarded. As a result, all links added based on this relationship table appear using the link text `https://www.dita-ot.org` rather than the actual desired link text `DITA Open Toolkit`. With this fix, the link text is preserved inside the key reference, and the link is correct.

This also fixes #1439 using the sample provided in that report -- because the `topicmeta` is preserved, the link text is preserved to be used in whatever way needed for the TOC, so we now get consistent TOC behavior for links defined with a URI and for the same links used via keys.

## How Has This Been Tested?

Test set here, based on #1439 and adding my relationship table case:
[1439.zip](https://github.com/dita-ot/dita-ot/files/2738007/1439.zip)

Also resulted in modifications to several existing unit / integration test cases that exhibited this behavior.

## Type of Changes

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -237,16 +237,22 @@ public final class KeyrefPaser extends AbstractXMLFilter {
 
     @Override
     public void endElement(final String uri, final String localName, final String name) throws SAXException {
-        if (keyrefLevel != 0 && empty && !elemName.peek().equals(MAP_TOPICREF.localName)) {
+        if (keyrefLevel != 0 && empty) {
             // If current element is in the scope of key reference element
             // and the element is empty
             if (!validKeyref.isEmpty() && validKeyref.peek()) {
                 final Element elem = keyDef.element;
                 // Key reference is valid,
                 // need to pull matching content from the key definition
-                // If current element name doesn't equal the key reference element
-                // just grab the content from the matching element of key definition
-                if (!name.equals(elemName.peek())) {
+                // If keyref on topicref, and no topicmeta, copy topicmeta from key definition
+                if (elemName.peek().equals(MAP_TOPICREF.localName)) {
+                    final Optional<Element> topicmetaNode = XMLUtils.getChildElement(elem, MAP_TOPICMETA);
+                    if (topicmetaNode.isPresent()) {
+                        domToSax(topicmetaNode.get(), true, false);
+                    }
+                } else if (!name.equals(elemName.peek())) {
+                    // If current element name doesn't equal the key reference element
+                    // just grab the content from the matching element of key definition
                     final NodeList nodeList = elem.getElementsByTagName(name);
                     if (nodeList.getLength() > 0) {
                         final Element node = (Element) nodeList.item(0);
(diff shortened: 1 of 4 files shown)


jelovirt:
Could you rebase this onto `develop` and change the target branch on this PR?

src/main/java/org/dita/dost/writer/KeyrefPaser.java
```diff
@@ -232,16 +232,22 @@ public void characters(final char[] ch, final int start, final int length) throw
...
             if (!validKeyref.isEmpty() && validKeyref.peek()) {
                 final Element elem = keyDef.element;
                 // Key reference is valid,
                 // need to pull matching content from the key definition
-                // If current element name doesn't equal the key reference element
-                // just grab the content from the matching element of key definition
-                if (!name.equals(elemName.peek())) {
+                // If keyref on topicref, and no topicmeta, copy topicmeta from key definition
+                if (elemName.peek().equals(MAP_TOPICREF.localName)) {
+                    final NodeList topicmetaNodes = elem.getElementsByTagName(MAP_TOPICMETA.localName);
```
jelovirt: This should use `XMLUtils.getChildElement(elem, MAP_TOPICMETA)` instead, because then we would not rely on element names.

robander:
Working on rebase to `develop` - I've got the code working, and the test case working, plus suggested switch away from using the topicmeta local name ... but for some reason it's failing with the key scope test case added for 3.3. Trying to get that worked out so this can be merged to develop.

PR #3182, opened 2019-01-12
Add support for SAX filter in pipeline modules

jelovirt:
Allow pipeline modules to be configured to use SAX filters. This allows configuring the module's behavior more at Ant level and add additional processing to a module.


raducoravu:
I do not quite understand how I could benefit from this feature, possibly a small example in the documentation would help.

PR #3187, opened 2019-01-19
Add extension points to contribute params to reader modules

jelovirt:
Add extension points to add parameters and filters to preprocess reader modules:

* `dita.preprocess.debug-filter.param`
* `dita.preprocess.map-reader.param`
* `dita.preprocess.topic-reader.param`

### DIFF
--- a/src/main/plugins/org.dita.base/plugin.xml
+++ b/src/main/plugins/org.dita.base/plugin.xml
@@ -51,6 +51,9 @@ See the accompanying LICENSE file for applicable license.
   <!-- Deprecated since 2.1 -->
   <extension-point id="depend.preprocess.copy-subsidiary.pre" name="Copy subsidiary pre-target"/>
   <extension-point id="depend.preprocess.post" name="Preprocessing post-target"/>
+  <extension-point id="dita.preprocess.debug-filter.param" name="Debug filter module parameters"/>
+  <extension-point id="dita.preprocess.map-reader.param" name="Debug filter module parameters"/>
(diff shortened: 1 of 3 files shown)


src/main/plugins/org.dita.base/plugin.xml
```diff
@@ -51,6 +51,9 @@ See the accompanying LICENSE file for applicable license.
   <!-- Deprecated since 2.1 -->
   <extension-point id="depend.preprocess.copy-subsidiary.pre" name="Copy subsidiary pre-target"/>
   <extension-point id="depend.preprocess.post" name="Preprocessing post-target"/>
+  <extension-point id="dita.preprocess.debug-filter.param" name="Debug filter module parameters"/>
+  <extension-point id="dita.preprocess.map-reader.param" name="Debug filter module parameters"/>
+  <extension-point id="dita.preprocess.topic-reader.param" name="Debug filter module parameters"/>
```
infotexture: @jelovirt Looks like the latter two `@name` values are copy/paste errors.

If so, we could override these values in the docs, but it would be better to fix them at the source.

robander: I saw that when approving this but thought it was intentional (now not sure if I was right) --
- `debug-filter.param` adds params to the Ant `debug-filter` task in `preprocess`, which covers everything
- `map-reader.param` adds params to `map-reader`, which does the initial parse / debug / filter of maps in `preprocess2`
- `topic-reader.param` adds params to `topic-reader`, which does the initial parse / debug / filter of topics in `preprocess2`

Issue #3150, opened 2018-11-13
Update Xerces to 2.12.0


eerohele:
Xerces 2.12.0 was released in May this year. Might be worth updating DITA-OT to use it.

robander:
I wonder if this fixes the grammar caching issue that drops DTD entities (we have to run every build with grammar caching turned off because of the occasional use of entities). The release notes aren't very explicit ... I see one crediting @raducoravu that just says "Fixed various bugs and made various improvements. "

raducoravu:
@robander No, grammar pool issue still opened:
https://issues.apache.org/jira/browse/XERCESJ-1205

robander:
Updated with #3191 .

PR #3191, opened 2019-01-24
Upgrade dependencies

jelovirt:
Update Java dependencies:

* Ant 1.10.5
* Jackson 2.9.8 
* Saxon-HE 9.8.0-14
* Xerces-J2 2.12.0


robander:
Looks like this should resolve #3150

infotexture:
Should we update Gradle while we’re at it?

jelovirt:
@infotexture Let's upgrade Gradle in a separate PR, because it may have dependencies to docs build.

PR #3194, opened 2019-01-26
Feature/2523 keyscopes in flattened map not handled

jelovirt:
_Migrated to new PR from #3141 in order to allow commiters to modify the code._

Fixes test cases reported in issue #2523.
Two issues were identified:

1. When collecting resolve tasks in KeyRefModule.WalkMap, files referenced through mapref but flattened in map-ref stage are not picked up. They are instead added in collectProcessingTopics. This means they are parsed with root scope instead of their correct scope.
2. When transforming a flattened map key using the KeyRefPaser module it's assumed that one file only has a single key scope. Scopes added outside topicrefs or maprefs are then ignored.

I refactored the elemName deque to keyRefElemNames to make the distinction between the two deques elemNames and keyRefElemNames more clear.


jelovirt:
Since I opened this PR as a copy of the original PR, I cannot approve this using the GitHub review. So approving in a comment.

PR #3197, opened 2019-01-28
Add support for truthy values in dita command property values

### LINKED ISSUE #2225: Build parameters should accept consistent true-ish and false-ish values
Currently, different build parameters accept true / false or yes / no values, and might be case-sensitive. This increases the likelihood of user errors when building content.

My recommendation is twofold:
- For consistency in usage and documentation, we standardize the default build parameters to use either true / false or yes / no (not a mishmash).
- For backward compatibility and to help users avoid errors, the build should parse, case-insensitively, "true-ish" values equivalently and likewise with "false-ish" values. That is to say, YES = yes = true = True.

jelovirt:
Add support for `true`, `yes`, `1`, and `on` values for truthy property values in _dita_ command.

Fixes #2225

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -246,16 +246,10 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
                         .map(XMLUtils::getText)
                         .collect(Collectors.toSet());
                 if (vals.size() == 2) {
-                    if (vals.contains("true") && vals.contains("false")) {
-                        return new BooleanArgument(name, "true", "false");
-                    } else if (vals.contains("TRUE") && vals.contains("FALSE")) {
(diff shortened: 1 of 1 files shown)


robander:
I sort of hate to ask, but would this also support the capitalized values `YES` and `NO` used for one of our params? https://www.dita-ot.org/dev/parameters/parameters-base.html#base__args.gen.task.lbl

If not, and you don't want to support those, then it may be time to break compatibility on that single param and change the parameter values to `yes` and `no`. I can't think of other all-caps values, though plugins could always have added them.

infotexture:
Was about to ask the same question, but @robander beat me to it.

_Testing now…_

jelovirt:
Ok, now both upper and lower case values are supported.

jelovirt:
No, we can't change the case of `args.gen.task.lbl` because that would not be backwards compatible.

PR #3207, opened 2019-02-08
Support hazard domain in html5, xhtml

robander:

Goal: produce styled default output for hazard domain that matches the style added for PDF in DITA-OT 3.2.

Still to do:
- Currently assumes a copy of the SVG image in output directory (used for testing), need to update to generate properly
- Need to fix case with `type="other"` (also broken for PDF)
- Once HTML5 is working, add support for XHTML using same (or nearly the same) extension

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
+++ b/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
@@ -22,3 +22,20 @@
 
 .typeofhazard {
 }
+
+.hazardstatement--caution {
+    background-color: #FFD100;
+}
+.hazardstatement--danger {
+    background-color: #C8102E;
+}
+.hazardstatement--warning {
+    background-color: #FF8200;
+}
+.hazardstatement--attention, .hazardstatement--note, .hazardstatement--notice {
+    background-color: #0072CE;
+}
+.hazardstatement--fastpath, .hazardstatement--important, .hazardstatement--other, 
+.hazardstatement--remember, .hazardstatement--restriction, .hazardstatement--tip,   {
+    @extend .hazardstatement--caution;
+}
--- a/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
+++ b/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
@@ -23,19 +23,20 @@
 .typeofhazard {
 }
 
-.hazardstatement--caution {
+.hazardstatement {
+  &--caution {
     background-color: #FFD100;
-}
-.hazardstatement--danger {
+  }
+  &--danger {
     background-color: #C8102E;
-}
-.hazardstatement--warning {
+  }
+  &--warning {
     background-color: #FF8200;
-}
-.hazardstatement--attention, .hazardstatement--note, .hazardstatement--notice {
+  }
+  &--attention, &--note, &--notice {
     background-color: #0072CE;
-}
-.hazardstatement--fastpath, .hazardstatement--important, .hazardstatement--other, 
-.hazardstatement--remember, .hazardstatement--restriction, .hazardstatement--tip,   {
+  }
+  &--fastpath, &--important, &--other, &--remember, &--restriction, &--tip {
     @extend .hazardstatement--caution;
+  }
 }
--- a/src/main/plugins/org.dita.html5/xsl/hazard-d.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/hazard-d.xsl
@@ -1,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<xsl:stylesheet version="2.0"
+     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+     xmlns:ditamsg="http://dita-ot.sourceforge.net/ns/200704/ditamsg"
+     xmlns:dita-ot="http://dita-ot.sourceforge.net/ns/201007/dita-ot"
+     xmlns:xs="http://www.w3.org/2001/XMLSchema"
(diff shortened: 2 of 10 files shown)


robander:
Per discussion at yesterday's DITA-OT contributor call - it's probably best / safest to embed the default SVG warning image directly, rather than using a local image (with problems knowing when to copy). Definitely better than using CSS, which would lose the hazard image for anyone viewing without the CSS.

Still deciding the best way to get the image embedded into the HTML. In PDF we find the image with a variable. The variable is a relative path within the PDF plugin. This works in PDF because that config directory is copied into temp, so the relative path (inside the FO file inside of temp) is correct.

- I don't want to copy the PDF config directory into temp for HTML - it's wrong for a bunch of reasons.
- Should we just copy the SVG image from the PDF plugin into temp, and then read it from there? Also feels wrong because HTML shouldn't depend on PDF.
- Should we copy the image into HTML plugin? Better, but means dual maintenance (more if also copied for XHTML).
- Do we move it from PDF to the base plugin? Possibly better, but now makes PDF processing weird.
- Do we pass a parameter to html / xhtml to let it find this image? 
- The SVG file is small ... do we just put a literal copy of it in the XSL for HTML / XHTML? This is probably easiest to code but trickier to customize. (Keeping in mind that this is only for the default image: you can always specify an alternate image within the hazard statement itself.)

src/main/plugins/org.dita.html5/xsl/hazard-d.xsl
```diff
@@ -0,0 +1,72 @@
...
+  <xsl:param name="default.hazard.image.filename" as="xs:string" select="'ISO_7010_W001.svg'"/>
+  <xsl:variable name="default.hazard.image" as="xs:string">
+    <xsl:sequence select="concat(/processing-instruction()[name()='path2rootmap-uri'], $default.hazard.image.filename)"/>
+  </xsl:variable>
+  
+  <xsl:template match="*[contains(@class,' hazard-d/messagepanel ')]" mode="get-element-ancestry"><xsl:value-of select="name()"/></xsl:template>
+  <xsl:template match="*[contains(@class,' hazard-d/messagepanel ')]/*" mode="get-element-ancestry"><xsl:value-of select="name()"/></xsl:template>
+  <xsl:template match="*[contains(@class,' hazard-d/hazardstatement ')]">
+    <xsl:variable name="type" select="(@type, 'caution')[1]" as="xs:string"/>
+    <xsl:variable name="hazardcolor">
```
jelovirt: I think this should be moved to CSS. Then the hazard statement will not be output with background colour if CSS is not loaded, but I think the image should be enough.

robander: :+1: forgot I'd started out that way, doing this in a plugin where I couldn't depend on the styles being in the default CSS. Definitely better to move.

jelovirt: Maybe generate `class="hazardstatement--{$type}"` and add matching styles to SCSS.

robander:
Attaching my test case; includes a topic with every hazard statement (default image) and a topic with every hazard statement (specified image). 
[hazards.zip](https://github.com/dita-ot/dita-ot/files/2853344/hazards.zip)

src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
```diff
@@ -22,3 +22,20 @@
 
 .typeofhazard {
 }
+
+.hazardstatement--caution {
```
jelovirt: Since this is SCSS, my preference would be use

```scss
.hazardstatement {
  &--caution {
    background-color: #FFD100;
  }
  &--danger {
    background-color: #C8102E;
  }
...
}
```

src/main/plugins/org.dita.html5/xsl/hazard-d.xsl
```diff
@@ -0,0 +1,102 @@
...
+-->
+<xsl:stylesheet version="2.0"
+     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+     xmlns:ditamsg="http://dita-ot.sourceforge.net/ns/200704/ditamsg"
+     xmlns:dita-ot="http://dita-ot.sourceforge.net/ns/201007/dita-ot"
+     xmlns:xs="http://www.w3.org/2001/XMLSchema"
+     exclude-result-prefixes="ditamsg dita-ot xs">
+  
+  <xsl:variable name="inline-hazard-svg" as="element()">
+    <svg class="hazardsymbol" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:cc="http://creativecommons.org/ns#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:svg="http://www.w3.org/2000/svg"
```
jelovirt: Could we stick to some `resources` directory as a separate file and read with `document()`?

robander: :+1: 

Since we're using the same image for both HTML5 and XHTML - would you prefer to create a new `org.dita.base/resources/` directory and stick it there? I don't want HTML5 to depend on XHTML (or vice versa), so I think the other option is just to have a `resources/` directory for each plugin with a copy of the image. Only one copy seems better, but having two copies would keep the HTML5 / XHTML plugins more separate ... I'm probably fine either way.

jelovirt: Currently I consider XHTML a legacy plugin and I don't really care about it. They have a lot of duplicated code, both having the image file is OK by me.

PR #3208, opened 2019-02-09
Added missing [1] predicate to key() lookups where as is element()?

drmacro:
Corrects a failure in 3.2.1 where a reference to a glossary entry that is referenced through an intermediate keyref caused an XSLT runtime failure.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/abbrev-domain.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/abbrev-domain.xsl
@@ -25,7 +25,7 @@ See the accompanying LICENSE file for applicable license.
   
   <xsl:template match="*[contains(@class,' abbrev-d/abbreviated-form ')]" name="topic.abbreviated-form">
     <xsl:variable name="keys" select="@keyref"/>
-    <xsl:variable name="target" select="key('id', substring(@href, 2))"/>
+    <xsl:variable name="target" select="key('id', substring(@href, 2))[1]"/>
     <xsl:choose>
       <xsl:when test="$keys and $target/self::*[contains(@class,' glossentry/glossentry ')]">
         <xsl:call-template name="topic.term">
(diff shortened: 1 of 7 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/abbrev-domain.xsl
```diff
@@ -25,7 +25,7 @@ See the accompanying LICENSE file for applicable license.
   
   <xsl:template match="*[contains(@class,' abbrev-d/abbreviated-form ')]" name="topic.abbreviated-form">
     <xsl:variable name="keys" select="@keyref"/>
-    <xsl:variable name="target" select="key('id', substring(@href, 2))"/>
+    <xsl:variable name="target" select="key('id', substring(@href, 2))[1]"/>
```
jelovirt: Could you add `as="element()?"` while you're at it?

drmacro:
Commit 3 included -s flag so should be good now.

jelovirt:
No, all commits need to have the sign-off. Could you squash the commits to a single commit that has the sign-off?

drmacro:
Does that require creating a new branch or can I just rebase the current branch and force push?

jelovirt:
When you have your `feature/fix-target-variable-pdf2` checked out:

```bash
git fetch origin
git rebase -i origin/developer
```

This will open an editor. There, let the first row be `pick` and the remaining rows change to `squash`. Then save and exit the editor. Then you will get another editor to change the concatenated commit messages. Edit that to contain a single commit message and add `Signed-off-by: Eliot Kimber <ekimber@contrext.com>` on it's own row.

src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
```diff
@@ -492,7 +492,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:param name="refID"/>
 
     <xsl:for-each select="$index-entries">
-      <xsl:variable name="entries" select="key('index-key',$value)" as="element()*"/>
+      <xsl:variable name="entries" select="key('index-key',$value)[1]" as="element()*"/>
```
jelovirt: Is it correct to only take the first entry here? In the next line we filter based on the ID and `value-of` will handle outputting only the first one.

drmacro: You are correct--should not have a [1] there. Missed the element()*

drmacro: That code still looks a little iffy on its face--doesn't look good to potentially get multiple elements and then apply value-of to them but I don't know what this code is really doing or why so will simply remove the [1]

PR #3210, opened 2019-02-09
Add [1] and as= for key() lookups as appropriate

drmacro:
Corrects a failure in 3.2.1 where a reference to a glossary entry that is referenced through an intermediate keyref caused an XSLT runtime failure.

Signed-off-by: Eliot Kimber <ekimber@contrext.com>

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/abbrev-domain.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/abbrev-domain.xsl
@@ -25,7 +25,7 @@ See the accompanying LICENSE file for applicable license.
   
   <xsl:template match="*[contains(@class,' abbrev-d/abbreviated-form ')]" name="topic.abbreviated-form">
     <xsl:variable name="keys" select="@keyref"/>
-    <xsl:variable name="target" select="key('id', substring(@href, 2))"/>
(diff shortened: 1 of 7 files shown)


karsten-klein:
Switching from DITA-OT 3.2.1 to DITA-OT 3.3 my abbreviated-form references do not work anymore. For each abbreviated form I get the message:
[pipeline] file:ae-reference-document/introduction/en/tpc_introduction.dita:6:67: [DOTX060W][WARN]: Key 'xyz' was used in an abbreviated-form element, but the key is not associated with a glossary entry. Abbreviated-form should ONLY be used to reference to a glossary entry.

All keys are correct, the Glossary is rendered and all entries are available, but the references are not rendered.

It seems that the toolkit can no longer resolve the glossary entries

May this be related to the change at hand?

drmacro:
Are your keys resource-only?

Can you provide me with your source DITA? Since I'm currently working on additional glossary processing in the context of the 3.3 OT this is definitely something I can and should look into.

karsten-klein:
I composed a little example and checked it again with 3.3. Reproduces my issue.
[example.zip](https://github.com/dita-ot/dita-ot/files/2948773/example.zip)

./bin/dita -i example/example.ditamap --transtype=pdf

Issue #3214, opened 2019-02-12
No support for revprop on <tm> element in PDF output


easirois:
## Expected Behavior
The content in a topic has an `@rev` with an associated revprop value in a ditaval.  In this case, the background for the content has a particular color associated with it.  So I would expect the PDF to display the appropriate background color for the text in the PDF output.

## Actual Behavior
The appriate color should be displayed for content within `<tm>` when `@rev`  and revprop specified in a ditaval.

## Possible Solution
Add call to commonattributes prior to base apply-template reference.
```
<fo:inline xsl:use-attribute-sets="tm">
			<xsl:call-template name="commonattributes"/>
			<xsl:apply-templates/>
```

## Steps to Reproduce
Content:
`<tm rev="test-new" status="new" tmtype="reg">Test</tm>`

Ditaval:
`<revprop action="flag" val="test-new" backcolor="#A6F28D" changebar="style:solid;width:1pt;offset:5pt;color:#A6F28D"/>`

## Environment

* DITA-OT version: 2.3.1 to develop
* Operating system and version:
  Window 10 x8664
* How did you run DITA-OT?
oXygen transformation scenario 
* Transformation type:
PDF2

easirois:
working on submitting a pull request for this issue.

robander:
Fixed with #3215

PR #3215, opened 2019-02-12
Fix for issue 3214 - fix for revprop support for tm element

easirois:
Just need to add a call to commonattributes in the template and the rest is taken care of by deafult.


robander:
Looks like this fix for #3214 will also fix broken links to `<tm>` elements [about as low a priority issue as I can think of, but I remember hitting that once when I constructed a "link to everything with an ID" test case].

Issue #2787, opened 2017-09-18
Chunk to-content does not preserve (HTML) or correctly reflect (PDF) topicheads within the chunk


drmacro:
## Expected Behavior

Topicheads are reflected in the resulting chunk as though they were title-only topics. This both the only reasonable result (omitting them would be data loss, which is unacceptable) and is consistent with the "to-content generates title-only topics" rule for topicheads.

## Actual Behavior

HTML: The topicheads are reflected in the ToC but not in the resulting chunk.

PDF: The topicheads are reflected in the PDF after all the non-topichead descendants of the chunk root.

## Steps to Reproduce

1. Process the root map https://github.com/dita-community/dita-test-cases/blob/master/topichead-chunking/topichead-chunking-test-03.ditamap to both XHTML and PDF
2. In the HTML observe that the topicheads are not reflected in the resulting HTML file for the chunk.
3. In the PDF observe that the topicheads occur after their subordinate topics in the flow and ToC.

## Copy of the error message, log file or stack trace

No unexpected messages from the chunk stage.

## Environment

* DITA-OT version: 2.5.3
* Operating system and version: macOS

* How did you run DITA-OT?
oXygen
* Transformation type: xhtml, PDF2

robander:
Have been chatting with @drmacro macro about this one and agree the current behavior is wrong ... not sure what the best result would be.

One idea that seems like a good enhancement is that if you specify `@copy-to` on a `<topichead>`, the preprocess should generate a stub topic in all cases, using that `@copy-to` value, and setting `@href` to the equivalent value. This would allow it to flow through the rest of the process, treated as a real heading-only topic, respecting links, _not disappearing during chunking_, etc.

- This seems logical -- the `<topichead>` means "This is a heading", but I'm not sure what `@copy-to` could possibly mean except "generate it as this topic". [Though ... it is likely that `@copy-to` existing there is just an oversight, in which case there _is_ no actual intent for it.]
- Slight worry though would be that later steps have explicit coding for `mapgroup-d/topichead` which complains about or ignores `@href`. This is certainly possible ... though I've always avoided coding for that, and instead look for "`map/topicref` with just a title". Also ... we shouldn't ever worry about checking for `topichead/@href` given that it's not valid in the source.

robander:
[...]

Issue #2843, opened 2017-11-15
NullPointerException thrown when publication contains nested levels of chunking


radu-pisoi:
## Expected Behavior

The XHTML transformation should work without errors when the publication contains nested levels of chunking.

The DITA Map used to reproduce the problem has the next simplified structure:
```
main.ditamap
<mapref href="ditamap_level2.ditamap" format="ditamap" type="map"/>
---
**  ditamap_level2.ditamap
**  <map chunk="to-content">
**    <title>DITA MAp Level 2</title>       
**    <mapref href="ditamap_level3.ditamap" format="ditamap" type="map"/>
**  </map>
---
****    ditamap_level3.ditamap
****	<map chunk="to-content">
****      <title>DITA Map Level 3</title>
****      <topicref href="Topics/test.dita" type="concept"/>       
****    </map>
```

## Actual Behavior
The XHTML transformation fails with ``java.lang.NullPointerException`` in chunk processing.

```
chunk:
    [chunk] Processing file:/D:/test-oxygen/dita-ot-chunk-issue/temp/xhtml/oxygen_dita_temp/main.ditamap
    [chunk] Processing file:/D:/test-oxygen/dita-ot-chunk-issue/temp/xhtml/oxygen_dita_temp/Chunk1025069204.dita
    [chunk] Processing file:/D:/test-oxygen/dita-ot-chunk-issue/temp/xhtml/oxygen_dita_temp/Topics/test.dita
    [chunk] Writing file:/D:/test-oxygen/dita-ot-chunk-issue/temp/xhtml/oxygen_dita_temp/Chunk456609017.dita

BUILD FAILED
D:\Program Files\Oxygen XML Editor 19.1\frameworks\dita\DITA-OT2.x\build.xml:45: The following error occurred while executing this line:
D:\Program Files\Oxygen XML Editor 19.1\frameworks\dita\DITA-OT2.x\plugins\org.dita.base\build_preprocess.xml:323: java.lang.NullPointerException
	at org.dita.dost.writer.ChunkTopicParser.processChunk(ChunkTopicParser.java:239)
	at org.dita.dost.writer.ChunkTopicParser.write(ChunkTopicParser.java:61)
	at org.dita.dost.reader.ChunkMapReader.processCombineChunk(ChunkMapReader.java:542)
	at org.dita.dost.reader.ChunkMapReader.processTopicref(ChunkMapReader.java:347)
[...]

Issue #2730, opened 2017-06-21
The XHTML transformation fails with NullPointerException when the entire map is chunked


balasaalin:
- I believe this to be a bug, not a question about using DITA-OT.
- I read the [CONTRIBUTING][] file.

## Actual Behavior
I have a DITA map with the @chunk='to-content' attribute set on the "map" element. The map contains a "topichead" which is also chunked. When I try to transform the map to XHTML, the transformation fails with the following NullPointerException:

```
BUILD FAILED
D:\projects\trunk\eXml\frameworks\dita\DITA-OT2.x\build.xml:45: The following error occurred while executing this line:
D:\projects\trunk\eXml\frameworks\dita\DITA-OT2.x\plugins\org.dita.base\build_preprocess.xml:314: java.lang.NullPointerException
	at org.dita.dost.writer.ChunkTopicParser.processChunk(ChunkTopicParser.java:238)
	at org.dita.dost.writer.ChunkTopicParser.write(ChunkTopicParser.java:60)
	at org.dita.dost.reader.ChunkMapReader.processCombineChunk(ChunkMapReader.java:536)
	at org.dita.dost.reader.ChunkMapReader.processTopicref(ChunkMapReader.java:341)
	at org.dita.dost.reader.ChunkMapReader.chunkMap(ChunkMapReader.java:227)
	at org.dita.dost.reader.ChunkMapReader.process(ChunkMapReader.java:136)
	at org.dita.dost.writer.AbstractDomFilter.read(AbstractDomFilter.java:55)
	at org.dita.dost.reader.ChunkMapReader.read(ChunkMapReader.java:118)
	at org.dita.dost.module.ChunkModule.execute(ChunkModule.java:80)
	at org.dita.dost.pipeline.PipelineFacade.execute(PipelineFacade.java:70)
	at org.dita.dost.invoker.ExtensibleAntInvoker.execute(ExtensibleAntInvoker.java:222)
	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:293)
	at sun.reflect.GeneratedMethodAccessor6.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:106)
	at org.apache.tools.ant.Task.perform(Task.java:348)
[...]

Issue #2428, opened 2016-07-13
Chunking ignores topicheads


stokecn:
Using DITA-OT 2.3. Similar to [https://github.com/dita-ot/dita-ot/issues/686](https://github.com/dita-ot/dita-ot/issues/686) I think, but that issue is closed since 2013.

When chunk="to-content" is set on an ancestor (in my case) the result chunk does not include generated content for a descendant topichead. Since PDF output does support topichead, this results in a topichead generated out of order.

The chunk attribute in my scenario is desired for another transtype but the map is shared and also used in PDF docs. Both transtypes share this issue but the PDF case is the easiest to demonstrate.

[topichead_behavior.zip](https://github.com/dita-ot/dita-ot/files/362445/topichead_behavior.zip)

drmacro:
In addition, if there are two levels of topichead, both of which specify chunk="to-content select-topic", the child topichead does not result in a chunk and any topicref children of the child topichead are treated as though they had been select-branch chunked into the parent topichead's chunk (even though the parent topichead's chunk is correctly created as a single topic chunk reflecting only the topichead's navigation title).

My test case is here: https://github.com/dita-community/dita-test-cases/tree/master/topichead-chunking

It exercises each possible combination of chunk values in a two-level topichead situation.

I get the same incorrect result with 1.8.5. With 2.3 I get an NPE failure in the chunk processing:

```
/Users/ekimber/dita-ot/dita-ot-2.3/plugins/org.dita.base/build_preprocess.xml:313: java.lang.NullPointerException
	at org.dita.dost.writer.ChunkTopicParser.processChunk(ChunkTopicParser.java:262)
	at org.dita.dost.writer.ChunkTopicParser.write(ChunkTopicParser.java:63)
	at org.dita.dost.reader.ChunkMapReader.processChunk(ChunkMapReader.java:520)
	at org.dita.dost.reader.ChunkMapReader.processTopicref(ChunkMapReader.java:312)
	at org.dita.dost.reader.ChunkMapReader.processChildTopicref(ChunkMapReader.java:497)
	at org.dita.dost.reader.ChunkMapReader.processTopicref(ChunkMapReader.java:336)
	at org.dita.dost.reader.ChunkMapReader.process(ChunkMapReader.java:140)
	at org.dita.dost.writer.AbstractDomFilter.read(AbstractDomFilter.java:48)
	at org.dita.dost.reader.ChunkMapReader.read(ChunkMapReader.java:104)
[...]

PR #3216, opened 2019-02-14
Fixes and improvements to chunk on topic groups

robander:
This fixes several problems with `@chunk` attributes that are specified on topic groups (`<topicgroup>` but also any `topicref` that specifies `@chunk` but does not specify a topic).

- When `@chunk` is specified on a topic group, `ChunkMapReader` will create a stub topic (or "stump" topic) and use that as the basis for any combined chunk.
- If `chunk="to-content"` is specified on any reference, processing is passed to  `ChunkTopicParser`, which reads child topic references and merges them.
- `ChunkTopicParser` assumes that any nested reference that specifies `@chunk` will also refer to a topic. If `@chunk` is specified on a group, it results in an NPE, as reported in #2730 and #2843.
- The first update here fixes the NPE. Before processing is passed over to `ChunkTopicParser`, it checks child topicrefs for the group condition + `@chunk`. If a group is chunked inside of the new chunk, this update will generate the stub ahead of time, avoiding the NPE entirely.
- The second trivial update -- I noticed that stub topics were generated that set the DITAArchVersion to 1.2, but this should be 1.3 at this point.
- The final enhancement is for topicgroups that need stub topics, but do not have a title. Today, for that condition, the name of the stub topic (such as `Chunk1234567`) becomes the actual title of the topic, so we generate output that includes the generated file name as a title. The update checks for stubs generated from a topic group with no title; if there is no title and no shortdesc, the stub is set only to `<dita></dita>`, which is all that we need for merging child content but does not have the side effect of bad titles.
- For the edge case condition of a topic group with no title (but with a shortdesc), this update generates the topic (with shortdesc) but leaves the title empty.

Tested the full build with the sample files in #2843; updated unit tests to include those same samples (as part of the existing topicgroup test); updated all test files to account for the version change and the group title change.


drmacro:
Shouldn't the shortdesc for a topicgroup be ignored just as its title is? It couldn't possibly make sense to generate an empty-title topic in that case.

If the spec doesn't explicitly handle this case then that's an oversight.

src/main/java/org/dita/dost/reader/ChunkMapReader.java
```diff
@@ -535,9 +541,24 @@ private void processCombineChunk(final Element topicref) {
         final ChunkTopicParser chunkParser = new ChunkTopicParser();
         chunkParser.setLogger(logger);
         chunkParser.setJob(job);
+        createChildTopicrefStubs(getChildElements(topicref, MAP_TOPICREF));
         chunkParser.setup(changeTable, conflictTable, topicref, chunkFilenameGenerator);
         chunkParser.write(currentFile);
     }
+    
+    //Before combining topics in a branch, ensure any descendant topicref with @chunk and no @href has a stub
```
jelovirt:
```suggestion
    /** Before combining topics in a branch, ensure any descendant topicref with @chunk and no @href has a stub */
```

robander:
@drmacro probably, if actually using `<topicgroup>`, though as you noted it's something the spec forgot to address. I'd consider that a lower-priority ugly edge case of "why are you marking up your documents this way", and don't feel a strong need to add exceptions for it. 

Today, if you have `<topicgroup>` with a short description inside of a branch that's already chunked, you get the NPE; if you create a new chunk with the same markup, you get both a generated title like "Chunk123456" *and* the shortdesc, which I think is worse than just the shortdesc.

In general though, I was using "topic group" in the more general sense of "grouping element that has no href / no navtitle / but has shortdesc". The test case for this that went into the unit test was basically the following; definitely still an edge case, but one where I think the short description should not be dropped:
```
<topicref chunk="to-content">
  <topicmeta>
    <shortdesc>This is an edge case, but shouldn't crash the build</shortdesc>
  </topicmeta>
  <topicref href="shortdesc-kid.dita"/>
</topicref>
```

robander:
Looking at other open bugs tagged with `preprocess/chunking`, I think this may also resolve #2787 and #2428  (@raducoravu you'd already suggested 2428 might be a duplicate of an earlier one, I'm still not sure if it was but I do think the larger fix here resolves any remaining issues there).

PR #3221, opened 2019-02-16
Support PDF images in PDF2 FOP

jelovirt:
Support using PDF as images in PDF2 FOP plugin via FOP PDF Images plugin.

### DIFF
--- a/src/main/plugins/org.dita.pdf2.fop/build.gradle
+++ b/src/main/plugins/org.dita.pdf2.fop/build.gradle
@@ -13,7 +13,7 @@ dependencies {
     runtimeOnly(group: 'org.apache.xmlgraphics', name: 'fop', version: '2.3') {
         exclude group: 'xalan'
     }
-    runtimeOnly group: 'org.apache.pdfbox', name: 'fontbox', version: '2.0.7'
+    runtimeOnly group: 'org.apache.pdfbox', name: 'fontbox', version: '2.0.13'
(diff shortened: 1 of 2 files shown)


jelovirt:
While the code itself works, now tests fail with some JAR dependency issue with DOM.

src/main/plugins/org.dita.pdf2.fop/build.gradle
```diff
@@ -13,12 +13,16 @@ dependencies {
...
     }
-    runtimeOnly group: 'org.apache.pdfbox', name: 'fontbox', version: '2.0.7'
+    runtimeOnly group: 'org.apache.pdfbox', name: 'fontbox', version: '2.0.13'
     runtimeOnly(group: 'org.apache.xmlgraphics', name: 'batik-all', version: '1.10') {
         exclude group: 'xalan'
     }
     runtimeOnly group: 'xml-apis', name: 'xml-apis-ext', version: '1.3.04'
     runtimeOnly group: 'org.slf4j', name: 'jcl-over-slf4j', version: '1.7.25'
+    runtimeOnly(group: 'org.apache.xmlgraphics', name: 'fop-pdf-images', version: '2.3') {
+        exclude group: 'org.apache.xmlgraphics'
```
jelovirt: This excludes transitive dependency to FOP which in turn brings a wrong version of XML Apis and DOM.

PR #3229, opened 2019-02-20
Fix for spaces in file/dir name #3213

### LINKED ISSUE #3213: empty nav when the topic has a space in its name for nav-toc=partial in HTML5
## Expected Behavior
Every topic accessed in the HTML5 output should display the nav when setting `nav-toc=partial`

## Actual Behavior
Whenever a topic has a space in its name (properly linked from the ditamap using a `%20`), the topic will not have any nav in the output.

## Steps to Reproduce
1. Take any content, put a space in a topic name and change the link from the topicref associated
2. Run the html5 output with `nav-toc=partial`
3. Open the topic from the index.html

## Environment

* DITA-OT version: 3.2.1
* Operating system and version: Windows 10
* How did you run DITA-OT? command line
* Transformation type: HTML5

robander:

Fixes #3213 

When a TOC is added into the `<nav>` section of HTML5 output, it is done by matching the file name (passed into XSLT as `$FILENAME`) against `@href` values in the map. When the operating system passes in `$FILENAME` with backslashes, we already normalize those to `/` to match expected entries in the map.

When the file name (or directory name) contains spaces, `$FILENAME` or `$FILEDIR` can be passed in with a space value in place of `%20`, but the map uses the URI value `%20`, so we need to do the same normalization here.

I've updated the matching code so that `$current-file` adds an additional normalization for spaces to the existing normalization for `\`. Also moved this into a general function for HTML5 in case more normalization is needed in the future; this also makes it easy to set up an XSPEC test for normalization.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/functions.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/functions.xsl
@@ -172,5 +172,10 @@ See the accompanying LICENSE file for applicable license.
       $table/@keycol and xs:integer($table/@keycol) eq count($entry/preceding-sibling::*) + 1
     "/>
   </xsl:function>
+  
+  <xsl:function name="dita-ot:normalize-href" as="xs:string?">
+    <xsl:param name="href" as="xs:string"/>
+    <xsl:value-of select="replace(translate($href, '\', '/'), ' ', '%20')"/>
+  </xsl:function>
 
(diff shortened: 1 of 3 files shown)


plus-:
It's #3213 :)

robander:
@plus- thanks, fixed that in the description.

PR #3231, opened 2019-02-21
Sync hazard statement presentation

infotexture:
The default HTML5 output for hazard statements from #3207 currently differs from the PDF style in several ways:

1. No padding in the table cells, so the text content nearly touches the borders
2. Poor contrast with black text in the blue & red header rows 
   _(PDF uses white text in these cases)_
3. Header rows are left-aligned, whereas PDF centers them
4. PDF uses all caps for _all_ header rows, whereas HTML only does that for CAUTION & DANGER.
5. PDF italicizes header text for notes & related variants
6. Colors differ for the following types _(blue in PDF, yellow in HTML)_:
   - fastpath
   - important
   - other
   - remember
   - restriction
   - tip

- [x] Sync header variants with PDF attribute sets
- [x] Sync Sass-generated CSS from HTML5 to XHTML
 
## HTML state on `develop` vs. this PR branch

| Before | After |
| ------ | -----
| html-hazards-develop | html-hazards-new |

## PDF

pdf-hazards

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/_variables.scss
+++ b/src/main/plugins/org.dita.html5/sass/_variables.scss
@@ -38,6 +38,9 @@ $hazard-iso-yellow:     #f9a800;
 $hazard-iso-green:      #237f52;
 $hazard-iso-blue:       #005387;
 
+// Text color for hazard statements with dark backgrounds: danger (red), note (blue)
+$hazard-heading-light:  #fff;
+
 //== Scaffolding
 //
 //## Settings for some of the most global styles.
--- a/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
+++ b/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
@@ -43,6 +43,7 @@
   &--notice {
     background-color: $hazard-ansi-blue;
     color: $hazard-heading-light;
+    font-style: italic;
   }
 }
 
--- a/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
+++ b/src/main/plugins/org.dita.html5/sass/domains/_hazard.scss
@@ -14,6 +14,10 @@
     padding: 0.5rem;
   }
 
+  th {
+    text-align: center;
+  }
+
   &--caution,
   &--fastpath,
   &--important,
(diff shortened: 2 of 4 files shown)


infotexture:
This PR addresses № 1–5 above. Regarding the colors in № 6, I assume the current PDF implementation is closer to the spec(s), but I wasn't certain.

I understand from @robander that #3207 re-uses capitalization code from the PDF implementation, but it seems better to delegate that to Sass/CSS here, to make it easier for users to override.

@jelovirt Any preferences?

robander:
I don't get quite the same default results from `develop` that you do, which makes me realize some of this is probably down to browser defaults -- for example all of the headers in my test from `develop` are centered (using Firefox). Pretty clear though that it's better to have that and similar defaults explicit in the CSS.

robander:
FWIW -- I have little knowledge of the specification here, I was generally just trying to give decent output that matched the PDF. Clearly I overlooked some differences (like the white text on blue background), missed some due to browser defaults (like centered text), and didn't even consider the extra padding around the text. Thanks for the close eye @infotexture when reviewing.

infotexture:
@robander & @jelovirt Thanks for the comments and approval, but the main question (№ 6) remains: 

How should we handle the 6 hazard statement types which currently have different colored headers in PDF & HTML? I don't know which is "more correct", but I think users would at least expect us to treat each type consistently between output formats.

Once that's clear, we would also need to push another commit to sync the modified styles to XHTML output.

robander:
@infotexture I think unless somebody says it's wrong, I'd go with PDF and make them default to blue. Best to go with what we've already shipped there and keep them consistent.

4everJang:
I purchased the ANSI Z535.6 standard years ago as I was working in machine industry. The hazard statement was meant to be compliant to this standard but got a couple of things wrong. ANSI only defines 4 levels of hazard statements, related to the effects of not avoiding these dangers: notice, caution, warning and danger. Any other type for hazard statement should not show a header at all, as they will have negative impact on the actual hazard statements. If a header is shown I would remove all coloring as well as the symbol from them - to take them away from the true hazard statement look and feel. I am attaching an image for the four ANSI level headers taken from the standard.
ansi-headers
The space left of the message panel is reserved for more specific hazard symbols, such as high voltage or sharp edges. In caution, warning or danger statements the standard symbol for that level can be repeated but it can also be a more specific version such as high voltage or sharp edges. In a NOTICE, this should NOT be a yellow triangle but can be a symbol like gas mask or gloves etc. (blue symbols indicating personal protective equipment). This area is something the OT does not need to handle as that is what the hazard symbol is supposed to do.

Issue #2581, opened 2017-01-27
Path to PDF resources referenced through keyref mechanism are not computed correctly in the XHTML output


radu-pisoi:
We have an issue report from a user that the links to PDF resources are broken in the XHTML output. The PDF references are done using the **keyref** mechanism.

The structure of a sample DITA project to reproduce the issue is:
![image](https://cloud.githubusercontent.com/assets/6837947/22362908/0ed08758-e46f-11e6-978d-4790acfe2200.png)

In the DITA map, the key is defined as follows:
`<keydef  keys="pdf-key" href="pdf/flowers.pdf" format="pdf"/>`

The key reference from **topic.dita** is:
`<p>Reference to the <xref keyref="pdf-key">PDF resource</xref>.</p>`
 
In the HTML document, the path to the PDF resource is generated like:
`<a class="xref" href="pdf/flowers.pdf" target="_blank">PDF resource</a>`

but it should be:
`<a class="xref" href="../pdf/flowers.pdf" target="_blank">PDF resource</a>`

A sample DITA project to reproduce the issue can be found in the next attachment: [pdf-keyref-issue.zip](https://github.com/dita-ot/dita-ot/files/734576/pdf-keyref-issue.zip)

robander:
I think this is the same issue that came up in #1951 (originally reported as an issue for peer links with relative paths, but in the comments, it was reproduced with local non-DITA links as well).

jelovirt:
From #2667

> Non-image and non-dita links should be resolved [here](https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/writer/KeyrefPaser.java#L427) against the `keyDef.source` as well.

robander:
This is a duplicate of #2620 -- local paths for non-DITA key targets are not adjusted when resolving the key.

robander:
Fixed with #3234

Issue #2250, opened 2016-03-07
xref with keydef generating wrong link


gabrielmcg:
I created a keydef for an image: 

`<keydef keys="caution_image" 
        href="../media/images/caution.png"
        format="png"/>
`

and then referenced it in content:

`<p><image keyref="caution_image"/></p>`  
 `<p><xref keyref="caution_image">Download</xref></p>
`

The url for the image is created correctly:

`<p class="p"><img class="image" src="../media/images/caution.png" /></p>`

but the url for the xref doesn't have the correct relative url:

`<p class="p"><a class="xref" href="media/images/caution.png">Download</a></p>`

Is this a bug in the toolkit? or is the source wrong? I have added the files below - I tested with 2.2.2
# 

[dita-bug.zip](https://github.com/dita-ot/dita-ot/files/161574/dita-bug.zip)

eerohele:
Sounds very much like what @shaneataylor describes [here](https://github.com/dita-ot/dita-ot/issues/1951#issuecomment-174681131).

drmacro:
Note that having a cross reference to an image doesn't make any sense--an image is not a thing you can usefully navigate to.

eerohele:
>Note that having a cross reference to an image doesn't make any sense--an image is not a thing you can usefully navigate to.

You could have a cross-reference to something like a PDF file, though, and this bug affects that scenario as well.

There's multiple issues here. One is that keydefs that point to non-external non-DITA resources aren't resolved correctly. Tacking something like this on top of #2223 should fix the issue (untested — this is more like a reminder-to-self type of thing):

```diff
diff --git a/src/main/java/org/dita/dost/writer/KeyrefPaser.java b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
index e344070..d47652f 100644
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -402,6 +402,12 @@ public final class KeyrefPaser extends AbstractXMLFilter {
                                     final URI f = currentFile.resolve(targetOutput);
                                     normalProcessingRoleTargets.add(f);
                                 }
+                            } else if (!ATTR_FORMAT_VALUE_DITA.equals(elem.getAttribute(ATTRIBUTE_NAME_FORMAT))
[...]

Issue #1951, opened 2015-06-24
Keyref breaks with peer topic, relative path, topic in sub-directory


robander:
I've got a key definition with `@scope="peer"`, linking to a topic with a relative path. I've also got a topic in a sub-directory.

When the key is referenced in a `<reltable>` and pushes a link into the subdirectory, the path is adjusted properly, link resolves. When the key is referenced directly from that topic, the link is not adjusted, and is broken.

To reproduce, add this to the end of `hierarchy.ditamap`:

```
<keydef keys="peerhtml" href="../peerdir/file.html" navtitle="peer html file in peerdir" scope="peer" format="html"/>
<keydef keys="peerdita" href="../peerdir/file.dita" navtitle="peer DITA file in peerdir" scope="peer" format="dita"/>
<reltable>
  <relrow>
    <relcell>
      <topicref href="tasks/garagetaskoverview.xml" type="concept"/>
    </relcell>
    <relcell>
      <topicref keyref="peerhtml"/>
      <topicref keyref="peerdita"/>
    </relcell>
  </relrow>
</reltable>
```

Then add this reference into `tasks\garagetaskoverview.xml`:

```
<related-links>
  <linklist>
    <title>links in the topic (broken paths)</title>
    <link keyref="peerdita"/>
    <link keyref="peerhtml"/>
  </linklist>
</related-links>
```

drmacro:
What is your expectation for this type of peer reference to a topic? (or to an HTML file for that matter). Given that the spec through 1.3 does not really say what it means to make a peer reference (except now for peer references to maps in 1.3), I'm genuinely curious what the behavior is that you want or expect. 

I'm not sure what I would expect in that case--a reference to a topic with no other context can only be interpreted as being part of the publication represented by the root map that contains the topicref to the topic, but "peer" suggests that the topic is not a first-class citizen of the publication. In that case, what is it? Is the topic rendered in isolation as a standalone publication? Is it treated as a provisional part of the current publication (since peerness tends to imply potential unavailability of the source file).

robander:
If I have a local file reference, such as `href="../otherdir/file.dita"`, and that reference is moved to another directory, the path is adjusted so that it remains valid. This happens whether that reference is moved around / generated based on map-based-linking or (in this case) keyref.

[...]

PR #3234, opened 2019-02-21
Handle relative paths in peer or non-DITA key references

### LINKED ISSUE #2620: Problem when resolving object datakeyref
[datakeyrefProblem.zip](https://github.com/dita-ot/dita-ot/files/821184/datakeyrefProblem.zip)

My DITA Map refers to a topic and has a key which points to a movie:

        <topicref href="topics/movies.dita"/>
        <keydef href="resources/Particles.mp4" keys="shining_dots" format="mp4"/>

My topic is in the folder "topics/movies.dita" and has a reference to a video like this:

     <object datakeyref="shining_dots"/>

When producing HTML output the datakeyref is converted to:

    <object data="resources/Particles.mp4"></object>

but it should actually be:

    <object data="../resources/Particles.mp4"></object>

in order to properly point to the video.

robander:
Fixes #2620, #2581, #2250, and #1951 (recently closed by probot).

Currently, when a key definition uses a relative path, references to that key are evaluated as follows:

- If the key reference is on `<image>`, adjust relative path based on location of the key definition / key references
- Else if it is for a local DITA topic, adjust relative path based on location of the key definition / key references
- Else use the path as-is. If the key is defined with a relative path, but referenced from a different directory, this breaks all of the following cases (as reported in the various issues above):
 - Keys for peer DITA topics like `../peer/some.dita`, when referenced from a directory different from the map, are no longer valid
 - Keys for local non-DITA file, such as `example.pdf`, are also not adjusted based on the directory
 - This has been reported several times for `@datakeyref` on `<object>`, which will break if the key definition refers to a local file but the key reference is in a different directory
 - If images are referenced by link (rather than from `<image>`), the same issue comes up, because the special handling for images only works on `<image>` element.

This fix corrects all of those cases by testing the `@href` value on the key definition:

- If the path is absolute, OR if `scope="external"`, leave it as-is (no change from today)
- Otherwise, it is a relative path and either local or peer, so adjust the path as needed to stay valid in any referencing location

Also updated keyref unit tests to cover all of these conditions, to ensure that we test for keys defined in one directory but referenced from a topic in a different directory.

### DIFF
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -553,8 +553,16 @@ public final class KeyrefPaser extends AbstractXMLFilter {
                                 }
                             } else {
                                 valid = true;
-                                final URI targetOutput = normalizeHrefValue(href, elementId);
-                                XMLUtils.addOrSetAttribute(resAtts, refAttr, targetOutput.toString());
+                                if (href.isAbsolute() || 
+                                        (keyDef.scope != null && keyDef.scope.equals(ATTR_SCOPE_VALUE_EXTERNAL))) {
+                                    final URI targetOutput = normalizeHrefValue(href, elementId);
+                                    XMLUtils.addOrSetAttribute(resAtts, refAttr, targetOutput.toString());
+                                } else { //Adjust path for peer or local references with relative path
+                                    final URI target = keyDef.source.resolve(href);
+                                    final URI relativeTarget = URLUtils.getRelativePath(currentFile, target);
+                                    final URI targetOutput = normalizeHrefValue(relativeTarget, elementId);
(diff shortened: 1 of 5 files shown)


raducoravu:
Fix looks good to me, I tested and it works for HTML5 output. I have a problem with the PDF output, it may not be related to this issue but a folder called "${_dita.map.output.dir}" gets created in the working directory (place where the Java process starts) and it contains the video inside it.

robander:
@raducoravu probably best to open a new issue so that doesn't get lost -- I'm guessing (?) the output today isn't working right for PDF anyway, so if anything that is just not-working-in-a-different-way?

raducoravu:
@robander I added a new issue with the PDF side-effect here: https://github.com/dita-ot/dita-ot/issues/3242

PR #3235, opened 2019-02-21
Change bar on table/fig should also appear in booklist

robander:

This matches a change in 3.2 that added change bar support for revised topics in the TOC. #3100 

This update does the same thing for tables and figures:

- When a revision is specified on a table (or figure), AND
- When that revision uses a change bar in the DITAVAL, AND
- When that table (or figure)appears in a generated table list (or figure list),
- The table (or figure) should be marked with that change bar in the table list (or figure list)


robander:
Sample bookmap for testing, with table/figure list, revised / not revised tables and figures, and ditaval: 
[pr3235.zip](https://github.com/dita-ot/dita-ot/files/2891897/pr3235.zip)

PR #3238, opened 2019-02-22
Bugfix issue 3162

kluba:
plugin files are copied instead of moved into the plugin directory


jelovirt:
Opened an alternative fix #3239 that uses Apache Commons IO utils to move files, instead of implementing copy with custom code.

PR #3239, opened 2019-02-23
Support different DITA-OT installation volume and temp volume

### LINKED ISSUE #3162: Installation of a custom plugin fails on linux with different partitions / file systems
## Expected Behavior
dita --install plugin.zip should install a new plugin

## Actual Behavior
On linux with several partitions / file systems (/tmp and /home are separated) the command above fails with the following exemplary error message:

Error: /tmp/2913888489706122428/plugin/com.scoop-software.plugin

The class PluginInstallTask tries to move a nonempty directory

`Files.move(tempPluginDir.toPath(), pluginDir.toPath());
`

what does not work if the source and the target are located on different partitions

## Possible Solution
I've replaced the Files.move line with the following code:

```
Files.walkFileTree(tempPluginDir.toPath(), new SimpleFileVisitor<Path>() {

	@Override
	public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
		Path targetPath = pluginDir.toPath().resolve(tempPluginDir.toPath().relativize(dir));
		if(!Files.exists(targetPath)) {
			Files.createDirectory(targetPath);
		}
		return FileVisitResult.CONTINUE;
	}

	@Override
	public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
		Files.copy(file, pluginDir.toPath().resolve(tempPluginDir.toPath().relativize(file)));
		return FileVisitResult.CONTINUE;
	}
});
```

which fixes the problem on my linux server (and also works on my Windows 10 PC)

## Steps to Reproduce
A linux server with two different partitions is needed 

1. Create a new folder on the first partition. Create a file in it
2. Adopt the paths in the following class. The second path has to point to a different (second) partition

```
import java.io.File;
import java.io.IOException;
import java.nio.file.Files;

public class MoveFailure {
	private static final String tempDir = "/tmp/plugin";
	private static final String homeDir = "/home/user/plugin";
	
	public static void main(String[] args) throws IOException {
            final File tempPluginDir = new File(tempDir);
            final File pluginDir = new File(homeDir);

            Files.move(tempPluginDir.toPath(), pluginDir.toPath());
	}
}
```

3. Run the class using java. It results in:

```
Exception in thread "main" java.nio.file.DirectoryNotEmptyException: /tmp/plugin
        at sun.nio.fs.UnixCopyFile.move(UnixCopyFile.java:498)
        at sun.nio.fs.UnixFileSystemProvider.move(UnixFileSystemProvider.java:262)
        at java.nio.file.Files.move(Files.java:1395)
        at TestMove.main(TestMove.java:17)
```

```
 java.nio.file.DirectoryNotEmptyException: /tmp/2913888489706122428/plugin/com.scoop-software.plugin
     at sun.nio.fs.UnixCopyFile.move(UnixCopyFile.java:498)
     at sun.nio.fs.UnixFileSystemProvider.move(UnixFileSystemProvider.java:262)
     at java.nio.file.Files.move(Files.java:1395)
     at org.dita.dost.ant.PluginInstallTask.execute(PluginInstallTask.java:133)
     at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
...
```

## Environment

* DITA-OT version: 3.2
* Operating system and version: Linux (SLES 11.4)
* How did you run DITA-OT: `dita` command
* Transformation type: custom (based on PDF)

jelovirt:
If DITA-OT installation directory and temporary directory are not on the same volume, simple Files.move will fail. Use Apache Commons IO to move directories.

Fixes #3162

Closes  #3238

### DIFF
--- a/src/main/java/org/dita/dost/ant/PluginInstallTask.java
+++ b/src/main/java/org/dita/dost/ant/PluginInstallTask.java
@@ -128,7 +128,7 @@ public final class PluginInstallTask extends Task {
                         throw new BuildException(new IllegalStateException(String.format("Plug-in %s already installed: %s", name, pluginDir)));
                     }
                 }
-                Files.move(tempPluginDir.toPath(), pluginDir.toPath());
+                FileUtils.moveDirectory(tempPluginDir, pluginDir);
             }
         } catch (IOException e) {
             throw new BuildException(e.getMessage(), e);


hcw70:
I would propose a higher priority here, since
* The bug affects many people (All people running on *nix like systems with a "normal" file system layout, where /tmp and /usr are on different file systems).
* The effect is that none of those people can install a plugin with the "dita" command.

PR #3243, opened 2019-02-25
Add support for project configuration

jelovirt:
Add support for project files that allow defining project input files, transformation and profiling conditions. Project files replace input properties file or manually setting properties.

* Project file defines deliverables that are a set of input resources, transtype and associated properties, and profiling conditions
* A project file should be able to define multiple deliverables
* Profile files do not take arguments that can be used to replace variables
* Project files should have at least XML syntax, but JSON and/or YAML format are also good candidates
* Reusable configuration structure can be names and can be referenced to reuse configuration
* Project file should be able to reference other project files to reuse configuration structures

## Usage

```bash
# Run all deliverables in a project file
$ dita -p site.xml
$ dita --project=site.xml
# Run deliverable netlify in project file
$ dita --project=site.xml --deliverable=netflify
```

## Example

JSON project file:

```json
{
  "deliverables": [
    {
      "name": "Deliverable name",
      "inputs": {
        "inputs": [
          {"href": "site.ditamap"}
        ]
      },
      "output": "site",
      "profiles": {
        "ditavals": [
          {"href": "site.ditaval"}
        ]
      },
      "publication": {
        "transtype": "html5",
        "params": [
          {
            "name": "args.gen.task.lbl",
            "value": "YES"
          },
          {
            "name": "args.rellinks",
            "value": "noparent"
          },
          {
            "name": "args.hdr",
            "href": "../../resources/header.xml"
          }
        ]
      }
    }
  ]
}
```

XML project file:

```xml
<?xml-model href="project.rnc" type="application/relax-ng-compact-syntax"?>
<project>
  <deliverable name="Deliverable name">
    <inputs>
      <input href="site.ditamap"/>
    </inputs>
    <output>site</output>
    <profile>
      <ditaval href="site.ditaval"/>
    </profile>
      <param name="args.gen.task.lbl" value="YES"/>
      <param name="args.rellinks" value="noparent"/>
      <param name="args.hdr" href="../../resources/header.xml"/>
    </publication>
  </deliverable>
</project>
```

YAML project file:

```yaml
---
deliverables:
- name: Deliverable name
  inputs:
    inputs:
    - href: site.ditamap
  output: site
  profiles:
    ditavals:
    - href: site.ditaval
  publication:
    transtype: html5
    params:
    - name: args.gen.task.lbl
      value: 'YES'
    - name: args.rellinks
      value: noparent
    - name: args.hdr
      href: "../../resources/header.xml"
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,6 +45,7 @@ dependencies {
     compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-core', version: '2.9.8'
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.9.8'
+    compile group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-xml', version: '2.9.8'
     compile group: 'org.relaxng', name: 'jing', version: '20181222'
     runtime group: 'org.apache.ant', name: 'ant-apache-resolver', version:'1.10.5'
     testCompile group: 'nu.validator.htmlparser', name: 'htmlparser', version:'1.4'
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -342,12 +342,6 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
     private boolean readyToRun = false;
 
     /**
-     * Whether or not we should only parse and display the project help
-     * information.
-     */
-    private boolean projectHelp = false;
-
-    /**
      * Whether or not a logfile is being used. This is used to check if the
      * output streams must be closed.
      */
@@ -575,9 +569,6 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
                 handleArgInputHandler(args);
             } else if (isLongForm(arg, "-emacs") || arg.equals("-e")) {
                 emacsMode = true;
-            } else if (isLongForm(arg, "-projecthelp") || arg.equals("-p")) {
-                // set the flag to display the targets and quit
-                projectHelp = true;
             } else if (isLongForm(arg, "-find") || arg.equals("-s")) {
                 searchForFile = true;
                 args.pop();
@@ -1110,9 +1101,7 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
                 System.setOut(new PrintStream(new DemuxOutputStream(project, false)));
                 System.setErr(new PrintStream(new DemuxOutputStream(project, true)));
 
-                if (!projectHelp) {
-                    project.fireBuildStarted();
-                }
+                project.fireBuildStarted();
 
                 // set the thread priorities
                 if (threadPriority != null) {
@@ -1152,12 +1141,6 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
 
                 ProjectHelper.configureProject(project, buildFile);
 
-                if (projectHelp) {
-                    printDescription(project);
-                    printTargets(project, msgOutputLevel > Project.MSG_INFO, msgOutputLevel > Project.MSG_VERBOSE);
-                    return;
-                }
-
                 // make sure that we have a target to execute
                 if (targets.size() == 0) {
                     if (project.getDefaultTarget() != null) {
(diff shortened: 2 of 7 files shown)


jelovirt:
If I have `output` field the deliverable, how should the CLI `--output` relate to it. Should it work so that

* for `{"output": "foo"}` the output is generated into`./foo`
* for `{"output": "foo"}` and `--output=./out` the output is generated into`./out/foo`
* for `{"output": "/foo"}` the output is generated into`/foo`

This way you could specify in your deliverable that the output should be generated into e.g. `site` directory, and when you run the project, you could specify a base directory for relative output directories.

drmacro:
I like having the project output spec be relative to --output. I would definitely want that ability to generate things into different absolute locations without having to modify the project file.

raducoravu:
Looks very nice, a question about:

         final URI base = new File(".").toURI();

shouldn't all relative references in the project.xml be resolved relative to the place where the project.xml or YAML or JSON) is placed?
Shouldn't each "Publication" define its own output folder? The default being the transtype name...
Or have a parent "Publications" defining the output folder and then each Publication could define its own output folder relative to the parent Publications' output folder.

raducoravu:
One more thing, for each Input there should be a Profile, so maybe the Profile object could be moved inside the Input object. Because distinct DITA Maps may have distinct DITAVAL profiles which could be applied on them.
Also shouldn't the Publication object also be inside the "Input"? So that for each input we can define a set of distinct publications?
Or allow for both cases? Allow for each Input to contain inside a Profile and Publication's and also allow some general Profile and Publication objects which would apply automatically on all defined inputs.

infotexture:
```yaml
- name: args.hdr
  href: "../../resources/header.xml"
```

This example suggests that the project file feature may offer welcome relief for various [issues](https://github.com/dita-ot/docs/pull/163#issuecomment-337163323) with the lack of relative path support in certain parameters, such as HTML headers/footers #2815. 

If so, this alone would seem to be a strong argument for use of project files to define truly portable build configurations that can run on multiple platforms and systems, which are not currently possible with the current `.properties` file limitations. 👍 

Would relative paths defined in the project file be relative to the location of the project file as Radu [suggests](https://github.com/dita-ot/dita-ot/pull/3243#issuecomment-467336070) above?

> Project file should be able to reference other project files to reuse configuration structures

If a referenced project file contains overlapping parameter entries with different definitions, which one wins?

jelovirt:
@infotexture all relative paths are relative to the project file path.

I suppose if you define `key=value` param in project file and pass `--key=override`, then the override will take priority.

jelovirt:
> Feedback from @georgebina: Somehow George would want the abstract model of the Project to be better defined. Somehow he considers that for a deliverable multiple publications should be added.

Mutliple deliverables are to be added later.

jelovirt:
> Working just with the "project.rnc" I'm trying to figure out the structure of the project XML configuration file:
> 
> A project should have one or more "deliverable" elements, right now only one is allowed in the RNC.

At this stage I want to only support a single deliverable. Support for multiple deliverables will added in a later PR of this epic.

> The schema needs more documentation.
> We should have a line with "##" annotation before every element=.... definition.
> Oxygen uses the annotation for example when creating an XML document based on the RNC.
> I would also enjoy having annotations for each attribute definition, for example what is the deliverable @name used for? If the @href value on the "input" is relative, what is it relative to? Ideally the schema would contain all the information about how the XML file should be built.

Docs will be added once the design and features have stabilized.

> That "output" element does not seem to have any attribute set on it. It only has a text value, I do not know what value I should set there. There seems to be a single "output" for all the "inputs". Does that mean that all the published content for all DITA Map inputs will be sent to the same output folder?

The reason for having text content in the XML serialization is a limitation on Jackson XML serializer. In the end we want it to be `<output href="dir"/>`

> Should we also allow defining a "temp" element pointing to a specific temporary folder which could be used for the publication?

Not in the first stage, maybe later.

> For each "deliverable" there is a "publication". This would mean that for my "PDF" output I would need one "deliverable" and for the HTML output another "deliverable" although they might have the same "profile" settings, so the "profile" would be duplicated in multiple parts of the project. Maybe in the "project" we could have multiple "profile" elements, each with a name/id, and then refer to them inside the "deliverable" element. Same for the "inputs" element, define it separately, set an ID on it and then refer to it in multiple places in the XML. This would make sense if we consider that most of the times people will manually create and maintain this XML file.

The current design it the opposite. You can reference reusable `profile` element, that way do you don't need duplication. Same will apply to `publication` or `inputs`, but support for that will be added later.

I don't want to make this a PR that is open for a year or longer, like I've done before. Thus I want to implement this in stages.

raducoravu:
One idea of implementing the project configuration idea in the DITA OT would be to dynamically create an ANT build file based on the project configuration, the ANT build file would be using "exec" to run a forked "dita" executable with various parameters. In this way we might be able in the future to run "exec" tasks in parallel. Also if a processing task fails because of an out of memory, it will not bring down the publishing process because there will be separate processes for the others..

raducoravu:
NullPointerException when the project configuration file is missing parameters in the transtype:

    java.lang.NullPointerException
        at org.dita.dost.invoker.Main.lambda$handleProject$4(Main.java:805)
        at java.util.ArrayList.forEach(ArrayList.java:1249)
        at org.dita.dost.invoker.Main.handleProject(Main.java:797)
        at org.dita.dost.invoker.Main.processArgs(Main.java:720)
        at org.dita.dost.invoker.Main.startAnt(Main.java:460)
        at org.apache.tools.ant.launch.Launcher.run(Launcher.java:284)
        at org.apache.tools.ant.launch.Launcher.main(Launcher.java:101)

sample project file:

        <project>
            <deliverable>
                <inputs>
                    <input href="flowers.ditamap"/>
                </inputs>
                <output>outputCustom/html5</output>
                <profile>
                    <ditaval href="flowers.ditaval"/>
                </profile>
                <publication transtype="html5"/>
            </deliverable>
        </project>

PR #3249, opened 2019-03-05
Fix table numbering when draft excluded

robander:

Currently, when tables or figures are included in a `<draft-comment>` or `<required-cleanup>`, and DRAFT output is not active, the PDF has a number of problems:

- Table / figure numbering is based on all tables / figures in the doc. So if tables 1 and 3 are included but table 2 is within `<required-cleanup>` and hidden, the captions still read **Table 1** and **Table 3**
- Links to the tables have the same problem (the number for the reference is calculated using the same XSLT mode)
- If using a bookmap, and a list of tables / list of figures is included, the hidden items still appear in the list, with page number 0 and a broken link. (This is actually how I became aware of the problem, with the FO processor reporting a broken link).

This pull request fixes all of the issues with a number of small updates:

- A new utility function checks if an element is inside of a draft-comment or required-cleanup element that will be excluded (used for all of the following updates)
- Hidden elements are now excluded from the `tableset` and `figureset` variables used to count numbers for the lists
- When creating the lists, tables and figures inside of hidden sections are not counted
- When calculating the number for inline content or cross references, hidden sections are now excluded form the count

Tested with and without draft by adding a minor update to the `taskbook.ditamap` sample -- adding figure/table lists, and adding tables/figures that will be hidden such that with draft there 6 of each, but without draft only the first and fourth appear (numbered 1 and 2). Samples attached.
[taskbook.zip](https://github.com/dita-ot/dita-ot/files/2933526/taskbook.zip)

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
@@ -914,4 +914,20 @@ See the accompanying LICENSE file for applicable license.
       <xsl:sequence select="$topicType"/>
     </xsl:function>
 
+    <xsl:function name="dita-ot:notExcludedByDraftElement" as="xs:boolean">
+      <xsl:param name="ctx" as="element()"/>
+      <xsl:choose>
+        <xsl:when test="$publishRequiredCleanup='yes' or $DRAFT='yes'">
+            <xsl:sequence select="true()"/>
+        </xsl:when>
+        <xsl:when test="$ctx/ancestor::*[contains(@class,' topic/draft-comment ') or 
+                                         contains(@class,' topic/required-cleanup ')]">
+            <xsl:sequence select="false()"/>
+        </xsl:when>
+        <xsl:otherwise>
+            <xsl:sequence select="true()"/>
+        </xsl:otherwise>
+      </xsl:choose>
+    </xsl:function>
+
 </xsl:stylesheet>
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/lot-lof.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/lot-lof.xsl
@@ -21,27 +21,31 @@ See the accompanying LICENSE file for applicable license.
   
   <xsl:variable name="tableset">
     <xsl:for-each select="//*[contains (@class, ' topic/table ')][*[contains(@class, ' topic/title ' )]]">
-      <xsl:copy>
-        <xsl:copy-of select="@*"/>
-        <xsl:if test="not(@id)">
-          <xsl:attribute name="id">
-            <xsl:call-template name="get-id"/>
-          </xsl:attribute>
-        </xsl:if>
-      </xsl:copy>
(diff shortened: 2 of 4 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
```diff
@@ -913,5 +913,14 @@ See the accompanying LICENSE file for applicable license.
       </xsl:variable>
       <xsl:sequence select="$topicType"/>
     </xsl:function>
+    
+    <xsl:function name="dita-ot:insideExcludedDraftElement" as="xs:boolean">
```
jelovirt: Because all uses of `dita-ot:insideExcludedDraftElement` are negated with `not()`, I think for readability this should be inverted so that we don't need to always use `not()` on the result. However, I can't come up with a good name for it.

drmacro: dita-ot:notInsideExcludedDraftElement() ?

src/main/plugins/org.dita.pdf2/xsl/fo/lot-lof.xsl
```diff
@@ -57,7 +61,7 @@ See the accompanying LICENSE file for applicable license.
           <fo:block start-indent="0in">
             <xsl:call-template name="createLOTHeader"/>
             
-            <xsl:apply-templates select="//*[contains (@class, ' topic/table ')][child::*[contains(@class, ' topic/title ' )]]" mode="list.of.tables"/>
+            <xsl:apply-templates select="//*[contains (@class, ' topic/table ')][child::*[contains(@class, ' topic/title ' )]][not(dita-ot:insideExcludedDraftElement(.))]" mode="list.of.tables"/>
```
jelovirt:
```suggestion
            <xsl:apply-templates select="//*[contains (@class, ' topic/table ')]
                                            [child::*[contains(@class, ' topic/title ' )]]
                                            [not(dita-ot:insideExcludedDraftElement(.))]"
                                 mode="list.of.tables"/>
```

src/main/plugins/org.dita.pdf2/xsl/fo/commons.xsl
```diff
@@ -913,5 +913,14 @@ See the accompanying LICENSE file for applicable license.
...
+      <xsl:param name="ctx" as="element()"/>
+      <xsl:choose>
+        <xsl:when test="$publishRequiredCleanup='yes' or $DRAFT='yes'"><xsl:sequence select="false()"/></xsl:when>
+        <xsl:when test="$ctx/ancestor::*[contains(@class,' topic/draft-comment ') or contains(@class,' topic/required-cleanup ')]"><xsl:sequence select="true()"/></xsl:when>
```
jelovirt:
```suggestion
        <xsl:when test="$ctx/ancestor::*[contains(@class,' topic/draft-comment ') or
                                         contains(@class,' topic/required-cleanup ')]">
          <xsl:sequence select="true()"/>
        </xsl:when>
```

PR #3256, opened 2019-03-14
Add support for context of input and profile sets

jelovirt:
Add support for input and profile set context. This abstraction is not strictly required by publishing engine, but is a logical unit for editors and CMSs.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -795,7 +795,8 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
         final org.dita.dost.project.Project project = readProjectFile();
 
         project.deliverables.forEach(deliverable -> {
-            final String input = base.resolve(deliverable.inputs.inputs.get(0).href).toString();
+            final Deliverable.Context context = deliverable.context;
+            final String input = base.resolve(context.inputs.inputs.get(0).href).toString();
             definedProps.put("args.input", input);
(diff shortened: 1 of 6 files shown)


src/main/java/org/dita/dost/invoker/Main.java
```diff
@@ -556,8 +556,6 @@ public Main() {
      * @since Ant 1.6
      */
     private void processArgs(final String[] arguments) {
-        String searchForThis = null;
```
jelovirt: Removed unused features of `Main` in the same PR.

jelovirt:
`inputs` is a wrapper for multiple input files, to be implemented.

`<profile>` element is used in XML serialization, not `<profiles>`

jelovirt:
We can remove `<inputs>` and allow e.g.

```xml
<context>
  <input href="first.topic"/>
  <input href="second.topic"/>
  <profile>
    <ditaval href="eu.ditaval"/>
    <ditaval href="beginner.ditaval"/>
  </profile>
</context>
```

I think the format of the serialization should be deferred to a later PR.

raducoravu:
> `<inputs>` is a wrapper for multiple input files, to be implemented.

I still do not understand what this means. So if I refer two DITA Maps do I get in the output a single PDF containing the merged content or two PDFs, one for each map?

jelovirt:
One output per `<deliverable>`.

raducoravu:
Ok, understood.

PR #3262, opened 2019-03-25
Add support for including project files in project files

jelovirt:
## Description
Add support for import/include functionality to compose project files.

## Motivation and Context
Being able to import/include common project file components into a project files enables project file reuse.


jelovirt:
Yeah, recursive include detection will need to be added. The code can just pass a Set of processed project file URIs, that's the easiest way to stop recursion and log it.

jelovirt:
@raducoravu Added recursion detection.

PR #3264, opened 2019-03-25
Use last-token-of-class att rather than local name to check target el…

drmacro:
…ement

Signed-off-by: Eliot Kimber <ekimber@contrext.com>

---
name: Pull request
about: Correct conref processing to resolve conrefs to specializations of the referencing element

---

## Description

Updates conrefImpl.xsl to use last token of @class attribute of referencing element in place of local name to select target element.

## Motivation and Context

Make behavior correct per DITA spec.

## How Has This Been Tested?

Tested with dita-test-case tests for this-topic references: 

https://github.com/dita-community/dita-test-cases/tree/master/this-topic-fragment-ids

## Type of Changes

- Bug fix 

## Checklist

  I will try to add a unit test.


src/test/java/org/dita/dost/IntegrationTest.java
```diff
@@ -346,6 +346,16 @@ public void testconref_topiconly() throws Throwable {
                 .test();
     }
 
+    @Test
+    public void testconref_to_specialization() throws Throwable {
+        builder().name("conref_to_specialization")
+                .transtype(XHTML)
```
robander: We could save a bit of time on the integration test process if we only run `PREPROCESS` (since this is testing something that's only part of preprocessing) -- or possibly just add this conref test into an existing integration test for conref. I'm not sure if any of the existing tests are a good fit though.

drmacro: I just used an existing conref test as a model. They seemed to be very narrow in their focus.

Limiting it to preprocess makes sense though.

PR #3265, opened 2019-03-25
Add support referencing publications

jelovirt:
Allow referencing publications from deliverables to allow reusable publications.

### DIFF
--- a/src/test/resources/org/dita/dost/project/product.xml
+++ b/src/test/resources/org/dita/dost/project/product.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="../../../../../../main/resources/project.rnc" type="application/relax-ng-compact-syntax"?>
+<project>
+  <include href="common.xml"/>
+  <deliverable name="name">
+    <context name="Site" id="site">
+      <inputs>
+        <input href="site.ditamap"/>
+      </inputs>
+      <profile>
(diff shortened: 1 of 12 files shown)


src/test/resources/org/dita/dost/project/product.xml
```diff
@@ -0,0 +1,17 @@
...
+    <context name="Site" id="site">
+      <inputs>
+        <input href="site.ditamap"/>
+      </inputs>
+      <profile>
+        <ditaval href="site.ditaval"/>
+      </profile>
+    </context>
+    <output>./site</output>
+    <publication idref="common-sitePub2"/>
```
jelovirt: For now, I'm using this syntax. Later on we can decide if we want to use e.g.

```xml
<publication-ref id="common-sitePub2" />
```

Instead and how we want to handle possible overriding of e.g. parameters.

jelovirt:
@raducoravu It's kinda a hack to add colour formatting to error messages and not get duplicated error messages. This could be improved in another PR that handles how to get the fatal error formatted with the red ERROR colour so that it's passed inside BuildException. One solution would be to handle the fatal case differently in verbose and normal logging levels.

src/test/resources/org/dita/dost/project/product.xml
raducoravu: Looks good to me. Do you also plan to add idrefs for "context" elements?

PR #3269, opened 2019-03-29
Fix bugs in project parsing and missing features

jelovirt:
Fix bugs in project parsing.

* Add intermediate builder object
* Resolve URIs against project file URI, except output URI
* Add support for YAML
* Allow omitting output field


jelovirt:
@raducoravu Because one set is for the builder and the other for the actual use. The builders need to work with Jackson, so they are not necessarily compatible with the actual API objects we want.

PR #3276, opened 2019-04-03
Reference individual deliverable

jelovirt:
Allow running project with only a given deliverable, defined by `--deliverable=<id>`. If no deliverable is defined, run all deliverables in the project. If a single deliverable fails, DITA-OT CLI exits and all remaining deliverables will skipped. All log messages will go to the same stream.

Deliverables in a project can be listed with `--deliverables`.

Added namespace `https://www.dita-ot.org/project` to XML syntax.

## Example

When `all.yaml` contains deliverable definitions for all deliverables, run only `project` deliverable:

``` bash
$ dita --project=all.yaml --deliverables
product-a  Product A name
product-b  Product B name
$ dita --project=all.yaml --deliverable=product-a
```

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -797,10 +797,12 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
                             props.put(param.name, value);
                         }
                     });
-                    final String filters = context.profiles.ditavals.stream()
-                            .map(ditaVal -> Paths.get(base.resolve(ditaVal.href)).toString())
-                            .collect(Collectors.joining(File.pathSeparator));
-                    props.put("args.filter", filters);
+                    if (!context.profiles.ditavals.isEmpty()) {
+                        final String filters = context.profiles.ditavals.stream()
+                                .map(ditaVal -> Paths.get(base.resolve(ditaVal.href)).toString())
+                                .collect(Collectors.joining(File.pathSeparator));
+                        props.put("args.filter", filters);
+                    }
 
                     return props;
                 })
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -457,7 +457,6 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
      */
     @Override
     public void startAnt(final String[] args, final Properties additionalUserProperties, final ClassLoader coreLoader) {
-
         try {
             processArgs(args);
         } catch (final BuildException exc) {
@@ -474,6 +473,10 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
             handleLogfile();
             printMessage(exc);
             exit(1);
+            return;
+        }
+
+        if (!readyToRun) {
             return;
         }
 
@@ -958,7 +961,8 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
                 .reduce(Integer::max)
                 .orElse(0);
         for (Map.Entry<String, String> pair : pairs) {
-            System.out.println(Strings.padEnd(pair.getKey(), length, ' ') + "  " + pair.getValue());
+            System.out.println(Strings.padEnd(pair.getKey(), length, ' ')
+                    + (pair.getValue() != null ? ("  " + pair.getValue()) : ""));
         }
     }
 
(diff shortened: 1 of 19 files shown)


raducoravu:
I started to work, defining a project and using it from the command line. The "profile" element should be made optional in the project file. Right now if I have no filter I need to add an empty "profile" element and the transformation fails trying to find a DITAVAL:

    C:\Users\radu_coravu\Documents\dita-ot\src\main\bin>dita -p D:\projects\eXml\samples\dita\project.xml
    [ditaval-merge] [Fatal Error] :1:1: Content is not allowed in prolog.
    [ditaval-merge] Failed to close merged ditaval file: Stream Closed
    Error: Failed to run pipeline: Failed to parse ditaval file: Content is not allowed in prolog.

raducoravu:
Answering my own questions about relative output references:

      <output href="test/out"/>

right now they seem relative to the place where the Java process started, I would probably have expected them to be relative to the project file.

raducoravu:
In the project.rnc all attributes named "idref" need to be defined with type "IDREF", right now they are defined with type ID:

           attribute idref { xsd:ID}

raducoravu:
Also project.rnc needs to be updated to allow multiple deliverables in the same project.

raducoravu:
I get a NullPointerException when printing deliverables:

    dita -p D:\projects\eXml\samples\dita\project.xml -deliverables

on a project like this:

            <project>
                <context name="ctxt" id="ctxtID">
                    <input href="flowers/flowers2.ditamap"/>
                    <profile>
                        <ditaval href="test.ditaval"/>
                    </profile>
                </context>
            
                <deliverable id="didi1">
                    <context idref="ctxtID"/>
                    <output href="test/outHTML5"/>
                        <param name="clean.temp" value="no"/>
                    </publication>
                </deliverable>
                
                <deliverable id="didi2">
                    <context idref="ctxtID"/>
                    <output href="test/outPDF"/>
                    <publication transtype="pdf">
                        <param name="args.draft" value="yes"/>
                    </publication>
                </deliverable>
            </project>

raducoravu:
I am trying to split my project in multiple modules but I cannot do that because the validation against the project.rnc disallows me to have a project containing for example only contexts:

    <project>
    <context name="ctxt" id="ctxtID">
        <input href="flowers/flowers2.ditamap"/>
        <profile>
            <ditaval href="test.ditaval"/>
        </profile>
    </context>
    </project>

Maybe the validation should be done only against the resolved main project.xml (with includes replaced by the included content) and not against each invidivual module.

jelovirt:
Ant has `-keep-going` flag. Maybe we could have `--keep-going`.

jelovirt:
Can't reproduce the issue with a single context per file.

jelovirt:
The null pointers with `--deliverable` were caused by early exit not working correctly. Fixed.

jelovirt:
Fixed RNC content models and removed unused elements

jelovirt:
Fixed project that has no profiling.

jelovirt:
@raducoravu could you try the latest code?

raducoravu:
I tested, most of what I reported works so that's great, a couple of things which still remain:

> The "profile" element should be made optional in the project file.

The project.rnc still compels me to add an empty profile element if I do not want to specify filtering:

    <context name="ctxt" id="ctxtID">
        <input href="flowers.ditamap"/>
        <profile>
            <ditaval/>
        </profile>
    </context>

but the transformation works (it did not previously work). But maybe we could make the entire "profile" element optional in the RNC file.

> relative output references

I still think that if I specify a relative output reference in the project file, it should be relative to that project file. It's still relative to the folder where the process starts.

> In the project.rnc all attributes named "idref" need to be defined with type "IDREF"

I see you made all ID references NCNames probably in order to allow the validation of modular project files. You also relaxed the overall content model of the project file, and I think it's a good thing.

If you want I can accept this pull request and perform the other changes some time later (maybe based in the feedback others will have as well).

jelovirt:
Allowed omitting `profile` and added a test case that only contains the required elements.

jelovirt:
@raducoravu I think this could be merged as is and we can fix which path is used as the base directory later based on what people find to be the best alternative and/or least surprising.

drmacro:
I feel pretty strongly that any relative reference should be resolved relative to the base URI of the element on which the URI reference occurs.

One of the big annoyances in DOCX files is that relative references are relative to (I think) the root of the package, not relative to the location of the file making the reference, which means you can't do normal URL resolution when processing them.

Issue #1697, opened 2014-06-24
Publish DITA topic or map in the context of a DITA root Map for resolving key and conkeyrefs


raducoravu:
Some of our users want to preview how a edited DITA topic would look like in XHTML. 
The problem is that the topic might use conkeyrefs so it cannot be published on its own, it somehow would need to be published by specifying also as a context for resolving keys the root map.
The behavior should be similar to having a root map which references only this topic and to publish it.

raducoravu:
:+1: Another Oxygen + DITA OT end user wanted this.

raducoravu:
👍 Another Oxygen user asked for this:

> We are converting all our images to keys and creating maps of keys. We imbed those maps in maps containing our topics. When we run a transformation scenario against those maps, it works perfectly. But when we run a transformation scenario for a single topic, the key is not found and the image does not appear in the output.

xephon2:
+1

jelovirt:
Making this work with old preprocess might be a bit complex, but with map-first processing it should be a bit easier. It might be possible to make this work by creating a wrapper map

```xml
<map>
  <mapref href="<map>" processing-role="resource-only"/>
  <topicref href="<topic>"/>
</map>
```

and then at the end of the preprocessing step remove all files except the topic (and linked images).

raducoravu:
Actually it would be great if this could work also if you publish a submap, you would point a parameter to the master map. Ideally the master map would be used just to extract the key mappings. 
About this remark:

> and then at the end of the preprocessing step remove all files except the topic (and linked images)

why remove them? They should be referenced as resource only anyway...

jelovirt:
@raducoravu I'm not 100% sure the idea that using the wrapper map will work as expected, so that's why I mentioned the idea of removing the other files. But, if we want to publish a submap, then we can't remove other files. Maybe the dynamically created map would work; a similar approach is already used by the list of filters feature, where a filter is created dynamically.

raducoravu:
👍 We had another end user request for publishing a submap in the context of a main DITA Map defining keys for each image

raducoravu:
👍 Another one of our end users:

      I am using keys in my DITA topics. When I apply the transformation scenario for PDF for a DITA map, the keys are being resolved and everything looks fine. When I apply the transformation scenario for PDF for a single DITA topic, the keys are not being resolved and do not show in the PDF.
      Is it possible to use keys and still be able to generate a PDF output for single topics?

[...]

PR #3286, opened 2019-04-20
Add support for multiple input files

jelovirt:
Add support for multiple input files per DITA-OT process.


jelovirt:
@robander `args.input` can be either system path or a URI, relative or absolute. I'd prefer to make `args.inputs` (notice the plural) always be absolute, to make things more simple. If a API in front of the Ant API allows relative paths, then it's up to that layer to resolve those to absolute paths.

However, because Ant doesn't have support for array types, do we want to use properties to be begin with for `args.inputs`? We could make it a resource collection instead, because Ant resources can be either files or e.g. URLs. But, I'm worried it will be too difficult to use and unnecessarily binds use to Ant.

robander:
@jelovirt Thinking about how I'm most likely to use this in our larger environment, I think URIs would be OK as long as it can take the absolute path:
`/test/inputone.ditamap /test/bad%20name.ditamap`

I think over time more and more people interact with the command through editors or build scripts, which should be able to hide or translate the values appropriately. Basically, adding that layer ahead of the build to translate paths in a more friendly way.

I'm also thinking about this (for now) as a more complex usage, where it seems more acceptable to enforce stricter rules about the input - maybe others would see it differently?

jelovirt:
This could resolve #1697

src/main/java/org/dita/dost/invoker/ArgumentParser.java
```diff
@@ -326,7 +327,9 @@ public Arguments processArgs(final String[] arguments) {
                 targets.addElement(arg);
             }
         }
-        if (!inputs.isEmpty()) {
+        if (inputs.size() > 1) {
+            definedProps.put("args.inputs", String.join(" ", inputs));
```
robander: Is this going to cause issues with inputs that have spaces in the file name or path? Or do those get normalized to `%20` before we reach this point?

jelovirt: It's up to the client to pass valid URIs. So if the end user user inputs an invalid URI (a URI with a space) to the client, the client will need to convert it to a valid URI (by escaping the space with a `%20`) when setting `args.inputs` property.

jelovirt:
Dropping from 3.4 release due to lack of interest. Maybe in 4.0

PR #3293, opened 2019-04-26
Fix #2576 (branch filtering errors)

### LINKED ISSUE #2576: Branch filtering error in 2.4.2
The build expects topics in a filtered branch to exist even when they should not be generated based on one of the ditaval files. This worked in 2.3.2.

## ditamap:
```
<map>
  <topicref href="test.dita">
    <ditavalref href="prod-a.ditaval">
      <ditavalmeta><dvrResourceSuffix>-a</dvrResourceSuffix></ditavalmeta>
    </ditavalref>
    <ditavalref href="prod-b.ditaval">
      <ditavalmeta><dvrResourceSuffix>-b</dvrResourceSuffix></ditavalmeta>
    </ditavalref>
    <topicref href="prod-b-info.dita" product="b"/>
  </topicref>
</map>
```
## prod-a.ditaval
```
<val>
    <prop action="exclude" att="product"/>
    <prop action="include" att="product" val="a" />
</val>
```
## prod-b.ditaval
```
<val>
    <prop action="exclude" att="product"/>
    <prop action="include" att="product" val="b" />
</val>
```
## Log errors
```
[topic-fragment] Failed to process XML filter: Failed to transform .../temp/test/prod-b-info-a.dita: .../temp/test/prod-b-info-a.dita (No such file or directory)
[topicpull] Error 
[topicpull]   java.io.FileNotFoundException: .../temp/test/prod-b-info-a.dita
[topicpull]   (No such file or directory)
[topicpull] Failed to transform document: java.io.FileNotFoundException: .../temp/test/prod-b-info-a.dita (No such file or directory)
     [xslt] Error 
     [xslt]   java.io.FileNotFoundException: .../temp/test/prod-b-info-a.dita
     [xslt]   (No such file or directory)
     [xslt] Failed to transform document: java.io.FileNotFoundException: .../temp/test/prod-b-info-a.dita (No such file or directory)
```
The **prod-b-info.dita** topic should be built for the _prod-b_ branch, but not for the _prod-a_ branch, so the build should never look for **prod-b-info-a.dita**.

shaneataylor:
name: Pull request
about: Fix #2576 

---

## Description
Add a new step at the end of the BranchFilterModule to remove DITA targets from further processing if they were added by BranchFilterModule and then filtered by BranchFilterModule. These topics—without this fix—are not built but are still included in the job and thus generate errors in later build steps.

Affected topics include:
* those potential topics that are added but then filtered by branch processing

## Motivation and Context
Fixes #2576 

## How Has This Been Tested?
- Tested on MacOS with builds to both the test ditamap in the bug and a larger (1000+ topics) production sample with multiple instances of branch filtering.
- Tested also with Jenkins integration builds of the production sample.

## Type of Changes

## Checklist

    - Tests have been updated and all pass.

---
## Test map to illustrate problem/resolution
```xml
<map>
    <topicref href="test.dita">
        <ditavalref href="prod-a.ditaval">
            <ditavalmeta>
                <dvrResourceSuffix>-a</dvrResourceSuffix>
            </ditavalmeta>
        </ditavalref>
        <ditavalref href="prod-b.ditaval">
            <ditavalmeta>
                <dvrResourceSuffix>-b</dvrResourceSuffix>
            </ditavalmeta>
        </ditavalref>
        <topicref href="prod-a-info.dita" product="a"/>
        <topicref href="prod-b-info.dita" product="b"/>
    </topicref>
</map>
```
After the fix, only the following topics are available to subsequent processing:
- test.dita (technically it probably shouldn't be available, but this would be a breaking change)
- test-a.dita
- prod-a-info-a.dita
- test-b.dita
- prod-b-info-b.dita

The following are _not_ available:
- prod-a-info-b.dita (added by branch filtering and then filtered)
- prod-b-info-a.dita (added by branch filtering and then filtered)


jelovirt:
The problem with this is that topics are allowed to link to other topics, without links from map. If we remove all topics that are not linked to from the map, we may break some content. So if this feature is implemented, it needs to be opt-in.

shaneataylor:
@jelovirt That realization led me to change my initial approach. While I started this fix intending to remove processing for all topics that weren't referenced by the map, I realized that approach was too broad. 

I ended up removing _only those candidate topics that were created and then filtered_ by the BranchFilterModule. These topics _already_ don't exist in the built output; the problem was that processing after this step still looked for them—and created many build errors—because they were still listed in the job.

infotexture:
@shaneataylor This PR was inadvertently auto-closed when the `hotfix/3.3.2` branch was removed. 

It looks like it can't be reopened without the original base branch, so at this point it may be best to rebase your `2576` branch onto the `hotfix/3.3.3` branch and submit a new PR.

_Sorry for the inconvenience…_ 😰

PR #3301, opened 2019-05-11
Add property to skip HTML5 TOC file generation

jelovirt:
Add `html5.toc.skip` property to disable separate TOC file generation in HTML5.

## Usage

Disable `index.html` generation.

```bash
$ dita -i root.ditamap -f html5 --html5.toc.skip=yes
```

### DIFF
--- a/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
+++ b/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
@@ -116,7 +116,7 @@ See the accompanying LICENSE file for applicable license.
       </filterchain>
     </loadfile>
     <makeurl property="html5.map.url" file="${dita.temp.dir}/${html5.map}"/>
-    <makeurl file="${dita.input.valfile}" property="dita.input.valfile.url" validate="no"/>    
+    <makeurl file="${dita.input.valfile}" property="dita.input.valfile.url" validate="no"/>
   </target>
 
   <target name="html5.image-metadata"
@@ -146,12 +146,19 @@ See the accompanying LICENSE file for applicable license.
           depends="html5.map.init,
                    html5.map.toc"/>
 
-  <target name="html5.map.init" unless="noMap">
+  <target name="html5.map.init">
     <property name="args.html5.toc.xsl" value="${dita.plugin.org.dita.html5.dir}/xsl/map2html5-cover.xsl"/>
     <property name="args.html5.toc" value="index"/>
+
+    <condition property="_html5.toc.skip" value="true">
+      <or>
+        <isset property="noMap"/>
+        <istrue value="${html5.toc.skip}"/>
+      </or>
+    </condition>
   </target>
 
-  <target name="html5.map.toc" unless="noMap" description="Build HTML5 TOC file">
+  <target name="html5.map.toc" unless="_html5.toc.skip" description="Build HTML5 TOC file">
     <html5.map>
       <dita:extension id="dita.conductor.html5.toc.param" behavior="org.dita.dost.platform.InsertAction"/>
     </html5.map>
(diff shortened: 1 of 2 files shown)


jelovirt:
The name `html5.toc.skip` is basically negated already. So should we instead use property name `html5.toc = yes | no` to control whether TOC is generated or not.

lief-erickson:
I like `html5.toc = yes | no` better (with `yes` being the default). As you point out, it has logic that is easier to understand. 

In the HTML5 plugin the TOC becomes `index.html`. What happens when `html5.toc=no`? What page becomes `index.html`, if any? The first topic? Is it (re)named `index.html`?

Finally, there is already an [`args.html5.toc`](https://www.dita-ot.org/dev/parameters/parameters-html5.html#html5__args.html5.toc). Will the be used for the same thing? If not, then it could be confusing and we should stay with `html5.toc.skip`.

jelovirt:
Unfortunately we already have [`args.html5.toc`](https://www.dita-ot.org/dev/parameters/parameters-html5.html#html5__args.html5.toc), so we can't really add `html5.toc`.

jelovirt:
> In the HTML5 plugin the TOC becomes `index.html`. What happens when `html5.toc=no`? What page becomes `index.html`, if any? The first topic? Is it (re)named `index.html`?

It disables the generation of the TOC files, so there will be no `index.html`.

infotexture:
An alternative approach would be to _only_ generate an external ToC file if the existing `args.html5.toc` parameter is set to specify a filename, and otherwise skip generation by default.

This would spare us an extra parameter and allow the current option to serve dual purpose, but modifying the behavior of an existing property could yield unexpected results in established workflows.

jelovirt:
I think we should not break backwards compatibility, which means when you set nothing, you will still get `index.html`. Thus using `args.html5.toc` is not possible.

We already have an unofficial convention that `*.file` property name is a file path or file name; just like `*.url` for URIs and `*.dir` for directory path. Thus would not like to use `args.html5.toc.file` for a boolean value.

The old use of `args.` prefix for argument properties is not consistent and just makes property names longer. Thus I would not like to use `args.` prefix.

robander:
Rather than the negative `html.toc.skip = yes | no` I suppose we could go with `html.toc.create = yes | no`, with default of `yes`? I like that a bit better than the `.skip` version. 

In general I try to avoid `yes | no` options because so often I end up needing a third alternative ... but this one is so straightforward that I'm not sure how, and with the existing `args.html5.toc*` parameters I have to assume this one would stay binary.

jelovirt:
I think `html5.toc.generate = yes | no` is the best suggestion thus far, even if it's not perfect. I'll use that for now.

infotexture:
> I think `html5.toc.generate = yes | no` is the best suggestion thus far, even if it's not perfect.

👍 That's probably good enough. We should be sure to clarify in the parameter description that it governs whether a separate ToC file is generated, to distinguish it from the existing [nav-toc](https://www.dita-ot.org/dev/parameters/parameters-html5.html#html5__nav-toc) parameter, i.e.

> Generate a Table of Contents file from the DITA map.

jelovirt:
Changed. Note that this currently only works when the input file is a map. So if you use a topic as an input file and set `--html5.toc.generate=yes`, you will not get an `index.html` file with a single topic reference.

infotexture:
@jelovirt I [assume](https://github.com/dita-ot/dita-ot/pull/3301#pullrequestreview-236390895) that when `html5.toc.generate = no`, any specified values of `args.html5.toc` & `args.html5.toc.class` are silently ignored.

What about `args.html5.contenttarget` and `args.html5.toc.xsl` — are they still relevant in cases where `nav-toc` is set to `full` or `partial`?

PR #3304, opened 2019-05-13
Replace pixelated warning icon with SVG image

infotexture:
## Description

The SVG image added for the PDF implementation of the hazard domain in #3039 for 3.2 is a vector representation of the warning icon used by other note types.

This PR replaces the old jaggy bitmap icon with the vector image used in hazard statements.

## How Has This Been Tested?

Built distribution docs PDF to verify results:

dita-ot-warning-icon-changes

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.pdf2/cfg/common/vars/commonvariables.xml
+++ b/src/main/plugins/org.dita.pdf2/cfg/common/vars/commonvariables.xml
@@ -117,13 +117,13 @@ See the accompanying LICENSE file for applicable license.
   <!-- Image path to use for a note of "notice" type. -->
   <variable id="notice Note Image Path"></variable>
   <!-- Image path to use for a note of "attention" type. -->
-  <variable id="attention Note Image Path">Configuration/OpenTopic/cfg/common/artwork/warning.gif</variable>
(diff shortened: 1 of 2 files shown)


infotexture:
In keeping with the precedent established when the [hand.gif](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2/cfg/common/artwork/hand.gif) image was de-referenced for 2.2, this PR does not yet remove the [warning.gif](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2/cfg/common/artwork/warning.gif) image itself, so it remains available for anyone who needs to restore it.

I think we _should_ eventually remove these obsolete icons and the related [comment](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2/cfg/common/vars/commonvariables.xml#L114), but that sort of change should probably not happen in a hotfix release, but rather in a separate PR that targets the `develop` branch.

infotexture:
3b0fc03 formats the icon based on the approach used for the hazard image, but I wonder if this sort of formatting would be better extracted to an attribute set? _(Probably also not in a hotfix release though.)_

PR #3313, opened 2019-05-29
Fix for Saxon crash when calculating row headers

robander:
---
name: Pull request

---

## Description

Table entries in XHTML and HTML5 currently add `@headers` attributes for accessibility purposes - it explicitly identifies header cells to help screen readers.

This pull request refactors code as part of fixing several problems:

- If an entry in the first column is a header today, and it specifies `@id`, the `@headers` value refers to a generated ID rather than to the explicit ID. This is fixed for XHTML and HTML5.
- Finding the appropriately aligned row header is horribly inefficient. It uses recursion to check every table entry for spanning, to ensure the correct position of the current entry is known, and then uses similar recursion to find header entries that line up. Last week a customer had a table with nearly 1,500 lines, which crashed Saxon due to too many recursive calls as part of this calculation. The old method is now replaced with much simpler, non-recursive tests that make use of `@dita-ot:x` and `@dita-ot:y`, which were added by `preprocess` long ago.
- Finding the appropriately aligned `thead` headers was not as bad because those headers never reach that size, but it has also been refactored to make use of `x`/`y` values.
- To reuse this logic as much as possible (without creating dependencies between XHTML and HTML5), the functions were added to existing HTML5 table functions, and those (table-only) functions were copied to XHTML for use in that code.

If this were for a major release, I'd remove the obsolete functions that used recursion. This is intended as a hotfix for a problem in the current release so I've left them in (marked deprecated) to avoid any compatibility issues.

## Motivation and Context

Fixes the crash that results from inefficient processing, and fixes smaller defects noticed while pursuing that bug.

## How Has This Been Tested?

Topic that includes tables with no headers, just row headers, just `thead` headers, and a 1,500 row table with spanning rows, columns, and headers.
[longtable.zip](https://github.com/dita-ot/dita-ot/files/3234191/longtable.zip)

## Type of Changes

## Checklist

Need to add `xspec` tests for the new table functions in HTML5 but wanted to open this first to get the  started on the pull request.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/functions.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/functions.xsl
@@ -151,7 +151,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:param name="ctx" as="element()"/>
     <xsl:choose>
       <xsl:when test="$ctx/@nameend">
-        <xsl:value-of select="xs:integer(count(table:get-ending-colspec($ctx)/preceding-sibling::*[contains(@class, ' topic/colspec ')])+1)"/>
+        <xsl:value-of select="xs:integer(table:get-ending-colspec($ctx)/@colnum)"/>
       </xsl:when>
       <xsl:otherwise>
         <xsl:value-of select="xs:integer($ctx/@dita-ot:x)"/>
--- a/src/main/plugins/org.dita.html5/xsl/functions.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/functions.xsl
@@ -146,6 +146,66 @@ See the accompanying LICENSE file for applicable license.
        table:get-current-tgroup($el)/@rowsep)[1]
     "/>
   </xsl:function>
+  
+  <xsl:function name="table:find-entry-end-column" as="xs:integer">
+    <xsl:param name="ctx" as="element()"/>
+    <xsl:choose>
+      <xsl:when test="$ctx/@nameend">
+        <xsl:value-of select="xs:integer(count(table:get-ending-colspec($ctx)/preceding-sibling::*[contains(@class, ' topic/colspec ')])+1)"/>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:value-of select="xs:integer($ctx/@dita-ot:x)"/>
+      </xsl:otherwise>
+    </xsl:choose>
+  </xsl:function>
+  
+  <!-- Return true if an entry is entirely within the X or Y span of the header.
+       If entry ends before the header, or starts after the header, no match, otherwise there is overlap. -->
+  <xsl:function name="table:entry-within-range" as="xs:boolean">
+    <xsl:param name="entrystart" as="xs:integer"/>
+    <xsl:param name="entryend" as="xs:integer"/>
+    <xsl:param name="headerstart" as="xs:integer"/>
+    <xsl:param name="headerend" as="xs:integer"/>
+    <xsl:sequence select="if ($entryend lt $headerstart or $entrystart gt $headerend) then (false()) else (true())"/>
+  </xsl:function>
+  
+  <xsl:function name="table:get-matching-thead-headers" as="xs:string*">
+    <xsl:param name="ctx" as="element()"/>
+    <xsl:variable name="startposition" select="xs:integer($ctx/@dita-ot:x)"/>
+    <xsl:variable name="endposition" select="table:find-entry-end-column($ctx)"/>
+    <xsl:for-each select="$ctx/../../../*[contains(@class,' topic/thead ')]/*[contains(@class,' topic/row ')]/*[contains(@class,' topic/entry ')]">
+      <xsl:variable name="headstart" select="xs:integer(@dita-ot:x)"/>
+      <xsl:variable name="headend" select="table:find-entry-end-column(.)"/>
+      <xsl:if test="table:entry-within-range($startposition, $endposition, $headstart, $headend)">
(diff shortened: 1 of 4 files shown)


jelovirt:
I would like to extract OASIS table processing into a preprocess step that normalizes the table and adds processing hints so that transtype specific code would not have to recalculate it.

robander:
@jelovirt I don't think I understand the comment -- I think we do that now (adding `@dita-ot:x` and `@dita-ot:y`, the logic here now is a much simpler set of logic that makes use of those hints. The old logic was extremely complex / ineffecient, and tried to calculate everything recursively without using those hints.

Do you mean the header links should be explicitly added in preprocessing, along with the `x` and `y` info? Right now X/HTML is the only output that makes use of it, but it's possible some other output could find a way to use it. But, that seems a bit like a new feature / enhancement -- all the current pull request does is strip out really complex old code and replace it with simple code that uses the hints.

src/main/plugins/org.dita.html5/xsl/tables.xsl
```diff
@@ -255,6 +255,7 @@ See the accompanying LICENSE file for applicable license.
   
   <!-- Find the end column of a cell. If the cell does not span any columns,
        the end position is the same as the start position. -->
+  <!-- DEPRECATED: use table:find-entry-end-column -->
```
jelovirt: Could you add "since 3.3" so it's easier to track when these can be removed.

src/main/plugins/org.dita.html5/xsl/tables.xsl
```diff
@@ -344,48 +348,19 @@ See the accompanying LICENSE file for applicable license.
...
-    </xsl:variable>
-     <xsl:if test="string-length($rowheader) > 0 or string-length($hdrattr) > 0">
-      <xsl:attribute name="headers" select="concat($rowheader, $hdrattr)"/>
+    <!-- Find the IDs of all headers that are aligned above this cell. May contain duplicates due to spanning cells. -->
+    <xsl:variable name="all-thead-headers" select="table:get-matching-thead-headers(.)" as="xs:string*"/>
+    <!-- Row header should be 0 or 1 today, but future updates may allow multiple -->
+    <xsl:variable name="all-row-headers" select="table:get-matching-row-headers(.)" as="xs:string*"/>
+    <xsl:if test="exists($all-row-headers) or exists($all-thead-headers)">
+       <xsl:attribute name="headers">
+         <xsl:for-each select="distinct-values($all-row-headers)">
```
jelovirt: This could be rewritten as something like

```xml
<xsl:attribute name="headers"
               select="distinct-values($all-row-headers), distinct-values($all-thead-headers)"
               separator=" "/>
```

src/main/plugins/org.dita.xhtml/xsl/xslhtml/tablefunctions.xsl
```diff
@@ -0,0 +1,205 @@
...
+  </xsl:function>
+  
+  <!-- Return true if an entry is entirely within the X or Y span of the header.
+       If entry ends before the header, or starts after the header, no match, otherwise there is overlap. -->
+  <xsl:function name="table:entry-within-range" as="xs:boolean">
+    <xsl:param name="entrystart" as="xs:integer"/>
+    <xsl:param name="entryend" as="xs:integer"/>
+    <xsl:param name="headerstart" as="xs:integer"/>
+    <xsl:param name="headerend" as="xs:integer"/>
+    <xsl:sequence select="if ($entryend lt $headerstart or $entrystart gt $headerend) then (false()) else (true())"/>
```
jelovirt:
```suggestion
    <xsl:sequence select="not($entryend lt $headerstart or $entrystart gt $headerend)"/>
```

src/main/plugins/org.dita.html5/xsl/functions.xsl
```diff
@@ -137,6 +146,66 @@ See the accompanying LICENSE file for applicable license.
...
+  </xsl:function>
+  
+  <!-- Return true if an entry is entirely within the X or Y span of the header.
+       If entry ends before the header, or starts after the header, no match, otherwise there is overlap. -->
+  <xsl:function name="table:entry-within-range" as="xs:boolean">
+    <xsl:param name="entrystart" as="xs:integer"/>
+    <xsl:param name="entryend" as="xs:integer"/>
+    <xsl:param name="headerstart" as="xs:integer"/>
+    <xsl:param name="headerend" as="xs:integer"/>
+    <xsl:sequence select="if ($entryend lt $headerstart or $entrystart gt $headerend) then (false()) else (true())"/>
```
jelovirt:
```suggestion
    <xsl:sequence select="not($entryend lt $headerstart or $entrystart gt $headerend)"/>
```

src/main/plugins/org.dita.html5/xsl/functions.xsl
```diff
@@ -137,6 +146,66 @@ See the accompanying LICENSE file for applicable license.
        table:get-current-tgroup($el)/@rowsep)[1]
     "/>
   </xsl:function>
+  
+  <xsl:function name="table:find-entry-end-column" as="xs:integer">
+    <xsl:param name="ctx" as="element()"/>
+    <xsl:choose>
+      <xsl:when test="$ctx/@nameend">
+        <xsl:value-of select="xs:integer(table:get-ending-colspec($ctx)/@colnum)"/>
```
jelovirt: IMO it's better to use `xsl:sequence` when returning an integer, because `xsl:value-of` will return a string value and the function return type will then convert it back to integer

Issue #3184, opened 2019-01-14
Image flagging in HTML5, XHTML, and TOCJS fails with error message about NIO channels


StanDoherty:
## Expected Behavior
The attached sample files (see ZIP) worked fine for PDF.  I tested HTML, XHTML, and TOCJS expecting them to work as weel. 

## Actual Behavior
My attempts to build the same sources in HTML5, XHTML, or TOCJS failed with a mysterious message about NIO channels.

## Possible Solution
No clue here.

## Steps to Reproduce
1. Detach the attached ZIP containing source files.

2. Run the following command from within that same source folder to build PDF and verify that there's nothing wrong with the sources themselves. 
c:\dita-ot\bin\dita --input=test_002_image-flags.ditamap --format=pdf --filter=test_002_image-flags.ditaval --output=output\pdf

3. Run the following command to attempt to build HTML5.
C:\git\dita-ot-feature-tests>c:\dita-ot\bin\dita --input=test_002_image-flags.ditamap --format=html5 --filter=test_002_image-flags.ditaval --output=output\html5

## Copy of the error message, log file or stack trace
Here is the error message. 

[dita-ot-copy] Attempt to copy images\end_rev.jpg to C:\git\dita-ot-feature-tests\output\html5\file:\C:\git\dita-ot-feature-tests\images\end_rev.jpg using NIO Channels failed due to 'The filename, directory name, or volume label syntax is incorrect'.  Falling back to streams.
Error: The filename, directory name, or volume label syntax is incorrect

## Environment
OS: Windows 10 64-bit
Java: java version "1.8.0_181"
DITA-OT version: DITA-OT version 3.2
Shells: Git CMD and Windows CMD (same error)

[test_002_image-flags.zip](https://github.com/dita-ot/dita-ot/files/2755290/test_002_image-flags.zip)

Issue #3257, opened 2019-03-15
Problems when path to DITAVAL file or path from DITAVAL file to flagging images are outside of the project folder


raducoravu:
It's hard to give a sample project, but it's like this, the DITAVAL contains an absolute reference to a flagging image located somewhere else on the hard drive:

      <startflag imageref="file:/D:/JAD/nwft_cue.gif">
         <alt-text>New feature</alt-text>
      </startflag>

in the .job.xml, this reference ends up like this:

        <file src="file:/D:/JAD/nwft_cue.gif" uri="file:/D:/JAD/nwft_cue.gif"
            path="D:\JAD\nwft_cue.gif" result="file:/D:/JAD/nwft_cue.gif" format="image"
            flag-image="true"/>

because in the .job.xml the @uri attribute is absolute, the dita-ot-copy task fails with the error:

       [dita-ot-copy] Attempt to copy D:\JAD\nwft_cue.gif to C:\Users\radu_coravu\Desktop\julie\R3S\Main\Dev\Help\temp\chm\temp_chm_dir\file:\D:\JAD\nwft_cue.gif using NIO Channels failed due to 'The filename, directory name, or volume label syntax is incorrect'.  Falling back to streams.

raducoravu:
Another way to encounter the same situation is if the ditaval file and the image file are in the same folder which is not in the DITA project folder.

PR #3316, opened 2019-06-01
Fix flag processing for x/html output

robander:
---
name: Pull request

---

## Description

Fixes several problems with DITAVAL based flag processing in HTML / XHTML output:

- Flag images are currently broken when a DITAVAL file is located outside of the map directory. The generated references in HTML use a relative path back to the literal source flag image (outside the scope of the output directory); the flag images are not copied.
- Today, the flag copy process results in build failures under some circumstances (if the image already exists at the target location, which is often the case when a DITAVAL is in the map directory) -- reported in #3184 
- Flag image references are not treated consistently; how we handle relative paths (and where the images are copied) varies based on whether the DITAVAL is outside the map directory, in the map directory, or in a subdirectory of the map, resulting in unstable output

This update makes several changes to fix the issues:

- Relative URIs (based on the input directory) are now created in `.job.xml` for flag images based on DITAVAL outside of the map directory. Previously these were missing; with only the full URI to the source image, it was impossible for processing to know where to put the image in the output.
- The old flag copy routine is almost as old as DITA-OT, and relied on file lists that are mostly obsolete. With multiple DITAVAL input files, those lists also made it impossible to know how to resolve relative paths. The new routine is much simpler and should be more stable; it relies on info in the Job file to copy files.
- The older copy flag is still there because it cannot be removed in a hotfix, but I've marked it deprecated.

## Motivation and Context

Customers with a large number of plugins build them with a DITAVAL located in a common peer directory. When upgrading from 2.5 to 3.x, the flag images all disappeared due to a change in how file paths are calculated. This update resolves those issues.

## How Has This Been Tested?

The existing filter list test in `integrationTest` was a very useful test; also tested with:

- Content in a map directory / sub-directory, all using flags;
- With those flags defined in the map directory, and with those flags defined in the map directory;
- Where flags in the ditaval are in the same directory as the ditaval, and in a subdirectory of the ditaval

[3316simple.zip](https://github.com/dita-ot/dita-ot/files/3243235/3316simple.zip)
[3316.zip](https://github.com/dita-ot/dita-ot/files/3243236/3316.zip)

## Type of Changes

This could break functionality if we removed the obsolete code, but for now it is left in because custom build scripts could be calling it.

## Checklist

Still want to add the complex test cases described above into `integrationTest`, but the code should otherwise be ready for review.


robander:
Also need to test this against #3257 (flag with absolute path) - I don't think we have a test case today for that so I'm not sure of the result.

src/main/java/org/dita/dost/module/CopyFlagModule.java
```diff
@@ -0,0 +1,58 @@
...
+        		
+        		final URI outputDirURI = URLUtils.toURI(outputDirFile);
+        		for (final FileInfo f: fis) {
+        			final File srcFile = new File(f.src);
+        			final File destFile = new File(outputDirURI.resolve(f.uri));
+        			if (srcFile.exists() && destFile != null && !destFile.exists()) {
+        				fileUtils.copyFile(srcFile, destFile);
+        			}
+        		}
+        	} catch (final IOException e) {
```
jelovirt: Could you format the code, the indentation is not consistent.

src/main/java/org/dita/dost/module/CopyFlagModule.java
```diff
@@ -0,0 +1,58 @@
...
+        		final FileUtils fileUtils = FileUtils.newFileUtils();
+        		final File outputDirFile = new File(input.getAttribute(ANT_INVOKER_EXT_PARAM_DITAOUTPUTDIR));
+        		if (!outputDirFile.exists()) {
+        			outputDirFile.mkdir();
+        		}
+        		
+        		final URI outputDirURI = URLUtils.toURI(outputDirFile);
+        		for (final FileInfo f: fis) {
+        			final File srcFile = new File(f.src);
+        			final File destFile = new File(outputDirURI.resolve(f.uri));
```
jelovirt: We should not use the temporary file name to generate the final output file. `FileInfo` has `result`, an absolute URI for the intended output file.

src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
```diff
@@ -748,7 +753,7 @@ private void outputResult() throws DITAOTException {
             getOrCreateFileInfo(fileinfos, file.filename).format = file.format;
         }
         for (final URI file: flagImageSet) {
-            final FileInfo f = getOrCreateFileInfo(fileinfos, file);
+  		  final FileInfo f = getOrCreateFlagImageFileInfo(fileinfos, file, flagImagesMap.get(file));
```
jelovirt: Fix indentation. Are you using 4 spaces?

src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
```diff
@@ -901,6 +906,27 @@ private FileInfo getOrCreateFileInfo(final Map<URI, FileInfo> fileInfos, final U
...
+    /** Treat flag image references as if they are relative to the input dir.
+     * @param fileInfos
+     * @param file
+     * @param file
+     * @return
+     */
+    private FileInfo getOrCreateFlagImageFileInfo(final Map<URI, FileInfo> fileInfos, final URI file, final URI relpath) {
+        assert file.getFragment() == null;
+        final URI f = file.normalize();
+        FileInfo.Builder b;
```
jelovirt: `b` should be `final`

src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
```diff
@@ -901,6 +906,27 @@ private FileInfo getOrCreateFileInfo(final Map<URI, FileInfo> fileInfos, final U
...
+        if (fileInfos.containsKey(f)) {
+            b = new FileInfo.Builder(fileInfos.get(f));
+        } else {
+            b = new FileInfo.Builder().src(file);
+        }
+                b = b.uri(tempFileNameScheme.generateTempFileName(baseInputDir.resolve(relpath)));
```
jelovirt: Indentation

src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
```diff
@@ -901,6 +906,27 @@ private FileInfo getOrCreateFileInfo(final Map<URI, FileInfo> fileInfos, final U
...
+        final FileInfo i = b.build();
```
jelovirt: Set `result` too. It's normally the same as `src`.

src/main/plugins/org.dita.base/build_init.xml
```diff
@@ -22,6 +22,7 @@ See the accompanying LICENSE file for applicable license.
   <taskdef name="pipeline" classname="org.dita.dost.ant.ExtensibleAntInvoker"/>
   <taskdef name="dita-ot-echo" classname="org.dita.dost.ant.DITAOTEchoTask"/>
   <taskdef name="dita-ot-fail" classname="org.dita.dost.ant.DITAOTFailTask"/>
+  <!-- DEPRECATED: dita-ot-copy replaced with org.dita.dost.module.CopyFlagModule -->
```
jelovirt: For the deprecated notes, add the version of OT when it was deprecated.

src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
robander: (sigh) might be, unintentionally - I just set up Eclipse on my new system last week and hadn't changed settings to match the old system.

src/main/plugins/org.dita.base/build_init.xml
robander: Meant to ask about that -- do I list a hotfix 3.3.3 or the next release, 3.4?

src/main/java/org/dita/dost/module/CopyFlagModule.java
robander: 😦 Yes, had not realize this new instance of my Java editor was set up to use tabs instead of spaces. Fixing...

src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
robander: This method was based on the more general `getOrCreateFileInfo` one right above it -- is there a reason that one doesn't use `final`, or should that be changed too? (When I change either of them to `final`, I get compiler warnings a few lines down setting `b.uri()`). Lines 893 and 899 in the latest: https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java#L893

robander: (Wondering if the `.uri()` should be moved up for each of these methods - seems they should be consistent)

src/main/java/org/dita/dost/module/CopyFlagModule.java
robander: I feel like I'm missing something -- `result` looks like it matches `src`, so it's an absolute URI for the source file to be generated, but not absolute for the output file, right? Basing this on what I see for both DITA topics and (non-flag) images in 3.3.2 output for HTML5. For example, with an image referenced in the source:
`<file src="file:/C:/dcs/test/jira226/simple/content/anotherpuff.png" uri="anotherpuff.png" path="anotherpuff.png" result="file:/C:/dcs/test/jira226/simple/content/anotherpuff.png" format="image">`

The output image from this build is actually 
`file:/C:/dcs/test/jira226/simple/content/output332/anotherpuff.png`

robander: (To be clear: I can see how using the temporary file name is wrong, if any other scheme is used we get bad output / incorrect image file names. Just not sure how to translate `result` into the correct output path.)

robander:
After a lot of running in circles on the `uplevels` test -- I've just checked in what the result _should_ be, but it's only working right now if the input map and input filter file do not share any common directories. Still need to work out what's going wrong with the relative path / base dir / input map variables.

PR #3317, opened 2019-06-01
Preserve flags for generated links

robander:

---
name: Pull request

---

## Description

Preserves style / image flagging for generated links. These worked in a previous release but was lost when filter/flag processing moved to the start of pre-processing.

In 2.x, flagging was done at the end of preprocess, after links were already added to topics. Generated links contained all flagging property attributes, so they were flagged at that point.

Now that flagging moved early in the pipeline, the flag information is added before links are created. Any flags associated with the attributes are present in the `<topicref>` that generates the link, but are ignored. This update copies any flag associated with the `<topicref>` into generated links to that `<topciref>`.

## Motivation and Context

Some customers rely heavily on flagging to help identify relevant links. This disappeared with our upgrade form 2.5 to 3.3.

## How Has This Been Tested?

Same test case attached to #3316 

## Type of Changes

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
@@ -596,9 +596,9 @@ See the accompanying LICENSE file for applicable license.
     </xsl:if>
   </xsl:template>
   
-  <xsl:template match="@*|*|text()" mode="add-props-to-link">
+  <xsl:template match="@*|node()" mode="add-props-to-link">
     <xsl:copy>
-      <xsl:apply-templates select="@*|*|text()" mode="add-props-to-link"/>
+      <xsl:apply-templates select="@*|node()" mode="add-props-to-link"/>
     </xsl:copy>
   </xsl:template>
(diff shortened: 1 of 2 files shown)


src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
```diff
@@ -590,10 +591,21 @@ See the accompanying LICENSE file for applicable license.
           <!-- add desc node and text -->
           <xsl:apply-templates select="*[contains(@class, ' map/topicmeta ')]/*[contains(@class, ' map/shortdesc ')]"/>
         </xsl:if>
+        <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-endprop ')]" mode="add-props-to-link"/>
       </link>
     </xsl:if>
   </xsl:template>
   
+  <xsl:template match="@*|*|text()" mode="add-props-to-link">
```
jelovirt: Could this be `@* | node()` to be a bit more generic. Same for the apply-templates expression below.

robander: Updated.

PR #3321, opened 2019-06-12
Preserve xml:lang when processing links

robander:
---
name: Pull request

---

## Description

Ensures that the active `@xml:lang` value is still available when processing links.

The code that processes links uses variables to group them into generated `<linklist>` elements, with locale-based headings such as "Related information". The context when generating those headings is the original document, so the closest set `@xml:lang` value is used to get that variable. However, the link elements themselves are processed from the context of the generated `<linklist>` element, which is the root of a variable -- any `xml:lang` setting on an ancestor element is unavailable. 

This update ensures that the closest `xml:lang` value is preserved on the generated `<linklist>`; in this way, processing any link _within_ those lists has access to the current language. (If the link itself sets an alternate language, that remains the active value.)

## Motivation and Context

We have generated variables for several types of links; in particular, every external link gets hover help like "Opens in a new tab or window". We recently noticed that for all non-English links, that text is broken -- it always comes out in the default language (English), because `getVariable` from that `xsl:variable` context never finds the right language.

Preserving the currently-active language on `linklist` ensures that any variable set from the `<link>` context uses the appropriate language.

## How Has This Been Tested?

Tested with the attached topic (and a few variations):

- Active language of Russian on the topic -- `getVariable` from the link context returns English variables in 3.3.2, but returns Russian for default links with this update
- Tested with alternate language set directly on `<link>` -- that language is used correctly in 3.3.2 and with this update
- Tested with a manually created `<linklist>` with its own language -- that language is used correctly in 3.3.2 and with this update

[jira413.zip](https://github.com/dita-ot/dita-ot/files/3279132/jira413.zip)

## Type of Changes

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
@@ -279,7 +279,7 @@ Each child is indented, the linktext is bold, and the shortdesc appears in norma
     <xsl:param name="links" as="node()*"/>
     <xsl:if test="exists($links)">
       <linklist class="- topic/linklist " outputclass="relinfo">
-        <xsl:sequence select="ancestor-or-self::*[@xml:lang][1]/@xml:lang"/>
+        <xsl:copy-of select="ancestor-or-self::*[@xml:lang][1]/@xml:lang"/>
         <title class="- topic/title ">
           <xsl:call-template name="getVariable">
             <xsl:with-param name="id" select="'Related information'"/>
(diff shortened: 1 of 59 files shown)


src/main/plugins/org.dita.html5/xsl/rel-links.xsl
```diff
@@ -578,6 +579,12 @@ Each child is indented, the linktext is bold, and the shortdesc appears in norma
   </xsl:template>
 
   <!--linklists-->
+  <xsl:template match="*[contains(@class, ' topic/linklist ')]/@xml:lang" priority="100">
```
robander: This section ensures that if the `@xml:lang` value on the link list matches an already-processed ancestor language (such as the language on a root `<topic>`, we don't need to reset `lang` to the same value in the generated HTML5. It also skips doing this if there is no ancestor language and this link list uses the default language. Basically, just a way to keep extra noise out of the HTML.

src/main/plugins/org.dita.xhtml/xsl/xslhtml/rel-links.xsl
```diff
@@ -600,6 +601,12 @@ Each child is indented, the linktext is bold, and the shortdesc appears in norma
   </xsl:template>
 
   <!--linklists-->
+  <xsl:template match="*[contains(@class, ' topic/linklist ')]/@xml:lang" priority="100">
```
robander: This should serve the same purpose as the change in the HTML file. (I didn't intend to check in the priority and should fix that.) However: our root `dita2xhtml.xsl` file overrides processing for `@xml:lang`, and regardless of what priority I set here, any stylesheet call entering with that file gets the processing from that root file. That's why all of the XHTML integration test results had to be updated. 

I originally considered moving that processing from the shell into `dita2htmlImpls.xsl`, which more or less matches the HTML5 organization, but there is a very slight chance of introducing an incompatibility with that.

Any shell using the base XHTML rules, without that `dita2xhtml.xsl` shell, will be able to use this override here; given the status of XHTML output, and that duplicating the lang attribute isn't actually a problem, I decided to just fix the tests.

Issue #3078, opened 2018-09-24
Errors caused by link to topic with xref in shortdesc


robander:
## Expected Behavior

When I have the following condition:

1. Topic contains `<xref>` to another topic `xref-in-shortdesc.dita`
1. The referenced topic has a short description, *and* that short description also contains an `<xref>` to `finaltarget.dita`
1. The xref inside that short description is missing link text or `<desc>`

The `topicpull` step ends up grabbing the short description from the middle file `xref-in-shortdesc.dita`, putting that in a variable, and then processing the variable. The result is that the `<xref>` to `finaltarget.dita` is processed again but in the context of that variable, which causes the processor to look for it in our XSLT directory, and ends with errors. Happens with both `preprocess` and `preprocess2`, here is the error log from `preprocess`:
```
C:\DCS\test\apiref>\dita-ot\dita-ot-3.1.2\bin\dita -i in.ditamap -f html5
[topicpull] I/O error reported by XML parser processing file:/C:/DITA-OT/dita-ot-3.1.2/xsl/preprocess/finaltarget.dita: C:\DITA-OT\dita-ot-3.1.2\xsl\preprocess\finaltarget.dita (The system cannot find the file specified.)
[topicpull] file:/C:/DCS/test/apiref/xref-in-shortdesc.dita:5:68: [DOTX031E][ERROR]: The file finaltarget.dita is not available to resolve link information.
```

This worked at some point in the past (at least, the `xref` inside shortdesc was processed to get text) but could have been broken as far back as 1.8 -- the text was pulled and no errors were thrown.

## Actual Behavior

Errors appear as shown above. `preprocess2` has the same issue:
```
C:\DCS\test\apiref>\dita-ot\dita-ot-3.1.2\bin\dita -i in.ditamap -f pdf
[topicpull] I/O error reported by XML parser processing file:/C:/DITA-OT/dita-ot-3.1.2/xsl/preprocess/dd3f451dd1448c00ddc872728d9573466227ad65.dita: C:\DITA-OT\dita-ot-3.1.2\xsl\preprocess\dd3f451dd1448c00ddc872728d9573466227ad65.dita (Thesystem cannot find the file specified.)
[topicpull] file:/C:/DCS/test/apiref/xref-in-shortdesc.dita:5:68: [DOTX031E][ERROR]: The file dd3f451dd1448c00ddc872728d9573466227ad65.dita is not available toresolve link information.
```

## Possible Solution

[...]

PR #3324, opened 2019-06-13
Pass base context URI when processing links

raducoravu:
Proposed fix for:
https://github.com/dita-ot/dita-ot/issues/3078

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
@@ -162,6 +162,7 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
     -->
   <xsl:function name="dita-ot:getTargetDoc" as="document-node()?">
     <xsl:param name="linkElement" as="element()"/>
+    <xsl:param name="baseContextElement" as="element()?"/>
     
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -696,7 +700,10 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
         <xsl:if test="exists($shortdesc) and not(normalize-space($shortdesc) = '')">
           <xsl:apply-templates select="." mode="topicpull:add-genshortdesc-PI"/>
           <desc class="- topic/desc ">
-            <xsl:apply-templates select="$shortdesc"/>
+            <xsl:apply-templates select="$shortdesc">
+              <!-- EXM-42721 Pass the base context element. -->
```
jelovirt: This should be removed

jelovirt:
DCO is missing.

jelovirt:
Could you squash the commits and use descriptive commit message, the current "Update topicpullImpl.xsl" is a bit generic.

infotexture:
@raducoravu Does this resolve #3078, or are there aspects of that issue unaddressed by these changes?

raducoravu:
@infotexture there are still cases in which the warning is shown, I gave more details and a sample in #3078. But the warning seems to be benign in that case so maybe we should leave this as such.

Issue #1899, opened 2015-04-23
Othermeta and source metadata should not cascade to child topicref


ghost:
Here is an example of the present behaviour:
The map:

``` xml
<bookmap>
    <booktitle>
        <mainbooktitle>Simple example</mainbooktitle>
    </booktitle>
    <bookmeta>
       <othermeta name="this_meta" content="should not be inherited"/>
    </bookmeta>
    <chapter href="child.dita"/>
</bookmap>
```

The topic:

``` xml
<topic id="topic-1">
    <title>The child</title>
    <body>
        <p>Should not have any meta.</p>
    </body>
</topic>
```

The result xhtml transformation of the topic:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html
  PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en-us" lang="en-us">
<head><meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />

<meta name="copyright" content="(C) Copyright 2015" />
<meta name="DC.rights.owner" content="(C) Copyright 2015" />
<meta name="DC.Type" content="topic" />
<meta name="DC.Title" content="The child" />
<meta name="this_meta" content="should not be inherited" />
<meta name="DC.Format" content="XHTML" />
<meta name="DC.Identifier" content="topic-1" />
<link rel="stylesheet" type="text/css" href="commonltr.css" />
<title>The child</title>
</head>
<body id="topic-1">

        <h1 class="title topictitle1" id="ariaid-title1">The child</h1>

        <div class="body">
                <p class="p">Should not have any meta.</p>

        </div>

</body>
</html>
```

[...]

PR #3326, opened 2019-06-18
Removed othermeta from the `cascadeset` as it should not cascade

lionelmoi:
##  `othermeta` and `source` should not cascade
I removed `othermeta` and `source` from the `cascadeset` in MapMetaReader.java.

## Motivation and Context
Based on the official DITA 1.3 specs http://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part3-all-inclusive/archSpec/base/reconciling-topic-and-map-metadata.html#reconciling-topic-and-map-metadata
Fixes issue #1899 

## How Has This Been Tested?
Transformed a map heavy with `othermeta` and with some `source` declarations. Cascading behaves as expected.

## Type of Changes
- Bug fix

Signed-off-by: Lionel Moizeau <lionel.moizeau@gmail.com>

### DIFF
--- a/src/main/java/org/dita/dost/reader/MapMetaReader.java
+++ b/src/main/java/org/dita/dost/reader/MapMetaReader.java
@@ -47,7 +47,6 @@ public final class MapMetaReader extends AbstractDomFilter {
     private static final Set<String> cascadeSet = Collections.unmodifiableSet(new HashSet<>(asList(
             TOPIC_AUDIENCE.matcher,
             TOPIC_AUTHOR.matcher,
-            TOPIC_SOURCE.matcher,
             TOPIC_CATEGORY.matcher,
             TOPIC_COPYRIGHT.matcher,
             TOPIC_CRITDATES.matcher,
--- a/src/main/java/org/dita/dost/reader/MapMetaReader.java
+++ b/src/main/java/org/dita/dost/reader/MapMetaReader.java
@@ -53,7 +53,6 @@ public final class MapMetaReader extends AbstractDomFilter {
             TOPIC_CRITDATES.matcher,
             TOPIC_PERMISSIONS.matcher,
             TOPIC_PRODINFO.matcher,
-            TOPIC_OTHERMETA.matcher,
             TOPIC_PUBLISHER.matcher
     )));
     private static final Set<String> metaSet = Collections.unmodifiableSet(new HashSet<>(asList(
@@ -307,7 +306,7 @@ public final class MapMetaReader extends AbstractDomFilter {
                     } else {
                         if (TOPIC_NAVTITLE.matches(classValue)) {
                             //Add locktitle value to navtitle so we know whether it should be pushed to topics
-                            final String locktitleAttr =  ((Element) meta.getParentNode()).getAttributeNode(ATTRIBUTE_NAME_LOCKTITLE) != null ? 
+                            final String locktitleAttr =  ((Element) meta.getParentNode()).getAttributeNode(ATTRIBUTE_NAME_LOCKTITLE) != null ?
                                                           ((Element) meta.getParentNode()).getAttributeNode(ATTRIBUTE_NAME_LOCKTITLE).getNodeValue() : "no";
                             elem.setAttributeNS(DITA_OT_NS, DITA_OT_NS_PREFIX + ":" + ATTRIBUTE_NAME_LOCKTITLE, locktitleAttr);
                         }
--- a/src/main/java/org/dita/dost/reader/MapMetaReader.java
+++ b/src/main/java/org/dita/dost/reader/MapMetaReader.java
@@ -50,6 +50,7 @@ public final class MapMetaReader extends AbstractDomFilter {
             TOPIC_CATEGORY.matcher,
             TOPIC_COPYRIGHT.matcher,
             TOPIC_CRITDATES.matcher,
+            TOPIC_METADATA.matcher,
             TOPIC_PERMISSIONS.matcher,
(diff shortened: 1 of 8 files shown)


lionelmoi:
*do not merge*

I'm still missing `<metadata>` elements.

lionelmoi:
@jelovirt  @robander I need your opinion.
 The DITA specs specifies that the `<metadata>` element must cascade: https://dita.fluidtopics.com/reader/7O8WA55ox7Lhg9IPioVbSQ/IYAm3JV0n_W9AAp1FfQj3w

Element | How does it apply to the topic? | Does it cascade to child topicref elements? | What is the purpose when set on the map element?
-- | -- | -- | --
metadata | Add to the topic | Yes | Specify metadata for the entire map

I interpret as "the `<metadata>` element and all its content must cascade". Otherwise metadata will never cascade as it is unwrapped somewhere during preprocessing.

robander:
I think the whole issue of `<metadata>` cascading is difficult -- largely because of how it evolved, and our inability to break compatibility between versions of DITA 1.x.

The `<metadata>` container didn't even exist in DITA 1.0 -- it was added later to make `<prolog>` and `<topicmeta>` work more alike, allowing copy/paste type reuse or moving between the elements.

When it says that `<metadata>` cascades, that definitely does not / cannot mean "everything inside of the container", because the container also allows quite a few things that explicitly do not cascade. For example, I've seen documents where index terms are defined in that container -- a habit that started when they were defined inside of topics:
```
<metadata>
  <keywords><indexterm>example</indexterm>
</metadata>
```
The `<keywords>` element does not cascade, based on the same topic you referenced. If everything in the `<metadata>` container cascades, then this index entry would also cascade, and would apply to every document in the branch -- index terms defined like this would become cumulative, where every topic is indexed with (index entries defined here or an any ancestor).

So ... this is unclear and needs to be fixed in the DITA 2.x spec but I'm not sure what to say right now, for DITA-OT behavior with 1.x. We definitely do not want everything inside of there to cascade, but when elements inside of `<metadata>` do cascade -- like `<audience>` -- it is appropriate for the container to cascade with that inside.

lionelmoi:
I think there is a nice use case for cascading the contents of the `<metadata>` tag.

If you put something in the metadata tag it means you want it to cascade. If you put something outside it means that it should only cascade if it is meant to.

It provides information architect with a way to define metadata *packages* that are inherited by the `<map>` or `<topicref>` children.

If we explain it, and I volunteer to do so, I think this behavior would be a nice addition to the DITA-OT.

robander:
I can understand that argument, but ... adding it would definitely break a lot of my documents for indexing (which is why I used that example). So I would not want this as a change to the default behavior -- by default, elements inside of `<metadata>` that are not supposed to cascade, should still not cascade.

lionelmoi:
I see your point.

Could we imagine implementing it behind a parameter? Or as a plugin?

I could dedicate some time for this but I would need some pointers as there is **absolutely no comments** in the java code...

robander:
I think a plugin would be best, but I'm not sure how best to implement. I've tried to fix a couple of items in that cascade code without success -- it is some of the oldest code still in DITA-OT, written by people who left the project many years ago. 

I know that the general design is to create a mapping table of (`href` values) associated with `metadata`. But this can cause problems - which I was not able to solve - when the same `href` appears more than once. In that case, each entry adds to the collected `topicmeta`, and then the merged result is written out for all cases. But this means that if I use metadata for one instance of `example.dita`, and I want different metadata in another context -- like a different navigation title or short description -- it does not work. I know there was an issue open for this several years ago but it is possible this was closed as stale.

lionelmoi:
So this is implemented in xslt? 

Can you point to where the ancient code is?

robander:
I was about to say "no" but ... I'm not actually sure, it depends on where you want / need to make a change. There is both Java and XSLT code involved.

Java code pushes the content to topics: https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/module/MoveMetaModule.java

XSLT implements attribute cascading, but it's been a long time since I looked ... I can't remember if this also handles element cascading, or if that's in `MoveMetaModule` -- https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/xsl/preprocess/mappullImpl.xsl

You can extend the `mappull` XSLT step, so that might be the easiest place to hook in and add new cascading (but it would not be a great spot to change / undo cascading that already happened in the Java module).

lionelmoi:
After some painful investigation, I think this is all handled in Java
here: https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/reader/MapMetaReader.java
and here: https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/writer/AbstractDitaMetaWriter.java + https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/writer/DitaMetaWriter.java and maybe here: https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/writer/DitaMapMetaWriter.java

🤕 -> 💥

lionelmoi:
If reviewers agree this PR can be merged as the '<metadata>' issue evoked in this discussion does not enter its scope.

jelovirt:
Rebased locally and because there were multiple commits that should have been together, squashed into ce905b4

jelovirt:
@lionelmoi sorry getting this in took so long, but it seems we are really slow.

lionelmoi:
It's fine @jelovirt :) 

I'm happy to see my first Java pull-request merged!

PR #3330, opened 2019-06-27
Update Saxon to 9.9.1.3

jelovirt:
## Description
Update Saxon dependency to latest stable version 9.9.1.3

## Motivation and Context
Older versions of Saxon are likely to not be updated anymore and newer version have additional features, XSLT 3.0 in particular.

## How Has This Been Tested?
Passes tests.
## Type of Changes

- Update dependency

## Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ dependencies {
     compile group: 'xerces', name: 'xercesImpl', version:'2.12.0'
     compile group: 'xml-apis', name: 'xml-apis', version:'1.4.01'
     compile group: 'xml-resolver', name: 'xml-resolver', version:'1.2'
-    compile group: 'net.sf.saxon', name: 'Saxon-HE', version: '9.8.0-14'
+    compile group: 'net.sf.saxon', name: 'Saxon-HE', version: '9.9.1-4'
     compile group: 'com.ibm.icu', name: 'icu4j', version:'61.1'
     compile group: 'org.apache.ant', name: 'ant', version:'1.10.5'
     compile group: 'org.apache.ant', name: 'ant-launcher', version:'1.10.5'


jelovirt:
@raducoravu will you have problems with this upgrade?

drmacro:
This will require reworking the code for integrating saxon extension functions and any libraries that provide extension functions or collators. Not a huge deal but the java API did change in a way that means the 9.8 version can't work with 9.9 and visa versa, as far as I know.

Not saying we shouldn't just indicating that there is a cost.

raducoravu:
@jelovirt maybe I can try next week our plugins with your Saxon upgrade and give some feedback.

drmacro:
Also, since I haven't been able to get the glossary processing code working the latest OT due to lack of time I don't really have a hard dependency on 9.8 if this upgrade is done soon, so I say go for it.

robander:
I'm happy with the upgrade, but would be good to know if it breaks that code we added in 3.3 first.

raducoravu:
I started looking into how our plugins are affected by Saxon 9.9, one thing which would have made my debug process easier (without the need to overwrite the entire XsltModule class) would have been this:
https://github.com/dita-ot/dita-ot/issues/3334

raducoravu:
@jelovirt Thanks for suggesting for me to look into how our plugins work with Saxon 9.9. I looked at how our WebHelp plugins interact with your DITA OT Saxon 9.9 branch. Made a change to an XSLT stylesheet which was subtracting an integer from a parameter which did not have a specified type (and somehow 9.9 had problems with this). Also made a small change in one of our Saxon extension Java function implementations as it was using an utility method from "net.sf.saxon.om.SequenceTool" which is now gone. 
So no problem on my side, I made things work on my side with small changes, the changes I made are also compatible with the older Saxon 9.8.

PR #3346, opened 2019-08-01
Fix link text for flagged links

robander:

---
name: Pull request

---

## Description

Fixes an issue with resolving link text when the cross reference or link is flagged.

## Motivation and Context

Encountered this recently in customer documentation. 

When an element is flagged, DITA-OT adds the flagging information using `ditaval-startprop` and `ditaval-endprop` elements. 

The code that pulls in link text does not take that into account, and assumes that anything other than `<desc>` equates to link text. So if a link or cross reference is empty, but contains start or end flagging elements, the `topicpull` module thinks it has link text and nothing is pulled in. As a result, only the flag appears in place of the link.

The updated code replaces the (anything but `desc`) test with (anything but `desc` or `ditaval-startprop` or `ditaval-endprop`). It also updates the test for text content to use `text()` rather than `normalize-space(.)` (which was also thrown off by flagging).

To simplify the code a bit, I spit the single `topicpull:get-stuff_get-linktext` match template into two. Previously the logic in that template was entirely split -- "If a link with text, do X, otherwise if not a link with text, do Y, otherwise if a link, do Z.1, otherwise do Z.2". It didn't seem to buy anything to keep those together when there was so little overlapping logic, especially when adding even more tests for flag elements.

## How Has This Been Tested?

Test case attached, with map, topic with links, target topic, and DITAVAL. Build the test to HTML; before the fix, every link or cross reference with `platform="aix"` is broken, and after the fix, they all have the same link text as links without flags.

## Type of Changes


src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -714,37 +714,55 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
   <xsl:template match="*" mode="topicpull:get-stuff_get-linktext">
     <xsl:param name="targetElement" as="element()"/>
     
     <xsl:choose>
-      <xsl:when test="contains(@class,' topic/link ') and *[not(contains(@class, ' topic/desc '))]">
-        <xsl:apply-templates select="." mode="topicpull:add-usertext-PI"/>
-        <xsl:apply-templates select="*[not(contains(@class, ' topic/desc '))]|comment()|processing-instruction()"/>
-      </xsl:when>
-      <xsl:when test="dita-ot:is-link(.) and (normalize-space(.) != '' or *[not(contains(@class, ' topic/desc '))])">
+      <xsl:when test="dita-ot:is-link(.) and (normalize-space(text()) != '' or *[not(contains(@class, ' topic/desc ') or contains(@class,' ditaot-d/ditaval-startprop ') or contains(@class,' ditaot-d/ditaval-endprop '))])">
```
robander: This line won't work, needs to be fixed.

I switched from `normalize-space(.)` to `normalize-space(text())` because using the current node will give the wrong result when there is a `<desc>` tag -- but `text()` fails when there are sub-elements and you get more than one text node.

jelovirt: Has this been fixed?

src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -714,37 +714,55 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
...
             <xsl:with-param name="targetElement" as="element()" select="$targetElement"/>
           </xsl:apply-templates>
         </xsl:variable>
+        <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-startprop ')]"/>
         <xsl:if test="exists($linktext) and dita-ot:is-link(.)">
           <xsl:apply-templates select="." mode="topicpull:add-gentext-PI"/>
-          <!-- FIXME: need to avoid flattening complex markup here-->
-          <xsl:value-of select="$linktext"/>
+          <xsl:sequence select="$linktext"/>
+          <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-endprop ')]"/>
```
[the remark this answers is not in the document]
robander: Yes - fixed now with latest commit, thanks.

src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
robander: Now fixed, and split into multiple lines.

src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -298,6 +298,17 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
     <xsl:sequence select="$result"/>
   </xsl:function>
   
+  <xsl:function name="dita-ot:textNodesHaveContent" as="xs:boolean">
+    <xsl:param name="ctx" as="element()"/>
+    <xsl:variable name="textNodes" as="text()*">
```
jelovirt: I'd prefer to simplify this to

```xslt
<xsl:sequence select="some $t in $ctx/text() satisfies normalize-space($t) ne ''"/>
```

or even just

```xslt
<xsl:sequence select="some $t in $ctx/text() satisfies normalize-space($t)"/>
```

jelovirt: Fixed this.

PR #3358, opened 2019-09-20
Add Docker configuration

jelovirt:
Add DITA-OT Docker image.

This image is not based on [dita-ot-project-docker](https://github.com/dita-community/dita-ot-project-docker), but instead is a independent image for DITA-OT that is under the [dita-ot organization](https://github.com/dita-ot).

The image will be published into [GitHub Package Registry](https://help.github.com/en/articles/about-github-package-registry) instead of [Docker Hub](https://hub.docker.com/).

## Documentation

Documentation stubs for using DITA-OT via Docker image:

### Running

```bash
docker run -it \
  -v /Volumes/tmp:/src docker.pkg.github.com/dita-ot/dita-ot/dita-ot:3.3.4 \
  -i /src/topic.dita -o /src/out -f html5 -v
```

Where you map your local `/Volumes/tmp` to the container’s `/src` directory. Then you pass it the `dita` command arguments, so that you reference files in the containers `/src` directory.

### Extending

Custom plug-ins can be installed by creating a new Docker image that's based on the DITA-OT image:

```Dockerfile
FROM docker.pkg.github.com/dita-ot/dita-ot/dita-ot:3.3.4

RUN dita --install https://example.com/plugin.zip
RUN dita --install from-plugin-registry
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ stages:
     if: repo = dita-ot/dita-ot AND branch = develop AND type != pull_request
   - name: release
     if: repo = dita-ot/dita-ot AND tag IS present AND type != pull_request
+  - name: deploy
+    if: repo = dita-ot/dita-ot AND tag IS present AND type != pull_request
 jobs:
   include:
     # Test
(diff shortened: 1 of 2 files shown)


jelovirt:
Building the image will be made part of the CI pipeline.

.travis.yml
```diff
@@ -91,6 +91,20 @@ jobs:
...
+    - state: deploy
+      name: "Deploy Docker"
+      install: skip
+      script: skip
+      deploy:
+        - provider: script
+          script: |-
+            docker login docker.pkg.github.com -u $GITHUB_USER -p $GITHUB_PACKAGE_REGISTRY
+            TAG=${TRAVIS_TAG} docker build -t docker.pkg.github.com/dita-ot/dita-ot/dita-ot:${TRAVIS_TAG} --build-arg VERSION=${TRAVIS_TAG} .
+            TAG=${TRAVIS_TAG} docker push docker.pkg.github.com/dita-ot/dita-ot/dita-ot:${TRAVIS_TAG}
```
jelovirt: Since this deployment doesn't build the package, only downloads it from GitHub, this can be ran locally to build image for older versions of DITA-OT.

docker-compose.yml
```diff
@@ -0,0 +1,5 @@
+version: '2.4'
```
jelovirt: The Docker Compose file isn't used by Travis, so I suppose it could be removed from the PR?

infotexture: I think the question is whether this might be potentially useful to anyone else as an example, even if it's not required by Travis.

jelovirt: Not really. Removing

jelovirt:
@thendarion Created an Alpine based image and it's 263 MB vs. the Debian based 343 MB. I would prefer to only publish a single image, because the difference isn't that big.

I switched to JRE from JDK.

PR #3363, opened 2019-09-23
Remove TocJS plugin from default distribution

jelovirt:
Remove [TocJS plug-in](https://github.com/dita-ot/com.sophos.tocjs/) from default DITA-OT distribution package.


robander:
The TocJS output was a really nice frame-based improvement on basic XHTML back when it was added, but at this point I think it's pretty much obsolete in favor of lightly customized HTML5. I think it's probably time to remove it from the default distribution, and those who still use it can get it from the registry.

PR #3364, opened 2019-09-23
Remove troff plugin from default distribution

jelovirt:
Remove [troff plug-in](https://github.com/dita-ot/org.dita.troff) from default DITA-OT distribution package.


robander:
For a select set of documentation, we still rely heavily on `troff`, and even extend it to create other varieties of output. But, I also know that it's now easy to get it from the registry, so for our installation this would not be a problem. If anyone else is 1) using it and 2) keeping up to date with DITA-OT, they should also be able to get it from the registry. 

Are others (who follow this repository) still using `troff` output, or are we the only ones? As long as it's a pretty small number, I think it's probably OK to remove it from the default distribution.

PR #3373, opened 2019-10-03
Fix index link when term has subterms #3320; fix index-see-also #3314

### LINKED ISSUE #3314: PDF index <index-see-also> and <index-see> issues
PDF FOP index produces the following issues related to `<index-see-also>`:
1. If there are multiple `<index-see-also>` elements as children of `<indexterm>` only the first one (alphabetically, after sorting occurs between *_MERGED.xml and stage1.xml) appears in the index. In the example below, only "See also DTD" is emitted. Both "RELAX NG" and "schema" are suppressed.
```xml
<indexterm>grammar files<index-see-also>schema</index-see-also>
                        <index-see-also>DTD</index-see-also>
                        <index-see-also>RELAX NG</index-see-also>
</indexterm>
```
Note: Changing the markup to put each `<index-see-also>` as a lone child of `<indexterm>` had no effect on the final "see also" entries; it does however create extra page number entries for "grammar files" in the index, which is likely related to the next issue.
```xml
<indexterm>grammar files<index-see-also>schema</index-see-also></indexterm>
<indexterm>grammar files<index-see-also>RELAX NG</index-see-also></indexterm>
<indexterm>grammar files<index-see-also>DTD</index-see-also></indexterm>
```

2. When emitting multiple "see also" entries, they should be separated by a comma. I don't see any processing in index.xsl or index_fop.xsl that would account for that currently.

3. The `<index-see-also>` entries show their page number twice (e.g., "schema 3, 3") -- once for each time the term is used in my test files. Only one entry per page should be emitted. Because the page number is emitted once for each time that a term appears in `<indexterm>` or `<index-see-also>` leads me to believe that each `<index-see-also>` is being processed, but there is no logic to remove duplicated page numbers.

Removing `[1]` from from certain `opentopic-index:index.entry[1]` in plugins\org.dita.pdf2\xsl\fo\index.xsl does emit all of the "see also" entries, albeit the linking is wrong (only to the alphabetically first ID) and there is no comma. 

I'm not sure if anything in plugins\org.dita.pdf2.fop\xsl\fo\index_fop.xsl is implicated.

Tested using
* The attached files
* DITA-OT 3.3.1 with no custom plugins
* Generated from oXygen.

The attached .zip file contains a DITA map and one topic file to demonstrate the issues.

Edit: I also confirmed that `<index-see>` is also affected. 

[see-also.zip](https://github.com/dita-ot/dita-ot/files/3238413/see-also.zip)

### LINKED ISSUE #3320: Index entries missing if child entries exist
# Description
The DITA specification is clear that an `<indexterm>` with children should not create locators (i.e., links or page numbers). 

> In the case of nested <indexterm> elements, each <indexterm> with no <indexterm> children (the "leaves") contributes a page number to the generated index; the ancestral <indexterm> elements for each leaf <indexterm> provide the higher levels for the multilevel entry.

But the specification appears to make an assumption that the index entries are all in the same file. That distinction is not clear when multiple files are involved. 

garagetaskoverview.xml & wwfluid.xml
`<indexterm>garage</indexterm>`

shelving.xml
`<indexterm>garage<indexterm>shelving</indexterm></indexterm>`

# Actual result
garage
&nbsp;&nbsp;&nbsp;shelving 5

# Expected result
garage 3, 7
&nbsp;&nbsp;&nbsp;shelving 5

Bonus: OT Info message suggestion. "This `<indexterm>` has the same text as at least one other `<indexterm>` that has children entries."

# Environment
OT used: DITA OT 3.3.1
Target: pdf
PDF Formatter: FOP
OT run from: Command line and oXygen

# Errors
None in log.

## Files to reproduce

[omitted-entries.zip](https://github.com/dita-ot/dita-ot/files/3278587/omitted-entries.zip)

robander:

---
name: Pull request

---

## Description

Current PDF output will drop all instances of an index term from the index, if even one instance of the term has a sub-term. This is done silently, so adding one secondary or tertiary term can erase any number of page links for the primary term.

Doing this ignores what the actual markup says, and might not even be the original intent (some of the code design is unclear). 

Some indexing tools follow a practice that when there are child terms, the author should not index the primary; however, our code should respect the markup and leave that restriction to local rules.

The update here adds a variable `$index.allow-link-with-subterm` which defaults to `true()`; if set to `false()` it will ignore the links on the parent term as we do today. However, that should not be the default.

## Motivation and Context

Fixes #3320 
Fixes #3314

## How Has This Been Tested?

Tested with the content in 3320; using the default `true()` setting keeps the expected links, while setting to `false()` removes them.

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.pdf2.xep/xsl/fo/index_xep.xsl
+++ b/src/main/plugins/org.dita.pdf2.xep/xsl/fo/index_xep.xsl
@@ -117,7 +117,7 @@ See the accompanying LICENSE file for applicable license.
                         <xsl:if test="contains($isNormalChilds,'true ')">
                           <xsl:apply-templates select="." mode="make-index-ref">
                             <xsl:with-param name="idxs" select="if ($index.allow-link-with-subterm and exists(key('index-leaves',@value)))
-                              then (opentopic-index:refID)
+                              then opentopic-index:refID
                               else ()"/>
                             <xsl:with-param name="inner-text" select="opentopic-index:formatted-value"/>
                             <xsl:with-param name="no-page" select="$isNoPage"/>
--- a/src/main/plugins/org.dita.pdf2.xep/xsl/fo/index_xep.xsl
+++ b/src/main/plugins/org.dita.pdf2.xep/xsl/fo/index_xep.xsl
@@ -116,6 +116,9 @@ See the accompanying LICENSE file for applicable license.
                         </xsl:variable>
                         <xsl:if test="contains($isNormalChilds,'true ')">
                           <xsl:apply-templates select="." mode="make-index-ref">
+                            <xsl:with-param name="idxs" select="if ($index.allow-link-with-subterm and exists(key('index-leaves',@value)))
+                              then (opentopic-index:refID)
+                              else ()"/>
                             <xsl:with-param name="inner-text" select="opentopic-index:formatted-value"/>
                             <xsl:with-param name="no-page" select="$isNoPage"/>
                           </xsl:apply-templates>
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
@@ -249,12 +249,17 @@ See the accompanying LICENSE file for applicable license.
                         <xsl:with-param name="id" select="'Index See String'"/>
                     </xsl:call-template>
                 </fo:inline>
-                <fo:basic-link>
-                    <xsl:attribute name="internal-destination">
-                        <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-destination"/>
-                    </xsl:attribute>
-                    <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-value"/>
-                </fo:basic-link>
+                <xsl:for-each select="opentopic-index:index.entry">
+                    <xsl:if test="not(position() eq 1)">
+                        <xsl:text>, </xsl:text>
+                    </xsl:if>
+                    <fo:basic-link>
+                        <xsl:attribute name="internal-destination">
+                            <xsl:apply-templates select="." mode="get-see-destination"/>
+                        </xsl:attribute>
+                        <xsl:apply-templates select="." mode="get-see-value"/>
+                    </fo:basic-link>
+                </xsl:for-each>
             </xsl:when>
             <xsl:otherwise>
                 <xsl:call-template name="output-message">
@@ -271,12 +276,17 @@ See the accompanying LICENSE file for applicable license.
                             <xsl:with-param name="id" select="'Index See Also String'"/>
                         </xsl:call-template>
                     </fo:inline>
-                    <fo:basic-link>
-                        <xsl:attribute name="internal-destination">
-                            <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-destination"/>
-                        </xsl:attribute>
-                        <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-value"/>
-                    </fo:basic-link>
+                    <xsl:for-each select="opentopic-index:index.entry">
+                        <xsl:if test="not(position() eq 1)">
+                            <xsl:text>, </xsl:text>
+                        </xsl:if>
+                        <fo:basic-link>
+                            <xsl:attribute name="internal-destination">
+                                <xsl:apply-templates select="." mode="get-see-destination"/>
+                            </xsl:attribute>
+                            <xsl:apply-templates select="." mode="get-see-value"/>
+                        </fo:basic-link>
+                    </xsl:for-each>
                 </fo:block>
             </xsl:otherwise>
         </xsl:choose>
@@ -306,8 +316,10 @@ See the accompanying LICENSE file for applicable license.
     <xsl:template match="opentopic-index:index.entry" mode="get-see-value">
         <fo:inline>
             <xsl:apply-templates select="opentopic-index:formatted-value/node()"/>
-            <xsl:text> </xsl:text>
-            <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-value"/>
+            <xsl:if test="opentopic-index:index.entry[1]">
+              <xsl:text> </xsl:text>
+              <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-value"/>
+            </xsl:if>
         </fo:inline>
     </xsl:template>
 
@@ -318,12 +330,17 @@ See the accompanying LICENSE file for applicable license.
                     <xsl:with-param name="id" select="'Index See Also String'"/>
                 </xsl:call-template>
             </fo:inline>
-            <fo:basic-link>
-                <xsl:attribute name="internal-destination">
-                    <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-destination"/>
-                </xsl:attribute>
-                <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-value"/>
-            </fo:basic-link>
+            <xsl:for-each select="opentopic-index:index.entry">
+                <xsl:if test="not(position() eq 1)">
+                    <xsl:text>, </xsl:text>
+                </xsl:if>
+                <fo:basic-link>
+                    <xsl:attribute name="internal-destination">
+                        <xsl:apply-templates select="." mode="get-see-destination"/>
+                    </xsl:attribute>
+                   <xsl:apply-templates select="." mode="get-see-value"/>
+                </fo:basic-link>
+            </xsl:for-each>
         </fo:block>
     </xsl:template>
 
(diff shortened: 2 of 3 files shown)


robander:
Not yet ready for review - still need to remove any page link for the primary term when it *does* specify a secondary.

robander:
Updated and added a better / more complete test case; tested for AH and FOP output. 
[3320.zip](https://github.com/dita-ot/dita-ot/files/3692243/3320.zip)

To validate output when publishing `3320.ditamap`:

- `EVERYWHERE` is only a primary term, should have page links. (FOP doesn't remove duplicates so I get one page twice.)
- `Issue3320` should have one page link because used as primary once; it should have 2 sub-terms each with a page link
- `Only subterms` should not have any page links, only subterms
- `PrimaryA` should have two page links and one child with a page link 
- `PrimaryB` should have one page link and one child with a page link
- `SeeALSOTest` should have a page link, with the See also
- `SeeTest` should not have a page link

robander:
Adding to this pull request a fix for #3314.

- If a term has more than one `index-see-also`, current processing ignores all but the first. Latest commit fixes that issue by preserving all see-also instances, separating them with a comma. If the two "see also" references are to `IMPORTANT` and `MORE IMPORTANT`, current processing will generate "See also IMPORTANT". With this fix, it will generate "See also IMPORTANT, MORE IMPORTANT"`. Each see-also target is linked correctly.
- If a term has more than one `index-see`, current processing ignores all but the first. Latest commit fixes that issue by preserving all see instances, separating them with a comma. If the two "see" references are to `IMPORTANT` and `MORE IMPORTANT`, current processing will generate "See IMPORTANT" after the term. With this fix, it will generate "See IMPORTANT, MORE IMPORTANT"`. Each see-also target is linked correctly. A better long term fix would be to convert one or all of these to see-also but it didn't seem like it was worth going that far (particularly having to handle the case of a term with both `index-see` and `index-see-also`.

Attaching test cases for #3314. 
[3314.zip](https://github.com/dita-ot/dita-ot/files/3692814/3314.zip)

infotexture:
Output files from FOP test above:
- [3320.pdf](https://github.com/dita-ot/dita-ot/files/3693410/3320.pdf)
- [3320book.pdf](https://github.com/dita-ot/dita-ot/files/3693411/3320book.pdf)

infotexture:
> The update here adds a variable `$index.allow-link-with-subterm` which defaults to `true()`; if set to `false()` it will ignore the links on the parent term as we do today. However, that should not be the default.

@robander Would users be able to adjust this default by passing a parameter on the CLI, or would changing this require an XSLT customization plug-in?

@lief-erickson With this change, the current docs index in https://github.com/dita-ot/docs/pull/264 would show a large number of (redundant) page numbers for certain parent entries:

docs-index_pr-264_pr-3373

lief-erickson:
@infotexture, that screen capture certainly exposes the duplicate page number issue in FOP, doesn't it?

robander:
@infotexture not sure why you are getting page links on the primary term when the secondary is indexed. That's not what I see so I need to double check if something wasn't checked in.

As currently defined this is not a CLI option - to me, this feels way too specific an option to expose on the CLI. Especially for this case -- the only reason to suppress them is if you have a rule that primary terms cannot be indexed when secondary terms are present, but in that case, I think our default should still be "Do what the markup says and if you don't like it, you can update the markup per your local rules"

And the duplicate page numbers -- yeah, that has always been the case with FOP output, just wasn't exposed for us in this case because we ignored all of those entries.

src/main/plugins/org.dita.pdf2.xep/xsl/fo/index_xep.xsl
```diff
@@ -116,6 +116,9 @@ See the accompanying LICENSE file for applicable license.
                         </xsl:variable>
                         <xsl:if test="contains($isNormalChilds,'true ')">
                           <xsl:apply-templates select="." mode="make-index-ref">
+                            <xsl:with-param name="idxs" select="if ($index.allow-link-with-subterm and exists(key('index-leaves',@value)))
+                              then (opentopic-index:refID)
```
jelovirt:
```suggestion
                              then opentopic-index:refID
```

src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
```diff
@@ -58,6 +59,11 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="warn-enabled" select="true()"/>
 
   <xsl:key name="index-key" match="opentopic-index:index.entry" use="@value"/>
+  <xsl:key name="index-leaves"
+    match="opentopic-index:index.entry
+              [empty(opentopic-index:index.entry|opentopic-index:see-childs)]
+              [not(ancestor::opentopic-index:index.group)]"
```
jelovirt:
```suggestion
              [empty(ancestor::opentopic-index:index.group)]" 
```

src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
```diff
@@ -263,12 +276,17 @@ See the accompanying LICENSE file for applicable license.
...
                         </xsl:call-template>
                     </fo:inline>
-                    <fo:basic-link>
-                        <xsl:attribute name="internal-destination">
-                            <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-destination"/>
-                        </xsl:attribute>
-                        <xsl:apply-templates select="opentopic-index:index.entry[1]" mode="get-see-value"/>
-                    </fo:basic-link>
+                    <xsl:for-each select="opentopic-index:index.entry">
+                        <xsl:if test="not(position() eq 1)">
```
jelovirt:
```suggestion
                        <xsl:if test="position() ne 1">
```

src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
```diff
@@ -220,16 +228,16 @@ See the accompanying LICENSE file for applicable license.
 
     <xsl:template match="opentopic-index:index.entry[not(opentopic-index:index.entry)]" mode="index-postprocess" priority="1">
         <xsl:variable name="page-setting" select=" (ancestor-or-self::opentopic-index:index.entry/@no-page | ancestor-or-self::opentopic-index:index.entry/@start-page)[last()]"/>
-    <xsl:variable name="isNoPage" select=" $page-setting = 'true' and name($page-setting) = 'no-page' "/>
+        <xsl:variable name="isNoPage" select=" $page-setting = 'true' and name($page-setting) = 'no-page' "/>
```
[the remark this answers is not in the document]
robander: Honestly - it looks like no, for the reason you describe. I didn't touch that code and haven't tested it. (Didn't even examine it while working on the rest of this.) I'm guessing it's related to page ranges, which I think have other open issues.

robander:
@infotexture testing again with code built from github and I reproduced the results you got with FOP -- investigating.

src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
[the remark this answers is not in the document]
robander: I've fixed the spacing on the variable declaration but would rather leave this alone until I have a test case that would demonstrate whether it works or not - I think you're right that it shouldn't work, but ...

lief-erickson:
Tested with FOP:
* Using the DITA-OT User Guide [1393f62b](https://github.com/dita-ot/docs/commit/1393f62bf089f65f58ca5c215f037cd22be4ae04). The dropped `see-also` terms are present. That is fixed.

I am seeing primary entries that have sub-terms show page numbers. Several examples: Ant, dita command, DITAVAL, and many more. This is good. 

In the case of DITAVAL, which is one example of several, the page numbers shown after the primary term are also shown after a sub-term. After some testing it appears as though the `<index-see-also>` markup is contributing a page number to the primary term. This doesn't seem correct to me. 

Both of these entries below contribute a page locator to the primary DITAVAL entry (even the second one). I would expect no page locator and the term (clickable) in the "see also" list (for the relevant primary term).

`<indexterm>DITAVAL<index-see-also>filters</index-see-also></indexterm>`
`<indexterm>profiling<index-see-also>DITAVAL</index-see-also></indexterm>`

* Using a private test document that showed results similar to the DITA-OT User Guide.

robander:
@lief-erickson That was not my understanding of `index-see-also`, based on this bit from the spec: http://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part1-base/langRef/base/index-see-also.html

> In addition to its "see also" redirection, an <index-see-also> functions as a pointwise index term, thereby typically generating a page reference as well as the "see also" indication.

As I've understood the specification and the implementation -- "see also" means "Index this point, and also put out a pointer to other terms that may be useful"; "index-see" does not put out a point, and is purely a redirect.

lief-erickson:
@robander, you're right. The spec does indeed say otherwise. It's not my expectation, but the output does match the spec.

robander:
@lief-erickson thanks for testing, and highlighting that issue. I've updated my test case to add this in the first topic:
`<indexterm>ChildA</indexterm>`
and these in the second topic:
`<indexterm>ChildB<index-see-also>ChildA</index-see-also></indexterm>`
`<indexterm>ChildBSee<index-see>ChildA</index-see></indexterm>`

As you noted -- both of those entries in the second topic incorrectly added page links to "ChildA", going to that second topic. I've got an update ready so that only the entry in the first topic creates a page link to "ChildA".

PR #3374, opened 2019-10-04
Add move before each element definition

raducoravu:
Oxygen's content completion gathers annotations only when they are placed just before the element definition. So I moved the annotations just before the element define.

### DIFF
--- a/src/main/resources/project.rnc
+++ b/src/main/resources/project.rnc
@@ -1,25 +1,26 @@
 default namespace = "https://www.dita-ot.org/project"
 
 
+project = 
 ## Publication project
-project = element project { (includes | deliverable | publication | context)* }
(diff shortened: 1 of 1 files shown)


infotexture:
You need to [sign your work](http://www.dita-ot.org/DCO) to give us permission to use  your contribution.

To sign off your last commit, use:

```
git commit --amend --signoff
```

Then force-push to update the branch on your fork with:

```
git push --force-with-lease
```

raducoravu:
@infotexture I feel like such an idiot when using git. I could not do the amend from inside SourceTree, I'm afraid I might break something using the command line so I'll probably delete the branch and open another pull request (hopefully not fogetting to sign that one :))

Issue #2914, opened 2018-03-13
<indexterm> elements create whitespace in PDF with FOP


xephon2:
## Expected Behavior

Index terms should not create whitespace.

## Actual Behavior
They create whitespace.
![screenshot_2018-03-13_12-35-50](https://user-images.githubusercontent.com/3266567/37339759-03b5a156-26bc-11e8-8d96-03a1fed82646.png)

## Possible Solution
Whitespaces are probably preserved somewhere (see also [FOP-2520](https://issues.apache.org/jira/browse/FOP-2520)), but I could not find out where this happens.

## Steps to Reproduce

1. In **topic-attr.xsl**, set a background-color in the attribute-set **div**:
    ```xml
    <xsl:attribute-set name="div">
        <xsl:attribute name="background-color">red</xsl:attribute>
    </xsl:attribute-set>
    ```
2. In **topic-attr.xsl**, set a background-color in the attribute-set **section.title**:
    ```xml
    <xsl:attribute-set name="section.title" use-attribute-sets="common.title">
        <xsl:attribute name="font-weight">bold</xsl:attribute>
        <xsl:attribute name="space-before">15pt</xsl:attribute>
        <xsl:attribute name="keep-with-next.within-column">always</xsl:attribute>
        <xsl:attribute name="background-color">green</xsl:attribute>
    </xsl:attribute-set>
    ```
3. Download [DITA-Whitespace Massacre.zip](https://github.com/dita-ot/dita-ot/files/1806701/DITA-Whitespace.Massacre.zip).
4. Publish the **map.ditamap**.

## Environment

* DITA-OT version: **2.5.4**
* Operating system and version: **Windows 10**
* How did you run DITA-OT? **dita command**
* Transformation type: **pdf2**
* PDF-Formatter: **FOP**

robander:
Fixed with #3381

Issue #3042, opened 2018-08-09
[PDF] Index terms create extra space between section title and content


raducoravu:
Reported here:
https://www.oxygenxml.com/forum/viewtopic.php?f=20&t=16970&p=48639#p48639

Let's say we have this DITA topic:

```xml
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
<topic id="introduction">
  <title>Introduction</title>
  <body>
    <section>
      <title>test section</title>
      <indexterm>flowers
        <indexterm>summer</indexterm></indexterm>
      <p>some text for the test section</p>
    </section>
  </body>
</topic>
```

which has some indexterms between the title and the paragraph in the section.

When publishing it to PDF using Antenna House or RenderX XEP the following equivalent XSL-FO output is generated:

```xml
<fo:block space-before="0.6em">
  <fo:block font-weight="bold" 
            keep-with-next.within-column="always" 
            space-before="15pt" 
            line-height-shift-adjustment="disregard-shifts" 
            font-family="Helvetica, Arial Unicode MS, Tahoma, Batang, SimSun">test section
  </fo:block>
  <fo:inline index-key="flowers:"/>
  <fo:inline index-key="flowers:summer:"/>
  <fo:block space-after="0.6em" 
            space-before="0.6em" 
            text-indent="0em">some text for the test section
  </fo:block>
</fo:block>
```

Those `<fo:inline index-key="flowers:"/>` add extra spacing when publishing using Antenna House so they should be replaced with `fo:wrapper`.

When publishing using Apache FOP, the constructed XSL-FO already has `fo:wrapper` elements there, but because of a bug in FOP the fo:wrappers also take vertical space, but that's another story: https://github.com/dita-ot/dita-ot/issues/2455

raducoravu:
The XSLT stylesheet "plugins/org.dita.pdf2/xsl/fo/index.xsl' is the place where those fo:inline elements with the @index-key attribute are emitted.

robander:
Sounds like the `fo:inline` is actually preferred in FOP to work around the known issue? In that case maybe the default should use `fo:wrapper` but the FOP plugin should override that to keep `fo:inline`.

raducoravu:
[...]

PR #3381, opened 2019-10-11
Use block for index to work around FOP issue #2455

### LINKED ISSUE #2455: Use fo:block instead of fo:wrapper for ID containers
I wrote this already as comment to the pull request #2229. But I noticed that one was already closed/merged. So here a new issue:

Using FOP I have additional space before a title and I could identify the empty fo:wrapper elements to cause this.

My sample fo code looks like this (the spaces for indention were added by hand to make it easier to read and are not present in the actual code):

```
<fo:block border="solid blue">
    <fo:marker marker-class-name="current-h2">Title</fo:marker>
    <fo:wrapper id="unique_3"/>
    <fo:wrapper id="_OPENTOPIC_TOC_PROCESSING_d89e342"/>
    <fo:block border="solid red">Title</fo:block>
</fo:block>
```

producing this output:
dita-ot-title-wrapper
So the fo:wrapper elements generate a blank line.

When I change the fo:wrapper to fo:inline the result doesn't change.

When I change the fo:wrapper to fo:block the additional space disappears:
dita-ot-title-block

So I think at least for FOP fo:block should be used. I can't test it with XEP or AH, though.

Maybe there could be a named template "idContainer" with different overriding for different engines?

However, when using fo:block the @keep-with-next behavior of two consecutive titles will no more work since the first title will only be kept together with the (empty) id container block. So an additional keep-together.within-column="always" for the while title element is required - which should have no negative side effects...

robander:

---
name: Pull request

---

## Description

Similar to #3379 but with focus on FOP, and more limited in scope.

Index terms use `fo:wrapper` to create points the index can link to.

In FOP, this causes extra space in a lot of contexts, as mentioned in both #3042 and #2455, due to a known bug in FOP. That bug can be worked around _in certain contexts_ by using `fo:block`, which does not add space. However, in other contexts, it will not remove the extra space, and may even cause new line breaks.

This fix focuses on index terms at the start of `<section>` or `<example>` elements, which I think is quite a common practice, and the specific place reported in both of those issues. I can also verify that the fix works in this location. If we want to extend this to other locations, we need to do so selectively, because it's easy to cause problems by adding new `fo:block` elements (for example, when one appears between two sentences in a paragraph).

Specifically, it will use `fo:block` for index anchors when:

- Parent of the primary index entry is section or example
- Only elements before the primary entry are title, other index terms, or the DITAVAL flagging content
- All text nodes before the primary entry are empty

I made this into a function to make all of that testing easier to follow. I also anticipate that we may want to add other conditions in the future that use `fo:block`, and having that long test in a function will make it easier to understand when we add more.

## Motivation and Context

Fixes #2455

## How Has This Been Tested?

Sample attached.
[3042.zip](https://github.com/dita-ot/dita-ot/files/3719623/3042.zip)

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.pdf2.fop/xsl/fo/index_fop.xsl
+++ b/src/main/plugins/org.dita.pdf2.fop/xsl/fo/index_fop.xsl
@@ -38,9 +38,42 @@ See the accompanying LICENSE file for applicable license.
 
     <xsl:apply-templates/>
   </xsl:template>
+  
+  <!-- At start of a section, fo:wrapper will add extra line breaks, but fo:block does not.
+          Workaround for FOP issue: https://issues.apache.org/jira/browse/FOP-2016
+  Use fo:block instead of fo:wrapper when:
+  * In a section or example
+  * No preceeding elements in the section other than title, other index terms, or DITAVAL flags
+  * No preceding text nodes with text -->
+  <xsl:function name="opentopic-index:use-block-in-section" as="xs:boolean">
+    <xsl:param name="ctx" as="element()"/>
+    <xsl:choose>
+      <xsl:when test="empty($ctx/ancestor::opentopic-index:index.entry[last()]/
+        parent::*[contains(@class,' topic/section ') or contains(@class,' topic/example ')])">
+        <xsl:sequence select="false()"/>
+      </xsl:when>
+      <xsl:when test="exists($ctx/ancestor::opentopic-index:index.entry[last()]/
+        preceding-sibling::*[not(self::opentopic-index:index.entry or contains(@class,' topic/title ') or contains(@class,' ditaot-d/ditaval-startprop '))])">
+        <xsl:sequence select="false()"/>
+      </xsl:when>
+      <xsl:when test="exists($ctx/ancestor::opentopic-index:index.entry[last()]/preceding-sibling::text()[normalize-space(.)!=''])">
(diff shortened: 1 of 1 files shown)


infotexture:
Also resolves https://github.com/dita-ot/dita-ot/issues/2914.

src/main/plugins/org.dita.pdf2.fop/xsl/fo/index_fop.xsl
```diff
@@ -38,9 +38,42 @@ See the accompanying LICENSE file for applicable license.
...
+  <!-- At start of a section, fo:wrapper will add extra line breaks, but fo:block does not.
+          Workaround for FOP issue: https://issues.apache.org/jira/browse/FOP-2016
+  Use fo:block instead of fo:wrapper when:
+  * In a section or example
+  * No preceeding elements in the section other than title, other index terms, or DITAVAL flags
+  * No preceding text nodes with text -->
+  <xsl:function name="opentopic-index:use-block-in-section" as="xs:boolean">
+    <xsl:param name="ctx" as="element()"/>
+    <xsl:choose>
+      <xsl:when test="empty($ctx/ancestor::opentopic-index:index.entry[last()]/
```
jelovirt: `$ctx/ancestor::opentopic-index:index.entry[last()]` could be saved to a variable so we don't need to query it for each when-branch.

PR #3382, opened 2019-10-13
Add support to interleave foreign content to XML project files

jelovirt:
## Description
Add support to interleave elements and attributes from foreign namespaces in project XML files.

## Motivation and Context
This allows custom elements and attributes in project files for e.g. when a single CMS export files contains both DITA-OT project configuration and CMS metadata.

## How Has This Been Tested?
Reader tests have been extended to include foreign content.

## Type of Changes

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -442,7 +442,9 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
             throw new BuildException("");
         }
         try {
-            return ProjectFactory.load(args.projectFile.toURI());
+            final ProjectFactory factory = ProjectFactory.getInstance();
+            factory.setLax(true);
+            return factory.load(args.projectFile.toURI());
         } catch (Exception e) {
             printErrorMessage(e.getMessage());
             throw new BuildException("");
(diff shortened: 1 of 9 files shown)


raducoravu:
Moving @georgebina 's feedback from Slack:

>  Other formats allow extensions using annotations, like Schematron or Relax NG itself, but they do not change the root element. 
>  The lax schema even forces a different root element, it flags a normal project as invalid, which is wrong in my opinion. 
>  The start is a foreign element and its content matches  “anything” so basically all the validation is skipped, so all the schema fragments that are there are not used.

raducoravu:
Probably the project.rnc could have been made a little bit more flexible in accepting foreign elements and attributes.
But indeed the 
Indeed the "project-lax.rnc" seems too lax, maybe it can be removed completely and the validation could be done individually in the Java code on each "project" element from inside the XML document using the "project.rnc".

raducoravu:
Also after I built a DITA OT from the sources, publishing a regular project file (which has the root "project" in that particular namespace) does not work anymore:

         [exec] Error: Failed to read project file: org.xml.sax.SAXParseException; systemId: file:/D:/projects/eXml/samples/dita/mobile-phone/mobilePhoneProjectFile.xml; lineNumber: 2; columnNumber: 50; element "project" not allowed here

PR #3390, opened 2019-10-23
Add extension point for PDF2 index processing

jelovirt:
Add extension point `depend.org.dita.pdf2.index` to PDF2. This allows adding custom index processing targets to PDF2. In order to disable to default index processing, a new property `pdf2.index.skip` is added. When a custom index target is used, the contract for the extension point requires that the output is written into `stage1.xml` in the temporary directory for backwards compatibility.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/build_template.xml
+++ b/src/main/plugins/org.dita.pdf2/build_template.xml
@@ -172,7 +172,9 @@ with those set forth herein.
   </target>
 
   <target name="transform.topic2fo"
-          depends="transform.topic2fo.init,
+          dita:extension="depends org.dita.dost.platform.InsertDependsAction"
+          dita:depends="transform.topic2fo.init,
+                   {depend.org.dita.pdf2.index},
                    transform.topic2fo.index,
                    transform.topic2fo.main,
                    transform.topic2fo.i18n"/>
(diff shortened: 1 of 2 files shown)


src/main/plugins/org.dita.pdf2/build_template.xml
```diff
@@ -172,8 +172,9 @@ with those set forth herein.
   </target>
 
   <target name="transform.topic2fo"
-          depends="transform.topic2fo.init,
-                   transform.topic2fo.index,
+          dita:extension="depends org.dita.dost.platform.InsertDependsAction"
+          dita:depends="transform.topic2fo.init,
+                   {depend.org.dita.pdf2.index},
```
jelovirt: Moved `transform.topic2fo.index` to be used via the `depend.org.dita.pdf2.index` extension point.

src/main/plugins/org.dita.pdf2/build_template.xml
```diff
@@ -242,7 +243,15 @@ with those set forth herein.
...
-  <target name="transform.topic2fo.index">
+  <target name="transform.topic2fo.index.init">
+    <condition property="_pdf2.index.skip" value="true">
+      <istrue value="${pdf2.index.skip}"/>
+    </condition>
+  </target>
+
+  <target name="transform.topic2fo.index"
+          depends="transform.topic2fo.index.init"
+          unless="_pdf2.index.skip">
```
jelovirt: Added `_pdf2.index.skip` property so we can accept truthy value for the skip property. This way the value can be falsy and the target will not be skipped.

PR #3391, opened 2019-10-23
Filter see index links that have no destination

jelovirt:
See index links that have no destination in create broken `fo:basic-link` elements because the `internal-destination` attribute will be an empty string. By filtering out all links where the destination is empty, the broken links are removed from output.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
@@ -256,7 +256,7 @@ See the accompanying LICENSE file for applicable license.
                         <xsl:with-param name="id" select="'Index See String'"/>
                     </xsl:call-template>
                 </fo:inline>
-                <xsl:for-each select="opentopic-index:index.entry">
+                <xsl:for-each select="opentopic-index:index.entry[opentopic-index:get-see-destination(.)]">
                     <xsl:if test="position() ne 1">
                         <xsl:text>, </xsl:text>
                     </xsl:if>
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/index.xsl
```diff
@@ -256,7 +256,7 @@ See the accompanying LICENSE file for applicable license.
                         <xsl:with-param name="id" select="'Index See String'"/>
                     </xsl:call-template>
                 </fo:inline>
-                <xsl:for-each select="opentopic-index:index.entry">
+                <xsl:for-each select="opentopic-index:index.entry[opentopic-index:get-see-destination(.)]">
```
robander: If there are no valid See destinations, does that mean we will end up with an empty _See_ in the result? Same for below with _See also_.

That may be OK, because this is basically an error condition, but it might be good to put an error in the log that there were see/see-also entries but none were valid? Or a more general "Dropping see/see-also from X to Y because Y is not defined"?

jelovirt: The validation should really be done in the index processing code, not this late in the chain. This is a quick fix to allow content to be processed in FOP, because FOP will crash on this type of bug.

jelovirt: The validation should really be done in the index processing code, not this late in the chain. This is a quick fix to allow content to be processed in FOP, because FOP will crash on this type of bug.

robander: OK - works for me, and agree it should be done earlier.

infotexture: > it might be good to put an error in the log that there were see/see-also entries but none were valid

We have `DOTJ050W`:

> Found an `<index-see>` or `<index-see-also>` reference to the term '%1', but that term is not defined in the index.

PR #3394, opened 2019-10-27
Remove lax project schema and instead filter foreign content on lax mode

jelovirt:
Remove lax project schema and instead filter foreign content on lax processing. This allows using the same schema for all project files.


jelovirt:
> On the "filterAttributes" you could avoid creating new attributes list instance if all attributes are in no namespace (which is most common scenario).

jelovirt:
> What is the purpose of the "org.dita.dost.project.XmlReader.lax" flag if the reader in practice works only in lax mode?

Because this code can be used in other context too, I wanted to allow strict processing. The `dita` command can be a bit more lax about it, but maybe some use case wants to be strict 🤷‍♀

> Instead of creating stream to project.rnc "getResourceAsStream("project.rnc")" why not get an URL to the resource "URL resource = getResource("project.rnc")" and pass it to the validator's input source? In this way the validator is responsible of opening and closing stream plus the RNC could in time import other RNCs and relative references inside the RNC would be properly resolved.

In this case I prefer to handle the IO because I don't know what validator is doing with a URI. Also, the schema file is inside a JAR, so it's likely/possible that the validator would not be able to read it just based on a URI.

> Ideally we would separately catch SaxParseException's and use their toString() or line/column information to enrich the error message.

That's a good idea.

raducoravu:
I also built and tested again the distribution and things still seem to work alright.

PR #3397, opened 2019-10-30
Move hard-coded indexshow styles to Sass partial

infotexture:
The inline style attributes prevent users from customizing the presentation of index terms when these are emitted to HTML5 output with `args.indexshow=yes`.

These changes move the default presentation rules to CSS to allow users to override presentation in custom stylesheets.

The output is visually equivalent to the results generated by previous toolkit versions.


infotexture:
**Note:** The same inline styles are duplicated in `org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl` and `org.lwdita/xsl/dita2markdownImpl.xsl`, but I assume we'd want to leave XHTML alone, and I'm not sure whether the LwDITA variant would pick up the styling from the HTML5 CSS files.

PR #3408, opened 2019-11-07
Update LwDITA plugin to 2.3.2

jelovirt:
Update bundled LwDITA plugin to version 2.3.2.


infotexture:
Also need to remember to publish this latest version to the plug-in registry.

Issue #3395, opened 2019-10-29
Treatment for UncheckedXPathException exceptions


raducoravu:
I'm converting a DITA topic to Word using the "com.elovirta.ooxml" plugin.
The transformation fails with:

    D:\projects\eXml\frameworks\dita\DITA-OT3.x\plugins\org.dita.base\build.xml:29: The following error occurred while executing this line:
    D:\projects\eXml\frameworks\dita\DITA-OT3.x\plugins\com.elovirta.ooxml\build.xml:72: net.sf.saxon.trans.UncheckedXPathException: Supplied element name is a zero-length string
	at net.sf.saxon.lib.DelegatingErrorListener.fatalError(DelegatingErrorListener.java:51)
	at net.sf.saxon.Controller.reportFatalError(Controller.java:466)
	at net.sf.saxon.trans.XsltController.applyTemplates(XsltController.java:775)
	at net.sf.saxon.s9api.AbstractXsltTransformer.applyTemplatesToSource(AbstractXsltTransformer.java:347)
	at net.sf.saxon.s9api.XsltTransformer.transform(XsltTransformer.java:349)
	at net.sf.saxon.jaxp.TransformerImpl.transform(TransformerImpl.java:71)
	at org.dita.dost.module.TopicMergeModule.execute(TopicMergeModule.java:99)
	at org.dita.dost.ant.ExtensibleAntInvoker.execute(ExtensibleAntInvoker.java:182)

We need to somehow individually catch such "UncheckedXPathException " exceptions and display instead their inner "net.sf.saxon.trans.UncheckedXPathException.getXPathException()" which probably has line and column information missing in the wrapper runtime exception.

raducoravu:
Added also an issue on the Saxonica list related to unchecked exceptions:
https://saxonica.plan.io/issues/4364

raducoravu:
The HTML5 output works but it has its quirks, the output folder contains both an "index.html" which is the actual topic with all keyrefs and conkeyrefs expanded (but which has a broken reference to the image file). There is also an HTML equivalent of the original topic in a "topics" subfolder and in this case the image reference is properly resolved.

raducoravu:
Also I've been having problems obtaining plain PDF output from a simple DITA Map, something like:

          dita -i D:\projects\eXml\samples\dita\flowers\flowers.ditamap -f pdf -o "d:\outputTest3" -t "d:\abcTMP3"  -d -v

[...]

PR #3412, opened 2019-11-16
Support for addtional input resources

jelovirt:
## Description
Add support for additional input resources that are referenced as resource-only.

## Motivation and Context
This allows processing e.g. individual topics with key references without needing to create a wrapper map that defines the keys.

## Example

`topic.dita`
```xml
<topic id="topic">
  <title>Installing <keyword key="product"/></title>
  …
</topic>
```

`keys.ditamap`
```xml
<map>
  <keydef keys="product">
    <topicmeta>
      <keywords>
        <keyword>Product</keyword>
      </keywords>
    </topicmeta>
  </keydef>  
</topic>
```

When processed with

```bash
dita -i topic.dita -r keys.ditamap -f html5
```

Will result 

> **Installing Product**
> …

### DIFF
--- a/src/main/java/org/dita/dost/exception/DITAOTException.java
+++ b/src/main/java/org/dita/dost/exception/DITAOTException.java
@@ -8,6 +8,8 @@
  */
 package org.dita.dost.exception;
 
+import net.sf.saxon.trans.UncheckedXPathException;
+import net.sf.saxon.trans.XPathException;
 import org.dita.dost.log.MessageBean;
 
 /**
@@ -29,7 +31,7 @@ public final class DITAOTException extends Exception {
      * as its detail message.
      */
     public DITAOTException() {
-        this(null, null);
+        this(null, (Throwable) null);
     }
 
     /**
@@ -38,7 +40,7 @@ public final class DITAOTException extends Exception {
      * @param message the detail message
      */
     public DITAOTException(final String message) {
-        this(message, null);
+        this(message, (Throwable) null);
     }
 
     /**
@@ -59,6 +61,26 @@ public final class DITAOTException extends Exception {
      */
     public DITAOTException(final String message, final Throwable cause) {
         super(message, cause);
+    }
+
+    /**
+     * Constructs a new exception with cause, using location information from a root {@link XPathException}
+     *
+     * @param cause the cause
+     */
+    public DITAOTException(final UncheckedXPathException cause) {
+        super(cause.getXPathException().getMessageAndLocation(), cause.getXPathException());
+    }
+
+    /**
+     * Constructs a new exception with the specified detail message and
+     * cause, using location information from a root {@link XPathException}
+     *
+     * @param message the detail message.
+     * @param cause the cause
+     */
+    public DITAOTException(final String message, final UncheckedXPathException cause) {
+        super(message + ": " + cause.getXPathException().getMessageAndLocation(), cause.getXPathException());
     }
 
     /**
--- a/src/main/java/org/dita/dost/invoker/ArgumentParser.java
+++ b/src/main/java/org/dita/dost/invoker/ArgumentParser.java
@@ -60,6 +60,8 @@ final class ArgumentParser {
         ARGUMENTS.put("--deliverable", new StringArgument("project.deliverable"));
         ARGUMENTS.put("-i", new FileOrUriArgument("args.input"));
         ARGUMENTS.put("--input", new FileOrUriArgument("args.input"));
+        ARGUMENTS.put("-r", new FileOrUriArgument("args.resources"));
+        ARGUMENTS.put("--resource", new FileOrUriArgument("args.resources"));
         ARGUMENTS.put("-o", new AbsoluteFileArgument("output.dir"));
         ARGUMENTS.put("--output", new AbsoluteFileArgument("output.dir"));
         ARGUMENTS.put("--filter", new AbsoluteFileListArgument("args.filter"));
@@ -178,6 +180,7 @@ final class ArgumentParser {
     private String uninstallId;
 
     private List<String> inputs = new ArrayList<>();
+    private List<String> resources = new ArrayList<>();
 
     /**
      * The build targets.
@@ -303,6 +306,8 @@ final class ArgumentParser {
                 handleArgNice(args);
             } else if (isLongForm(arg, "-input") || arg.equals("-i")) {
                 handleArgInput(arg, args, ARGUMENTS.get(getArgumentName(arg)));
+            } else if (isLongForm(arg, "-resource") || arg.equals("-r")) {
+                handleArgResource(arg, args, ARGUMENTS.get(getArgumentName(arg)));
             } else if (ARGUMENTS.containsKey(getArgumentName(arg))) {
                 definedProps.putAll(handleParameterArg(arg, args, ARGUMENTS.get(getArgumentName(arg))));
             } else if (getPluginArguments().containsKey(getArgumentName(arg))) {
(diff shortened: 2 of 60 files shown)


jelovirt:
Example of cases where this will not work as expected. If we have `root.ditamap`

```xml
<map keyscope="root">
  <mapref href="keys.ditamap">  
  <topicref href="topic.dita"> 
</topic>
```

In `topic.dita` we reference a key in the scope `root`

```xml
<topic id="topic">
  <title>Installing <keyword key="root.product"/></title>
  …
</topic>
```

If we try to convert the topic alone with keys as the resource

```bash
dita -i topic.dita -r keys.dita …
```

We will get an error message

> [DOTJ047I][INFO] Unable to find key definition for key reference "root.key" in root scope…

This is because the root map uses the scope `root` and that's what the topic expects, but the key map just inherits it. This is not a bug in the code, but rather just a limitation. This could be fixed by adding yet another property, like `root-key-keyscope`. That would allow you to manually change the key scope all scopes will be added to. So if you set `--root-key-keyscope=root` for the case above, that would work. But I'm not sure if it has any other real use cases, because I don't know how people really use key scopes in the wild.

raducoravu:
So the result is as if the entire context keys map is referenced with processing role resource only from another map which also includes the published topic? That’s clever. Does this work with both preprocess and preprocess2 targets? 
Will try to build off the branch and look more into the code.
From the list of possible types of key refs:

Keyword with Keyref,
Image with keyref
Conkeyref
Xref with keyref

Probably the last one does not work, especially in a PDF but I would not expect it to.
Curious about what is shown in place of link though...

jelovirt:
It should work fine with any type of keyref usage, since after the key is resolved, other preprocess steps handle the rest. Like conrefs or link texts.

robander:
Curious about a couple of things, and I don't know the answer to all of them:

- If you have multiple input resources, which key will win? I'm assuming they are evaluated in order (so if the first and second resource maps define `keys="example"`, the first one would win)?
- That might get uglier if you have to go into submaps to get keys, due to the evaluation rules about depth-of-definition. For example, if your input resources are `first.ditamap` and `second.ditamap`, where `first` defines after several levels of nested maps, and `second` defines the same key directly.
- If the key resolves to a relative path, what do you evaluate that against? The usual DITA-OT issue with paths, if the resource and input topic are not in the same directory, where does the path go?
- Can I use a topic as an input resource, or just maps? (Not sure what the reason for using a topic would be, since we will already pull in `conref` and link targets as needed.)

raducoravu:
The "args.resources" new param probably needs to be added to the base plugin.xml as a new param. In this way Oxygen will be able to show it in the Parameters list for the transformation scenarios.

I built and stared using the DITA OT with the changes on a topic which has image with keyref, plain keyref, link with keyref and conkeyref.
I'm attaching a sample zip below:
[testSingleTopicOutput.zip](https://github.com/dita-ot/dita-ot/files/3862310/testSingleTopicOutput.zip)

When publishing with plain HTML5 the conkeyref is the only one not getting resolved for some reason, the target for it seems to be searched for in the wrong folder.

I also published a topic from the Oxygen's user manual DITA project which contained only keyrefs to the "product" key. The "branch-filter, keyref, move-meta-entries" stages are lightweight and performed only on the published topic, all other steps including the "html5.topics.common.inner" and "image-metadata" seem to be performed on all topics.

jelovirt:
Adding `args.resources` to plugin.xml means we also have to add support for parameter type `files` or `path`, because `args.resources` is a list of files delimited by path separators.

jelovirt:
In order to only get HTML output for the single file in @raducoravu example, we need to dynamically add `processing-role="resource-only"` to resource map root.

jelovirt:
@raducoravu in your testSingleTopicOutput.zip, are you running `topics/install-eclipse.dita` as the input file and `simple.ditamap` as a resource?

raducoravu:
@jelovirt I'm actually publishing "test_topic.dita" with the "simple.ditamap" set as additional resource, sorry that I did not state this from the beginning.

jelovirt:
Fixed conkeyref resolution.

jelovirt:
@raducoravu, are you able to test this again? I've been testing with your testSingleTopicOutput.zip and that seems to work now. Would appreciate a bit more testing before merging this.

raducoravu:
I'm running this command line:

          dita -i C:\Users\radu_coravu\Downloads\testSingleTopicOutput\testSingleTopicOutput\topics\test_topic.dita -f pdf -o "d:\outputTest" --args.resources=C:/Users/radu_coravu/Downloads/testSingleTopicOutput/testSingleTopicOutput/simple.ditamap

and I get this error:

         Error: net.sf.saxon.trans.UncheckedXPathException: Supplied element name is a zero-length string

which is similar to what I reported here: https://github.com/dita-ot/dita-ot/issues/3395
and again it's hard to get the stack trace of the error or the template where this occurs.

jelovirt:
@raducoravu added catches for unchecked XPath exception. Are you able to test again, because I'm not able to reproduce that issue.

raducoravu:
@jelovirt added your latest changes, compiled the project. Now I cannot reproduce the "Supplied element name is a zero-length" when generating the PDF. Generation works, PDF looks good inside, keyrefs, conkeyrefs expanded. So I don't really undestand why this did not work for me before.

raducoravu:
I've been having problems obtaining plain PDF output from a simple DITA Map, something like:

          dita -i D:\projects\eXml\samples\dita\flowers\flowers.ditamap -f pdf -o "d:\outputTest3" -t "d:\abcTMP3"  -d -v

no longer works on a DITA Map which has a plain topic referenced in it, I get a "topic.fo" which is invalid, contains nothing in the "xsl-region-body" flow:

          [fop] [ERROR] net.sf.saxon.trans.XPathException: org.apache.fop.fo.ValidationException: "fo:flow" is missing child elements. Required content model: marker* (%block;)+ (See position 1:10681)
      [fop] [ERROR] Anttask - Error rendering fo file: d:\abcTMP3\topic.fo <org.apache.fop.apps.FOPException: org.apache.fop.fo.ValidationException: "fo:flow" is missing child elements. Required content model: marker* (%block;)+ (See position 1:10681)

raducoravu:
The HTML5 output on the topic from the original sample project works but it has its quirks, the output folder contains both an "index.html" which is the actual topic with all keyrefs and conkeyrefs expanded (but which has a broken reference to the image file). There is also an HTML equivalent of the original topic in a "topics" subfolder and in this case the image reference is properly resolved.

jelovirt:
PDF topic still has issues.

raducoravu:
@jelovirt also not sure if you've seen my comment above about regular DITA Map to PDF not working anymore: https://github.com/dita-ot/dita-ot/pull/3412#issuecomment-561512604

jelovirt:
@raducoravu PDF works now, either with or without resources.

raducoravu:
I  can confirm things now work out on the sample project I initially tested with.
Then I went back to the Oxygen user's manual and tried to publish a topic to PDF, it seems that if you have submaps things no longer work that well.
For example the zip below has a submap which refers to topics and defines keys. The conkeyref inside the "test_topic.dita" no longer resolves on my side. Neither the keyref link from the test_topic.dita to the other topic.
[testSingleTopicOutputReusablesSubmap.zip](https://github.com/dita-ot/dita-ot/files/3938282/testSingleTopicOutputReusablesSubmap.zip)

jelovirt:
@raducoravu with the submap example, what did you convert with what arguments?

raducoravu:
@jelovirt I run this command line:

      dita -i C:\Users\radu_coravu\Downloads\testSingleTopicOutput\topics\test_topic.dita -f pdf -o "d:\outputTest" --args.resources=C:/Users/radu_coravu/Downloads/testSingleTopicOutput/simple.ditamap -d -v

jelovirt:
@raducoravu I created https://github.com/dita-ot/dita-ot/pull/3429 to fix this. The root cause is that when you have topic input file, then we set `noMap`. It was used to disable mapref processing, but now that we can have maps from additional resources, that guard/optimization needs to be relaxed.

Based on my testing this fixes your problem with additional resources with submaps.

PR #3413, opened 2019-11-16
Feature to dynamically adjust result file URI

jelovirt:
Add feature to dynamically specify the result file URI for preprocess2 and optionally for preprocess1.

### DIFF
--- a/src/main/plugins/org.dita.base/plugin.xml
+++ b/src/main/plugins/org.dita.base/plugin.xml
@@ -162,8 +162,8 @@ See the accompanying LICENSE file for applicable license.
       <val>true</val>
       <val default="true">false</val>
     </param>
-    <param name="result.rewrite-rule.xsl" desc="Filename rewrite rules in XSLT" type="string"/>
-    <param name="result.rewrite-rule.class" desc="Filename rewrite rules in Java" type="string"/>
+    <param name="result.rewrite-rule.xsl" desc="Specifies the name of the XSLT file used to rewrite filenames" type="string"/>
+    <param name="result.rewrite-rule.class" desc="Specifies the name of the Java class used to rewrite filenames" type="string"/>
   </transtype>
   <feature extension="dita.image.extensions" value=".gif"/>
   <feature extension="dita.image.extensions" value=".eps"/>
--- a/src/main/plugins/org.dita.base/plugin.xml
+++ b/src/main/plugins/org.dita.base/plugin.xml
@@ -162,6 +162,8 @@ See the accompanying LICENSE file for applicable license.
       <val>true</val>
       <val default="true">false</val>
     </param>
+    <param name="result.rewrite-rule.xsl" desc="Filename rewrite rules in XSLT" type="string"/>
+    <param name="result.rewrite-rule.class" desc="Filename rewrite rules in Java" type="string"/>
   </transtype>
   <feature extension="dita.image.extensions" value=".gif"/>
   <feature extension="dita.image.extensions" value=".eps"/>
--- a/src/main/java/org/dita/dost/module/AbstractPipelineModuleImpl.java
+++ b/src/main/java/org/dita/dost/module/AbstractPipelineModuleImpl.java
@@ -37,7 +37,10 @@ public abstract class AbstractPipelineModuleImpl implements AbstractPipelineModu
         this.job = job;
     }
 
-    abstract public AbstractPipelineOutput execute(AbstractPipelineInput input) throws DITAOTException;
+    @Override
+    public AbstractPipelineOutput execute(AbstractPipelineInput input) throws DITAOTException {
+        return this.execute(input.getAttributes());
+    }
 
     @Override
     public void setFileInfoFilter(Predicate<FileInfo> fileInfoFilter) {
--- a/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
+++ b/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
@@ -113,6 +113,7 @@ public class CleanPreprocessModule extends AbstractPipelineModuleImpl {
         if (useResultFilename) {
             // collect and relativize result
             final Collection<FileInfo> original = job.getFileInfo().stream()
+                    .filter(fi -> fi.result != null)
                     .map(fi -> FileInfo.builder(fi).result(base.relativize(fi.result)).build())
                     .collect(Collectors.toList());
             original.forEach(fi -> job.remove(fi));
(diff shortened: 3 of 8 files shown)


jelovirt:
While the code that actually does the file renaming and link rewriting is simple, we have to come up with a way to define the new file names. Since this cannot be a single property, it has to be some code. Many DITA-OT users already know XSLT, so passing in a stylesheet to do the rewriting could be one option:

```xslt
<xsl:template match="node() | @*">
  <xsl:copy>
    <xsl:apply-template select="node() | @*"/>
  </xsl:copy>
</xsl:template>

<xsl:template match="file[@format = 'image']/@result">
   <xsl:attribute name="{local-name()}" select="concat('images/', .)"/>
</xsl:template>
```

This would move all images to subdirectory `images`.

However, I would prefer to do this in Java or JavaScript, with something like

```java
List<FileInfo> rewrite(final List<FileInfo> fis) {
  return fis.stream()
    .map(fi -> {
      final FileInfo.Builder builder = FileInfo.builder(fi);
      if (fi.format.equals("image")) {
        builder.source("images/" + fi.source)
      }
      return builder.build();
    })
    .collect(Collections.toList())
} 

```

jelovirt:
One option for this would be to initially support XSLT and Java, because those are pretty straightforward to add. Because choosing which rewrite rule to use should customizable at run time (as opposed to by installing the one plugin that implements it), we could have a few properties

* `result.rewrite-rule.xsl` — absolute path to an XSLT stylesheet that can rewrite the `file` elements in a Job configuration file.
* `result.rewrite-rule.class` — Java class that implements a rewrite processor interface that can rewrite `FileInfo` objects' `result` field.

If you specify either one of those, then that is used for the rewrite. If you specify both, that's an error. If you specify none, then the default rewrite rule is used, i.e. the current processing. With this approach, we could add e.g. `result.rewrite-rule.javascript` later on and it would just be another way to do implement the rewrite without breaking backwards compatibility.

The Ant way of doing this would be to use [file name mapper](https://ant.apache.org/manual/Types/mapper.html), but it would require adding an extension point for this and enabling them with [conditional attributes](https://ant.apache.org/manual/ifunless.html). Also, it would only be able to use a single path as the input to the mapping, which might not be enough information to work on. Because we don't want to be more bound to Ant, IMO we could allow mappers, but not really make them the solution we want to build on top of.

raducoravu:
I agree, for Java you will probably need to supply the "FileInfo" maybe through a minimal API interface which will need to remain unchanged from version to version.
After you have the Java extension a plugin could be built with support for what's in general required (flat folder of HTMLs, all images in images folder).

jelovirt:
The code now allows you to use `result.rewrite-rule.class=com.example.CustomRewriteRule` or `result.rewrite-rule.xsl=plugin:com.example:xsl/rewrite.xsl`.

A limitation of `result.rewrite-rule.xsl` is that it's defined as a string in the plugin configuration. It's a string, so we can use values like `plugin:com.example:xsl/rewrite.xsl`. If it were a file type, the users would need to set either an absolute value, or then in Ant using plugin directy properties like `${dita.plugin.com.exampledir}/xsl/rewrite.xsl`, and it would not be possible to use `plugin` URIs.

The current configuration allows you to basically give it any type of path or URI and it will work. But, e.g. in Oxygen people will not have a file selector available when configuring the value int transformation scenarios. I don't see this as a particularly big problem, because this isn't a property that is normally set from the command line, i.e. from "outside". It's normally set by a transtype's initialization target.

jelovirt:
@lionelmoi are you able to test this? Build this branch and run it with `--result.rewrite-rule.xsl=plugin:com.example:xsl/rewrite-rule.xsl` property and add `xsl/rewrite-rule.xsl` stylesheet to `com.example` plugin

```xslt
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
                version="2.0">

  <xsl:template match="node() | @*">
    <xsl:copy>
      <xsl:apply-templates select="node() | @*"/>
    </xsl:copy>
  </xsl:template>

  <xsl:template match="file[@format = 'image']/@result">
      <xsl:attribute name="{local-name()}" select="concat('images/', .)"/>
  </xsl:template>

</xsl:stylesheet>
```

src/main/plugins/org.dita.base/plugin.xml
```diff
@@ -161,6 +161,8 @@ See the accompanying LICENSE file for applicable license.
       <val>true</val>
       <val default="true">false</val>
     </param>
+    <param name="result.rewrite-rule.xsl" desc="Filename rewrite rules in XSLT" type="string"/>
```
infotexture:
```suggestion
    <param name="result.rewrite-rule.xsl" desc="Specifies the name of the XSLT file used to rewrite filenames" type="string"/>
```

src/main/plugins/org.dita.base/plugin.xml
```diff
@@ -161,6 +161,8 @@ See the accompanying LICENSE file for applicable license.
...
+    <param name="result.rewrite-rule.class" desc="Filename rewrite rules in Java" type="string"/>
```
infotexture:
```suggestion
    <param name="result.rewrite-rule.class" desc="Specifies the name of the Java class used to rewrite filenames" type="string"/>
```

lionelmoi:
I tested and it works but only on images with a href. When an image is linked with a keyref it is not processed by the xslt.

image
keyref
image
href
image

I used this map
[flowers.zip](https://github.com/dita-ot/dita-ot/files/3936080/flowers.zip)

My plugin 
[com.example.zip](https://github.com/dita-ot/dita-ot/files/3936081/com.example.zip)

The logs
[log.txt](https://github.com/dita-ot/dita-ot/files/3936082/log.txt)

The result 
[out.zip](https://github.com/dita-ot/dita-ot/files/3936083/out.zip)

The temp dir
[temp20191208110912524.zip](https://github.com/dita-ot/dita-ot/files/3936084/temp20191208110912524.zip)

jelovirt:
@lionelmoi, you hit a feature of DITA-OT that's not a bug, but a side-effect of us trying to do less during processing. You have

```xml
<keydef href="somethingelse/Chrysanthemums.jpg" format="jpg" keys="iris"/>
```

And that sets the format of `Chrysanthemums.jpg` to `jpg`, even if it's used as an image.

-   The rewrite code needs to then match that format too
    
    ```xsl
    <xsl:template match="file[@format = ('image', 'jpg')]/@result">
    ```

-   The other and IMO better solution is to use `image` format in key declarations instead of the format
    
    ```xml
    <keydef href="somethingelse/Chrysanthemums.jpg" format="image" keys="iris"/>
    ```

If you fix the problem with either solution, it should work.

lionelmoi:
Indeed it werked 😄

jelovirt:
Closing this and we can handle code polish and adding functionality in separate PRs.

PR #3429, opened 2019-12-09
Run mapref even if input file is not a map

jelovirt:
Fixes [additional resource related bug](https://github.com/dita-ot/dita-ot/pull/3412#issuecomment-563100812) but also any processing that could be affected by `noMap` guard. The old code assumed that if the input file is a topic, then no map related processing needs to be done. Now that we have additional resource support, the input file can be a topic but we still have maps in the process. Thus `noMap` guards need to be relaxed to fix this. Preprocess2 already has moved on from using `noMap` to allow modules to check whether maps are available. There we need to allow processing maps that are not input files.


raducoravu:
This still does not work for me on the "testSingleTopicOutputReusablesSubmap.zip".
Neither the conkeyref nor the link with keyref get expanded.

       dita -i C:\Users\radu_coravu\Downloads\testSingleTopicOutputReusablesSubmap\testSingleTopicOutputReusablesSubmap\topics\test_topic.dita --args.resources=C:\Users\radu_coravu\Downloads\testSingleTopicOutputReusablesSubmap\testSingleTopicOutputReusablesSubmap\simple.ditamap --output.dir=d:\someFolder -f pdf

jelovirt:
Now PDF2 should, meaning both preprocess and preprocess2 should work.

`xref` with keyref doesn't show the target title in PDF because it's converting only a single topic and the merged file only contains the single topic. Thus the title of the target is not available and PDF2 does its own link title pull processing.

This is something PDF has always had and to fix it we would have to change PDF2. I'm not sure if I want to go there, because then we would have to start adding linked content to merged file. This is basically the same problem as if you had a map you convert to PDF, but then you don't have a topicref to a topic in map, only an xref from a topic. The link will not work nor have pulled link text.

raducoravu:
👍  I can confirm this works better both on the small sample and on a bigger topic from the Oxygen user's guide, filtering also seems to properly work as well. On the real-world example lots of topics (probably directly and indirectly referenced) end up being processed along the way in various stages so the processing is not very fast but at least the final result looks Ok.

src/main/java/org/dita/dost/module/MaprefModule.java
```diff
@@ -57,10 +57,6 @@ private void init(final AbstractPipelineInput input) {
         } catch (TransformerConfigurationException e) {
             throw new RuntimeException("Failed to compile " + styleFile + ": " + e.getMessageAndLocation(), e);
         }
-
-        if (fileInfoFilter == null) {
-            fileInfoFilter = fileInfo -> fileInfo.format != null && fileInfo.format.equals(ATTR_FORMAT_VALUE_DITAMAP);
-        }
```
jelovirt: This is moved to a later stage, no need to do this in module init.

src/main/java/org/dita/dost/module/reader/AbstractReaderModule.java
```diff
@@ -845,16 +845,31 @@ private String getPrefix(final File relativeRootFile) {
...
+                }
+                i = b.build();
+            } else {
+                i = new FileInfo.Builder()
+                        .src(f)
+                        .uri(tempFileNameScheme.generateTempFileName(f))
+                        .build();
+            }
+            fileInfos.put(i.src, i);
+            return i;
```
jelovirt: Use current Job to check if we have already encountered this file in a _previous module_. The old code assume e.g. topic reader could not encounter maps.

src/main/java/org/dita/dost/writer/MapCleanFilter.java
```diff
@@ -46,7 +44,7 @@ public void startElement(final String uri, final String localName, final String
             stack.addFirst(Keep.DISCARD_BRANCH);
         } else if (SUBMAP.matches(cls)) {
             stack.addFirst(Keep.DISCARD);
-        } else if (MAPGROUP_D_KEYDEF.matches(cls)) {
+        } else if (DITA_OT_D_KEYDEF.matches(cls)) {
```
jelovirt: Fixing a bug that no one has found thus far.

src/test/java/org/dita/dost/AbstractIntegrationTest.java
```diff
@@ -683,7 +683,7 @@ public void messageLogged(BuildEvent event) {
                     // out.println(event.getMessage());
                     break;
                 default:
-                    err.println(message);
+//                    err.println(message);
```
jelovirt: No need to log errors in CI, it doesn't help at all.

src/main/java/org/dita/dost/util/XMLUtils.java
```diff
@@ -690,42 +697,16 @@ public static String escapeXML(final char[] chars, final int offset, final int l
...
-                    Class.forName(SAX_DRIVER_CRIMSON_CLASS);
-                    reader = XMLReaderFactory.createXMLReader(SAX_DRIVER_CRIMSON_CLASS);
-                } catch (final ClassNotFoundException exc) {
-                    reader = XMLReaderFactory.createXMLReader();
-                }
-            }
-        }
-        if (Configuration.DEBUG) {
-            reader = new DebugXMLReader(reader);
+            final XMLReader reader = saxParserFactory.newSAXParser().getXMLReader();
```
jelovirt: We use Xerces, no point having this chooser anymore.

Issue #2957, opened 2018-05-11
Output character not available in this encoding - when file name has certain character


raducoravu:
Tested with DITA OT 3.0.4. 
Originally reported here:

https://groups.yahoo.com/neo/groups/dita-users/conversations/topics/43544

So if a DITA map refers to a file name which has the character "ﬁ" in its name the publishing fails with something like:

            C:\Users\radu_coravu\Desktop\DITA OT newest\dita-ot-3.0.4\plugins\org.dita.base\build_init.xml:42: Fatal error during transformation using C:\Users\radu_coravu\Desktop\DITA OT newest\dita-ot-3.0.4\xsl\job-helper.xsl: Output character not available in this encoding (xfb01); SystemID: file:/C:/Users/radu_coravu/Desktop/DITA%20OT%20newest/dita-ot-3.0.4/xsl/job-helper.xsl; Line#: 101; Column#: 5

The XSLT stylesheet "job-helper.xsl" does not specify an encoding to write the text output and on Windows the default platform encoding is not UTF-8 (although it is UTF-8 for Linux and Mac), it's more restrictive. So maybe we can force specify an "UTF-8" encoding on it's xsl:output.

raducoravu:
👍 One more of our end users encountered this problem:
https://www.oxygenxml.com/forum/post47819.html#p47817

jelovirt:
Because the file lists are for backwards support in Ant, we should have a way to disable their generation and make sure we really don't rely on those files anywhere anymore.

raducoravu:
How exactly were the old ANT tasks reading the properties files?
The ANT "loadproperties" task has an attribute to specify the encoding, otherwise it's probably the platform encoding, UTF-8 on Linux/Mac and probably CP1252 on Windows (depending possibly on the region where the installation is performed and the used Windows kit, the language packs it has).

raducoravu:
One more similar report from one of our clients:

     [31mError: The following error occurred while executing this line:
       ....dita-ot/plugins/org.dita.base/build_init.xml:39: Fatal error during transformation using ...dita-ot/plugins/org.dita.base/xsl/job-helper.xsl: 
      Output character not available in this encoding (x2014); SystemID: file:/..dita-ot/plugins/org.dita.base/xsl/job-helper.xsl; Line#: 101; Column#: 5[0m

At a glance it seems one of her files uses an em-dash instead of a plain "-" in the file name.

raducoravu:
[...]

PR #3434, opened 2019-12-28
Remove unused and deprecated preprocessing configuration

jelovirt:
## Description
Remove unused and deprecated preprocess and preprocess2 Ant configuration.

## Motivation and Context
The code is no longer in use and in some cases has been deprecated multiple versions ago.

## Changes to documentation

Migration guide _may_ list the properties and files that are no longer being generated and how to recreate them in custom plug-ins. However, the properties and files are no longer being generated because alternative and better ways have been taken into use to achieve the same, namely the `ditafileset` to choose files to process.


robander:
@infotexture I think our migration notes for other releases already mention getting off of list files -- and it's in our best practices for plugins -- but the migration guide for this should also note that it is a "must do" update for those who might have continued relying on the deprecated list files.

raducoravu:
If the "job-helper.xsl" will not be necessary anymore probably this issue can be closed as well #2957

jelovirt:
@raducoravu I think we will keep `job-helper.xsl`, in case people need it. They can then recreate the list files with just few lines of Ant configuration, instead of needing to reimplement them from scratch.

PR #3438, opened 2020-01-03
Implement configurable label/title/description separator strings

chrispy-snps:
Signed-off-by: Chris Papademetrious <chrispy@synopsys.com>

---
name: Pull request

---

## Description
This is a proposed change to make the ". " separator between figure/table title labels, <title> text, and <desc> elements configurable via a string.

## Motivation and Context
Our content uses a colon to separate figure and table titles:

Figure 1: Some Figure
Table 1: Some Table

After creating a plugin to implement this single-character change, I realized it would be more user-friendly to configure this via a single configurable string (in org.dita.base/xsl/common/strings-common.xml). For example,

  <str name="label-separator">: </str>
  <str name="desc-separator"> -- </str>

## How Has This Been Tested?
I created a testcase to confirm that the default HTML5 and XHTML output remains unchanged, then I reran it with a different string values to confirm that the output was adjusted as desired.

## Type of Changes
This new feature was implemented by identifying the places where the title content was generated, then changing the hardcoded text strings to equivalent variables.

## Checklist
Please let me know if unit tests are needed for this type of enhancement. (I would need to learn how to add and run them.) Thanks!

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/common/strings-common.xml
+++ b/src/main/plugins/org.dita.base/xsl/common/strings-common.xml
@@ -8,6 +8,7 @@ See the accompanying LICENSE file for applicable license.
 -->
 <strings xml:lang="">
   <str name="figure-number-separator"> </str>
+  <str name="label-separator">. </str>
   <str name="OpenQuote">“</str>
   <str name="CloseQuote">”</str>
   <str name="ColonSymbol">:</str>
--- a/src/main/plugins/org.dita.base/xsl/common/strings-common.xml
+++ b/src/main/plugins/org.dita.base/xsl/common/strings-common.xml
@@ -7,12 +7,22 @@ Copyright 2017 IBM Corporation
 See the accompanying LICENSE file for applicable license.
 -->
 <strings xml:lang="">
+
+  <!-- "Figure 1. Figure Title. Description here"
+              ^ ^^            ^^
+              | ||            ||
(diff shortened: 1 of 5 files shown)


robander:
Personally - I have rather mixed feelings on this. I think the end result is probably good but it builds on a clunky design. (Admission: that clunky design came from my early code.)

I think with figure/table titles, a better way to go is to use variables like PDF -- which can build up with tokens like the separator, but more importantly, they let you pass in the figure/table number and title text, then determine how to present them as a whole. I like that better than adding several new variables for specific contexts, like `desc.separator`. 

But ... I'm also not sure it's worth going that far. We already have `ColonSymbol` which combines with other text ... I feel the same way about that, it would be better to get a variable that (for example) returns **Note:** rather than one to return *Note** and another to return a colon (or for French, a space and a colon). This basically uses the design. I think ideally both should use the PDF style, which came to the toolkit well after that **Note:** process went into HTML.

So, yeah, vaguely uncomfortable with the clunky design, but I agree it follows the current design in HTML and probably gives the right result.

Thoughts from others?

chrispy-snps:
Hi Robert,

I don’t know the PDF style (I’m new to all this) but if the consensus is it would be preferable to adopt that approach, I’m happy to try to copy that approach too!

I basically just wanted an easier way to configure this than overriding entire templates to change one character.


  *   Chris

PR #3439, opened 2020-01-06
Add in-memory Store implementation

jelovirt:
## Description
Add in-memory `Store` implementation that keeps XML files as s9api's `XdmNode` instances.

## Motivation and Context
Possible performance improvement in some cases.

## How Has This Been Tested?
Currents tests pass.

## Type of Changes

### DIFF
--- a/src/main/java/org/dita/dost/ant/DITAOTCopy.java
+++ b/src/main/java/org/dita/dost/ant/DITAOTCopy.java
@@ -17,10 +17,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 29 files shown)


robander:
I pulled this branch on Friday and ran my largest doc set through it (some amount of filtering, heavy on chunking and conkeyref, not as heavy on other features). Also have a small number of extensions to HTML style and after preprocess. Happy to report that comparing output of the latest `develop` with the output of this branch, the only changes I saw for HTML were around some generated IDs (and those from my own extension that uses `generate-id()` to ensure some elements always get anchors).

robander (approved):
Hard to evaluate as a whole but verified that it works for my large sample case.

Issue #3443, opened 2020-01-22
pdf2: filtering of nested topicrefs runs into 'SXXP0003: Premature end of file' error


skulle:
## Expected Behavior
A nested topic in a submap (refer to topic3.xml in the attached sample) is filtered out by a ditavalref and the audience attribute

## Actual Behavior
The build aborts with an error:
```
[topic-fragment] Processing file:/C:/git/dita-sample-data/dita-ot-bug/temp/e91279b807935ed88f940757f71ca197b1a82a26.xml
[topic-fragment] Error 
[topic-fragment]   SXXP0003: Error reported by XML parser: Premature end of file.
[topic-fragment] Error 
[topic-fragment]   Error reported by XML parser processing
[topic-fragment]   file:/C:/git/dita-sample-data/dita-ot-bug/temp/e91279b807935ed88f940757f71ca197b1a82a26.xml: Premature end of file.
```

## Possible Solution
It seems that the problem is related to a `keys` attribute and the position of the `audience` attribute. 

The attached sample runs through if you remove the unused key k1 (document.ditamap) or if you move the audience attribute from the concept element (topic3.xml) to the topicref element (map.ditamap).

Currently I don't know how to fix the issue but I am looking into it. If anyone has some ideas I am glad to hear them.

## Steps to Reproduce
Please build `document.ditamap` with transtype `pdf2`
[sample.zip](https://github.com/dita-ot/dita-ot/files/4096293/sample.zip)

## Copy of the error message, log file or stack trace
[console.log](https://github.com/dita-ot/dita-ot/files/4096269/console.log)

## Environment

* DITA-OT version: 3.3.3
* Operating system and version: Windows 10 1909
* How did you run DITA-OT?  Ant, via ant-launcher.jar
* Transformation type: pdf2

raducoravu:
The DITA OT never worked too well when you have profiling attributes set on the root element:
https://github.com/dita-ot/dita-ot/issues/3352

Maybe you can remove that 'audience="Internal"' attribute from the root element of the Topic3 and add it on the topicref instead:

            <topicref href="topic3.xml" audience="Internal" />

skulle:
Thx for the quick reply Radu.
Yes I think we are going to move the audience attribute to the topicref. As far as I know it should not have been put there in the first place since our company guidlines prefer to have it on the topicref. Hopefully we can do this for all affected documents, including published documents, lets hope...

raducoravu:
[...]

PR #3445, opened 2020-01-29
Bug 3443

skulle:
---
name: Work in progres, bug-3443
about: Enable filtering via root elements
---

## Motivation and Context

#3443 

## How Has This Been Tested?
Currently i am testing locally with a dita sample map only

## Type of Changes
For map-first pre-processing:
1. map-branch-filter step: Add an additional step to the FilterUtils.needExclude method to check if a filtering attribute exists at the referenced map/topic. Remove filtered maps/topics from the job's file section

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/module/FilterModule.java
+++ b/src/main/java/org/dita/dost/module/FilterModule.java
@@ -55,6 +55,7 @@ final class FilterModule extends AbstractPipelineModuleImpl {
             filterUtils = new FilterUtils(printTranstype.contains(transtype));
         }
         filterUtils.setLogger(logger);
+        filterUtils.setJob(job);
 
         final ProfilingFilter writer = new ProfilingFilter();
         writer.setLogger(logger);
@@ -100,6 +101,8 @@ final class FilterModule extends AbstractPipelineModuleImpl {
                 logger.error("Failed to profile " + file.getAbsolutePath() + ": " + e.getMessage());
             }
         }
+
+        job.removeFiltered();
 
         try {
             job.write();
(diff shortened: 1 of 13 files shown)


.gitignore
```diff
@@ -1,3 +1,14 @@
+#Eclipse
```
jelovirt: You can ignore these in your local `.git/info/exclude`, you don't need to add these to `.gitignore`

skulle: That is practical, i didn't know that feature. I will revert the gitignore changes...

jelovirt:
If the issue is supporting filter cases where the whole topic content is being filtered out, why do we need to changed key resolution code?

skulle:
The key resolution is only changed so far that it checks if a keyref points to a filtered map/topic or if a keydef is filtered out itself. Otherwise dead links could appear for example.

jelovirt:
We already allow dead links, so IMO we should not touch key processing. It's up to the logger to filter out duplicate dead link messages.

jelovirt:
An alternative would be to store unavailable file into into the `Job.FileInfo` metadata, so we know not to try to read them. But, it should be done in a separate PR.

skulle:
> 
> 
> An alternative would be to store unavailable file into into the `Job.FileInfo` metadata, so we know not to try to read them. But, it should be done in a separate PR.

I am doing this currently in the map-branch-filter because i need the info later for the dead links. 

What do you mean by 'It's up to the logger to filter...'  Is that something a user of the toolkit should do with a custom class?

jelovirt:
But why do you need information for the dead links when the issue is about supporting filtering out the whole topic? If filtering operations remove a file and create dead links, that's a different issue that should be part of a different PR. This PR for #3443 should only implement the support for filtering out the whole topic.

skulle:
Initially, only the filtered topic was removed but then I still had my links and i thought I doesn't make sense. It is just that, there was no more profound need or so. 
Probably I can make a PR for only removing the topics and storing the info in the job file. Not sure if the first commit of mine will run through with all the tests. I will check that and make a new PR.

jelovirt:
It's preferred to keep PR scope to a minimum, thus one PR for supporting filtering out topic content. When the whole topic has been filtered out, it should be removed from `Job` instead of marking it as deleted.

You can use this PR for the work and just rebase, instead of closing this and opening a new PR.

skulle:
Of course, let me do the changes and I will come back to you... 

Thx for having a look at my PR.

skulle:
Hi @jelovirt, the PR is refactored to remove filtered topics only. If you like you can have another look.

PR #3450, opened 2020-02-07
Fix indexterms with PDF using RenderX XEP

### LINKED ISSUE #3447: PDF Index term generation broken when using XEP
This is a side effect of the fix @robander made in: https://github.com/dita-ot/dita-ot/issues/2110
When generating a PDF from a DITA Map which uses index terms, the XEP processor throws some errors like this:

          [xep] [XEPJ002E][ERROR] no entries for index key 'Specs:'

and the PDF is no longer generated.
Publishing with FOP and Antenna House still works.

Probably the XSLT stylesheet "plugins/org.dita.pdf2.xep/xsl/fo/index_xep.xsl" needs to be modified to take the new "@indexid" attribute into account.

raducoravu:
Fixes #3447. Indexterm generation in PDF using XEP was broken.


infotexture:
Should this also be rebased onto `hotfix/3.4.1` for inclusion in the next maintenance release?

PR #3453, opened 2020-02-07
Add DITA 2.0 include element

jelovirt:
Add support for DITA 2.0 `<include>` element. The implementation is effectively the same as for the old `<coderef>`, thus `CoderefResolver` is retrofitted to support `<include>`.


jelovirt:
@infotexture Yes, we'll retain the old `coderef` as is.

infotexture:
I figured the existing `<coderef>` extensions would be retained for backwards compatibility, but what I meant is _“will the same extensions be available for the new `<include>` element?”_

I imagine line range extraction or similar capabilities would be as welcome in this more generalized context as they are in the existing `<coderef>` use case:

```xml
<include href="file.ext#line-range(5,10)" format="something"/>
```

jelovirt:
Yes, the `include` with `parse=text` will have those extensions.

jelovirt:
@infotexture Now that this is in, should I generate an issues for the docs, or are you able to pick these up?

infotexture:
As a rule, it would be good to provide docs input with each PR per https://github.com/dita-ot/.github/commit/e76ba8b097e1eb38a275a58b4e10006f59af456b.

In cases where changes require extensive docs updates, a dedicated docs issue might be better.

In this particular case, I'm not sure whether the changes warrant documentation of their own, or whether the `<include>` support would be better described in the context of evolving support for DITA 2.0 features.

PR #3466, opened 2020-02-23
preserve dita domains via option

chrispy-snps:
## Description
This pull request adds new arguments to the HTML5 and XHTML transformations that preserve the full domain-qualified class values ($domain/$element) in the output \@class attribute values.

- existing arguments (on by default)
    - args.html5.classattr
    - args.xhtml.classattr
- new arguments (OFF by default)
    - args.html5.domainattr
    - args.xhtml.domainattr

For example,

```
<strong class="ph b topic/ph hi-d/b">This is bold.</strong>
```

The controls are independent and can be enabled/disabled in any combination.

## Motivation and Context
Oxygen XML Author uses domain-qualified CSS selectors. Preserving domain-qualified @class values in the output allows better CSS reuse between the editing and output environments.

## How Has This Been Tested?
I ran the following commands:

```
./gradlew
./gradlew test
./gradlew integrationTest
```

then ran manual tests to confirm the functionality.

## Type of Changes
A new Boolean $includedomain parameter is added to the get-element-ancestry templates that allow it to return results that are/are not domain-qualified. This is then called conditionally based on the value of both the "classattr" and "domainattr" arguments.

When creating this change, I converted the XHTML string-appending @class attribute construction to the newer tokenization-based approach used by HTML5. However, this causes some integrationTest failures due to duplicate class values being uniquified:

```
Caused by: java.lang.AssertionError: Expected attribute value 'ph cmd ph' but was 'ph cmd'
```

I think the tokenized behavior is better, but I don't know enough to update the tests accordingly.

### DIFF
--- a/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
+++ b/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
@@ -185,6 +185,7 @@ See the accompanying LICENSE file for applicable license.
         <param name="ARTLBL" expression="${args.artlbl}" if:set="args.artlbl"/>
         <param name="GENERATE-TASK-LABELS" expression="${args.gen.task.lbl}" if:set="args.gen.task.lbl"/>
         <param name="PRESERVE-DITA-CLASS" expression="${args.html5.classattr}" if:set="args.html5.classattr"/>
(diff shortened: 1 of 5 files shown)


jelovirt:
IMO this should be plugin, not part of the HTML5 plugin.

PR #3473, opened 2020-03-10
Use singleton XML utils

jelovirt:
## Description
Use singleton XML utils throughout a single DITA-OT process.

## Motivation and Context
Sharing the XML utils allows using the same Saxon processors. This in turn allows using a shared name pool and other XML related resources.

No documentation changes needed.

### DIFF
--- a/src/main/java/org/dita/dost/log/DITAOTAntLogger.java
+++ b/src/main/java/org/dita/dost/log/DITAOTAntLogger.java
@@ -120,17 +120,32 @@ public final class DITAOTAntLogger extends MarkerIgnoringBase implements DITAOTL
 
     @Override
     public void error(String format, Object arg) {
-        log(MessageFormat.format(format, arg), null, Project.MSG_ERR);
+        if (arg instanceof Throwable) {
+            log(format, (Throwable) arg, Project.MSG_ERR);
+        } else {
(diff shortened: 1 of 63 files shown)


src/main/java/org/dita/dost/log/DITAOTAntLogger.java
```diff
@@ -120,17 +120,32 @@ public void error(final String msg) {
 
     @Override
     public void error(String format, Object arg) {
-        log(MessageFormat.format(format, arg), null, Project.MSG_ERR);
+        if (arg instanceof Throwable) {
+            log(format, (Throwable) arg, Project.MSG_ERR);
+        } else {
+            log(MessageFormat.format(format, arg), null, Project.MSG_ERR);
+        }
```
jelovirt: Small fix to logging that has nothing to do with XML IO

infotexture:
> No documentation changes needed.

@robander Thanks for the mention, but I'm not sure any of the changed code Is mentioned in the docs, as it seems to be rather low-level plumbing most users wouldn't touch. 

If these changes are likely to affect existing customizations, I'd need input for the migration topic, as I'm not entirely sure what's going on here and what users would need to change to ensure their plug-ins remain compatible.

PR #3476, opened 2020-03-12
Allow extending eclipsehelp

### LINKED ISSUE #3468: Allow extending eclipsehelp
## Description
currently you check 
        INDEX_TYPE_ECLIPSEHELP.equals(transtype)
and
         transtype.equals(INDEX_TYPE_ECLIPSEHELP)
but then if I create my eclipsehelp plugin, transtype becomes my_eclipsehelp and suddenly a lot of features are disabled

## Possible Solution
         transtype.contains(INDEX_TYPE_ECLIPSEHELP)

thendarion:
Signed-off-by: thendarion <16006640+thendarion@users.noreply.github.com>

---
name: #3468 Allow extending eclipsehelp
about: allow eclipsehelp trastypes to use standard eclipsehelp features

---

## Description
instead of equals, use contains, so transtype="my-eclipsehelp" is picked up too

## Motivation and Context
resolves #3468 Allow extending eclipsehelp

## How Has This Been Tested?
currently DOST doesn't seem to have related tests

## Type of Changes
- New feature 

nothing
nope

### DIFF
--- a/src/main/java/org/dita/dost/module/ChunkModule.java
+++ b/src/main/java/org/dita/dost/module/ChunkModule.java
@@ -64,7 +64,7 @@ final public class ChunkModule extends AbstractPipelineModuleImpl {
         final ChunkMapReader mapReader = new ChunkMapReader();
         mapReader.setLogger(logger);
         mapReader.setJob(job);
-        mapReader.supportToNavigation(INDEX_TYPE_ECLIPSEHELP.equals(transtype));
+        mapReader.supportToNavigation(transtype.contains(INDEX_TYPE_ECLIPSEHELP));
         if (input.getAttribute(ROOT_CHUNK_OVERRIDE) != null) {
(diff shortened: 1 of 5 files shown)


robander:
This one makes me uncomfortable, for a couple of reasons:

- While simple to understand, I don't really like the design of switching from `equals` to `contains` because of the fact that Eclipse help transforms can pick up the behavior but only if they include `eclipsehelp` in the `transtype` name. Ideally it would work for any extension of Eclipse help.
- As you know, Eclipse help output is now maintained in a separate repository in its own plugin. Ideally, all of the code for Eclipse help would be stripped out of the core DITA-OT and integrated into that plugin. Unfortunately with few of us making use of the code, it has not risen high enough in the priority list. But with that in mind, it's hard to think about extending the core code that we'd like to move out.

I know your linked issue #3468 mentions indexing -- is that the Eclipse feature you were hoping to pick up with this? I know the code in the core repo also contains some support for the "export anchors" markup, but I don't think I've ever seen that markup used outside of test documents.

PR #3477, opened 2020-03-12
Automatically bump Homebrew formula on release

infotexture:
## Description

This PR adds a GitHub Action script to automatically bump the [DITA-OT Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/dita-ot.rb) whenever a new release tag is pushed to the DITA-OT repo.

## Motivation and Context

Currently the Homebrew formula is updated manually after each release.

This configures an existing action from the GitHub Actions Marketplace to update the DITA-OT formula: https://github.com/mislav/bump-homebrew-formula-action.

A new `COMMITTER_TOKEN` secret has been added to the core repo to authenticate the action.

## How Has This Been Tested?

I'm afraid there may be no other way to test this than to merge it and verify whether it works when a new tag is created. 

## Type of Changes

- What documentation changes are needed for this feature? _NONE_
- Will this change affect backwards compatibility or other users' overrides? _NO_

### DIFF
--- a/.github/workflows/bump-homebrew-formula.yml
+++ b/.github/workflows/bump-homebrew-formula.yml
@@ -1,0 +1,34 @@
+# Automatically bump Homebrew formula after each release
+# https://github.com/mislav/bump-homebrew-formula-action
+
+# Controls when the action will run. Triggers the workflow when new tags are pushed
+on:
+  push:
+    tags: '*'
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+
+  # This workflow contains a single job called "homebrew"
+  homebrew:
+    name: Bump Homebrew formula
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v1.6
+        if: "!contains(github.ref, '-')" # skip prereleases
+        with:
+          formula-name: dita-ot
+          homebrew-tap: Homebrew/homebrew-core
+          base-branch: master
+          download-url: https://github.com/dita-ot/dita-ot/releases/download/3.4/dita-ot-3.4.zip
+          commit-message: |
+            {{formulaName}} {{version}}
+            
+            Created by https://github.com/mislav/bump-homebrew-formula-action
+        env:
+          # Access token added as secret to repo settings
+          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
+          # GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/bump-homebrew-formula.yml
+++ b/.github/workflows/bump-homebrew-formula.yml
@@ -17,13 +17,17 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - uses: mislav/bump-homebrew-formula-action@v1.6
+      - name: Set download URL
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-homebrew-formula.yml
```diff
@@ -0,0 +1,34 @@
...
+  # This workflow contains a single job called "homebrew"
+  homebrew:
+    name: Bump Homebrew formula
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v1.6
+        if: "!contains(github.ref, '-')" # skip prereleases
```
jelovirt: This is redundant in our case, but we can keep it.

.github/workflows/bump-homebrew-formula.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          formula-name: dita-ot
+          homebrew-tap: Homebrew/homebrew-core
+          base-branch: master
+          download-url: https://github.com/dita-ot/dita-ot/releases/download/3.4/dita-ot-3.4.zip
```
jelovirt: This needs to be fixed. The `download-url` should have the correct version numbers. If we omit it, the action will attempt to [generate](https://github.com/mislav/bump-homebrew-formula-action/blob/master/src/main.ts#L9) the URL but it will not work correctly in our case.

jelovirt:
Per https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions, we should add a step

```yaml
      - name: Set env
        run: echo ::set-env name=RELEASE_VERSION::${GITHUB_REF:10}
```

and then change the download URL to be 

```yaml
download-url: https://github.com/dita-ot/dita-ot/releases/download/${{ env.RELEASE_VERSION }}/dita-ot-${{ env.RELEASE_VERSION }}.zip
```

Not 100% sure this works, though.

.github/workflows/bump-homebrew-formula.yml
jelovirt: Fixed

jelovirt:
The only concern here I have is that this GA completes before Travis has completed the deployment. If the Brew CI tests this PR by trying to run it, it will fail because the artifact is not available yet.

We may need to update the GA code so that we run it manually, somehow.

.github/workflows/bump-homebrew-formula.yml
```diff
@@ -17,13 +17,17 @@ jobs:
...
+        id: download-url
+      - name: Create pull-request
+        uses: mislav/bump-homebrew-formula-action@v1.6
         if: "!contains(github.ref, '-')" # skip prereleases
         with:
           formula-name: dita-ot
           homebrew-tap: Homebrew/homebrew-core
           base-branch: master
-          download-url: https://github.com/dita-ot/dita-ot/releases/download/3.4/dita-ot-3.4.zip
+          download-url: ${{ steps.download-url.outputs.url }}
```
infotexture: @jelovirt Thanks. I realized later that this should probably be made dynamic, but wasn't sure how.

infotexture:
> The only concern here I have is that this GA completes before Travis has completed the deployment.

Not sure if there's any way to introduce a delay in the workflow, so when a new tag is pushed, the workflow would run 10 minutes later, or 30, or something. 🤔

infotexture:
🚧 **Deferred to after 3.4.1 release**

- [ ] Investigate options to delay execution until Travis build artifacts are available

jelovirt:
Since GitHub has added support for [manual triggers](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/), this should be reopened.

Issue #3464, opened 2020-02-23
Add spanning and title to simpletable


jelovirt:
Per https://github.com/oasis-tcs/dita/issues/292

-   add `<title>` as the first element of `<simpletable>`
-   add `@colspan` and `@rowspan` attributes to `<stentry>`
-   add `@headers` and `@scope` to `<stentry>`

Plugins:

- [x] HTML5
- [x] PDF2

PR #3479, opened 2020-03-13
Add spanning and title to simpletable in PDF2 #3464 

jelovirt:
Per https://github.com/oasis-tcs/dita/issues/292 add support to PDF2:

- [x] add `<title>` as the first element of `<simpletable>`
- [x] add `@colspan` and `@rowspan` attributes to `<stentry>`
- [x] add `@headers` and `@scope` to `<stentry>`

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
@@ -312,7 +312,9 @@ See the accompanying LICENSE file for applicable license.
     </xsl:template>
   
   <xsl:template match="*[contains(@class, ' topic/table ')]/*[contains(@class, ' topic/title ')]" mode="table.title-number">
-    <xsl:value-of select="count(key('enumerableByClass', 'topic/table')[. &lt;&lt; current()][dita-ot:notExcludedByDraftElement(.)])"/>
+    <xsl:value-of select="count((key('enumerableByClass', 'topic/table') | key('enumerableByClass', 'topic/simpletable'))
+                                [. &lt;&lt; current()]
+                                [dita-ot:notExcludedByDraftElement(.)])"/>
   </xsl:template>
 
     <xsl:template match="*[contains(@class, ' topic/tgroup ')]" name="tgroup">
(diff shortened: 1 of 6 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -962,61 +989,82 @@ See the accompanying LICENSE file for applicable license.
...
-      <xsl:param name="number-cells" as="xs:integer">
+      <xsl:param name="number-cells" as="xs:integer" tunnel="yes">
             <xsl:apply-templates select="../*[1]" mode="count-max-simpletable-cells"/>
         </xsl:param>
         <fo:table-header xsl:use-attribute-sets="sthead">
             <xsl:call-template name="commonattributes"/>
             <fo:table-row xsl:use-attribute-sets="sthead__row">
                 <xsl:apply-templates select="*[contains(@class, ' topic/stentry ')]"/>
-                <xsl:variable name="row-cell-count" select="count(*[contains(@class, ' topic/stentry ')])" as="xs:integer"/>
+                <!--
```
robander: Was this commented section (and the one below) left in intentionally?

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
```diff
@@ -1042,17 +1090,25 @@ See the accompanying LICENSE file for applicable license.
...
+          <xsl:apply-templates select="../.." mode="count-max-simpletable-cells"/>
+        </xsl:param>
         <fo:table-cell xsl:use-attribute-sets="strow.stentry">
             <xsl:call-template name="commonattributes"/>
-            <xsl:variable name="entryCol" select="count(preceding-sibling::*[contains(@class, ' topic/stentry ')]) + 1"/>
+            <xsl:call-template name="simpletableApplySpansAttrs"/>
+            <xsl:variable name="entryCol" select="@dita-ot:x"/>
             <xsl:variable name="frame" as="xs:string" select="(ancestor::*[contains(@class, ' topic/simpletable ')][1]/@frame, $table.frame-default)[1]"/>
+            <xsl:variable name="last-column" as="xs:integer">
+              <xsl:variable name="span" select="if (exists(@colspan)) then (xs:integer(@colspan) - 1) else 0"/>
```
robander: (Maybe a thought for the future) Should we be doing validation of `@colspan` as integer, or do we do that earlier? I think using `xs:integer(@colspan)` will throw messy errors if it's not an integer - I've seen that with code that assumed an integer with proportional column widths, and failed when it got `.8*` instead of `8*` ... but that seems less likely here

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
jelovirt: Yeah, I don't want to throw away the code, if we end up wanting to use it.

src/main/plugins/org.dita.pdf2/xsl/fo/tables.xsl
jelovirt: We have clean-up filters that we could use to implement the integer validation. I'll leave it for future.

PR #3482, opened 2020-03-21
Add outputclass support to DITAVAL

### LINKED ISSUE #3463: Add outputclass to ditaval
Add `@outputclass` to DITAVAL per https://github.com/oasis-tcs/dita/issues/252.

jelovirt:
Add `@outputclass` to DITAVAL per https://github.com/oasis-tcs/dita/issues/252.

- [x] Content profiling in preprocess
- [x] HTML5
- [x] XHTML

Fixes #3463

### DIFF
--- a/src/main/java/org/dita/dost/reader/DitaValReader.java
+++ b/src/main/java/org/dita/dost/reader/DitaValReader.java
@@ -238,7 +238,8 @@ public final class DitaValReader implements AbstractReader {
                 style != null ? style.trim().split("\\s+") : null,
                 getValue(elem, ATTRIBUTE_NAME_CHANGEBAR),
                 readFlagImage(elem, "startflag"),
-                readFlagImage(elem, "endflag"));
+                readFlagImage(elem, "endflag"),
+                getValue(elem, ATTRIBUTE_NAME_OUTPUTCLASS));
     }
 
     private FlagImage readFlagImage(final Element elem, final String name) {
--- a/src/main/java/org/dita/dost/util/FilterUtils.java
+++ b/src/main/java/org/dita/dost/util/FilterUtils.java
@@ -223,7 +223,7 @@ public final class FilterUtils {
             return res.stream()
                     .map(f -> new Flag(ELEMENT_NAME_PROP, conflictColor ? foregroundConflictColor : f.color,
                             conflictBackcolor ? backgroundConflictColor : f.backcolor,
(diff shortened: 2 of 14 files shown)


jelovirt:
Ok, now it should work in HTML5

jelovirt:
Removing XHTML from planned support.

jelovirt:
The problem here is that now that we process `@outputclass` and no longer use `@outputclass` to carry `@style`, XHTML output no longer flags content correctly. If we don't want to remove exiting functionality, we have to implement this in XHTML too.

infotexture:
Given the  sample `<revprop>` [above](https://github.com/dita-ot/dita-ot/pull/3482#pullrequestreview-378952221):

```xml
<revprop action="flag" val="3.3.3" color="green" style="underline" outputclass="mark"/>
```

the final code here currently outputs the green text `@color` to the `@style` attribute, but moves the "underline" style keyword to the `@class` hierarchy along with the `mark` keyword:

```html
<p style="color:green;" class="p mark underline">
```

⚠️ This means that the flagged text will **not** appear underlined in output _unless_ the user defines a corresponding CSS rule that targets the `.underline` class selector in a custom CSS file and implements `text-decoration: underline` there.

This would be fine if the `underline` keyword had been added to the `@outputclass` attribute of the `<revprop>` element, but if specified as `@style`, I would expect it to be output to `@style`.

I understand that oasis-tcs/dita#252 adds support for `@outputclass` as an _alternative_ method of defining flagging styles, but I see nothing there that suggests existing `@style` entries should be implemented as class keywords _instead of as inline styles_.

@robander Is there another proposal for DITA 2.0 that mandates this behavior?

@jelovirt If this change is intentional, which of the existing DITA 1.3 `<revprop>` attributes will still be output to `@style`, and which as class keywords?

I understand that inline HTML `@style` attributes are problematic due to their specificity, as they cannot be readily overridden with external stylesheets, and welcome the addition of the `@outputclass` option in DITA 2.0, but not implementing `@style` as `@style` seems to violate both backwards-compatibility and the "principle of least astonishment".

One option to preserve backwards compatibility in cases like this would be to add corresponding rules to the default `common[ltr|rtl]` stylesheets, but I see no such changes here.

robander:
Nothing about DITA 2.0 mandates that behavior (making those into CSS values on `@style` was entirely a DITA-OT implementation thing). 

I missed that some values moved - I saw the examples where values were preserved and didn't realize that styles like underline didn't move.

PR #3483, opened 2020-03-22
Recognize new class atts for link text, short description, search title

robander:
---
name: Pull request

---

## Description

DITA 2.0 changes the class attributes for 3 elements in maps, which also appear in topics:
`map/linktext` ==> `topic/linktext`
`map/shortdesc` ==> `topic/shortdesc`
`map/searchtitle` ==> `topic/searchtitle`

This update allows both the 1.3 and 2.0 values to function. For map processing link in `mappull` it recognizes either value. It continues to write out `map/*` values for now when creating new elements, so that the code is backwards compatible / still based on the latest stable release.

## Motivation and Context

Ensures that testing / formatting with the new DITA 2.0 grammar files works properly for these elements in a map.

## How Has This Been Tested?

Created XSPEC tests for new XSL functions that check both class attributes.

Also: 
* Made a simple update to `hierarchy.ditamap` to add in link text and short descriptions
* Verified output with current processing (child links use the link text and short desc from map)
* Made explicit attributes of `topic/linktext` and `topic/shortdesc` on those elements
* Without the code update, those elements are ignored; with the code update, the child links work the same as with the default class values

## Type of Changes

Minor enhancement

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/map-elements.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/map-elements.xsl
@@ -8,9 +8,11 @@ See the accompanying LICENSE file for applicable license.
 -->
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
   xmlns:fo="http://www.w3.org/1999/XSL/Format"
+  xmlns:dita-ot="http://dita-ot.sourceforge.net/ns/201007/dita-ot"
+  exclude-result-prefixes="dita-ot"
   version="2.0">
 
-  <xsl:template match="*[contains(@class,' map/topicmeta ')]/*[contains(@class,' map/searchtitle ')]"/>
+  <xsl:template match="*[contains(@class,' map/topicmeta ')]/*[dita-ot:matches-searchtitle-class(@class)]" priority="10"/>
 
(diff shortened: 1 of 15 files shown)


src/main/plugins/org.dita.pdf2/xsl/fo/map-elements.xsl
```diff
@@ -8,9 +8,11 @@ See the accompanying LICENSE file for applicable license.
 -->
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
   xmlns:fo="http://www.w3.org/1999/XSL/Format"
+  xmlns:dita-ot="http://dita-ot.sourceforge.net/ns/201007/dita-ot"
+  exclude-result-prefixes="dita-ot"
   version="2.0">
 
-  <xsl:template match="*[contains(@class,' map/topicmeta ')]/*[contains(@class,' map/searchtitle ')]"/>
+  <xsl:template match="*[contains(@class,' map/topicmeta ')]/*[dita-ot:matches-searchtitle-class(@class)]" priority="10"/>
```
robander: I don't usually like priorities in our code, but this one is added so that `topicmeta/searchtitle` (with either class att) takes priority over the generic `searchtitle` rule that already exists elsewhere.

PR #3485, opened 2020-03-23
Update to Saxon 10.0

jelovirt:
Update Saxon-HE dependency from 9.9 to latest version 10.5.

Highlights of [changes between Saxon 9.9 and 10](https://www.saxonica.com/documentation10/index.html#!changes/all/9.9.1-10):

* Higher order functions and the xsl:evaluate instruction are now available in Saxon-HE.
* Two new APIs have been introduced to make construction of XDM trees easier.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ dependencies {
     implementation group: 'xerces', name: 'xercesImpl', version:'2.12.0'
     implementation group: 'xml-apis', name: 'xml-apis', version:'1.4.01'
     implementation group: 'xml-resolver', name: 'xml-resolver', version:'1.2'
-    implementation group: 'net.sf.saxon', name: 'Saxon-HE', version: '9.9.1-7'
+    implementation group: 'net.sf.saxon', name: 'Saxon-HE', version: '10.5'
     implementation group: 'com.ibm.icu', name: 'icu4j', version:'61.1'
     implementation group: 'org.apache.ant', name: 'ant', version:'1.10.9'
     implementation group: 'org.apache.ant', name: 'ant-launcher', version:'1.10.9'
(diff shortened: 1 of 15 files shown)


jelovirt:
Opened issue for this in Saxon https://saxonica.plan.io/issues/4752

The problem is that Saxon 10 implements XSLT 3.0 that no longer has recoverable errors. Instead, stylesheets should use [try-catch](https://www.w3.org/TR/xslt-30/#try-catch). Saxon currently has no way of configuring this.

jelovirt:
https://saxonica.plan.io/issues/4752?pn=1#note-5 will likely make it possible for use to move to Saxon 10 when this feature is released.

jelovirt:
Saxon 10.5 contains new `net.sf.saxon.lib.EmptySource` to support backwards compatibility.

jelovirt:
@raducoravu What's Oxygen's current take on this PR? When this is merged, you can no longer use Saxon 9.9. Will this prevent Oxygen from updating DITA-OT version?

raducoravu:
> @raducoravu What's Oxygen's current take on this PR? When this is merged, you can no longer use Saxon 9.9. Will this prevent Oxygen from updating DITA-OT version?

For plain XML + XSLT transformations and debugging Oxygen uses Saxon 9 libraries present in its "lib" folder. We plan to update those to 10 as well when time will allow.
For DITA OT publishing by default we use the Saxon libraries bundled with DITA OT, so this should work. People could by editing the transformation scenario to change the order of the Saxon libraries so that the Saxon 9 libs bundled with Oxygen would be used with the DITA OT but this was not documented anywhere and not a common request, it was useful only to allow directly calling Java code from XSLT as Oxygen's Saxon 9 is fully licensed. But we'll probably also update to Saxon 10 anyway.

robander (approved):

PR #3490, opened 2020-03-28
Add language mapping for Norwegian Bokmål

### LINKED ISSUE #3478: Add more mappings for Norway language
One of our end users uses the "nb-NO" language code in their DITA topics: 

    We work here with many different languages in Oxygen. And now our translation software (MemoQ, Trados) uses the code nb-NO for Norwegian (Norwegian-bokmål).

Maybe we can add in the translation strings.xml a mapping also for "nb-NO" similar to what we do for "no-NO"

infotexture:
This change appears simple enough to consider for 3.5.

Fixes #3478.

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/common/common-strings.xml
+++ b/src/main/plugins/org.dita.base/xsl/common/common-strings.xml
@@ -77,6 +77,7 @@ See the accompanying LICENSE file for applicable license.
   <lang xml:lang="nl-nl" filename="strings-nl-nl.xml"/>
   <lang xml:lang="no"    filename="strings-no-no.xml"/>
   <lang xml:lang="no-no" filename="strings-no-no.xml"/>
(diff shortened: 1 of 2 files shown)


infotexture:
1. 3717fd4 added the initial mapping for non-PDF output formats
2. bc871c8 adds the corresponding mapping for PDF output

The change in bc871c8 duplicates the previous line to explicitly map `xml:lang="nb-no"` to the existing Norwegian strings file `no.xml`, since the automated fallback to the macrolanguage won't work in this case (in contrast to something like `de-DE`, where a `de.xml` file exists). 

I suppose we could copy `no.xml`  → `nb.xml` instead, but since we don't have dedicated Bokmål localization at this point, declaring Norsk as fallback seems more appropriate, and mimics the approach for other output formats in `common-strings.xml`.

@jelovirt @robander Does that sound like the right solution?

PR #3491, opened 2020-03-29
Copyedit 3.5 CLI help

infotexture:
## Description

Now that the `dita help` subcommand includes the list of available parameters (🎉) , output appears inconsistent with regard to initial capitalization of each line. 

The majority of entries are parameter descriptions in sentence case, so c7529a7 capitalizes the remaining arguments & options to align with this precedent, and clarifies a few descriptions in the process.

1bf9c0d: Copyedit parameter descriptions

- Reduce maximum line length to minimize line wrapping
- Clarify purpose of "outer" options

### DIFF
--- a/src/main/java/org/dita/dost/invoker/ConversionArguments.java
+++ b/src/main/java/org/dita/dost/invoker/ConversionArguments.java
@@ -353,16 +353,16 @@ public class ConversionArguments extends Arguments {
                 .subcommands("transtypes")
                 .subcommands("uninstall")
                 .subcommands("version")
-                .arguments("i", "input", "file", "input file")
-                .arguments("f", "format", "name", "output format (transformation type)")
-                .arguments("p", "project", "name", "run project file")
-                .options("r", "resource", "file", "resource file")
-                .options(null, "filter", "files", "filter and flagging files")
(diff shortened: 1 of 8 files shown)


src/main/java/org/dita/dost/invoker/ConversionArguments.java
```diff
@@ -353,16 +353,16 @@ void printUsage() {
...
                 .subcommands("uninstall")
                 .subcommands("version")
-                .arguments("i", "input", "file", "input file")
-                .arguments("f", "format", "name", "output format (transformation type)")
-                .arguments("p", "project", "name", "run project file")
-                .options("r", "resource", "file", "resource file")
-                .options(null, "filter", "files", "filter and flagging files")
-                .options("l", "logfile", "file", "use given file for log")
-                .options("o", "output", "dir", "output directory")
+                .arguments("i", "input", "file", "Input file")
```
jelovirt, line 356: For the next release, should we extract these strings into a properties resource file? We don't translate them, but would it be easier if they were not in Java code?

infotexture: Yes, that's an excellent idea, thanks! 👍

PR #3493, opened 2020-04-02
Fix doc build errors with new class att matches

robander:
---
name: Pull request

---

## Description

Fixes two issues with doc builds that arose from #3483 

* It looks like something in our process is adding elements into `topicmeta` that do not have a `@class` attribute. This is unrelated to that pull request; however, the function that sends `@class` as a parameter threw an error because it's being called with a missing parameter. Explicitly matching only elements with `@class` resolves the error, but longer term we should figure out what is being added without a `@class` att.
* There is an ambiguous rule match for `topic/shortdesc` in `topic.xsl` for PDF. I've removed the extra generic match, which is already handled above by an existing match; I've also made that match more explicit, so `topic.xsl` has one match for shortdesc-in-topic and one for shortdesc-not-in-topic.

## Type of Changes


jelovirt:
Travis has completed and passed, just not visible here.

PR #3497, opened 2020-04-06
Fix mapper with external keys in subdirectory

robander:
---
name: Pull request

---

## Description

Fixes a defect with external keys that are defined in a map in a subdirectory.

For example, I have a map inside `common/` with this key definition:
`<keydef href="/this/is/absolute" scope="external" format="html">`

This is an external link, which resolves properly when the published HTML is loaded onto a web server. I've marked it as absolute, and it uses an absolute path. However, when `mapref` merges maps from the `common/` subdirectory, it treats this as a relative path and adds a `common/` prefix -- the merged map is updated to this, which breaks my keyref links:
`<keydef href="common/this/is/absolute" scope="external" format="html">`

The mapref code already checks for `://` in a URI as a way to avoid updating it during the merge; my update also checks for `@scope="external"`, for which we avoid updates in a lot of other places.

I moved that check (for `://` and scope) into a function so that we are consistent about which paths do not get updated, in normal map merging + relationship table merging.

These absolute paths beginning with `/` still result in quirky output if the scope is set to `peer` but I think that's OK (at least, the result is arguable, while we've always treated `scope="external"` as static).

## How Has This Been Tested?

Passes integration tests; test case included here.
[mapref.zip](https://github.com/dita-ot/dita-ot/files/4441179/mapref.zip)

## Type of Changes

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/common/functions.xsl
+++ b/src/main/plugins/org.dita.base/xsl/common/functions.xsl
@@ -60,6 +60,13 @@ See the accompanying LICENSE file for applicable license.
            else $doc
     "/>
   </xsl:function>
+  
+  <!-- $href (might be @href or @copy-to) has absolute or external value that should not be changed -->
+  <xsl:function name="dita-ot:is-external" as="xs:boolean">
+    <xsl:param name="href" as="attribute()"/>
+    <xsl:sequence select="contains($href,'://') or 
+        $href/../@scope = 'external'"></xsl:sequence>
+  </xsl:function>
 
   <!-- ID -->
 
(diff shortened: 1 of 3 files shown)


src/main/plugins/org.dita.base/xsl/common/functions.xsl
```diff
@@ -60,6 +60,13 @@ See the accompanying LICENSE file for applicable license.
            else $doc
     "/>
   </xsl:function>
+  
+  <!-- $href (might be @href or @copy-to) has absolute or external value that should not be changed -->
+  <xsl:function name="dita-ot:unmodifiable-uri" as="xs:boolean">
```
jelovirt, line 65: The name could instead be `dita-ot:is-external`, because that's what we're testing for.

src/main/plugins/org.dita.base/xsl/common/functions.xsl
```diff
@@ -60,6 +60,13 @@ See the accompanying LICENSE file for applicable license.
...
+    <xsl:param name="href" as="attribute()"/>
+    <xsl:sequence select="contains($href,'://') or 
+        (exists($href/../@scope) and $href/../@scope = 'external')"></xsl:sequence>
```
jelovirt, line 68:
```suggestion
        $href/../@scope = 'external'"/>
```

PR #3498, opened 2020-04-06
Fix build crash when term refs missing doc

robander:
---
name: Pull request

---

## Description

I have a `<term>` element that uses `@keyref`, but the key is a broken link (the topic is missing).

Our key processing code for terms assumes that the document exists, tries to access it, and then the entire build crashes with I/O errors and unchecked XPATH exceptions, for trying to examine nodes inside a document that did not resolve.

The fix just updates the common function `resolved-href-path` to check that the document is available before loading it into a variable; this allows the build to finish (with expected errors about the missing doc / unresolved term reference).

## Motivation and Context

This actually came up in a case that likely highlights a bug in the chunking code -- a branch of the map was chunked, and for some reason moved the topic from a subdirectory into the map directory with a generated `Chunk1345.dita` type name. The resolved key on `term` was not updated, meaning the path was now broken; that broken path resulted in the issue described above. It was easily reproducible by creating a broken key in the root map.

## How Has This Been Tested?

Passes tests; actual test case attached.
[term.zip](https://github.com/dita-ot/dita-ot/files/4441243/term.zip)

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/common/functions.xsl
+++ b/src/main/plugins/org.dita.base/xsl/common/functions.xsl
@@ -48,9 +48,12 @@ See the accompanying LICENSE file for applicable license.
   <xsl:function name="dita-ot:retrieve-href-target" as="node()?">
     <xsl:param name="href" as="attribute(href)"/>
 
-    <xsl:variable name="doc" as="document-node()"
-      select="doc(dita-ot:resolve-href-path($href))"/>
-
+    <xsl:variable name="resolved-href-path" select="dita-ot:resolve-href-path($href)" as="xs:anyURI"/>
+    <xsl:variable name="doc" as="document-node()?"
+      select="if (doc-available($resolved-href-path))
+      then (doc(dita-ot:resolve-href-path($href)))
+      else ()"/>
+    
     <xsl:sequence select="
         if (dita-ot:has-element-id($href))
       then $doc/key('id', dita-ot:get-element-id($href))


src/main/plugins/org.dita.base/xsl/common/functions.xsl
```diff
@@ -48,9 +48,12 @@ See the accompanying LICENSE file for applicable license.
   <xsl:function name="dita-ot:retrieve-href-target" as="node()?">
     <xsl:param name="href" as="attribute(href)"/>
 
-    <xsl:variable name="doc" as="document-node()"
```
jelovirt, line 51: Because [`doc` will return an empty node sequence when the resource cannot be read](https://www.w3.org/TR/xquery-operators/#func-doc), can't we just change the type to `document-node)? `. Then we would not have to use `doc-available`.

src/main/plugins/org.dita.base/xsl/common/functions.xsl
```diff
@@ -60,6 +63,13 @@ See the accompanying LICENSE file for applicable license.
            else $doc
     "/>
   </xsl:function>
+  
+  <!-- $href (might be @href or @copy-to) has absolute or external value that should not be changed -->
+  <xsl:function name="dita-ot:unmodifiable-uri" as="xs:boolean">
```
jelovirt, line 68: This is already part of another PR. The other one needs to be merged first and then this be rebased before merge.

src/main/plugins/org.dita.base/xsl/common/functions.xsl
robander: That's what I started with - but it still crashed. Just running `doc(dita-ot:resolve-href-path($href))` in the `select` attribute crashed the build when the document was not available.

jelovirt: Ah, sorry. I got confused. You should be using [`document()` function from XSLT](https://www.w3.org/TR/xslt20/#document), not `doc()` from XPath. The former can handle missing document, the latter throws an error.

src/main/plugins/org.dita.base/xsl/common/functions.xsl
robander: Yeah, I was fixing both issues in the same file on the same day, I knew some rebasing would be needed if the other PR had changes.

src/main/plugins/org.dita.base/xsl/common/functions.xsl
robander: Ah, yeah, changing to `document-node()?` and to `document()` also fixed it. On the other hand, using `document()` also added a lot more XSLT warnings -- so using `doc-available` might be better? The `document()` approach gives warnings like this (and we've already warned earlier in the build about the missing file):

```
     [xslt] Warning at char 35 in xsl:variable/@select on line 53 column 60 of functions.xsl:
     [xslt]   FODC0002: Document has been marked not available:
     [xslt]   file:/Users/robander/Documents/github/GITHUBTESTS/term/temp/missing.dita
```

Issue #3499, opened 2020-04-08
RNG default attributes expansion causes the mathml namespaces to be removed


raducoravu:
I have  a RNG topic containing an embedded MathML equation:
[oxygen_eqn.zip](https://github.com/dita-ot/dita-ot/files/4448722/oxygen_eqn.zip)

If it gets published to PDF and the temporary files folder is kept from being deleted we can notice in the temporary files folder that the processed topic no longer has the mathml namespace set on the "math" element, thus the MathML processor will no longer be able to process the equation in the PDF.

raducoravu:
The problem seems to be in the "org.ditang.relaxng.defaults.RelaxNGDefaultsComponent", I will compare it with the one we have in the Oxygen main library, see if it needs to be updated in some way.

raducoravu:
I would like to add a fix for this, should I create a pull request on the "develop" branch? Are there any automatic tests with RNG topic processing?

jelovirt:
@raducoravu I don't think we have any tests for RNG processing. Please make the PR against `release/3.5` branch. So first check out `release/3.5` and then create your feature branch from there.

PR #3501, opened 2020-04-09
#3499 Proper namespace processing for mathml content.

raducoravu:
Made changes to RNG pre processing stage so that namespace declarations in mathml or svg elements are no longer stripped for RNG based topic


jelovirt:
@raducoravu See https://www.dita-ot.org/contributing for adding sign-off to 
a38a30e

raducoravu:
@jelovirt that is just a resource which was not part of the fix and which was pushed because of a glitch in SourceTree, I checked it by mistake, unchecked it and then commit, and somehow SourceTree ended up adding it, then I removed it with that separate commit. Can you accept the pr as it is? Otherwise I can try tomorrow again the steps to add a signoff to that commit as well.

jelovirt:
@raducoravu the DCO check checks every commit, regardless of the content of the commit. Thus fixup commits also need to have a sign-off. Looks like your rebase worked and the test now passes.

raducoravu:
@jelovirt thanks I swear I don't do this on purpose (the lack of signoff), this last time GitHub conveniently had a button allowing me to delete that extra file from its web interface so I did not go through the local git client.

PR #3507, opened 2020-04-25
Ignore non-input maps after map merge

jelovirt:
## Description
After map merge, ignore non-input maps in processing. 

## Motivation and Context
After map merge only the input maps are used in processing. Submaps are no longer used for anything. This reduces the number of files that need to be processed.

Preprocess2 has already used this approach in most places, adding this to preprocess makes sense for consistency and performance improvement.

This can be backwards incompatible change if custom plug-ins use the submaps after map merge. We have never explicitly forbidden this, but it has never been something that makes sense for plug-ins to do.


jelovirt:
Due to conref and conrefpush between maps, this is somewhat problematic. The non-merged cannot be ignored immediately. The keyref, conref, and conrefpush stages need to be processed first, only then non-merged maps can be ignored or removed.

Issue #3433, opened 2019-12-19
No warning for broken in-topic link / unresolved link text.


robander:
## Expected Behavior

The attached topic has three invalid `@href` attributes that refer to content within the same file. When building HTML5 output, the result is a broken link and link text like `#jira680__twwo` and `#badtopic__twwo`

The build should warn about these unresolved links during `topicpull`.

## Actual Behavior

No message, clean build, bad output.

## Steps to Reproduce

Build the following topic to HTML5:

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN"  "topic.dtd">
<topic id="jira680" xml:lang="en-us">
<title>Jira680</title>
<shortdesc>Why does this link not report an error</shortdesc>
<body>
<ol>
<li>First item! look at the second! <xref href="#./tvo"/> and <xref href="#jira680/twwo"/> and <xref
href="#badtopic/twwo"/></li>
<li id="two">whoops I broke my ID when I linked here</li>
</ol>
</body>
</topic>
```

## Environment

* DITA-OT version: 3.3.4
* Operating system and version: Windows
* How did you run DITA-OT? `dita` command 
* Transformation type: HTML5

kirkilj:
@robander, you raised this a couple years ago. It looks like it was fixed in a forked OT repo, but not in the main repo. Do you know if this is still an issue in 3.7.1 or dev branch?

PR #3510, opened 2020-05-03
Issue #3433 No warning for broken in-topic link

ktsiligkiris:
Added error DOTX032E. Seemed most relevant for this issue.
The error is raised only when the problematic xref doesn't have a text
node, and is not possible to reference in current topic. I don't know
if a warning should be created for this situation, I just found the
position where the error message could be generated for this issue.

Signed-off-by: Konstantinos Tsiligkiris <ktsiligkiris@outlook.com>

---
name: Pull request
about: Propose changes to fix an issue

---

## Description
Added the error message DOTX032E in xslt code, during the rendering of an xref in html5, where the error is activated when the rendered link has a local href that doesn't return a title and the xref itself doesn't have a text node.

## Motivation and Context
This is about issue #3433 which seemed easy to fix, and this is my first pull request.
## How Has This Been Tested?
I rendered the dita example given in the issue and checked that the error message is reported only when the link is broken, not when the link is correct.

## Type of Changes

- I am not sure if error messages are included in the documentation, and I don't think that this breaks previous releases.

## Checklist

- My code follows the code style of this project. YES
- I have updated the unit tests to reflect the changes in my code. (Not sure about how to do this, it's my first pull request.)

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
@@ -79,6 +79,10 @@ See the accompanying LICENSE file for applicable license.
                 </xsl:when>
                 <xsl:otherwise>
                   <xsl:call-template name="href"/><!--use href text-->
+                  <xsl:call-template name="output-message">
+                    <xsl:with-param name="id" select="'DOTX032E'"/>
+                    <xsl:with-param name="msgparams">%1=<xsl:value-of select="@href"/></xsl:with-param>
+                  </xsl:call-template>
                 </xsl:otherwise>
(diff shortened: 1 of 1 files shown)


jelovirt:
Closing as this can be implemented already with

```xsl
<xsl:templates match="*" mode="determine-final-href">
  <xsl:next-match/>
  <xsl:call-template name="output-message">
    <xsl:with-param name="id" select="'DOTX032E'"/>
    <xsl:with-param name="msgparams">%1=<xsl:value-of select="@href"/></xsl:with-param>
  </xsl:call-template>
</xsl:template>
```

Issue #3376, opened 2019-10-07
NullPointerException in Saxon when using ANT XMLCatalog implementation


raducoravu:
When the DITA OT is placed in a folder containing spaces and French accents, Saxon throws a NullPointerException. I originally added the issue on the Saxonica issues list:

https://saxonica.plan.io/issues/4326

I think this can be fixed from the Saxon side but it would be better if our XsltModule code would not use the ANT "XMLCatalog" implementation for anything and always use the Apache Xerces Resolver instead.

Maybe on the method "org.dita.dost.module.XsltModule.setXMLCatalog(XMLCatalog)" we could just get the catalog path from the ANT XMLCatalog (which has various bugs) and create a Xerces Catalog Resolver over that path.
Or remove the method "org.dita.dost.module.XsltModule.setXMLCatalog(XMLCatalog)" completely because the "org.dita.dost.module.XsltModule.init()" method already sets the proper XML catalog resolver to the class.

PR #3516, opened 2020-05-17
Add dependency org.xmlresolver

jelovirt:
## Description
Replace dependency Apache Resolver with org.xmlresolver.

## Motivation and Context
The [org.xmlresolver](https://github.com/ndw/xmlresolver) library has features the old Apache Resolver lacks and is under more active development.

## How Has This Been Tested?
Old tests pass.
## Type of Changes

- Breaking change because dependencies change from `CatalogResolver` to `Resolver` at Java level. This only affects plug-ins that use custom Java code that use the catalog resolver.

Plug-in developer migration documentation to use `Resolver` class instead of `CatalogResolver`.

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/common/output-message.xsl
+++ b/src/main/plugins/org.dita.base/xsl/common/output-message.xsl
@@ -25,7 +25,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:param name="msg" select="'***'"/>
     <xsl:param name="msgparams" select="''"/>    
     
-    <xsl:variable name="msgdoc" select="document('platform:/config/messages.xml')" as="document-node()?"/>
+    <xsl:variable name="msgdoc" select="document('platform:config/messages.xml')" as="document-node()?"/>
     <xsl:variable name="msgcontent" as="xs:string*">
       <xsl:choose>
         <xsl:when test="$msg != '***'">
--- a/src/test/resources/bookmap7/src/p.dita
+++ b/src/test/resources/bookmap7/src/p.dita
@@ -6,7 +6,7 @@
   | (C) Copyright 2005 IBM Corporation - 2006
   *-->   
 <!DOCTYPE concept PUBLIC "-//OASIS//DTD DITA 1.3 Concept//EN"
- "concept.dtd">
+ "\concept.dtd">
 <concept id="newelementdata" xml:lang="en-us">
 <title>New element &lt;data&gt;</title>
 <conbody>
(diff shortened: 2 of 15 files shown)


jelovirt:
This is proving to be somewhat difficult to implement. Ant relies on the old catalog resolver and and we may have to keep on shipping the old catalog resolver for plug-in backwards compatibility.

raducoravu:
Some questions, what are those extra features the org.xmlresolver would have? As it may not be used in lots of implementations, it may have various issues that the stable Apache resolver does not.
About the ANT resolver which builds on top of the Apache resolver, it is indeed buggy, at some point I suggested removing the need for the ANT resolver: https://github.com/dita-ot/dita-ot/issues/3376

robander:
Removing from 3.6 project board based on discussion at contributor call: https://github.com/dita-ot/dita-ot/wiki/Meeting-minutes-2020-10-01

jelovirt:
It looks like too large and out of scope for DITA-OT to write a wrapper that would reimplement the old resolver's URI to be backed by the new resolver. So it's easier to just have _both_ the new and old resolver as dependency.

* Old resolver is used by Ant when you explicitly use `<xmlcatalog>`
* New resolver is used in all other cases.

src/main/plugins/org.dita.base/xsl/common/output-message.xsl
```diff
@@ -25,7 +25,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:param name="msg" select="'***'"/>
     <xsl:param name="msgparams" select="''"/>    
     
-    <xsl:variable name="msgdoc" select="document('platform:/config/messages.xml')" as="document-node()?"/>
+    <xsl:variable name="msgdoc" select="document('platform:config/messages.xml')" as="document-node()?"/>
```
jelovirt, line 28: Processing for custom URI schemes has changed slightly. If we use the old `platform:/config/messages.xml`, the catalog will try to resolve this to `file:/config/messages.xml`. We may want to raise this as a bug in xmlresolver.

src/test/resources/bookmap7/src/p.dita
```diff
@@ -6,7 +6,7 @@
   | (C) Copyright 2005 IBM Corporation - 2006
   *-->   
 <!DOCTYPE concept PUBLIC "-//OASIS//DTD DITA 1.3 Concept//EN"
- "concept.dtd">
+ "\concept.dtd">
```
jelovirt, line 9: Adding an invalid URI to cover invalid system identifiers in topic tests too.

jelovirt:
This will be included in #4130

jelovirt:
Implemented when updating to Saxon 12.3

infotexture:
> Implemented when updating to Saxon 12.3

@jelovirt Does this mean that we can remove “Apache XML Commons Resolver” from the list of [Third-party software](https://www.dita-ot.org/dev/reference/third-party-software)?

PR #3523, opened 2020-05-31
Extract CLI string into resource bundle

### LINKED ISSUE #3495: Extract user-facing text to strings file(s)
## Description

Currently, user-facing text in error messages, CLI help and parameter descriptions are embedded in various places in the Java code and elsewhere, which makes it difficult to find and review.

## Possible Solution

If these strings were extracted to dedicated .properties files (or another suitable format), it would make it much easier to review them all at once to ensure consistency and align the tone of messages that were created years apart by different authors.

_(This would also provide a foundation for localization, if that were ever considered for the CLI itself.)_

It would be ideal if all user-facing text could be consolidated in a single format, but given the extensible plug-in–based architecture, and variety of source code formats, I'm not sure that's practical.

As a first step, extracting strings from the core Java code would certainly be a significant improvement.

## Additional Context

Per discussion in https://github.com/dita-ot/dita-ot/pull/3491#discussion_r402084653.

jelovirt:
Extract hard-coded strings from CLI Java code into a resource bundle for easier editing.

Fixes #3495

### DIFF
--- a/src/main/resources/cli_en_US.properties
+++ b/src/main/resources/cli_en_US.properties
@@ -1,0 +1,56 @@
+#
+# This file is part of the DITA Open Toolkit project.
+#
+# Copyright 2020 Jarno Elovirta
+#
+# See the accompanying LICENSE file for applicable license.
+#
+error_msg=Error: %s
+help.option.help=Print help information
+help.option.debug=Enable debug logging
+help.option.verbose=Enable verbose logging
+# Version subcommand
+version=DITA-OT version %s
+version.usage=dita version [options]
+# Plugins subcomand
+plugins.usage=dita plugins [options]
+# Deliverables subcommand
+deliverables.usage=dita deliverables <file> [options]
+deliverables.argument.file=Project file
+deliverables.error.project_not_defined=Project file not defined
+# Install subcommand
+install.usage=dita install [<id> | <url> | <file>]
+install.argument.id=Install specified plug-in ID from registry
+install.argument.url=Install plug-in from a URL
+install.argument.file=Install plug-in from a local ZIP file
+install.option.force=Force-install plug-in (overwrite existing version)
+# Uninstall subcommand
+uninstall.usage=dita uninstall <id>
+uninstall.argument.id=Uninstall plug-in with the specified ID
+uninstall.error.identifier_not_defined=You must specify plug-in identifier when using the uninstall subcommand
+# Transtypes subcommand
+transtypes.usage=dita transtypes [options]
+# Conversion command
+conversion.usage.input=dita deliverables <file> [options]
+conversion.usage.project=dita --project=<file> [options]
+conversion.subcommand.deliverables=Print list of deliverables in project file
+conversion.subcommand.install=Install or reload plug-ins
+conversion.subcommand.plugins=Print list of installed plug-ins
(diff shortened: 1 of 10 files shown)


jelovirt:
@infotexture Do you think this would be an ok format? JSON or YAML with nested structure instead of period delimited keys would be nice, but this takes less effort to adopt.

jelovirt:
The only thing missing from the property file format is the support for array types. So that you could have an array of strings under one key, for example for usage strings. I need to think about this.

jelovirt:
Reminder to self, add key for command name and generate usages using that and subcommand name, instead of hard coding the whole usage string.

infotexture:
If we think ahead to the idea of one day extracting _all_ user-facing text to strings files (not only from the Java code), other established formats for string resources might be worth considering, such as:

- Apple's [.strings files](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html)
- Android's [strings.xml](https://developer.android.com/guide/topics/resources/string-resource) format
- [.po files](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) or
- XLIFF

That's beyond the scope of this PR of course, but it's probably worth implementing the initial changes with a view toward future use cases, such as localized toolkit versions. 🔮 

Given the mix of source formats that currently contribute text to the command-line experience (DITA, Java, XML, XSL), I'm not sure how easy it would be to consolidate to a single format for strings, but moving these strings out of the Java code is definitely a step in the right direction. 👍

jelovirt:
Would it be ok if our strings were in some XML format, but then the compilation/integration process would flatten them into a properties file. That would allow runtime to simply read a Java Resource Bundle, but development/authoring experience might be nicer?

jelovirt:
@infotexture could you check the few updates?

src/main/resources/cli_en_US.properties
```diff
@@ -0,0 +1,56 @@
...
+install.argument.file=Install plug-in from a local ZIP file
+install.option.force=Force-install plug-in (overwrite existing version)
+# Uninstall subcommand
+uninstall.usage=dita uninstall <id>
+uninstall.argument.id=Uninstall plug-in with the specified ID
+uninstall.error.identifier_not_defined=You must specify plug-in identifier when using the uninstall subcommand
+# Transtypes subcommand
+transtypes.usage=dita transtypes [options]
+# Conversion command
+conversion.usage.input=dita deliverables <file> [options]
```
jelovirt, line 34: Because we can't have an array of strings for usage, I used named usage strings.

infotexture:
> could you check the few updates?

Since this was force-pushed with a new single commit, I can't see what's changed since my previous review. 🤔 

I see the result has been merged, but I'll be happy to take another look if that still helps.

infotexture:
> Would it be ok if our strings were in some XML format, but then the compilation/integration process would flatten them into a properties file

That would be fine with me, though we'd probably want to make it clear which format is the source, and which is the generated output, so people don't try to edit the wrong file(s).

jelovirt:
The generated file would not be in Git and these files are not something that are intended to be edited by end users.

infotexture:
Right, I get that, but would the distribution package contain both, or only the generated file?

PR #3535, opened 2020-06-14
Throw warning on file reference that uses incorrect case

jelovirt:
## Description
Throw warning/error when file reference in DITA uses different case than the actual file in the file system. In strict processing mode this is considered a fatal error; in lax processing mode the file reference is rewritten to use the same case as the file system.

## Motivation and Context
When DITA map references e.g. `Topic.dita`, but the filename of the target on file system is actually `topic.dita`, we currently don't throw any error. However, this can lead to unexpected problems later on in the processing chain. By throwing a warning/error, we report issues early.

Oxygen has the same functionality in the editor, warning about the differente case.

### DIFF
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -414,7 +414,7 @@ See the accompanying LICENSE file for applicable license.
   </message>
 
   <message id="DOTJ082E" type="ERROR">
-    <reason>Resource %1 is referenced with incorrect path capitalization</reason>
+    <reason>The resource referenced as %1 is capitalized differently on disk</reason>
     <response></response>
   </message>
 
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -413,6 +413,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ082E" type="ERROR">
+    <reason>Resource %1 is referenced with incorrect path capitalization</reason>
+    <response></response>
+  </message>
+
   <!-- End of Java Messages -->
     
   <!-- Start of XSL Messages -->  
--- a/src/main/java/org/dita/dost/writer/ValidationFilter.java
+++ b/src/main/java/org/dita/dost/writer/ValidationFilter.java
@@ -17,8 +17,13 @@ import org.xml.sax.SAXException;
 import org.xml.sax.helpers.AttributesImpl;
 
 import javax.xml.namespace.QName;
+import java.io.File;
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.LinkOption;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.*;
 
 import static javax.xml.XMLConstants.XML_NS_URI;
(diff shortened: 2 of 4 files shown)


jelovirt:
@infotexture what do you think about the error message? It matches Oxygen.

@infotexture @robander In a case like this, should the skip and lax processing modes throw a warning instead, not error? Should the error code be different between the two?

robander:
My first thought is warning -- but then I remember that this has caused build failures (NPEs) on Linux while working fine on Mac/Windows. 3.5 should have fixed the NPE I know of, but that memory lingers.

I guess if it is correcting the reference, then it should probably be a warning. I can't remember right now if strict mode fails on warnings, or just errors ... if it fails on warnings then they should be able to use the same message number?

raducoravu:
It's good that you took the decision to issue the warning only when the paths are equal ignore case:

    if (!Objects.equals(absPath, canPath) && Objects.equals(absPath.toLowerCase(), canPath.toLowerCase())) {

this avoids reporting errors if a file is actually a symbolic link pointing to some other place.

src/main/config/messages_template.xml
```diff
@@ -413,6 +413,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ082E" type="ERROR">
+    <reason>Resource %1 is referenced with incorrect path capitalization</reason>
```
infotexture, line 417: Is the capitalization mismatch in the path, the filename, or either one? 
Will `%1` be expanded to the full path, or just the filename?

Error messages are better when they're specific, rather than re-using the same message for different cases. 
If we can't separate the causes (or we're not sure where the problem lies), this might be better:

> The resource referenced as %1 is capitalized differently on disk

Can we throw separate errors for strict vs. lax processing? 

Good error messages emphasize the solution and how to complete the action, so ideally, we'd add more info, like:

> Adjust the reference or filename to correct the case.

_or_

> The reference will be rewritten to match the file system case.

jelovirt: `%1` will expand to absolute path. The log message in different processing modes:

- strict: `[DOTJ082E][ERROR] Resource file:/Volumes/tmp/src/Topic.dita is referenced with incorrect path capitalization`
- skip: `[DOTJ082E][ERROR] Resource file:/Volumes/tmp/src/Topic.dita is referenced with incorrect path capitalization, using invalid value.`
- lax: `[DOTJ082E][ERROR] Resource file:/Volumes/tmp/src/Topic.dita is referenced with incorrect path capitalization, using file:/Volumes/tmp/src/topic.dita value.`

jelovirt: My preference is to use the same error code regardless of the processing mode, because the actual failure/error is the same in all of them. If our message format supported different messages per processing mode, then we could have different messages. Right now the `, using invalid value.` comes from the Java code. It's not pretty, but gets the job done.

infotexture, line 417:
```suggestion
    <reason>The resource referenced as %1 is capitalized differently on disk</reason>
```

src/main/java/org/dita/dost/writer/ValidationFilter.java
```diff
@@ -190,7 +195,35 @@ private AttributesImpl validateReference(final String attrName, final Attributes
...
+                    try {
+                        final File canFile = p.getCanonicalFile();
+                        final String absPath = p.getAbsolutePath();
+                        final String canPath = canFile.toString();
+                        if (!Objects.equals(absPath, canPath) && Objects.equals(absPath.toLowerCase(), canPath.toLowerCase())) {
+                            switch (processingMode) {
+                                case STRICT:
+                                    throw new RuntimeException(MessageUtils.getMessage("DOTJ082E", abs.toString()).setLocation(locator).toString());
+                                case SKIP:
+                                    logger.error(MessageUtils.getMessage("DOTJ082E", abs.toString()).setLocation(locator).toString() + ", using invalid value.");
```
infotexture, line 211:
```suggestion
                                    logger.error(MessageUtils.getMessage("DOTJ082E", abs.toString()).setLocation(locator).toString() + ", using authored value.");
```

src/main/java/org/dita/dost/writer/ValidationFilter.java
```diff
@@ -190,7 +195,35 @@ private AttributesImpl validateReference(final String attrName, final Attributes
...
+                                    break;
+                                case LAX:
+                                    final URI corrected = URLUtils.setFragment(currentFile.relativize(canFile.toURI()), uri.getFragment());
+                                    if (res == null) {
+                                        res = new AttributesImpl(atts);
+                                    }
+                                    res.setValue(res.getIndex(attrName), currentFile.toString());
+                                    logger.error(MessageUtils.getMessage("DOTJ082E", abs.toString()).setLocation(locator).toString() + ", using " + corrected + " value.");
```
infotexture, line 219:
```suggestion
                                    logger.error(MessageUtils.getMessage("DOTJ082E", abs.toString()).setLocation(locator).toString() + ", using " + corrected + ".");
```

_(No need to add the word "value" here, showing the corrected path is enough.)_

jelovirt:
Annoying when you accept a PR review suggestion, you have to manually add the sign-off, or fix it by rebasing.

PR #3541, opened 2020-06-28
Add rotate outputclass to entry output in HTML5

### LINKED ISSUE #3448: HTML - Support rotate on entry element
A continuation of:

https://github.com/dita-ot/dita-ot/issues/2717

We have users who want this for the Web-based outputs:

https://www.oxygenxml.com/forum/viewtopic.php?f=20&t=21076

So maybe we can propagate the attribute as an HTML "data-*" attribute and then have the builtin CSS with a rule matching this attribute and applying some kind of transform.

jelovirt:
Fixes #3448

## Description
Add `@rotate` support for OASIS table entry in HTML5 plug-in. The rotation is implemented using CSS `writing-mode:sideways-lr` because `transform:rotate(270deg)` would make the entry content overflow.

## Motivation and Context
Add parity with PDF2 in supporting the `@rotate` attribute.

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/_tables.scss
+++ b/src/main/plugins/org.dita.html5/sass/_tables.scss
@@ -97,7 +97,7 @@ $valign: top bottom middle;
 
 .entry {
   &.rotate {
-     writing-mode: vertical-lr;
+     writing-mode: sideways-lr;
   }
 }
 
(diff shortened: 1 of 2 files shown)


jelovirt:
Good point about matching PDF. IMHO `vertical-lr` looks better, but this is easy for users to override.

PR #3548, opened 2020-07-20
Refactor to use store for all temp IO

jelovirt:
## Description
Refactor to use `Store` for all temporary IO processing. Enable contributing `Store` implementations via service loader. Changes are not be visible to end users.

## Motivation and Context
This separates refactoring needed for #3439. Then the in-memory and caching `Store` implementation can be merged without refactoring changes.

## How Has This Been Tested?
All existing tests pass.

## Type of Changes

Contributing alternative `Store` implementations via `org.dita.dost.store.StoreBuilder` service loader _may_ be documented, but it's very unlikely that plugin contributed `Store` implementations will be used.

### DIFF
--- a/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
+++ b/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
@@ -307,7 +307,7 @@ public final class ExtensibleAntInvoker extends Task {
         Job job = project.getReference(ANT_REFERENCE_JOB);
         if (job != null) {
             if (job.isStale()) {
-                project.log("Reload stale job configuration reference", Project.MSG_VERBOSE);
+                project.log("Reload stale job configuration reference", Project.MSG_ERR);
                 try {
                     job = new Job(tempDir, job.getStore());
                 } catch (final IOException ioe) {
@@ -316,10 +316,16 @@ public final class ExtensibleAntInvoker extends Task {
                 project.addReference(ANT_REFERENCE_JOB, job);
             }
         } else {
+            XMLUtils xmlUtils = project.getReference(ANT_REFERENCE_XML_UTILS);
+            if (xmlUtils == null) {
+                project.log("XML utils not found from Ant project reference", Project.MSG_ERR);
+                xmlUtils = new XMLUtils();
+                xmlUtils.setLogger(new DITAOTAntLogger(project));
+            }
             Store store = project.getReference(ANT_REFERENCE_STORE);
             if (store == null) {
-                project.log("Store not found from Ant project reference", Project.MSG_VERBOSE);
-                store = new StreamStore( new XMLUtils());
+                project.log("Store not found from Ant project reference", Project.MSG_ERR);
+                store = new StreamStore(job.tempDir, xmlUtils);
             }
             project.log("Job not found from Ant project reference", Project.MSG_VERBOSE);
             try {
--- a/src/main/java/org/dita/dost/ant/InitializeProjectTask.java
+++ b/src/main/java/org/dita/dost/ant/InitializeProjectTask.java
@@ -12,14 +12,14 @@ import org.apache.tools.ant.Project;
 import org.apache.tools.ant.Task;
 import org.dita.dost.log.DITAOTAntLogger;
 import org.dita.dost.store.Store;
-import org.dita.dost.store.StreamStore;
+import org.dita.dost.store.StoreBuilder;
 import org.dita.dost.util.CatalogUtils;
 import org.dita.dost.util.XMLUtils;
 
 import java.io.File;
+import java.util.ServiceLoader;
 
-import static org.dita.dost.util.Constants.ANT_REFERENCE_STORE;
-import static org.dita.dost.util.Constants.ANT_REFERENCE_XML_UTILS;
+import static org.dita.dost.util.Constants.*;
 import static org.dita.dost.util.URLUtils.toFile;
 
 /**
@@ -28,6 +28,11 @@ import static org.dita.dost.util.URLUtils.toFile;
  * @since 3.5
  */
 public final class InitializeProjectTask extends Task {
+
+    private static ServiceLoader<StoreBuilder> storeBuilderLoader = ServiceLoader.load(StoreBuilder.class);
+
+    private String storeType = "file";
+
     @Override
     public void execute() throws BuildException {
         log("Initializing project", Project.MSG_INFO);
@@ -42,10 +47,28 @@ public final class InitializeProjectTask extends Task {
             xmlUtils.setLogger(new DITAOTAntLogger(getProject()));
             getProject().addReference(ANT_REFERENCE_XML_UTILS, xmlUtils);
         }
+        final Store store = getStore(xmlUtils);
+        getProject().addReference(ANT_REFERENCE_STORE, store);
+    }
+
+    private Store getStore(XMLUtils xmlUtils) {
         Store store = getProject().getReference(ANT_REFERENCE_STORE);
-        if (store == null) {
-            store = new StreamStore(xmlUtils);
-            getProject().addReference(ANT_REFERENCE_STORE, store);
+        if (store != null) {
+            return store;
         }
+        File tempDir = toFile(getProject().getUserProperty(ANT_TEMP_DIR));
+        if (tempDir == null) {
+            tempDir = toFile(getProject().getProperty(ANT_TEMP_DIR));
+        }
+        for (StoreBuilder storeBuilder : storeBuilderLoader) {
+            if (storeBuilder.getType().equals(storeType)) {
+                return storeBuilder.setTempDir(tempDir).setXmlUtils(xmlUtils).build();
+            }
+        }
+        throw new BuildException(String.format("Unsupported store type %s", storeType));
+    }
+
+    public void setStoreType(final String storeType) {
+        this.storeType = storeType;
     }
 }
--- a/src/main/java/org/dita/dost/ant/types/JobSourceSet.java
+++ b/src/main/java/org/dita/dost/ant/types/JobSourceSet.java
@@ -20,6 +20,7 @@ import org.apache.tools.ant.types.ResourceCollection;
 import org.apache.tools.ant.types.resources.FileResource;
 import org.apache.tools.ant.types.resources.URLResource;
 import org.dita.dost.ant.ExtensibleAntInvoker;
+import org.dita.dost.store.ant.types.StoreResource;
 import org.dita.dost.util.Constants;
 import org.dita.dost.util.FileUtils;
 import org.dita.dost.util.Job;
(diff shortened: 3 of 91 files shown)


src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
```diff
@@ -316,10 +316,16 @@ public static Job getJob(final Project project) {
                 project.addReference(ANT_REFERENCE_JOB, job);
             }
         } else {
+            XMLUtils xmlUtils = project.getReference(ANT_REFERENCE_XML_UTILS);
```
jelovirt, line 319: Store `XMLUtils` into Ant project reference, because we want to reuse the same Saxon `Processor` everywhere. This allows reusing the same name pool etc. that's needed for in-memory `Store` implementation.

src/main/java/org/dita/dost/ant/InitializeProjectTask.java
```diff
@@ -28,6 +28,11 @@
  * @since 3.5
  */
 public final class InitializeProjectTask extends Task {
+
+    private static ServiceLoader<StoreBuilder> storeBuilderLoader = ServiceLoader.load(StoreBuilder.class);
```
jelovirt, line 32: Use service loader to load `StoreBuilder` which in turn can be used to create the actual `Store` implementations.

src/main/java/org/dita/dost/ant/InitializeProjectTask.java
```diff
@@ -28,6 +28,11 @@
...
+
+    private String storeType = "file";
```
jelovirt, line 34: Default `StoreBuilder` name, plain old temporary directory in disk implementation.

src/main/java/org/dita/dost/ant/types/JobSourceSet.java
```diff
@@ -58,9 +59,9 @@ public JobSourceSet() {
             for (final FileInfo f : job.getFileInfo(this::filter)) {
                 log("Scanning for " + f.file.getPath(), Project.MSG_VERBOSE);
                 final File tempFile = new File(job.tempDir, f.file.getPath());
-                if (tempFile.exists()) {
+                if (job.getStore().exists(tempFile.toURI())) {
```
jelovirt, line 62: All cases where we used to test file existance with `File.exists()`, we use `Store.exists(URI)` instead. The default `StreamStore` implementation will just test if the file exists. The added abstraction here allows in-memory `Store` to first check if the resource exists in memory, then fall back to checking if the resource exists in disk.

src/main/java/org/dita/dost/ant/types/JobSourceSet.java
```diff
@@ -58,9 +59,9 @@ public JobSourceSet() {
...
                     log("Found temporary directory file " + tempFile, Project.MSG_VERBOSE);
-                    res.add(new FileResource(job.tempDir, f.file.toString()));
+                    res.add(new StoreResource(job, job.tempDirURI.relativize(f.uri)));
```
jelovirt, line 64: Custom Ant resource that can read from any `Store`, not just disk.

src/main/java/org/dita/dost/module/BranchFilterModule.java
```diff
@@ -374,10 +375,6 @@ private void generateCopies(final Element topicref, final List<FilterUtils> filt
                 writer.setCurrentFile(dstAbsUri);
                 final List<XMLFilter> pipe = singletonList(writer);
 
-                final File dstDirUri = new File(dstAbsUri.resolve("."));
```
jelovirt, line 377: No need to create directories into temporary directory in disk, the `Store` will handle that when resource is created.

src/main/java/org/dita/dost/module/DebugAndFilterModule.java
```diff
@@ -167,9 +161,9 @@ private void processFile(final FileInfo f) {
 
             in = new InputSource(f.src.toString());
 
-            final Serializer result = processor.newSerializer(outputFile);
+            final ContentHandler result = job.getStore().getContentHandler(outputFile.toURI());
```
jelovirt, line 164: Instead of manually creating a S9api `Serializer` ask for a `ContentHandler` from the `Store`; all IO to temp goes through `Store`.

src/main/java/org/dita/dost/module/MoveLinksModule.java
```diff
@@ -60,6 +62,8 @@ public AbstractPipelineOutput execute(final AbstractPipelineInput input) throws
 
             final XsltTransformer transformer = xsltCompiler.compile(new StreamSource(styleFile)).load();
             transformer.setErrorListener(toErrorListener(logger));
+            transformer.setURIResolver(new DelegatingURIResolver(CatalogUtils.getCatalogResolver(), job.getStore()));
```
jelovirt, line 65: The `DelegatingURIResolver` allow chaining XML catalog resolver and `Store` (which implements `URIResolver`), so that in the end `Store` will return the `Source` that is used to read the resource.

src/main/java/org/dita/dost/module/XsltModule.java
```diff
@@ -169,40 +151,62 @@ private void transform(final File in, final File out) throws DITAOTException {
             logger.debug("Set parameter " + filedirparameter + " to '" + v + "'");
             t.setParameter(new QName(filedirparameter), new XdmAtomicValue(v));
         }
+
+        if (properties.isEmpty()) {
```
jelovirt, line 155: If serialization properties (encoding, indentation etc.) from `<xslt>` element is empty, we can defer the transformation to `Store`. This means `Store` decides how to serialize the files. This allows in-memory `Store` to just create the resources and replace them in the internal cache, instead of first creating them in the cache with a temporary resource URI and then replacing them into the final resource URI.

src/main/java/org/dita/dost/module/XsltModule.java
```diff
@@ -169,40 +151,62 @@ private void transform(final File in, final File out) throws DITAOTException {
...
+            final Source source = job.getStore().getSource(in.toURI());
             t.setSource(source);
-            final Serializer serializer = processor.newSerializer(tmp);
-            for (final String key: properties.stringPropertyNames()) {
-                serializer.setOutputProperty(new QName(key), properties.getProperty(key));
+            final Destination destination = job.getStore().getDestination(tmp.toURI());
+            if (same) {
+                destination.setDestinationBaseURI(out.toURI());
+            }
+            if (destination instanceof Serializer) {
```
jelovirt, line 186: Only set output properties if the destination is a Serializer.

`Destination` can be e.g. a tree builder that results in a S9api `XdmNode`, or then it can be something that serializes XML into a byte stream.

src/main/java/org/dita/dost/module/reader/TopicReaderModule.java
```diff
@@ -180,54 +184,43 @@ public void readStartFile() throws DITAOTException {
...
-                            if (format == null) {
-                                format = in.getAttributeValue(null, ATTRIBUTE_NAME_FORMAT);
-                            }
-                            res.add(new Reference(fi.src, format));
-                            nonConrefCopytoTargetSet.add(fi.src);
-                        }
-                        break;
-                    default:
-                        break;
+            final XdmNode source = job.getStore().getImmutableNode(tmp);
```
jelovirt, line 188: Change reading resource from XML stream reader to S9api `XdmNode`. `XdmNode` is the XSLT/XPath document model that offers a fluent query API.

src/main/java/org/dita/dost/module/reader/TopicReaderModule.java
```diff
@@ -180,54 +184,43 @@ public void readStartFile() throws DITAOTException {
...
+            logger.info("Reading " + tmp);
+            final Predicate<? super XdmNode> isTopicref = xdmItem -> MAP_TOPICREF.matches(xdmItem.getAttributeValue(QNAME_CLASS));
```
jelovirt, line 190: Off-topic, In cases like this, local variable in Java 10 would be nice for readability:

```java
var isTopicref = xdmItem -> MAP_TOPICREF.matches(xdmItem.getAttributeValue(QNAME_CLASS));
```

src/main/java/org/dita/dost/reader/ChunkMapReader.java
```diff
@@ -232,7 +231,7 @@ private void chunkMap(final Element root) {
      * Create the new topic stump.
      */
     private void createTopicStump(final URI newFile) {
-        try (final OutputStream newFileWriter = new FileOutputStream(new File(newFile))) {
+        try (final OutputStream newFileWriter = job.getStore().getOutputStream(newFile)) {
```
jelovirt, line 234: Chunking code still uses byte streams to write the temporary XML files. Until chunking code is refactored or replaces, the `Store` API needs to support `InputStream` and `OutputStream`.

src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
```diff
@@ -697,12 +705,12 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
       <xsl:when test="empty($targetElement)"/>
       <!--otherwise try pulling shortdesc from target-->
       <xsl:otherwise>
-        <xsl:variable name="shortdesc">
+        <xsl:variable name="shortdesc" as="node()*">
```
jelovirt, line 708: Without typing this variable will create a new document node as the root and we don't want that.

src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
robander: Just curious -- in the past we had memory crashes in steps where Saxon ended up reading 5,000 or 10,000 topics (like the `mappull` step). Is reusing the same Saxon `Processor` going to have any impact on that sort of case, either better or worse, or is there no change?

src/main/java/org/dita/dost/store/Store.java
```diff
@@ -62,7 +97,7 @@
     Source getSource(URI path);
 
     /**
-     * Get immutable DOM document for file. If mutating methods are called,
+     * Get immutable DOM document for resouce. If mutating methods are called,
```
robander, line 100: Really trivial comment, would be good to fix the typo "resouce", here and a few other spots

src/main/java/org/dita/dost/store/ant/types/StoreResource.java
```diff
@@ -0,0 +1,160 @@
...
+    public String getName() {
+        return file.getPath();
+    }
+
+    /**
+     * Learn whether this file exists.
+     * @return true if this resource exists.
+     */
+    @Override
+    public boolean isExists() {
```
robander, line 45: Another trivial comment that `isExists` feels like a weird because you're not checking if a file is exists, you're checking whether `fileExists` or `uriExists` or whatever?

robander:
General question ... some of the most frustrating failures I've had with DITA-OT over the last couple years come from URI Exceptions, when the input is not a valid URI and the code crashes (without telling you what URI it was trying to validate).

I see a lot of `URI` variables in the new code -- is there any new testing for bad input URIs? Are external links going to be rewritten in any way, or are those left alone? For example:
`<object data="https://www.youtube.com/watch?v=5gdIvjF99ac">` (object can't set scope=external)
`<xref href="www.youtube.com/watch?v=5gdIvjF99ac">` (forgot the https)
`<topicref href="missing file.dita">` (space instead of `%20` )

jelovirt:
All/most the places where we've added a `URI` field, are to replace direct `InputStream` use. So I don't see it as a risk that we start breaking old content that has invalid links. All those invalid links should be fixed during initial read, in the normalization/fix filters.

src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
jelovirt: No, the `Processor` class is what maintains the name pool etc. The `XsltExecutable` is the one that would be a problem and we don't reuse it by default.

src/main/java/org/dita/dost/store/ant/types/StoreResource.java
jelovirt: It overrides a method from Ant's `Resource` class, so we can't rename it to anything that makes more sense.

src/main/java/org/dita/dost/store/Store.java
jelovirt: Fixed

PR #3549, opened 2020-07-21
Use unfiltered source topic as basis for topic fileinfo features

jelovirt:
When map contains branch filtered topics with preprocess2, in some cases the filtered topic is used as the base topic when storing topic features into Job state. In this case only the filtered topic gets the topic features info and the actual source topic is left with the defaults.

This PR attempts to fix the issue by by using topic features from the resource that has the same the source URI and result URI and matches the topic being processed.

### DIFF
--- a/src/main/java/org/dita/dost/module/reader/AbstractReaderModule.java
+++ b/src/main/java/org/dita/dost/module/reader/AbstractReaderModule.java
@@ -30,6 +30,7 @@ import javax.xml.namespace.QName;
 import java.io.*;
 import java.net.URI;
 import java.util.*;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -51,7 +52,7 @@ public abstract class AbstractReaderModule extends AbstractPipelineModuleImpl {
 
     Predicate<String> formatFilter;
     /** FileInfos keyed by src. */
-    private final Map<URI, FileInfo> fileinfos = new HashMap<>();
+    private final Map<URI, Collection<FileInfo>> fileinfos = new HashMap<>();
     /** Set of all topic files */
     final Set<URI> fullTopicSet = new HashSet<>(128);
     /** Set of all map files */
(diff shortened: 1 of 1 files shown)


src/main/java/org/dita/dost/module/reader/AbstractReaderModule.java
```diff
@@ -639,84 +640,91 @@ void outputResult() throws DITAOTException {
         resourceOnlySet.addAll(res);
 
         for (final URI file: outDitaFilesSet) {
-            getOrCreateFileInfo(fileinfos, file).isOutDita = true;
+            createOrUpdateFileInfo(file, fi -> fi.isOutDita = true);
```
jelovirt, line 643: Instead of getting the created or old `FileInfo` and updating it directory, we pass an update function that does the updating.

src/main/java/org/dita/dost/module/reader/AbstractReaderModule.java
```diff
@@ -51,7 +52,7 @@
 
     Predicate<String> formatFilter;
     /** FileInfos keyed by src. */
-    private final Map<URI, FileInfo> fileinfos = new HashMap<>();
+    private final Map<URI, Collection<FileInfo>> fileinfos = new HashMap<>();
```
jelovirt, line 55: Instead of mapping _some_ `FileInfo` here by source URI, the value is a list of all `FileInfo` objects that has this source URI.

src/main/java/org/dita/dost/module/reader/AbstractReaderModule.java
```diff
@@ -813,33 +822,44 @@ private String getPrefix(final File relativeRootFile) {
...
     }
 
-    private FileInfo getOrCreateFileInfo(final Map<URI, FileInfo> fileInfos, final URI file) {
+    private void createOrUpdateFileInfo(final URI file, final Consumer<FileInfo> consumer) {
+        for (final FileInfo fi : getOrCreateFileInfo(fileinfos, file)) {
+            consumer.accept(fi);
+        }
+    }
+
+    private Collection<FileInfo> getOrCreateFileInfo(final Map<URI, Collection<FileInfo>> fileInfos, final URI file) {
```
jelovirt, line 831: Get all `FileInfo` objects by source URI, not just the first (in unstable order).

PR #3556, opened 2020-08-05
Add support for multiple filter options in CLI 

jelovirt:
## Description
Allow defining multiple filters in CLI by repeating `--filter` option.

## Motivation and Context
Repeating is already supported by `--resource` option. For consistency, all multi-value options should be supported via repeated options.

## Type of Changes

The current documentation for `--filter` has:
> Note:
>
> To specify multiple filter files, use the system path separator character to delimit individual file paths (semicolon ‘;’ on Windows, and colon ‘:’ on macOS and Linux) and wrap the value in quotes:
>
> `--filter="filter1.ditaval;filter2.ditaval;filter3.ditaval"`

This should be amended with the possibility of repeating the option, e.g.

```
--filter=filter1.ditaval --filter=filter2.ditaval --filter=filter3.ditaval
```
We also need some way in the CLI help to show that multiple repetitions are supported.


jelovirt:
@infotexture Could you update https://github.com/dita-ot/dita-ot/blob/feature/multiple-filter-argument/src/main/resources/cli_en_US.properties#L45..L46 with the copy changes discussed in Slack?

infotexture:
@jelovirt Done. https://github.com/dita-ot/dita-ot/pull/3556/commits/ae242bbc7c36f3bb38f0a7ad3670d2b42ce12a0b ☝️

infotexture:
If we end up re-using this more, we might want to add a variable for `This option can be passed multiple times.`

But I figured it's best to keep it simple for now. Feel free to optimize if you prefer.

jelovirt:
Looks good and I added tests for these.

PR #3562, opened 2020-08-18
Remove deprecated msgprefix parameter

robander:
## Description

The `msgprefix` variable used by `output-message` was deprecated in 2.3, along with `magnum`, `msgcat`, and `msgsev` parameters in that template. Until it's removed, every program including or importing `output-message.xsl` has to declare the deprecated variable.

This pull request finally removes the deprecated parameters. It also fixes one lingering message in `conref` processing that still used the old method.

## Motivation and Context

Minor annoyance for many years finally caught up with me when I had to make a new step recently, and someone not familiar with DITA-OT asked why I was declaring that variable but not using it.

## How Has This Been Tested?

Tests initially failed because of the message in `conref` that still used the old parameters; tests all pass after that message was updated to use the proper syntax.

## Type of Changes

I think this counts as a breaking change; even though the old style was deprecated in 2.3, it still worked before this change, as demonstrated by the old call in conref processing.

Requires an item in the migration topic for the next release.

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/common/output-message.xsl
+++ b/src/main/plugins/org.dita.base/xsl/common/output-message.xsl
@@ -8,21 +8,10 @@ See the accompanying LICENSE file for applicable license.
 -->
 
 <!--
-  Standard error message template for DITA processing in XSL. This
-  file should be included by any XSL program that uses the standard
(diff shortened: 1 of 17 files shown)


jelovirt (approved):
LGTM.

PR #3563, opened 2020-08-20
Retain mapref keyscope in retable

### LINKED ISSUE #2393: Reltable links not properly reflected in output when using key scopes.
I'm attaching a sample project here:
https://www.oxygenxml.com/forum/files/reltableKeyScopes.zip

The relathioship table maps a topic to 2 topicrefs, each topicref pointing to the same topic but in a different key scope:

				<reltable>
					<relrow>
						<relcell>
							<topicref keyref="key_ref-common" />
						</relcell>
						<relcell>
							<topicref keyref="keyscope_prod-A.key_topic-reused_prod-A" />
							<topicref keyref="keyscope_prod-B.key_topic-reused_prod-B" />
						</relcell>
					</relrow>
				</reltable>

but in the published output the common.html only contains a link to the "Product A" topic and the link to the "Product B" topic is missing.

jelovirt:
## Description
Cascade `@keyscope` from `<mapref>` into `<reltable>` of the target map.

## Motivation and Context
Fixes #2393

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
@@ -174,7 +174,17 @@ See the accompanying LICENSE file for applicable license.
             <!-- retain key definition as a separate element -->
             <xsl:if test="@keys">
               <keydef class="+ map/topicref mapgroup-d/keydef ditaot-d/keydef " processing-role="resource-only">
-                <xsl:apply-templates select="@* except (@class | @processing-role)"/>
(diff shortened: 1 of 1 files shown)


jelovirt:
@raducoravu I refactored the code in the #2393 comment and it seems to work against the test set that was in the issue. Are you able to verify that this fix still works as expected and fixes the issue #2393?

raducoravu (approved):
I can confirm the relative links between the topics are present and that the NullPointerException is gone. There are still some errors in the branch-filter stage reported during the HTML5 publishing but they can be probably handled separately.

PR #3567, opened 2020-09-01
Fix sibling keyscope cascading

### LINKED ISSUE #2410: Keyref with inner key scope fails in XHTML output
I uploaded a set of resources here:
https://www.oxygenxml.com/forum/files/keyRefInnerScope.zip

Basically the DITA Map looks like this:

```
 <topicref href="test1.dita" keyscope="scope1">
  <topicref href="test2.dita" keyscope="scope2">
   <topicref href="test3.dita" keyscope="scope3">
    <topicref href="reused.dita" keys="key"/>
   </topicref>
   <topicref href="reusing.dita"/>
  </topicref>
 </topicref>
```

and inside the "reusing.dita" I have this conkeyref:

```
 <p conkeyref="scope3.key/reuse-me"/>
```

but the publishing engine cannot resolve it, it reports:

```
  [DOTJ046E][ERROR] Conkeyref="scope3.key/reuse-me" can not be resolved because it does not contain a key or the key is not defined. The build will use the conref attribute for fallback, if one exists.
```

Is it a bug in the engine or in how I understand the specification?

jelovirt:
## Description
Fix cascading child key scopes during key processing to correctly generate key definitions from sibling key scopes

## Motivation and Context
Fixes #2410

## How Has This Been Tested?
Added unit tests. Old test integration tests and test case in the issue passes.

## Type of Changes

No documentation changes needed.

### DIFF
--- a/src/main/java/org/dita/dost/reader/KeyrefReader.java
+++ b/src/main/java/org/dita/dost/reader/KeyrefReader.java
@@ -16,9 +16,11 @@ import static org.dita.dost.util.XMLUtils.*;
 import java.io.File;
 import java.net.URI;
 import java.util.*;
+import java.util.stream.Collectors;
 
 import javax.xml.parsers.DocumentBuilder;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.dita.dost.log.MessageBean;
 import org.dita.dost.log.MessageUtils;
 import org.dita.dost.util.Job;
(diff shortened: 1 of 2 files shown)


jelovirt:
Yes, those `FIXME` cases were added when I originally implemented this and should have worked pre DITA spec. Now they seem to work, so this improves spec compliance.

PR #3569, opened 2020-09-03
Cache DitaClass instances

jelovirt:
## Description
Add a caching `DitaClass.getInstance(cls)` method to replace direct `DitaClass` instance generation via constructor.

## Motivation and Context
By caching immutable `DitaClass` instances fewer instances need to be created.

## How Has This Been Tested?
Tests pass.
## Type of Changes

Documented in Javadoc.

Migration documentation should state the `DitaClass` constructor has been changed to package-private and plug-ins should use `getInstance` factory method.


raducoravu (approved):
It would have been safer to have a private constructor for ditaclass. Maybe update tests to use new method. Or add public method createInstanceFromTests(classVal) used only in tests?

jelovirt:
Added test and made the constructor package-private.

PR #3575, opened 2020-09-07
Fix table normalization for multiple groups and broken tables

### LINKED ISSUE #3566: Unhandled java error: java.lang.IndexOutOfBoundsException
## Expected Behavior

I expect the toolkit to rescue this Java error and return a DITA-OT error telling me that the file could not be transformed. The DITA build should either continue without this file, or exit at this point.

The issue I came across is that this failed silently in our build pipeline and ended up uploading an empty folder. 

Admittedly, the file that's causing the issue is a total mess. Cleaning up the many DITA validation warnings in the file solved the issue. I've attached a map with the original file and a fixed version. Because there are so many issues with the file, it was difficult to ascertain which of the issues is causing the java error. Our previous pipeline used 3.3.3 and did not produce the error.

## Actual Behavior
Running this map with a clean version of 3.5.3 (no plugins) causes a java error:
```
Error: java.lang.IndexOutOfBoundsException: Index 3 out-of-bounds for length 3
``` 
The build exits with a 0, so if I'm running multiple builds in a script, there is no indication that it failed without consulting the DITA log.

## Steps to Reproduce

1. Unzip the attached archive
2. Run an HTML transformation on the file: `path/to/dita -i test.ditamap -o html -f html5 -d`

* DITA-OT version: 3.5.3
* Operating system and version: MacOS 10.15.6
* How did you run DITA-OT? `path/to/dita -i test.ditamap -o html -f html5 -d`

Attachement: [java_error.zip](https://github.com/dita-ot/dita-ot/files/5152581/java_error.zip)

### LINKED ISSUE #3573: PDF multi-group table publishing fails
## Expected Behavior
Each `tgroup` must be rendered correctly.

image

## Actual Behavior
When the first `tgroup` has `cols` greater than then next one, the publishing fails. However, if you switch the two it publishes fine.

## Steps to Reproduce

Sample Topic:
```
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
<topic id="topic_id">
    <title></title>
    <body>
        <table frame="all" rowsep="1" colsep="1" id="table_id_1">
            <title/>
            <tgroup cols="2">
                <colspec colname="c1" colnum="1" colwidth="1*"/>
                <colspec colname="c2" colnum="2" colwidth="1*"/>
                <tbody>
                    <row>
                        <entry>A</entry>
                        <entry>B</entry>
                    </row>
                </tbody>
            </tgroup>
            <tgroup cols="3">
                <colspec colname="c1" colnum="1" colwidth="1*"/>
                <colspec colname="c2" colnum="2" colwidth="1*"/>
                <colspec colname="c3" colnum="3" colwidth="1*"/>
                <tbody>
                    <row>
                        <entry>A</entry>
                        <entry>B</entry>
                        <entry>C</entry>
                    </row>
                </tbody>
            </tgroup>
        </table>
    </body>
</topic>
```
[sample.zip](https://github.com/dita-ot/dita-ot/files/5172678/sample.zip)
1. Running it in with pdf format should give you the error.

Error: java.lang.IndexOutOfBoundsException: Index: 2, Size: 2

## Environment

* DITA-OT version: DITA-OT 3.5.2
  _(Windows)_
* How did you run DITA-OT?
  _(`dita` command)_
* Transformation type:
  _(PDF)_

jelovirt:
Fix table normalization for multiple groups. Add rudimentary recovery from broken tables.

Fixes #3566
Fixes #3573

### DIFF
--- a/src/main/java/org/dita/dost/writer/NormalizeTableFilter.java
+++ b/src/main/java/org/dita/dost/writer/NormalizeTableFilter.java
@@ -141,41 +141,49 @@ public class NormalizeTableFilter extends AbstractXMLFilter {
         getContentHandler().endElement(NULL_NS_URI, TOPIC_COLSPEC.localName, TOPIC_COLSPEC.localName);
     }
 
-    private void processEntry(AttributesImpl res) {
-        tableState.columnNumber = getStartNumber(res, tableState.columnNumberEnd);
-        final int colspan = getColSpan(res);
-        final int rowspan = getRowSpan(res);
-        Span prev;
-        if (tableState.previousRow != null) {
-            for (prev = tableState.previousRow.get(tableState.currentColumn); prev != null && prev.y > 1; prev = tableState.previousRow.get(tableState.currentColumn)) {
-                for (int i = 0; i < prev.x; i++) {
-                    tableState.currentColumn = tableState.currentColumn + 1; //prev.x - 1;
-                    grow(tableState.currentRow, tableState.currentColumn + 1);
-//                    tableState.currentRow.set(tableState.currentColumn, null);
+    private void processEntry(AttributesImpl res) throws SAXException {
+        try {
+            tableState.columnNumber = getStartNumber(res, tableState.columnNumberEnd);
+            final int colspan = getColSpan(res);
+            final int rowspan = getRowSpan(res);
+            Span prev;
+            if (tableState.previousRow != null) {
+                for (prev = tableState.previousRow.get(tableState.currentColumn); prev != null && prev.y > 1; prev = tableState.previousRow.get(tableState.currentColumn)) {
+                    for (int i = 0; i < prev.x; i++) {
+                        tableState.currentColumn = tableState.currentColumn + 1; //prev.x - 1;
+                        grow(tableState.currentRow, tableState.currentColumn + 1);
+//                        tableState.currentRow.set(tableState.currentColumn, null);
+                    }
                 }
+            } else {
+                prev = new Span(1, 1);
             }
-        } else {
-            prev = new Span(1, 1);
-        }
-        grow(tableState.currentRow, tableState.currentColumn + colspan);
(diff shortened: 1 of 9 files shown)


src/main/java/org/dita/dost/writer/NormalizeTableFilter.java
```diff
@@ -142,48 +141,56 @@ private void generateColSpec() throws SAXException {
...
+            }
+            // Add extensions
+            XMLUtils.addOrSetAttribute(res, DITA_OT_NS, ATTR_X, DITA_OT_NS_PREFIX + ":" + ATTR_X, "CDATA", Integer.toString(tableState.currentColumn + 1));
+            XMLUtils.addOrSetAttribute(res, DITA_OT_NS, ATTR_Y, DITA_OT_NS_PREFIX + ":" + ATTR_Y, "CDATA", Integer.toString(tableState.rowNumber));
 
-        tableState.currentColumn = tableState.currentColumn + colspan;
-        tableState.columnNumberEnd = getEndNumber(res, tableState.columnNumber);
+            tableState.currentColumn = tableState.currentColumn + colspan;
+            tableState.columnNumberEnd = getEndNumber(res, tableState.columnNumber);
+        } catch (IndexOutOfBoundsException e) {
```
jelovirt, line 180: This is kinda shotgun solution to broken tables that may cause exceptions in the normalization code. But it's better than failing or not logging anything.

raducoravu: Why not add an extra stop condition in the for loop where the problem originates instead of catching a runtime exception?

raducoravu (approved):

raducoravu (approved):

raducoravu:
I tested and the problems reported in #3566 and #3573 no longer exist.

PR #3578, opened 2020-09-08
Use separate cache maps by source and result in Job

jelovirt:
## Description
Use separate cache maps by source and result for`FileInfo`.

## Motivation and Context
Improve lookups by source or result.

## Type of Changes

None needed

### DIFF
--- a/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
+++ b/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
@@ -182,12 +182,12 @@ public class CleanPreprocessModule extends AbstractPipelineModuleImpl {
         if (rewriteTransformer != null) {
             try {
                 final DOMSource source = new DOMSource(serialize(fis));
-                final Map<URI, FileInfo> files = new HashMap<>();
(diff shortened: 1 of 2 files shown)


jelovirt:
@raducoravu are you able to test this? Note this is against `develop` branch, so it doesn't have the grammar caching changes. What is the difference between this and the base. Is the difference in `Job.getFileInfo(URI file)` significant? This makes the code more complex, thus I want to merge this only if the performance difference is noticeable.

raducoravu:
@jelovirt the problem with "Job.getFileInfo" seemed to be more evident in the "filter" stage when one of those huge DITA Maps was processed. I'm attaching some before and after profiling snapshots. You can see the time it takes for "getFileInfo" is faster. But it's hard for me to give you timings, like how many seconds it takes to parse the largest map with or without the fix. In the large scheme of things, taking into account that probably the XSLT stage make take more than the pre-processing any way... I don't know. Maybe you can create a wrapper class over those 3 hashmaps to hide the complexity.

Screen Shot 2020-09-04 at 2 10 01 PM
Screen Shot 2020-09-09 at 12 51 41 PM

PR #3585, opened 2020-09-14
Use S9api in keyref processing

jelovirt:
## Description
Use S9api's `XdmNode` instead of DOM in keyref processing.
## Motivation and Context
`XdmNode` is an easier object model to work with.
## How Has This Been Tested?
Existing tests pass.
## Type of Changes

No documentation needed.


robander (dismissed):
I'll admit it's hard to review the whole thing here but looks good. I can help test the change by running against some large doc sets and compare the output.

robander:
FWIW, verified that with our plugins and a single large content set, the output matches between this branch and `develop` (apart from changes to a handful of generated IDs).

My PDF build failed with `java.util.NoSuchElementException` but it also fails with that on `develop`, so that should be unrelated to this update.

PR #3594, opened 2020-09-27
Add generator=DITA-OT metadata to HTML header

jelovirt:
## Description
Add `<meta name="generator" content="DITA-OT">` metadata to HTML header.

## Motivation and Context
This allows us to get a better understanding how much DITA-OT is used to publish to HTML. Users can remove the `generator` metadata using a plug-in.

## Type of Changes

This doesn't affect HTML output consumption, so it requires no documentation from that perspective.


infotexture (dismissed):
👍 I think this is a good idea.

If we add this, it would probably make sense to preserve it in site output per https://github.com/dita-ot/org.dita-ot.html/pull/10#issuecomment-687387244.

PR #3595, opened 2020-09-27
Extract Dublin Core metadata field generation to own plug-in

jelovirt:
## Description
Extract Dublin Core metadata field generation to own [plug-in](https://github.com/dita-ot/org.dita.html5.dublin-core/) in HTML5.

## Motivation and Context
Make it easier to remove Dublin Core output and modularize Dublin Core specific code.

## How Has This Been Tested?
Existing tests pass.

## Type of Changes

Change is backwards compatible.


src/main/plugins/org.dita.html5/xsl/dublin-core.xsl
```diff
@@ -0,0 +1,216 @@
...
+          <xsl:if test="position() > 2">
+            <xsl:text>,</xsl:text>
+          </xsl:if>
+          <xsl:value-of select="normalize-space(text()[1])"/>
+        </xsl:if>
+      </xsl:for-each>
+    </xsl:variable>
+
+    <xsl:if test="string-length($keywords-content) > 0">
+      <meta name="DC.subject" content="{$keywords-content}"/>
```
robander, line 45: I think this means that the `<meta name="keywords">` will be dropped?

robander: Nevermind, I see the `next-match` now.

robander:
For everything that moves to the DC file, is there a no-op template they're overriding in `get-meta.xsl`? Just thinking it would be nice to be able to delete that line from your own HTML5 shell and have that drop the DC items, but it would be a pain if you had to add a no-op `match="*" mode="gen-metadata"` to prevent random text from showing up.

jelovirt:
@robander this already works like that, by default you get no extra metadata and only if you install the DC plugin you get that DC metadata.

robander (approved):

PR #3598, opened 2020-09-29
Read SVG images to collect linked bitmap images

jelovirt:
## Description
Read SVG images to collect linked bitmap images.

## Motivation and Context
Support images with linked bitmap images.

## Type of Changes


raducoravu:
There is a similar need expressed here:
https://www.oxygenxml.com/forum/post65675.html#p65675

In this case the end user has the SVG content placed directly inside the DITA topic.

chrispy-snps:
@jelovirt - this would be a nice feature to have. (Maybe this would be a nice addition to 4.0 if you feel it's ready to go?)

We have screen captures of our software that we add captions to (boxes, arrows, captions, etc.). Currently we embed the screen captures as bitmaps in the SVG file, which is quite inefficient. It would be cleaner to keep the screen captures as native PNG files, then annotate them in an annotations-only SVG that links to them.

PR #3602, opened 2020-10-03
Remove obsolete flag stylesheet and serialization of schemekeydef.xml

jelovirt:
## Description
Remove obsolete`flag.xsl` stylesheet and generated `schemekeydef.xml`.

## Motivation and Context
Remove obsolete and unused code.

## How Has This Been Tested?
Tests pass.

## Type of Changes
- New feature _(non-breaking change which removes old functionality)_


robander:
I'm pretty sure these were used before there was a preprocess step for flagging, so ... probably unused since early 2.x timeframe. Sounds like it is past time to remove them.

robander (approved):

robander:
The X/HTML code did still refer to the `schemekeydef.xml` file until much more recently, I remember that causing me some problems -- but it was not doing anything other than causing problems by looking for the file.

PR #3606, opened 2020-10-10
Add support for running module processing in parallel

jelovirt:
## Description
Add support for running preprocessing module code in parallel when enabled.

## Motivation and Context
May make processing faster some types and sizes of input sets, depending on the processing platform.

Parallel processing is enabled by setting `parallel` property to `true`; default is `false`.

## Type of Changes

Default for this feature is backwards compatible because parallel processing is disabled.

### DIFF
--- a/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
+++ b/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
@@ -264,7 +264,6 @@ public final class ExtensibleAntInvoker extends Task {
                 }
             }
             final AbstractPipelineModule module = factory.createModule(m.getImplementation());
-            module.setParallel(m.parallel);
             module.setProcessingPipe(m.getFilters());
             if (!m.fileInfoFilters.isEmpty()) {
                 module.setFileInfoFilter(combine(m.fileInfoFilters));
--- a/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
+++ b/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
@@ -245,6 +245,7 @@ public final class ExtensibleAntInvoker extends Task {
         } else if (m instanceof SaxPipeElem) {
             final SaxPipeElem fm = (SaxPipeElem) m;
             final XmlFilterModule module = new XmlFilterModule();
+            module.setParallel(fm.parallel);
             final List<FileInfoFilterElem> predicates = new ArrayList<>(fm.getFormat());
             predicates.addAll(m.fileInfoFilters);
             module.setFileInfoFilter(combine(predicates));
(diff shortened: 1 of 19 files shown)


src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
```diff
@@ -199,6 +198,12 @@ public void execute() throws BuildException {
     private AbstractPipelineModule getPipelineModule(final ModuleElem m, final PipelineHashIO pipelineInput) throws DITAOTException {
         if (m instanceof XsltElem) {
             final XsltElem xm = (XsltElem) m;
+            if (xm.reloadstylesheet && xm.parallel) {
+                throw new BuildException("Both reloadstylesheet and parallel cannot be true");
```
jelovirt, line 202: When in parallel mode, we can't reuse the XSLT executable, we have to get a new one from the XSLT template object for each thread.

src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
```diff
@@ -199,6 +198,12 @@ public void execute() throws BuildException {
...
+            }
+            if (xm.parallel && xm.xmlcatalog != null) {
+                throw new DITAOTException("Pipeline XSLT task with parallel=true cannot be used with Ant's xmlcatalog");
```
jelovirt, line 205: Ant's XML catalog implementation is not thread safe.

robander (dismissed):
Ran this several times with our large doc source and it worked; initially had different results in the log but that is now fixed, was caused by the Ant catalog not being thread safe. Actual processing improvement seems to vary widely based on the doc set, features used, and computer doing the processing, but some are likely to see a nice improvement.

raducoravu:
Adding here the timings I got on the Oxygen Users Manual:

 - Without parallel enabled - 255 seconds
 - With parallel enabled - 156 seconds

So great, the parameter seems to take effect

jelovirt:
Rebased and cleaned out commented code. @robander could you rereview?

src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
```diff
@@ -261,14 +264,12 @@ private AbstractPipelineModule getPipelineModule(final ModuleElem m, final Pipel
...
-                module.setProcessingPipe(m.getFilters());
-            } catch (IllegalAccessException | InstantiationException e) {
-                throw new BuildException(e);
-            }
+            module.setParallel(m.parallel);
+            module.setProcessingPipe(m.getFilters());
             if (!m.fileInfoFilters.isEmpty()) {
                 module.setFileInfoFilter(combine(m.fileInfoFilters));
             }
+            module.setParallel(m.parallel);
```
robander, line 272: Curious - why do we need to run `module.setParallel(m.parallel);` on lines 267 and 272? Does the FileInfoFilter reset it?

robander (approved):

src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
jelovirt: Good catch, duplicate setter call.

robander (dismissed):

PR #3609, opened 2020-10-13
Fix setting html5.map.url property

jelovirt:
## Description
Fix `html5.map.url` property declaration.

## Motivation and Context
Property was incorrectly set with `<makeurl>` that caused a URL to be resolved against a file path.

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
+++ b/src/main/plugins/org.dita.html5/build_dita2html5_template.xml
@@ -110,12 +110,11 @@ See the accompanying LICENSE file for applicable license.
                    html5.topics.common.inner"/>
 
   <target name="html5.topic.init">
-    <pathconvert property="html5.map">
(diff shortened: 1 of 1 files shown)


infotexture (approved):
Tested with docs site output. Fixes the error reported earlier after #3439 was merged:

```
[ant:xslt] Warning at char 9 in xsl:apply-templates/@select on line 24 column 82 of nav.xsl:
  FODC0002: Document has been marked not available:
  file:/…/docs/file:/var/folders/rq/…/T/dita-ot/1602421791683/site.ditamap
```

✅ With these changes, the build completes without errors.

jelovirt:
@infotexture Because this is was introduced in 3.6 code changes, from release notes' perspective this is not a bug? I'd like to keep the bug tag here because this was a bug, but this should be omitted from the release notes from 3.6.

PR #3616, opened 2020-10-25
Add repeat option to CLI to run process N times

jelovirt:
## Description
Add `repeat` option to CLI to run conversion multiple times.

The output for repeated conversions is output to standard out after all conversions have been run, e.g.

```
1 17531ms
2 9199ms
3 8077ms
4 7832ms
5 7693ms
```
Because this is a measurement tool, the output should not be part of the log output.

## Motivation and Context
This option is useful for performance measurement, since timings for the first transformation are often dominated by Java warm-up time.

## How Has This Been Tested?
Old tests pass.

## Type of Changes

Option is documented in the CLI help, but docs could mention that this option is intended for plug-in developers for measuring performance.


jelovirt:
Do you think we should improve this by calculating the average time and outputting it as the last time? Because of the JVM warm-up, we'd need to ignore the warm-up phase from the average, so the option would need to be something like `--repeat=5,10` where the `5` is the number of warm-up conversions and ´10´ is how many actual conversions are run. The average would then be calculated from the `10` conversions. Or do we just leave this to the user and only output the raw data?

jelovirt:
Or should be output be

raducoravu:
@jelovirt maybe you can output both the individual times and the average.

robander (approved):

robander:
It does seem that average would not mean a lot if it includes the first. Also (based on trying similar things today), it messes up when I've got 15 builds running and hand to go use Slack during one, which increased that one build time by 20%...

PR #3624, opened 2020-10-31
Fix workdir-uri generation bug

jelovirt:
## Description
Make sure URI for a directory will always end in a trailing slash. 

## Motivation and Context
If a `File` points to a path that doesn't exist, `File.toURI()` will not output a trailing slash.

## How Has This Been Tested?
Tests pass and added a new test for `toDirURI` function.

## Type of Changes

None


robander (approved):

lief-erickson:
> If a File points to a path that doesn't exist, File.toURI() will not output a trailing slash.

This statement conflicts with the title of this issue. I'm guessing "not" is a typo and should be "now." The code seems to reflect the title. Just want to confirm.

PR #3631, opened 2020-11-09
Add missing terminal punctuation

infotexture:
Information about default values is appended to the parameter description in the documentation, so it's important that each description is a complete sentence.


infotexture:
Amends https://github.com/dita-ot/dita-ot/commit/77a0f5ba32a6c7c74baeabd1344c37bae19814a0 from https://github.com/dita-ot/dita-ot/pull/3606.

robander (approved):

PR #3632, opened 2020-11-10
HTML5: Move hard-coded styles to Sass partials

infotexture:
In several places, the HTML5 code still hard-codes inline style attributes, which prevent users from customizing the presentation of the corresponding elements when these are emitted to HTML5 output, including:

- line-through and overline elements
- syntax diagrams
- long quote citations
- Boolean states

These changes move the default presentation rules to CSS to allow users to override presentation in custom stylesheets.

The output is visually equivalent to the results generated by previous toolkit versions.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/hi-d.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/hi-d.xsl
@@ -63,7 +63,7 @@ See the accompanying LICENSE file for applicable license.
   </xsl:template>
 
   <xsl:template match="*[contains(@class,' hi-d/line-through ')]" name="topic.hi-d.line-through">
-    <span style="text-decoration:line-through">
+    <span>
       <xsl:call-template name="commonattributes"/>
       <xsl:call-template name="setidaname"/>
       <xsl:apply-templates/>
@@ -71,7 +71,7 @@ See the accompanying LICENSE file for applicable license.
   </xsl:template>
 
   <xsl:template match="*[contains(@class,' hi-d/overline ')]" name="topic.hi-d.overline">
-    <span style="text-decoration:overline">
+    <span>
       <xsl:call-template name="commonattributes"/>
       <xsl:call-template name="setidaname"/>
       <xsl:apply-templates/>
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -585,7 +585,7 @@ See the accompanying LICENSE file for applicable license.
             </xsl:choose></cite></a></div>
         </xsl:when>
         <xsl:when test="@reftitle"> <!-- Insert citation text -->
-          <br/><div style="text-align:right"><cite><xsl:value-of select="@reftitle"/></cite></div>
+          <br/><div><cite><xsl:value-of select="@reftitle"/></cite></div>
         </xsl:when>
         <xsl:otherwise><!--nop - do nothing--></xsl:otherwise>
       </xsl:choose>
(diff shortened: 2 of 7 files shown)


infotexture:
@robander There a few other cases for titles & keywords in `syntax-svg.xsl`, but I'm not sure how to test these:

- https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/xsl/syntax-svg.xsl#L37
- https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/xsl/syntax-svg.xsl#L43
- https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/xsl/syntax-svg.xsl#L52

Same for spectitle attributes:

- https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/xsl/topic.xsl#L1910-L1916

These are probably seldom-used edge cases, so I'm not sure it's worth holding up the other changes, but open to any suggestions at this point.

robander:
@infotexture I'd probably leave the other ones alone. I don't really consider the syntax output very usable, which means there's little benefit from refactoring it. (If I'm wrong and someone else is using them, I have no problem moving those to CSS but it doesn't really seem worth the change otherwise.)

src/main/plugins/org.dita.html5/xsl/hi-d.xsl
```diff
@@ -63,15 +63,15 @@ See the accompanying LICENSE file for applicable license.
   </xsl:template>
 
   <xsl:template match="*[contains(@class,' hi-d/line-through ')]" name="topic.hi-d.line-through">
-    <span style="text-decoration:line-through">
+    <span>
```
robander, line 66: I'm a little conflicted on this, just because the style is really part of the element itself, so it sort of makes sense to have it in the HTML rather than the CSS. I'm thinking of a case like my own, where we don't use the toolkit's CSS at all, so updating to this version of OT would look like a regression.

Generally ambivalent though because I also don't like having hard coded style attributes; one way to counter the concern might be to list this among the "migrating to 3.6" instructions?

robander (approved):
Not sure about the line-through/overline changes but the others all look good; assuming we go ahead with those two as well, I think it should be listed in the 3.6 migration doc.

src/main/plugins/org.dita.html5/xsl/hi-d.xsl
infotexture: I can see the line-through/overline cases either way as well, and I understand this requires extra work in cases where the default CSS is not used.

The main issue I see is that the current code makes it hard to override the default presentation for these elements _at all_, since inline `@style` attributes take precedence over any CSS rules in external style sheets.

So even if a frontend developer wanted to replace overlining with something like animated top borders (as many do with link _underlining_), the hard-coded overlines would still be visible.

Either way, these changes should certainly be documented in the 3.6 migration info.

jelovirt: I'd prefer to move this to CSS and not hard-code it in HTML.

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -569,7 +569,7 @@ See the accompanying LICENSE file for applicable license.
       <xsl:apply-templates/>
       <xsl:choose>
         <xsl:when test="@href">
-          <br/><div style="text-align:right"><a>
+          <br/><div><a>
```
jelovirt, line 572: There's one more on line 588.

jelovirt (changes requested):

src/main/plugins/org.dita.html5/xsl/topic.xsl
infotexture: @jelovirt Thanks, good catch. Resolved in de7542f.

PR #3636, opened 2020-11-20
Fix repeat count in install and uninstall subcommands

jelovirt:
## Description
Fix repeat count initialization in install and uninstall subcommands.

## Motivation and Context
Repeat count was set to zero in install and uninstall subcommands and resulted in process not being run a single time. This initializes the count to one, so the process is run once.

## Type of Changes

## Documentation

No need to add to release notes for 3.6, because this was broken in PR for repeated conversion that is part of 3.6.


infotexture (approved):
Verified that these changes allow plug-ins to be installed without the message that currently appears when installing plug-ins from the tip of the `develop` branch:

    Searching for integrator.xml ...

PR #3642, opened 2020-11-28
Remove disabled pdf2.index.skip param and deprecate old PDF indexing

jelovirt:
## Description
Remove disabled `pdf2.index.skip` parameter and deprecate old PDF indexing code.

## Motivation and Context
Original PDF indexing implementation was extracted to `org.dita.index` plug-in in DITA-OT 3.4. The code can still be enabled by reverting configuration, but the original implementation is deprecated. Deprecation comments are added to reflect this because they were missing from the original PR that enabled `org.dita.index`.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/build_template.xml
+++ b/src/main/plugins/org.dita.pdf2/build_template.xml
@@ -243,12 +243,14 @@ with those set forth herein.
     <property name="args.tablelink.style" value="NUMTITLE"/>
   </target>
 
+  <!-- Deprecated since 3.4 -->
   <target name="transform.topic2fo.index.init">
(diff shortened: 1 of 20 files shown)


infotexture (approved):
👍 LGTM:

- The `pdf2.index.skip` entry is removed from the PDF parameter listing in docs
- The [org.dita.index.skip](https://www.dita-ot.org/dev/parameters/parameters-pdf.html#pdf__org.dita.index.skip) entry added with https://github.com/dita-ot/docs/pull/319 is preserved

This should be the final piece required to resolve the confusion flagged in https://github.com/dita-ot/docs/issues/316.

:shipit:

PR #3645, opened 2020-12-03
Add RELAX NG schemas to resources catalog

### LINKED ISSUE #3641: Add RELAX NG schemas to catalog
We currently ship 2 RELAX NG schemas that can be used to validate plug-ins and project files:

- `resources/project.rnc`
- `resources/plugin.rnc`

When these schemas are referenced in files with **relative** URIs as in our own docs [/samples/project-files/html.xml](https://github.com/dita-ot/docs/blob/develop/samples/project-files/html.xml#L2), validation may fail if these files are opened or maintained outside the toolkit directory _(as is common in many custom development projects)_. 

## Potential solution: add to catalog

In discussion with @raducoravu, he commented:

> if for example the `catalog-dita.xml` would have had an URI mapping
> for the `project.rnc` schema like:
>
>     <uri name="http://www.dita-ot.project.com/project.rnc
>     <http://www.dita-ot.project.com/project.rnc>"
>     uri="../../resources/project.rnc"/>
>
> you could have used that absolute URI when referencing the
> `project.rnc` from your XML and Oxygen would have found the target
> by going through the XML catalog support.

### — but to _which_ catalog?

> So the RNC file is located in `dita-ot/resources/project.rnc`.
>
> The `dita-ot/catalog-dita.xml` is deprecated and should not be modified further by adding an extra mapping in our case.
>
> The `dita-ot/plugins/org.dita.base/catalog-dita.xml` seems to be intended to map schemas or plugin IDs, XSLTs. Adding in it a reference to a folder 2 levels up 
> `../../resources/project.rnc` seems ugly to me.
>
> So I do not know where that catalog entry should be added, maybe an extra `catalog.xml` could be added to `dita-ot/resources` with some mappings for `plugin.rnc` and `project.rng`?

The catalog approach seems like the right way to handle this in a vendor-neutral way, as any XML editor with catalog support _(at least those aware of DITA-OT catalogs)_ would then be able to validate files based on these schemas regardless of file location.

The `project.rnc` file [declares a namespace](https://github.com/dita-ot/dita-ot/blob/develop/src/main/resources/project.rnc#L1), so we might want to do something similar for the plugin schema:

    default namespace = "https://www.dita-ot.org/project"

@jelovirt @robander Any thoughts?

infotexture:
## Description

We currently ship 2 RELAX NG schemas that can be used to validate plug-ins and project files:

- `resources/project.rnc`
- `resources/plugin.rnc`

In https://github.com/dita-ot/website/pull/455, these schemas were published to public URLs, and https://github.com/dita-ot/docs/commit/08e4f4ea6aab69d18a089b98a1ae77cac7aeac62 updated the corresponding `<?xml-model>` processing instructions in the docs to allow editors to validate these files when stored outside of DITA-OT.

## Motivation and Context

This PR adds a catalog to the `resources` folder that maps these public URLs to the local resources and extends the base `catalog-dita.xml` with this new catalog entry so that catalog-aware editors will load the local resource instead of accessing the public URL over the network.

## How Has This Been Tested?

Ran the distribution build in a [fresh worktree](https://github.com/dita-ot/dita-ot/wiki/Fresh-dist-worktree), and verified that the catalog is added and included in `plugins/org.dita.base/catalog-dita.xml`.

## Type of Changes

- Enhancement _(non-breaking change which improves existing functionality)_

  _Release Notes_ entry
  No.

Fixes #3641.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,7 @@ dependencies {
 
 jar {
     archiveName = "${project.name}.jar"
+    exclude "catalog.xml"
     exclude "plugin.rnc"
     exclude "project.rnc"
 }
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,11 @@ dependencies {
     testCompile group: 'org.xmlunit', name: 'xmlunit-core', version: '2.6.0'
 }
 
-jar.archiveName = "${project.name}.jar"
+jar {
+    archiveName = "${project.name}.jar"
+    exclude "plugin.rnc"
+    exclude "project.rnc"
+}
 
 processResources {
     filter ReplaceTokens, tokens: [
--- a/build.gradle
+++ b/build.gradle
@@ -256,6 +256,7 @@ task copyDistTemp(type: Copy, dependsOn: initDist) {
         include "plugins/org.oasis-open.dita.v1_3/**"
         exclude "temp"
         include "config/**"
+        include "resources/catalog.xml"
         include "resources/plugin.rnc"
         include "resources/project.rnc"
         // legacy build scripts
--- a/src/main/plugins/org.dita.base/catalog-dita_template.xml
+++ b/src/main/plugins/org.dita.base/catalog-dita_template.xml
@@ -14,9 +14,11 @@ See the accompanying LICENSE file for applicable license.
   <dita:extension id="dita.catalog.plugin-info"
     behavior="org.dita.dost.platform.ImportPluginCatalogAction"/>
 
-   <dita:extension id="dita.specialization.catalog"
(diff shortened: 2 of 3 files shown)


jelovirt (approved):

infotexture:
@raducoravu These changes should allow you to remove your previous catalog workarounds from oXygen's DITA framework, as the new public IDs are now referenced in `plugins/org.dita.base/catalog-dita.xml`.

raducoravu:
@infotexture right, thanks, I will add a note to look over this once I integrate the DITA OT 3.6 in our Oxygen project.

PR #3650, opened 2020-12-13
Restore RELAX NG schemas to 'dita-ot.jar'

infotexture:
Without these schemas in the .jar file, builds with project files fail with the following error:

> Error: invalid input for CompactParseable

- Amends 09c1d94 from #3645


jelovirt:
This is my bad. I forgot that we use RNC to validate the XML project file, but the RNC for plugin configuration is just for convenience. It's likely better to keep both RNC, for consistency.

jelovirt (approved):

PR #3652, opened 2020-12-17
Add support for multimedia oasis-tcs/dita#351

robander:
Signed-off-by: Robert D Anderson <gorodki@gmail.com>

---
name: Pull request

---

## Description

Adds basic support for oasis-tcs/dita#351 the new `<audio>` and `<video>` elements. Markup is already supported, this creates the expected HTML5 content. PDF already (correctly) uses the fallback element content, and does not generate an error.

* `href` is rewritten to `src`
* `format` is rewritten to `type` (DITA 2.0 spec clarifies that a MIME type is acceptable in `@format`)
* media attributes are either copied as-is or (for boolean) added if true

## Motivation and Context

These elements were redesigned over the last few months to become base DITA elements in DITA 2.0, which maps closely to HTML5 structure. Without this PR, they generate a bunch of "unknown element" warnings in HTML5.

This basically just maps the DITA markup to the HTML5 markup for `<audio>`, `<video>`, `<source>`, and `<track>`.

The DITA order does not match HTML5, so it explicitly processes source, then track, then any fallback content.

## How Has This Been Tested?

Attaching test case here with the simple case (only href + format) and the complex case (all the multimedia attributes, plus source and track info). Tested HTML is as expected. The attached audio/video files are fake content (copy of a DITA file).
[media.zip](https://github.com/dita-ot/dita-ot/files/5712760/media.zip)

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -1550,6 +1550,90 @@ See the accompanying LICENSE file for applicable license.
    </span>
   </xsl:template>
   
+  <!-- =========== DITA 2.0 multimedia elements -->
+  <xsl:template match="*[contains(@class,' topic/audio ')]">
+    <audio>
+      <xsl:call-template name="commonattributes"/>
+      <xsl:apply-templates select="@autoplay | @controls | @loop | @muted" mode="boolean-media-attribute"/>
+      <xsl:apply-templates select="@tabindex | @href | @format"/>
+      <xsl:call-template name="setid"/>
+      <xsl:apply-templates select="*[contains(@class,' topic/media-source ')],
+        *[contains(@class,' topic/media-track ')]"/>
+      <xsl:apply-templates select="text() | 
+        * except *[contains(@class,' topic/media-source') or contains(@class,' topic/media-track')]"/>
+    </audio>
+  </xsl:template>
+  
+  <xsl:template match="*[contains(@class,' topic/video ')]">
+    <video>
+      <xsl:call-template name="commonattributes"/>
+      <xsl:apply-templates select="@autoplay | @controls | @loop | @muted" mode="boolean-media-attribute"/>
+      <xsl:apply-templates select="@tabindex | @href | @format"/>
+      <xsl:apply-templates select="@poster"/>
+      <xsl:call-template name="setid"/>
+      <xsl:apply-templates select="*[contains(@class,' topic/media-source ')],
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -1550,6 +1550,90 @@ See the accompanying LICENSE file for applicable license.
...
+  <xsl:template match="*[contains(@class,' topic/audio ')]">
+    <audio>
+      <xsl:call-template name="commonattributes"/>
+      <xsl:apply-templates select="@autoplay | @controls | @loop | @muted" mode="boolean-media-attribute"/>
+      <xsl:apply-templates select="@tabindex | @href | @format"/>
+      <xsl:call-template name="setid"/>
+      <xsl:apply-templates select="*[contains(@class,' topic/media-source ')],
+        *[contains(@class,' topic/media-track ')]"/>
+      <xsl:apply-templates select="text() | 
+        *[not(contains(@class,' topic/media-source') or contains(@class,' topic/media-track'))]"/>
```
jelovirt, line 1563:
```suggestion
        * except *[contains(@class, ' topic/media-source') or contains(@class, ' topic/media-track')]"/>
```

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -1550,6 +1550,90 @@ See the accompanying LICENSE file for applicable license.
...
+    <video>
+      <xsl:call-template name="commonattributes"/>
+      <xsl:apply-templates select="@autoplay | @controls | @loop | @muted" mode="boolean-media-attribute"/>
+      <xsl:apply-templates select="@tabindex | @href | @format"/>
+      <xsl:apply-templates select="@poster"/>
+      <xsl:call-template name="setid"/>
+      <xsl:apply-templates select="*[contains(@class,' topic/media-source ')],
+        *[contains(@class,' topic/media-track ')]"/>
+      <xsl:apply-templates select="text() | 
+        *[not(contains(@class,' topic/media-source') or contains(@class,' topic/media-track'))]"/>
```
jelovirt, line 1577:
```suggestion
        * except *[contains(@class,' topic/media-source') or contains(@class,' topic/media-track')]"/>
```

jelovirt (changes requested):

jelovirt (approved):

raducoravu:
How about support to automatically copy audio and video files to the output folder? Also the [gen-list] stage right now attempts to parse the mp4 file as DITA XML.

Issue #3664, opened 2021-01-01
Question: Adding GitHub Actions. DITA-OT as a static website generator?


jason-fox:
I have been transfering my CI from Travis to GitHub Actions and I created a [GitHub Action for the unit-test plugin](https://github.com/jason-fox/dita-unit-test-action) for the DITA-OT plugin unit test framework. You can see a sample run [here](https://github.com/jason-fox/fox.jason.translate.xliff/runs/1618598953?check_suite_focus=true).

Obviously with the specific Unit Test GitHub Action based on my own `fox.jason.unit-test` plugin, it makes sense to be hosted  within my own `jason-fox` GitHub, but I have also created a more generic version to run any `dita` command. This repo currently hosts the generic DITA-OT GitHub Action: https://github.com/jason-fox/dita-build-action

![action](https://user-images.githubusercontent.com/3439249/103441576-3d546300-4c4f-11eb-9692-b5528c457c59.png)

The `README` in the repo already follows the Standard GitHub Action format and describes proper usage.

### Example in action

This -  https://github.com/jason-fox/test-dita-build - is a sample repository which uses the action on a test run:  it includes a few extra plugins and prerequisites to show what can be done - it builds a PDF and a custom HTML build. It combines with other actions to create a static website.

The outputs are fully customizable, but in this case it is a PDF document [here](https://jason-fox.github.io/test-dita-build/pdf/document.pdf) and an HTML website [here](https://jason-fox.github.io/test-dita-build/html5-bootstrap/topics/examples.html)

### Questions:

1) Before I add this generic DITA-OT Action to the GitHub Actions marketplace myself, I was wondering should it be transfered to be hosted under the `dita-ot` organization?

2) Regardless of whether the `dita-ot` organization wants to host this code or not,  what sort of documentation would you like added to the docs to support this?

jelovirt:
This is something we probably should do under the dita-ot organization. So 👍🏽 there. However, the actual implementation here is not IMO ready yet for public consumption.

The API right now not that user friendly. The example in the README is

```yaml
dita-ot-version: 3.6
setup-script: 'startup.sh'
prerequisites: 'org.doctales.xmltask,fox.jason.extend.css,fox.jason.prismjs'
ditamap: 'docsrc/document.ditamap'
transtypes: 'html5,pdf'
properties: '--filter=filter1.ditaval' # Any additional command line properties
output-dir: 'out' # The folder the action should write to.
```

IMO it should be more like

```yaml
[...]

PR #3665, opened 2021-01-03
Update Dockerfile

jason-fox:
## Description

- Add `SHELL` command
- Add Metadata `LABELS` following opencontainers.org
- Add maintainer `LABEL`
- Install locales and tzdata as part of the base image.

Related #3664

## Motivation and Context

I want to reduce the GitHub Docker container action to just some scripting hooks. At the moment I need to do the following installs (based on my unit-test GitHub Action experiences):

```bash 
apt-get install -y --no-install-recommends openjdk-11-jre ant curl unzip git locales tzdata
``` 

Most of these will be unnecessary if I can just use `ghcr.io/dita-ot/dita-ot/dita-ot:3.6` as a base image.

- `openjdk-11-jre` - present in `adoptopenjdk:11-jre-hotspot`
-  `ant` - not needed in the base image (Unit test coverage requirement)
-  `curl` - present in adoptopenjdk:11-jre-hotspot
-  `unzip` -  present in `ghcr.io/dita-ot/dita-ot/dita-ot:3.6`
-  `git` - not needed in the base image (Unit test coverage requirement)
- `locales`  - **added**
- `tzdata`   - **added**

I've discovered that `locales`  is necessary in the following weird edge case:

```xml
<ph props="phoneme alphabet(ipa) ph(t&#x259;mei&#x27E;ou&#x325;)">tomato</ph>
	This example shows how a pronunciation for "tomato" is specified using the SPR phonology:
```

I want this to display as International Phonetic Alphabet:

```xml
 <phoneme alphabet="ipa" ph="təmeiɾou̥">tomato </phoneme>
```

However the default in the docker image is `LC_POSIX_UNIX` and that doesn't encode these UTF-8 chars properly. Changing the locale to `en_US.UTF-8` fixes the issue. Now I'm not saying the base image should set the locale directly, but it would be very useful to include the tools to do so for other images downstream without needing to use `apt-get` again e.g.:

```Dockerfile
RUN export DEBIAN_FRONTEND=noninteractive && \
	ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime && \
	dpkg-reconfigure --frontend noninteractive locales tzdata && \
	locale-gen en_US.UTF-8
```

Obviously I could equally want to change to French or German or Vietnamese or Finnish or whatever. `tzdata` is kind of a companion to `locales` as I may want to add date stamps according to my preferred local time.

There are other minor improvements included:

- The unzip is now less noisy.
- The base image currently needs a `build-arg` to work - defaults have been added.

The following build statement now works:

```console
docker build -t dita-ot .
```

- The base image currently has no metadata `LABEL` statements - these have added (see [opencontainers.org](https://github.com/opencontainers/opencontainers.org) )

To properly build a specific version and even add more metadata `LABELS` you can do something like this :

```console
docker build -t ghcr.io/dita-ot/dita-ot/dita-ot:3.6 --build-arg VERSION=3.6 \
        --label "org.opencontainers.image.created"=$(date +%Y-%m-%d) \
        --label "org.opencontainers.image.version"=3.6 .
docker push ghcr.io/dita-ot/dita-ot/dita-ot:3.6 
```

## How Has This Been Tested?

Build and run locally.

## Type of Changes

Documentation is part of the discussion of #3664. 
This will reduce the proposed GitHub Container Action Dockerfile to:

```Dockerfile
FROM ghcr.io/dita-ot/dita-ot/dita-ot:3.6

SHELL ["/bin/bash", "-o", "pipefail", "-c"]

USER root
WORKDIR /

COPY entrypoint.sh entrypoint.sh
COPY script.sh install_script.sh
COPY script.sh build_script.sh

RUN export DEBIAN_FRONTEND=noninteractive && \
	ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime && \
	dpkg-reconfigure --frontend noninteractive locales tzdata && \
	locale-gen en_US.UTF-8 && \
    chmod +x /entrypoint.sh && \
	chmod +x /build_script.sh  && \
	chmod +x /install_script.sh

ENV LANG en_US.UTF-8  
ENV LANGUAGE en_US:en  
ENV LC_ALL en_US.UTF-8

ENTRYPOINT ["/entrypoint.sh"]
```

Which is basically just saying I'm US English (by default) and I need some bash scripts.


Dockerfile
```diff
@@ -1,12 +1,27 @@
...
-RUN apt-get update \
-    && apt-get install -y unzip \
-    && rm -rf /var/lib/apt/lists/*
+ARG VERSION=3.6
+ARG JAVA_VERSION=11-jre-hotspot
 
+FROM adoptopenjdk:${JAVA_VERSION}
 ARG VERSION
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
```
jelovirt, line 7: Why is `-c` used here?

jelovirt, line 7:
```suggestion
SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
```

Adds additional error handling we could have

Dockerfile
```diff
@@ -1,12 +1,27 @@
-FROM adoptopenjdk:11-jre-hotspot
-
-RUN apt-get update \
-    && apt-get install -y unzip \
-    && rm -rf /var/lib/apt/lists/*
+ARG VERSION=3.6
+ARG JAVA_VERSION=11-jre-hotspot
 
+FROM adoptopenjdk:${JAVA_VERSION}
```
jelovirt, line 4: Not sure this is really needed.

Dockerfile
```diff
@@ -1,12 +1,27 @@
-FROM adoptopenjdk:11-jre-hotspot
-
-RUN apt-get update \
-    && apt-get install -y unzip \
-    && rm -rf /var/lib/apt/lists/*
+ARG VERSION=3.6
```
jelovirt, line 1: Defaulting the version means we we would have to update this file every time we release a new version. My current preference is to not have a default for this.

Dockerfile
```diff
@@ -1,12 +1,27 @@
...
+ARG JAVA_VERSION=11-jre-hotspot
 
+FROM adoptopenjdk:${JAVA_VERSION}
 ARG VERSION
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+LABEL "maintainer"="DITA Open Toolkit Development Team"
```
jelovirt, line 9: There is no _DTA Open Toolkit Development Team_, so I'm not sure what this should be.

Dockerfile
```diff
@@ -1,12 +1,27 @@
...
+LABEL "org.opencontainers.image.authors"="Robert D Anderson, Jarno Elovirta, Roger W. Fienhold Sheen"
```
jelovirt, line 10: My preference would be to not put our names here, just because none of us wants people contacting us directly.

Dockerfile
```diff
@@ -1,12 +1,27 @@
...
+LABEL "org.opencontainers.image.documentation"="https://www.dita-ot.org/"
+LABEL "org.opencontainers.image.vendor"="DITA Open Toolkit Development Team"
+LABEL "org.opencontainers.image.licenses"="Apache-2.0"
+LABEL "org.opencontainers.image.title"="DITA Open Toolkit"
+LABEL "org.opencontainers.image.description"="The open-source XML publishing engine for content authored in the Darwin Information Typing Architecture."
```
jelovirt, line 15:
```suggestion
LABEL "org.opencontainers.image.description"="Publishing engine for content authored in the Darwin Information Typing Architecture."
```

Dockerfile
```diff
@@ -26,4 +41,4 @@ USER dita-ot
 ENV DITA_HOME=/opt/app
 ENV PATH=${PATH}:${DITA_HOME}/bin
 WORKDIR $DITA_HOME
-ENTRYPOINT ["/opt/app/bin/dita"]
+ENTRYPOINT ["/opt/app/bin/dita"]
```
jelovirt, line 44: Add missing line feed back

jelovirt (changes requested):

Dockerfile
jelovirt: Maybe it's safer to not add `-eu`, because those might be too strict.

Dockerfile
jason-fox: Typically this would be the legal entity that owns the copyright as found in the LICENCE file - The DITA-OT LICENSE hasn't listed the owner explicitly, so I would assume that the contributors to the repository - any suggestion as to how to rephrase it?

Dockerfile
jason-fox: Taken from `https://www.dita-ot.org/who_we_are` - I'll replace with the link.

Dockerfile
jason-fox: Reverted - 4b7c819

Dockerfile
jason-fox: Reverted - 4b7c819

Dockerfile
jason-fox: This `SHELL` is sanity checking, if we're going to fail let's fail hard. See https://github.com/docker/for-linux/issues/408

jason-fox: I have always used `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` - I think Hadolint prefers it as well.

Dockerfile
infotexture: In the docs, we use _“DITA Open Toolkit project”_ when we need a collective noun for the team. 

For [example](https://www.dita-ot.org/dev/topics/plugin-coding-conventions.html):

> … the DITA Open Toolkit project recommends …

Dockerfile
jelovirt: `-o pipefail` returns the first non-zero exit code in a pipe, but what about `-c`. The manual says

> -c
>    Read and execute commands from the first non-option argument command_string, then exit. If there are arguments after the command_string, the first argument is assigned to $0 and any remaining arguments are assigned to the positional parameters. The assignment to $0 sets the name of the shell, which is used in warning and error messages.

I have trouble parsing that. Could you elaborate what change it introduces?

Dockerfile
jelovirt: _DITA Open Toolkit project_ is good, since the maintainers really are just the people who file pull requests and people can come and go.

Dockerfile
jason-fox: Well you ain't going to need `-c` until you need it. It really only saves you from some weird edge cases. It helps with properly unquoting the command line parameters, say you want to invoke the container passing in the transtype which has been read in using some elaborate bash conversion function or something.

The best explanation I could find was here: https://ask.metafilter.com/316796/Using-piping-with-nix-find-without-getting-mired-in-string-escaping

Dockerfile
jason-fox: _DITA Open Toolkit project_ it is then. 50b6181

jason-fox:
Hmmm. I've just had a thought, maybe switching from `LC_POSIX_UNIX` to `en_US.UTF-8` actually  *is* necessary. What about indexing?

Take the following list - `bailey', 'boffey', 'böhm', 'brown'` - in the `en` locale these would sort ignoring diacritics using the following:

```text
bailey: 2927373d2f57010a010a
boffey: 294331312f57010a010a
böhm:   2943353f01859d060109
brown:  294943534101090109
```

But in `LC_POSIX_UNIX` any special chars will sort after standard ASCII, so `böhm` will come after `brown`.

What do you think?

Dockerfile
jelovirt: A concrete example would be that now you can run

```bash
$ docker run docker.pkg.github.com/dita-ot/dita-ot/dita-ot:3.6 install foo
```

and with the `-c` shell you can both

```
```bash
$ docker run docker.pkg.github.com/dita-ot/dita-ot/dita-ot:3.6 install foo
$ docker run docker.pkg.github.com/dita-ot/dita-ot/dita-ot:3.6 "install foo"
```

Seems useful.

jelovirt (approved):

Issue #3620, opened 2020-10-29
HTML5 fails with schema-based map having chunk=to-content


takesson:
## Expected Behavior
Transformation into HTML5 should work and produce a single chunk.

## Actual Behavior
Fails with error. The temporary ditabase does not have the xsi namespace declaration.

## Possible Solution
I suspect the declaration of xsi is not propagating to the pseudo-topic created when chunk=to-content is set on the map root element (assuming those declarations are propagating to the ditabase...).
```xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"```

## Steps to Reproduce

1. Create Sample.ditamap using xsi:noNamespaceSchemaLocation
2. ```bin/dita -f html5 -Droot-chunk-override=to-content -i Sample.ditamap```

## Copy of the error message, log file or stack trace
    [chunk] Failed to transform file:/var/folders/_d/qw_ml1h95vd2d6kmf1krj3600000gp/T/temp20201030002053316/Sample.dita: The prefix "xsi" for attribute "xsi:noNamespaceSchemaLocation" associated with an element type "topic" is not bound.
[move-meta] Error on line 1 column 1397 of Sample.dita:
[move-meta]   SXXP0003: Error reported by XML parser: The prefix "xsi" for attribute
[move-meta]   "xsi:noNamespaceSchemaLocation" associated with an element type "topic" is not bound.
Error: Failed to run pipeline: Failed to transform document: org.xml.sax.SAXParseException; systemId: file:/var/folders/_d/qw_ml1h95vd2d6kmf1krj3600000gp/T/temp20201030002053316/Sample.dita; lineNumber: 1; columnNumber: 1397; The prefix "xsi" for attribute "xsi:noNamespaceSchemaLocation" associated with an element type "topic" is not bound.

## Environment

* DITA-OT version: 3.5.4
* Operating system and version: macOS
* How did you run DITA-OT? bash, see above
* Transformation type: HTML5

**Comment by takesson** [comment]
Actually, it seems to fail when chunk=to-content is set on topicref elements within the ditamap as well. The issue is not specific to setting chunk on the map root element.

**Comment by takesson** [comment]
The following patch avoids the failure.
```
diff --git a/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java b/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
index bb4d209d5..36b5c9599 100644
[...]

PR #3684, opened 2021-01-21
Fix for issue 3620

easirois:
Signed-off-by: Eric Sirois <easirois@rogers.com>

---
name: Pull request for 3620
about: The current chunk mechanism is omitting the xmlns namespace declaration when merging content from individual topics contain namespaced content, like MathML, into the merged one.

---

## Description
Add some checks to the code to see if the element in question is in a different URI space.  If it is and it's the first level for that element, then add the @xmlns attribute declaration to that element.

## Motivation and Context

https://github.com/dita-ot/dita-ot/issues/3620  This used to work on 2.5.4. Now, If any DITA content included embeded content from another namespace, then the build fails.

## How Has This Been Tested?

Execute the same map that was broken previously and ensure the xmlns:<prefix> is defined on the appropriate element.
<m:math xmlns:m="http://www.w3.org/1998/Math/MathML">

## Type of Changes

Add methods that checks if the element in question is in a URI space and adds appropriate attribute.

No

  _(Provide links to existing documentation topics that will require updates)_

## Checklist

- There are no unit tests to reflect the changes in my code.

### DIFF
--- a/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
+++ b/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
@@ -21,7 +21,9 @@ import org.w3c.dom.Node;
 import org.w3c.dom.Text;
 import org.xml.sax.Attributes;
 import org.xml.sax.SAXException;
+import org.xml.sax.XMLReader;
 import org.xml.sax.helpers.AttributesImpl;
+import org.xml.sax.helpers.NamespaceSupport;
 
 import java.io.*;
 import java.net.URI;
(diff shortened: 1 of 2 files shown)


easirois:
Need to add constants to handle schema validation on topics.  I will need to make an update to this pull request.

jelovirt:
Could you add tests that don't work with the current code and are fixed by these changes.

jelovirt:
I'm a bit worried how this deals with namespace declarations, by writing them as normal attributes. It's a code smell that leads me to think that either this PR has issues or the original code doesn't handle namespace correctly to begin with. @robander any thoughts?

easirois:
@jelovirt  Sure, I can work on a test case.

easirois:
Yeah not sure what happened between 2.5.4 and now.  But 2.5.4 handles them without issue.  For the XSD namespace, I handle it the same as the other namespaces on the root attribute.  With SAX when you iterate through the attributes, none of the namespace declaration attributes are in the list you get back.  That's why I needed to implement the other methods that let you know you're in a namespaced element.

easirois:
@jelovirt  I think I have a test case created based on what already exists, but how do I run those test cases.  Not sure if that happens with the gradlew.bat file.

jelovirt (approved):

PR #3709, opened 2021-02-25
Fix: 'Warning' notes could not be styled separately

cpkio:
## Description

This commit fixes DITA-OT incapability to style «Warning» notes separatedly.

## Motivation and Context

IMO all the elements have to be styled explicitly (hidden dependencies are bad). Elements implicitly inheriting other styles («Danger» styling in this case) seem wrong.

## How Has This Been Tested?

Fixed my styling issues.

## Type of Changes

- Bug fix

I'm not aware of existing DITA-OT documentation that needs to be changed. But, this should be explicitly noted on new release changelog.

**Yes, it will affect workflow for those who have styled «Danger» notes and used «Warning» notes instead.**

## Checklist

- I have **NOT** updated the unit tests.

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
@@ -861,7 +861,7 @@ See the accompanying LICENSE file for applicable license.
                         </fo:inline>
                     </xsl:when>
                     <xsl:when test="@type='warning'">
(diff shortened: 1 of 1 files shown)


robander:
Sorry for the delay on looking at this - can you add the DCO signoff, and we can approve and merge this? The change looks good. Details on adding the sign-off to an existing commit are in here: https://www.dita-ot.org/contributing#adding-signoff-to-earlier-commits

cpkio:
Added DCO to commit message.

Issue #3720, opened 2021-03-25
preprocess.copy-html.skip is missing from org.dita.pdf2/build_template.xml


thendarion:
there is a
https://github.com/dita-ot/dita-ot/blob/a14677b00d5b7b3a3f47f8ddba15fce432b6d3b6/src/main/plugins/org.dita.pdf2/build_template.xml#L57
and there is a
https://github.com/dita-ot/dita-ot/blob/a14677b00d5b7b3a3f47f8ddba15fce432b6d3b6/src/main/plugins/org.dita.pdf2/build_template.xml#L58
but there is no 
```    <property name="preprocess.copy-html.skip" value="true"/>```

jelovirt:
@thendarion cound you add information to the issue description what this omission causes in PDF2 processing?

thendarion:
if I use
https://github.com/dita-ot/dita-ot/blob/578e100ed742917d8f55b77ce75578fae4cc2e26/src/main/plugins/org.dita.base/plugin.xml#L49
https://github.com/dita-ot/dita-ot/blob/ef8e64be901f4135ea7814b50915954c672ee4ec/src/main/plugins/org.dita.base/build_preprocess_template.xml#L472

then my custom html-copy target is executed too in PDF even though it was supposed to copy files for HTML only :(

jelovirt:
If you have a map that references a local non-DITA resource, like another PDF or Excel file, will setting this default prevent those files from being copied to output directory?

thendarion:
but then if you have a map that references a non-DITA resource like a PNG file, will the current default prevent those files from being copied to output directory too?

I guess this part is fundamentally broken, but at least I see the error of my ways now :D

robander:
(Just happened upon this) I think this is probably the original design, but I don't know that it's the right design. If you build a PDF, the images are not copied because they are embedded. If you link to a local HTML file, that's copied, and turning on this param would keep it from being copied. I don't think you could turn it back off, because this is an "unless set" test not "unless true". 

In my past PDF extensions, I always set this param to avoid copying the HTML -- assumption was that the PDF was self-contained and would be viewed / downloaded from other locations, not the output directory, so that any "local file" links did not need the files copied. 

So I kind of don't like the current behavior, but _for anyone who relies on it_, setting this parameter for all PDF transforms as in #3721 would possibly be a backwards incompatible change that is a real pain for most people to undo.

jelovirt:
[...]

PR #3721, opened 2021-03-25
#3720 [fix] preprocess.copy-html.skip is missing from org.dita.pdf2

thendarion:
Signed-off-by: thendarion <16006640+thendarion@users.noreply.github.com>


jelovirt (changes requested):
See https://github.com/dita-ot/dita-ot/issues/3720#issuecomment-896675473, IMO it would be too much work for users to revert back to current behaviour.

jelovirt:
Close as won't fix because of the high cost for existing users.

PR #3739, opened 2021-04-30
Retain @id in conreffed content when possible

### LINKED ISSUE #3736: ID processing of con(key)ref-ed elements
## Expected Behavior
Elements keep their ID as specified. Links to conrefed element IDs should work

## Actual Behavior
Element IDs are randomized to prevent duplicate IDs. PDF and HTML result differs. HTML can link to the topic but not the actual ID, PDF link does not work at all.

## Possible Solution

1. Only randomize IDs if they are not unique in the context
2. Do not randomize IDs and only accept first ID occurrence in any given context ingoring any other equal IDs. Similar to key behaviour in keyscope context.

## Steps to Reproduce
Create PDF / HTML from https://gist.github.com/Snafu/fe99e12669ae98707722631764d623e8

Not releveant

## Environment
Any DITA-OT environment

* DITA-OT version: DITA-OT version 3.6.0
  Windows
* How did you run DITA-OT?
  Directly or via oXygen
* Transformation type:
  HTML 5, PDF, HTML PDF

jelovirt:
## Description
Change conref  processing to retain original `@id` in conreffed content when possible. Duplicate IDs are rewritten as before and IDs with the same value from conreffing topic take priority. 

## Motivation and Context
Fixes #3736.

## How Has This Been Tested?
Added XSpec tests and existing tests pass.

## Type of Changes
- Enhancement

Mention in release notes. This changes the `@id` and `@href` attribute values in output, but now they should be more correct and align with Principle of Least Surprise.


src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -898,4 +895,37 @@ See the accompanying LICENSE file for applicable license.
...
+        </xsl:for-each>
+      </xsl:variable>
+      <xsl:choose>
+        <xsl:when test="empty($ids) and $orig-id is $first">
+          <xsl:attribute name="id" select="$orig-id"/>
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:message>
+            <xsl:processing-instruction name="level" select="'DEBUG'"/>
+            <xsl:value-of>Rewrite conref content @id from "<xsl:value-of select="@conref:orig-id"/>" to "<xsl:value-of select="@id"/>"</xsl:value-of>
```
jelovirt, line 917: @robander @infotexture Should we have ID for every message like this? We don't really need it for localization of the message text, but it would make it easier to find code that threw the message, be it error or debug.

src/test/xsl/preprocess/conrefImpl.xspec
```diff
@@ -55,5 +55,114 @@
     </x:scenario>    
     
   </x:scenario>
+  
+  <!--
```
jelovirt, line 59: This works in Oxygen but not in our unit test driver. Since this PR doesn't change this template, keeping as out commented code

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
jelovirt: Looking at XSLT code, we actually have only few cases where we throw ad hoc messages like this. This should be moved to messages.xml and given a message ID.

src/main/config/messages_template.xml
```diff
@@ -808,6 +808,11 @@ See the accompanying LICENSE file for applicable license.
     <reason>A content reference in a constrained document type cannot be resolved because it would violate one of the document constraints "%1".</reason>
     <response>The current constrained document may only reuse content from documents with equivalent constraints.</response>
   </message>
+  
+  <message id="DOTX077D" type="DEBUG">
```
jelovirt, line 812: @robander @infotexture Should this be info level instead? Is this something that you care about? With content that has a lot of duplicate `@id` in resulting conreffed content, this can cause a lot of messages.

src/main/config/messages_template.xml
```diff
@@ -808,6 +808,11 @@ See the accompanying LICENSE file for applicable license.
...
+    <reason>Rewrite conref content @id from "%1" to "%2".</reason>
+    <response></response>
```
jelovirt, line 814: @infotexture Do you have suggestion for better copy? I don't think we should have anything in the response, because it would be something like

> If you don't want @id to be rewritten, change the conreffed source to use some other ID value that is not already in use in the conreffing or other conreffed content.

😬

robander: I don't think the current message will be meaningful to most people -- maybe something like "Resolving content references in [file] results in duplicate ID [id]. Rewriting resolved version to [newid]"?

src/main/config/messages_template.xml
robander: I think this should be INFO level, could even say warning. For most sources this will not come up, but when it does it has the potential to change how your links resolve.

Snafu:
Don't know if you want to label this as user error, but if you have a reference to another element in the same topic but don't use the same topic identifier `#./targetId` but rather explicity state the current topic id `#topicId/targetId`, then this link will be broken when used in a conref, since it is not handled in generate-href.
You could argue 

See here for an example: https://github.com/Snafu/dita-keyref
Problematic part is here: https://github.com/Snafu/dita-keyref/blob/master/content_topic.dita#L25

src/main/config/messages_template.xml
jelovirt: Converted this to info.

src/main/config/messages_template.xml
jelovirt: Change the copy but the "in [file]" bit comes from log context so removed that. Now the message is e.g.

> [conref] file:/Volumes/tmp/src/conref/b.dita:7:20: [DOTX077I][INFO]: Resolving content references results in duplicate ID 'bar'. Rewriting resolved version to 'd22e17'.

Snafu:
@jelovirt I can create a new issue but I think it is related to this PR since before everything got randomized.
I'm not sure if I consider this a bug or not though. I would say `#./targetId` is working as expected, the case where you explicity write the topic ID would be the broken case in my mind if anything is broken at all.
I think it is fair to say that this should not work with conref because there is an explicitly stated ID that is not present after resolving whereas the same topic identifier does what it is supposed to do. The ID of the referenced element would never be taken over according to the spec, even if you conref the entire topic it would not end up in the resolved content because ID attribute is excluded, see [here](http://docs.oasis-open.org/dita/v1.2/os/spec/archSpec/conref.html) and [here](http://docs.oasis-open.org/dita/dita/v1.3/os/part1-base/archSpec/base/conref-attributes-specified-on-elements.html).
As far as I can tell the same topic identifier was only added in [DITA 1.3](http://docs.oasis-open.org/dita/dita/v1.3/os/part1-base/langRef/attributes/thehrefattribute.html#thehrefattribute), at least it is not mentioned in [DITA 1.2](http://docs.oasis-open.org/dita/v1.2/os/spec/common/thehrefattribute.html#thehrefattribute).
This means you cannot have links within the same topic with DITA 1.2 when you use conref to include the topic somwhere if I understand it correctly.
Should I create a new issue?

Snafu:
Hi @jelovirt, does the status "closed" mean that this fix won't be implemented or will it be included in version 3.7?
From what I followed your fix was just missing a review.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -487,55 +494,54 @@ See the accompanying LICENSE file for applicable license.
     <xsl:param name="current-relative-path" tunnel="yes" as="xs:string"/>
     <xsl:param name="conref-filename" tunnel="yes" as="xs:string?"/>
     <xsl:param name="conref-ids" tunnel="yes" as="xs:string*"/>
-    <xsl:attribute name="href">
```
jelovirt, line 490: Use `xsl:copy` to output the attribute with the same value instead of generating all cases inside a common `xsl:attribute`

src/test/resources/lang/exp/xhtml/lang-areg.html
```diff
@@ -47,21 +47,21 @@ <h1 class="title topictitle1" id="ariaid-title1">Topic elements with generated t
 
 <p class="p">Xref to the concept in this file: <a class="xref" href="lang-common.html#concept">Concept elements with generated text</a>. Now do it again, <a class="xref" href="lang-common.html#concept">with link text</a>.</p>
 
-<p class="p">Xref to the fig in this topic: <a class="xref" href="#topic__d7e166">شكل 1</a>. Now do it again, <a class="xref" href="#topic__d7e166">with link text</a>.</p>
+<p class="p">Xref to the fig in this topic: <a class="xref" href="#topic__d22e167">#topic__d22e167</a>. Now do it again, <a class="xref" href="#topic__d22e167">with link text</a>.</p>
```
jelovirt, line 50: Multiple files in the  `lang` test change because it uses conref to build up the source documents.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -34,18 +34,32 @@ See the accompanying LICENSE file for applicable license.
...
+          <xsl:with-param name="conref-ids" tunnel="yes" select="()"/>
+        </xsl:apply-templates>
+      </xsl:document>
+    </xsl:variable>
+    <xsl:variable name="resolved-ids" as="document-node()">
+      <xsl:document>
+        <xsl:apply-templates select="$resolved/node()" mode="id"/>
+      </xsl:document>
+    </xsl:variable>
+    <xsl:apply-templates select="$resolved-ids/node()" mode="href"/>
```
jelovirt, line 52: So the processing is split into three stages.

1. Resolve conrefs recursively and output attributes in conreffed content using `conref` namespace for resolution in following steps.
2. Resolve `conref:orig-id` that are ID:s in conreffed content. If the same ID is not used in non-conreffed content, use the original ID; otherwise use `generate-id` to create a new ID. The latter option was the only output in the old code.
3. Resolve `conref:href` which checks if some link types in conreffed content should resolve to the original ID or generated ID.

Steps 2 and 3 cannot be merged because we need to know the effective ID values first, before the know what element ID to use in the links.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -194,7 +208,7 @@ See the accompanying LICENSE file for applicable license.
           <xsl:value-of select="$conref-source-topicid"/>
         </xsl:when>
         <xsl:otherwise>
-          <xsl:value-of select="ancestor-or-self::*[contains(@class, ' topic/topic ')][1]/@id"/>
+          <xsl:value-of select="dita-ot:get-closest-topic(.)/@id"/>
```
jelovirt, line 211: Utility function we've had before.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -487,102 +501,50 @@ See the accompanying LICENSE file for applicable license.
...
-      </xsl:when>
-      <xsl:when test="$conref-gen-id = $href-gen-id">
-        <xsl:text>#</xsl:text>
-        <xsl:value-of select="$conref-source-topicid"/>
-        <xsl:text>/</xsl:text>
-        <xsl:value-of select="$conref-gen-id"/>
-      </xsl:when>
+      <xsl:when test="exists($conref-filename)">
+        <xsl:attribute name="id" select="generate-id(..)"/>
+        <xsl:attribute name="conref:orig-id" select="."/>
```
jelovirt, line 543: We generate a new ID for conreffed content as well as retain the original. We do this is step 1 so we don't have to generate any IDs in steps 2 or 3.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -487,102 +501,50 @@ See the accompanying LICENSE file for applicable license.
...
-          <xsl:value-of select="generate-id(..)"/>
-        </xsl:when>
-        <xsl:otherwise>
-          <xsl:value-of select="."/>
-        </xsl:otherwise>
-      </xsl:choose>
-    </xsl:attribute>
-  </xsl:template>
-
-  <xsl:template name="generate-href">
```
jelovirt, line 541: This code has been replaced with step 3, mode `href`.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -897,4 +859,87 @@ See the accompanying LICENSE file for applicable license.
     </xsl:call-template>
   </xsl:template>
 
+  <xsl:template match="*[@conref:orig-id]" mode="id">
```
jelovirt, line 862: Mode `id` is step 2.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -897,4 +859,85 @@ See the accompanying LICENSE file for applicable license.
...
+  
+  <xsl:template match="node() | @*" mode="id" priority="-10">
+    <xsl:copy>
+      <xsl:apply-templates select="node() | @*" mode="#current"/>
+    </xsl:copy>
+  </xsl:template>
+  
+  <xsl:key name="conref-id" match="*[@conref:orig-id]" use="@conref:orig-id"/>
+
+  <xsl:template match="@conref:href" mode="href">
```
jelovirt, line 897: Mode `href` is step 3

robander (approved):

src/test/resources/conrefbreaksxref/exp/preprocess/conrefbreaksxref.dita
```diff
@@ -3,20 +3,20 @@
 <shortdesc class="- topic/shortdesc ">This topic conrefs to another, which uses xref internally.</shortdesc>
 <body class="- topic/body ">
 <section class="- topic/section "><title class="- topic/title ">Section that is pulled to another topic.</title><p class="- topic/p ">Look
-as the figure in this section: <xref class="- topic/xref " href="#conrefbreaksxref/d7e48" type="fig"><?ditaot gentext?>Figure 1</xref>.</p><p class="- topic/p ">Look
-at a figure in this topic, outside of this section: <xref class="- topic/xref " href="#conrefbreaksxref/d7e56"/></p><p class="- topic/p ">Now
+as the figure in this section: <xref class="- topic/xref " href="#conrefbreaksxref/d8e48" type="fig"><?ditaot gentext?>Figure 1</xref>.</p><p class="- topic/p ">Look
+at a figure in this topic, outside of this section: <xref class="- topic/xref " href="conrefbreaksxreftarget.dita#this-target/out" type="fig"><?ditaot gentext?>Figure 2</xref></p><p class="- topic/p ">Now
```
jelovirt, line 7: This is an example of a case where the old output was broken local link and the new output is link to the original conref target. That may or may not actually be live based on whether the conref target is also output. So in most cases, the link will be broken in e.g. HTML5 output but in a different way than it used to be.

PR #3741, opened 2021-05-02
Add support for disabling colour in CLI using TERM or NO_COLOR

jelovirt:
## Description
Support disabling colour in CLI output with `TERM=dumb` or `NO_COLOR` environment variables and `--no-color` option.

## Motivation and Context
Follow conventions suggested in [12 Factor CLI Apps](https://medium.com/@jdxcode/12-factor-cli-apps-dd3c227a0e46#a808) for playing nice with common conventions.

## How Has This Been Tested?
Manual tests.

## Type of Changes

TBD

## Checklist

### DIFF
--- a/src/main/resources/cli_en_US.properties
+++ b/src/main/resources/cli_en_US.properties
@@ -9,6 +9,7 @@ error_msg=Error: %s
 help.option.help=Print help information
 help.option.debug=Enable debug logging
 help.option.verbose=Enable verbose logging
+help.option.no-color=Disable color in terminal
 # Version subcommand
 version=DITA-OT version %s
 version.usage=dita version [options]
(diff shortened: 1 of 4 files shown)


src/main/resources/cli_en_US.properties
```diff
@@ -9,6 +9,7 @@ error_msg=Error: %s
 help.option.help=Print help information
 help.option.debug=Enable debug logging
 help.option.verbose=Enable verbose logging
+help.option.no-color=Disable color in terminal logging
```
jelovirt, line 12: @infotexture is the copy ok?

infotexture: I think that's fine. 👍 

— “…in terminal _output_” might be an alternative, but not sure it’s better 🤔

jelovirt: @infotexture I'm ok with either, you choose.

lief-erickson: Not sure this any better or of any help, but what about "Disable color in terminal"?

raducoravu:
Maybe it would make sense to sometimes force enable the console coloring also on Windows. For example if I start the DITA OT on Windows from Oxygen XML, we control the console view there and maybe in time we can add coloring to it. Or somebody may use some more advanced console on Windows which already supports the coloring scheme.

robander (approved):

PR #3759, opened 2021-06-02
Fix integrator path generation on Windows to output chimera paths

### LINKED ISSUE #3755: Cannot run integrator on Windows and then use the DITA-OT on Linux
Let's say I have a DITA-OT folder on Windows and run the DITA-OT integrator in it, it re-generates various resources among which the `config/org.dita.dost.platform/plugin.properties`. All references in the re-generated resources contain Windows-style `\` separators.
Then I take the entire DITA-OT, copy it on a Linux or Mac machine and try to use it to publish a small DITA Map, it will not work, I get an error that the public ID of the DITA Map cannot be mapped through the XML catalogs.
I need to run the DITA-OT integrator again on Linux to make the DITA-OT work again.

In my opinion all resources generated by the DITA-OT integrator process should use `/` style path separators which work both on Windows and on Linux/Mac.

easirois:
## Description
When writing the configuration properties to the plugin.properties file make sure to always use the Linux path separator ("/").  So we can easily use the OT on both Windows and Linux without 

## Motivation and Context

Fixes #3755

## How Has This Been Tested?

Run the `dita install` command on a Windows machine.  Copy the DITA-OT over to a Linux machine and run an build on the Linux side.  I also run an DITA-OT build on the Windows machine with the Linux path and it worked fine too.

## Type of Changes

When putting the properties make sure to replace '\\" with '/'.

None.

## Checklist


src/main/java/org/dita/dost/platform/Integrator.java
```diff
@@ -321,7 +321,7 @@ private void integrate() throws Exception {
                 //configuration.put(name, ditaDir.getAbsolutePath());
                 configuration.put(name, ".");
             } else {
-                configuration.put(name, FileUtils.getRelativePath(new File(ditaDir, "dummy"), f.getPluginDir()).getPath());
+                configuration.put(name, FileUtils.getRelativePath(new File(ditaDir, "dummy"), f.getPluginDir()).getPath().replace("\\", "/"));
```
jelovirt, line 324: If we convert this to always use `/`, then it might as well be a URI.

easirois: There is getRelativeUnixPath that does the trick but it's deprecated.  Any specific reason?  
configuration.put(name, FileUtils.getRelativeUnixPath(aFile.toURI().toString(), f.getPluginDir().toURI().toString()));

Even if we were to create a method that takes URIs, it would still be calling the following underneath the covers.
private static String getRelativePath(final String basePath, final String refPath, final String sep) 

Suggestions?

easirois: Ahh found this method in URLUtils

easirois: The last commit is the good one.

jelovirt:
@easirois in the end I think it's better to use the deprecated `getRelativeUnixPath` method; the deprecation is to signal "don't use this unless you really know what you're doing".

Are you able to test this on Windows?

jelovirt (dismissed):

easirois:
yes, I can test this on both sides.

jelovirt (approved):

PR #3770, opened 2021-07-01
Load custom CSS last to facilitate overrides

infotexture:
## Description

The HTML5 `chapterHead` template currently inserts the custom CSS file reference specified via `args.css` _before_ the contents of the custom header file specified via `args.hdf`.

This means that if the header file _(or any of the 3 preceding* template calls)_ is [used](https://github.com/infotexture/dita-bootstrap/blob/develop/includes/bootstrap.hdf.xml) to insert references to external CSS stylesheets for frameworks like [Bootstrap](https://getbootstrap.com/docs/5.0/getting-started/introduction/#css), the framework styles take precedence over any equivalent rules in the user's custom stylesheet.

This PR adjusts the order to ensure that custom stylesheets can be used to override any references included in the custom header file without resorting to increased selector specificity or `!important` hacks in the custom CSS.

### * Related question

If I'm reading the code right, in addition to the `processHDF` template that inserts the contents of the custom header file specified via `args.hdf`, there are **4** additional named templates that are all empty by default, but can be customized to insert content into the `<head>` element of the generated HTML files for each topic:

1. `copyright`
2. `gen-user-head`
4. `gen-user-scripts`
5. `gen-user-styles`

The same redundant approach appears to be used for running headers and footers that are inserted into the HTML `<body>` element:

- an empty `gen-user-header` template, followed by the `processHDR` template that inserts the contents of the custom header file specified via `args.hdr`
- an empty `gen-user-footer` template, followed by the `processFTR` template that inserts the contents of the custom footer file specified via `args.ftr`

@robander I assume the empty templates preceded the ability to insert content via external files and are retained for backwards compatibility, but is there any advantage to customizing these empty templates over the use of external `args.*` files?

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -2358,12 +2358,12 @@ See the accompanying LICENSE file for applicable license.
       <xsl:call-template name="generateDefaultMeta"/> <!-- Standard meta for security, robots, etc -->
       <xsl:apply-templates select="." mode="getMeta"/> <!-- Process metadata from topic prolog -->
       <xsl:call-template name="copyright"/>         <!-- Generate copyright, if specified manually -->
-      <xsl:call-template name="generateCssLinks"/>  <!-- Generate links to CSS files -->
       <xsl:call-template name="generateChapterTitle"/> <!-- Generate the <title> element -->
       <xsl:call-template name="gen-user-head" />    <!-- include user's XSL HEAD processing here -->
       <xsl:call-template name="gen-user-scripts" /> <!-- include user's XSL javascripts here -->
       <xsl:call-template name="gen-user-styles" />  <!-- include user's XSL style element and content here -->
       <xsl:call-template name="processHDF"/>        <!-- Add user HDF file, if specified -->
+      <xsl:call-template name="generateCssLinks"/>  <!-- Generate links to CSS files -->
     </head>
   </xsl:template>
 


infotexture:
😞 Looks like the integration tests are failing here. 

Might need help fixing those if my change requires modifications to the tests.

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -2358,12 +2358,12 @@ See the accompanying LICENSE file for applicable license.
       <xsl:call-template name="generateDefaultMeta"/> <!-- Standard meta for security, robots, etc -->
       <xsl:apply-templates select="." mode="getMeta"/> <!-- Process metadata from topic prolog -->
       <xsl:call-template name="copyright"/>         <!-- Generate copyright, if specified manually -->
-      <xsl:call-template name="generateCssLinks"/>  <!-- Generate links to CSS files -->
       <xsl:call-template name="generateChapterTitle"/> <!-- Generate the <title> element -->
       <xsl:call-template name="gen-user-head" />    <!-- include user's XSL HEAD processing here -->
       <xsl:call-template name="gen-user-scripts" /> <!-- include user's XSL javascripts here -->
       <xsl:call-template name="gen-user-styles" />  <!-- include user's XSL style element and content here -->
       <xsl:call-template name="processHDF"/>        <!-- Add user HDF file, if specified -->
+      <xsl:call-template name="generateCssLinks"/>  <!-- Generate links to CSS files -->
```
jelovirt, line 2366: Should we instead move `processHDF` up instead of moving `generateCssLinks` down. Then this would allow you to declared _everything_ before the normal `gen*` template output. So move `generateHDF` to after `generateCharset`?

infotexture: If we consider **only** `processHDF` and `generateCssLinks`, then yes: that would solve the problem just as well — but if anyone were to embed an external stylesheet using `gen-user-head` or `gen-user-styles` the same problem would appear — the custom CSS file would not override the framework stylesheets, since they'd be listed _later_.

Though the template names _suggest_ certain usage, any one of them could theoretically be used to insert arbitrary code into the `<head>` element, including linked stylesheets.

Since the `gen-user-*` templates are all empty by default, it's hard to say whether these should precede or follow  `processHDF`, but I'd be inclined to preserve the original order there in case there's a reason for that sequence.

Apart from charset first, I'm not sure if there's any rule or best practice convention regarding the order of elements within an HTML `<head>`, but I think it's fairly common for the `<title>` element to precede linked stylesheets.

robander: I can't say for sure, but my guess is that there is not a reason for the original sequence with those gen-* stubs; I think it's likely historical and just hasn't changed. I think it's probably safe to say if you specify a CSS file with your parameter, that's the one you want to have the most weight. I wonder though, if you've decided to go with `gen-user-styles` to put in one or more CSS links, those would previously override `commonltr.css` but after this move, the common one would win, right?

As for order within `<head>`, the title is supposed to come after the character set metadata so that any tool reading it already knows the character set. Otherwise I haven't run into rules or best practices for the order of these.

infotexture:
@robander Fixed the integration tests with afad60422a7fd419e7193a8e74c7db0dc0935148, they now pass.

>  if you've decided to go with gen-user-styles to put in one or more CSS links, those would previously override `commonltr.css` but after this move, the common one would win, right?

Yes, there's apparently no way to move just the custom CSS file call, as `generateCssLinks` includes both common & custom CSS.

Wondering if we should formally deprecate those old `gen-user-*` things? 🤔

jelovirt (approved):

PR #3777, opened 2021-07-27
dita-ot:generate-stable-id(): Handle input element is not within a topic

drmacro:
## Description

Added check for empty $topic and in that case use generate-id() on the input element.

## Motivation and Context

Fixes runtime error resulting from generation of simple table head for properties tables.

## How Has This Been Tested?

Tested in 3.6.1 using test document that fails with unmodified code.

## Type of Changes

Fixes runtime failure for properties tables (and possibly all simple tables) that do not have explicit headers. 
More generally, fixes failure for any case that generated content ends up using dita-ot:generate-stable-id().

  No documentation change
  No
## Checklist

  Added unit test generate-stable-id.xspec


drmacro:
Redoing pull request with signed commits.

PR #3780, opened 2021-08-04
Support fileset for style configuration

jelovirt:
## Description
Support `<style>` element for pipeline XSLT task and `<param>` element with [Resource](https://ant.apache.org/manual/Types/resources.html) value.

## Motivation and Context
The built-in XSLT task supports defining XSLT with a `<style>` element and this PR adds support to match it. Supporting `<style>` enables passing the stylesheet as a Resource serves as initial support for classpath based plugin resources.

## How Has This Been Tested?
Existing tests pass.
## Type of Changes

[Use custom <pipeline> elements](https://www.dita-ot.org/dev/topics/plugin-coding-conventions.html) documents use of `<pipeline>` element for XSLT transformations but doesn't include details about how it works. This PR makes pipeline XSLT module more consistent with Ant's XSLT task, so no new documentation is needed.

### DIFF
--- a/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
+++ b/src/main/java/org/dita/dost/ant/ExtensibleAntInvoker.java
@@ -8,12 +8,10 @@
 package org.dita.dost.ant;
 
 import com.google.common.collect.ImmutableSet;
-import org.apache.tools.ant.BuildException;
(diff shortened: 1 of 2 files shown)


jelovirt:
@infotexture we don't really have a label for changes like these that change plugin configuration syntax etc.

robander (approved):

PR #3792, opened 2021-08-17
Populate ancestor links during map pull step

jelovirt:
## Description
Populate ancestor links during map pull step in preprocessing.

## Motivation and Context
Other link role types are already populated so this adds a missing role processing.

## How Has This Been Tested?
Current tests and new tests for `maplinkImpl.xsl`.

## Type of Changes

Release notes to list the new feature. https://www.dita-ot.org/dev/reference/preprocess-maplink.html doesn't really go into details what role types are supported.


jelovirt:
This still needs to take `include.rellinks` into consideration when generating the links.

jelovirt:
@robander A side effect of this is that now that we generate those ancestor links, the [`include.rellinks`](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/build_init.xml#L136) contains `ancestor`. This in turn [activates breadcrumbs creation](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/xsl/rel-links.xsl#L102). Should we combine this with a change in HTML5 and XHTML that either

-   Add a guard to [`generateBreadcrumbs`](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/xsl/topic.xsl#L2576) or remove the default processing that generates the breadcrumbs, or
-   Change [`args.rellinks=nofamily`](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/build_init.xml#L139) to not include `ancestor`.

My preference is the latter, because the only reason `ancestor` and `descendant` were not excluded from `nofamily` was that `ancestor` was not automatically generated by preprocess at the time and I missed it. So IMO this is a feature and a bug fix.

robander:
I agree `nofamily` should include ancestor, that sounds like a fix.

If someone currently generates HTML with links set to `all`, they'll now start getting these links where they didn't before. Do we have an easy way to turn off just these, to keep the old behavior? Many of the past systems I've worked with had links set to `all`, but then loaded the HTML into a framework that handled all of the ancestor links based on TOC location.

I also wonder what the result will be if the topic is in the map in two locations. I guess the answer there is probably the same as "what is the result for parent linking", you either get two parents/ancestor paths or you force unique copies of the topic.

jelovirt:
Those uses who now use `include.rellinks="all"` and don't want those `ancestor` links have to change to `include.rellinks="#default parent child sibling friend next previous cousin descendant sample external other"`. This will have to be in the release notes/migration docs.

robander (approved):

PR #3794, opened 2021-08-20
Use modes instead of named templates in HTML5 stylesheets

jelovirt:
## Description
Use modes instead of named templates in HTML5 stylesheets.

## Motivation and Context
Using modes allows plug-ins to override templates and use `xsl:next-match`, where named templates can only be overridden.

## How Has This Been Tested?
Test pass
## Type of Changes

Existing plugins will not break. Plugins should however be changed to use modes instead of overriding named templates for

- ` copyright` 
- ` gen-endnotes` 
- ` generateDefaultMeta` 
- ` generateCssLinks` 
- ` generateChapterTitle` 
- ` processHDF` 
- ` generateBreadcrumbs` 
- ` processHDR` 
- ` processFTR` 
- ` generateCharset`

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -1990,8 +1990,12 @@ See the accompanying LICENSE file for applicable license.
   <xsl:template name="spec-title-nospace">
    <xsl:if test="@spectitle"><div style="margin-bottom: 0;"><strong><xsl:value-of select="@spectitle"/></strong></div></xsl:if>
   </xsl:template>
-    
+
+  <!-- Deprecated since 3.7, use "copyright" mode instead -->
   <xsl:template name="copyright">
+    <xsl:apply-templates select="." mode="copyright"/>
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -1990,8 +1990,12 @@ See the accompanying LICENSE file for applicable license.
   <xsl:template name="spec-title-nospace">
    <xsl:if test="@spectitle"><div style="margin-bottom: 0;"><strong><xsl:value-of select="@spectitle"/></strong></div></xsl:if>
   </xsl:template>
-    
+
+  <!-- Deprecated since 3.7, use "copyright" mode instead -->
```
jelovirt, line 1994: However, we cannot rewrite current uses of

```xml
<xsl:call-template name="copyright"/>
```

into

```xml
<xsl:apply-templates select="." mode="copyright"/>
```

Because that would not longer work with current plugins that have overridden `copyright` named template.

robander (approved):

Issue #3795, opened 2021-08-23
Variable references inside <navtitle> do not resolve inside a multiple-value @keyscope region


chrispy-snps:
## Expected Behavior
If the current `@keyscope` region has multiple values (1) and a variable is defined (2), that variable should resolve both in a `<navtitle>` element (3) as well in topic content (4).

![multiple_value_keyscope](https://user-images.githubusercontent.com/50950969/130520837-5650c03c-600c-493b-8ef2-873b1c484945.png)

## Actual Behavior
The variable does not resolve in the `<navtitle>` element (3). The HTML output is as follows:

```
Map Title
- Navigation Title for ""
  - Topic Title for "My Product"
```

This issue occurs when there are multiple space-separated values in `@keyscope`, whether they are the same value (`"top top"` or `"local local"`) or different (`"top local"`).

This testcase might seem improbable, but it is reduced for simplicity. In our production environment, a top-level `<mapref>` with a top-level scope instantiates a submap's `<map>` with its own local scope. The `@keyscope` values are concatenated during `mapref` preprocessing per ["2.3.4.2 Key scopes"](http://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part3-all-inclusive/archSpec/base/keyScopes.html), thus triggering this bug and causing a product variable to fail evaluation in a `<navtitle>` within the submap.

## Steps to Reproduce
Testcase:
[multiple-value-keyscope.zip](https://github.com/dita-ot/dita-ot/files/7034995/multiple-value-keyscope.zip)

Command:
`dita -i map.ditamp -f html5`

## Environment

* DITA-OT version: 3.6.1
* Operating system and version: Linux (Ubuntu 20.04)
* How did you run DITA-OT? `dita` command
* Transformation type: `html5`

jelovirt:
@chrispy-snps  #3796 has been merged. Has it affected this issue?

chrispy-snps:
@jelovirt - I confirmed that the fix for #3796 works around this bug when the `<mapref>` and `<map>` share the same `@keyscope` value:

```
<!-- map.ditamap -->
<?xml version="1.0" encoding="utf-8"?>
<?xml-model href="urn:oasis:names:tc:dita:rng:map.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
<map>
  <title>Map Title</title>
  <mapref href="submap.ditamap" keyscope="my_book"/>
[...]

PR #3796, opened 2021-08-24
merge duplicate @keyscope values between <mapref> and <map>

chrispy-snps:
## Description
Per ["2.3.4.2 Key scopes"](http://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part3-all-inclusive/archSpec/base/keyScopes.html), when a `<mapref>` element and its target `<map>` element both define `@keyscope` values, the resulting `@keyscope` value is a union of the original values.

In `mapref` preprocessing, this is currently implemented with simple string concatenation, which results in duplicate values when the `<mapref>` and `<map>` scopes are the same:

```
<submap keyscope="bookA bookA">
                  ^^^^^^^^^^^
```

This duplication can be avoided by combining the values using the `distinct-values()` XSLT function instead:

```
<submap keyscope="bookA">
                  ^^^^^
```

## Motivation and Context
I noticed this behavior while debugging another issue (#3795). We are adding local keyscopes to all our books, and thus this duplication will occur in all of our collection maps.

The motivation is (1) the desirability of keeping unique keyscope values through downstream processing, (2) the simplicity of the change, and (3) the low risk of the proposed change.

In addition, (4) this elegantly works around #3795 for our case.

The runtime improvement is perhaps negligible for Java-based processing of keyscopes, but perhaps more significant for XSLT-based processing of keyscopes.

## How Has This Been Tested?
I ran the `gradlew` tests and observed no new errors.

I published multiple book collections in our environment to HTML5 with and without the change, then compared the results.

## Type of Changes
- Coding style _(non-breaking change to code structure)_

- No documentation updates are needed for this change.
- No template interfaces have been altered, so other users' template overrides should be unaffected.

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
@@ -204,7 +204,7 @@ See the accompanying LICENSE file for applicable license.
                     <xsl:text> </xsl:text>
                     <xsl:value-of select="$target[contains(@class, ' map/map ')]/@keyscope"/>
                   </xsl:variable>
-                  <xsl:value-of select="string-join(distinct-values(tokenize($keyscope, '\s+')), ' ')"/>
+                  <xsl:value-of select="string-join(distinct-values(tokenize(normalize-space($keyscope), '\s+')), ' ')"/>
                 </xsl:attribute>
               </xsl:if>
               <xsl:apply-templates select="$target/@chunk"/>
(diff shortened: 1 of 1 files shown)


src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
```diff
@@ -204,7 +204,7 @@ See the accompanying LICENSE file for applicable license.
                     <xsl:text> </xsl:text>
                     <xsl:value-of select="$target[contains(@class, ' map/map ')]/@keyscope"/>
                   </xsl:variable>
-                  <xsl:value-of select="normalize-space($keyscope)"/>
+                  <xsl:value-of select="string-join(distinct-values(tokenize($keyscope, '\s+')), ' ')"/>
```
jelovirt, line 207: If the input is `"  foo    bar"`, this will output `" foo bar"`

```suggestion
                  <xsl:value-of select="string-join(distinct-values(tokenize(normalize-space($keyscope), '\s+')), ' ')"/>
```

jelovirt (changes requested):

jelovirt (approved):

chrispy-snps:
@jelovirt - thank you for the `normalize-space()` improvement!

PR #3803, opened 2021-09-19
Update to Gradle 7.2

jelovirt:
## Description
Update to build scripts to work with Gradle 7.2.

## Motivation and Context
Keep up with Gradle development and make it easier for docs repo to also use new Gradle version.
## How Has This Been Tested?
Build and distribution build works.
## Type of Changes

None


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.7-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
```
infotexture, line 3: This is the minimal upgrade path per [Gradle docs](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper):

> Note that running the wrapper task once will update `gradle-wrapper.properties` only, but leave the wrapper itself in `gradle-wrapper.jar` untouched. This is usually fine as new versions of Gradle can be run even with ancient wrapper files. If you nevertheless want all the wrapper files to be completely up-to-date, you’ll need to run the `wrapper` task a second time.

In https://github.com/dita-ot/docs/pull/355, I also [upgraded the wrapper](https://github.com/dita-ot/docs/pull/355/files#diff-e9721dc750619a21053ddea8a5d04929a608877d8c5daec1b57d243d3424e745).

Shall we do that here too?

jelovirt: Yeah, I'll update the wrapper just in case

PR #3806, opened 2021-09-26
Add commonattributes mode in HTML5 to allow for easier extension

jelovirt:
## Description
Retrofit `commonattributes` named template to be backed up by `commonattributes` mode in HTML5 plug-in.

## Motivation and Context
Modes allow for easier extension combined with `<xsl:next-match>` than named templates.

## How Has This Been Tested?
Existing tests pass.

## Type of Changes

Existing plug-ins continue to work without migration.

Recommended change is to replace

```xml
<xsl:template name="commonattributes">
  <!-- whole copy of commonattributes named template with customizations -->
</xsl:template>
```
with
```xml
<xsl:template match="@* | node()" mode="commonattributes">
  <xsl:param name="default-output-class" as="xs:string*"/>
  <xsl:next-match>
    <xsl:with-param name="default-output-class" select="$default-output-class"/>
  </xsl:next-match>
  <!-- customizations -->
</xsl:template>
```

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -1794,11 +1794,18 @@ See the accompanying LICENSE file for applicable license.
   <!-- Process standard attributes that may appear anywhere. Previously this was "setclass" -->
   <xsl:template name="commonattributes">
     <xsl:param name="default-output-class"/>
+    <xsl:apply-templates select="." mode="commonattributes">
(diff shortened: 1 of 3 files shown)


infotexture (commented):
Similar named templates are used in the XHTML and PDF plug-ins.

Would it make sense to use the mode-based approach there as well for consistency?

infotexture (approved):

chrispy-snps:
This is a great enhancement - thank you!

Issue #1188, opened 2012-04-26
Localization should be independent of transtype.


jelovirt:
Converted from [SourceForge issue 3516644](http://sourceforge.net/support/tracker.php?aid=3516644), submitted by @jannylund

Localization should be configured in one place. Currently PDF2 uses an "insertVariable" functionality and XHTML uses the "getString" functionality. This would require splitting out the variable functionality so it would not be part of an output type.

A first step against a better solution would be to make the variable files in an generic format and generate the transtype specific localizations using integrator.xml  (or maybe on the fly?).

macksol:
I am really confused at where we are with this enhancement, any clarification would be helpful.

1/ In the documentation doc/dev_ref/plugin-addgeneratedtext.html - it refers to using "getString" but "getString" is deprecated in favour of "getVariable".
I presume this just means that the documentation needs updated?

2/ the org.dita.pdf2 plugin uses "insertVariable" but this is also deprecated and using standard code produces lots of these deprecated messages in the log.
I presume over time the pdf plugin will be rewritten to use "getVariable" and until then we live with the log messages?

3/ the org.dita.pdf2 plugin (and my heavily customised version of it) used to support string customizations by having the base strings in the cfg/common/vars sub folder with the client specific version in the  customization sub folder. The old code used to check in the cusomtization file first and if it could not find that content then it would look in the base file.
Is this no longer supported?

PR #3817, opened 2021-10-30
Convert old strings files to new format

jelovirt:
## Description
Convert old strings files using `<strings>` or `<vars>` elements to the new format (#1188) that uses `<variables>` element.

## Motivation and Context
The new format has more features and configuration in a single format is easier to maintain.

## How Has This Been Tested?
Existing tests pass.
Mention in release notes.


infotexture (commented):
> Mention in release notes.

For the _Release Notes_ and migration topics, how shall we describe the necessary changes for those who have customized strings in custom plug-ins built for previous versions?

- Will existing overrides in a `/cfg/common/vars/` folder of a custom PDF plug-in just work?
- If those overrides begin with a `<vars xmlns="http://www.idiominc.com/opentopic/vars">` root element as in previous versions, will that still work, or would they need to change that to the new `<variables>` root element used here?
- If an existing custom HTML plug-in has an override file at `xsl/common/strings-de-de.xml`, would it still override the common German strings as in previous versions if it contains `strings/str` constructs, or would those need to be rewritten to the `variables/variable` format used here?

jelovirt:
@infotexture,

> Will existing overrides in a /cfg/common/vars/ folder of a custom PDF plug-in just work?

Variables under PDF2 are not overrides. They are just PDF2's variable definitions. If you list variables under org.dita.base and org.dita.pdf2, there are no duplicates. So PDF2 doesn't override anything.

> If those overrides begin with a <vars xmlns="http://www.idiominc.com/opentopic/vars"> root element as in previous versions, will that still work, or would they need to change that to the new <variables> root element used here?

PDF2 has no overrides but some other plug-in might have overrides. The format of the override doesn't matter. You can still mix all _three_ formats and the one specified last in the configuration will win, just like before.

> If an existing custom HTML plug-in has an override file at xsl/common/strings-de-de.xml, would it still override the common German strings as in previous versions if it contains strings/str constructs, or would those need to be rewritten to the variables/variable format used here?

Plug-ins don't need to change anything. The logic how strings are resolved has not changed, just the format.

jelovirt:
@infotexture This PR could serve as basis for deprecating the two old formats and suggesting users to move over to new (Since 2014!) format.

infotexture (commented):
Tested by making a simple change to the figure labels and building the docs output to confirm that the modified strings appear in the output.

- ✅ Changes to `<variable id="Figure">` in `org.dita.base/xsl/common/strings-en-us.xml` are applied to HTML5 output.
- ❌ Changes to `<variable id="Figure Number">` in `org.dita.pdf2/cfg/common/vars/en.xml` are not applied to PDF output.

jelovirt:
@infotexture Can't reproduce the problem of not seeing the modifications when strings are changed.

infotexture (approved):
Sorry, PDF issue in my [previous comment](https://github.com/dita-ot/dita-ot/pull/3817#pullrequestreview-799573609) was user error.

- ✅ Changes to `<variable id="Figure.title">` in `org.dita.pdf2/cfg/common/vars/en.xml` are applied to PDF.

PR #3829, opened 2021-11-17
Replace attribute-set reflection in PDF2

### LINKED ISSUE #3827: Replace attribute reflection in PDF2
## Description
Remove [attribute reflection](https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.pdf2/xsl/common/attr-set-reflection.xsl#L319) and replace it with utility function that extracts attributes generated from attribute sets. This is related to #3826 because attribute reflection has support for the legacy customization `cfg` URIs.

## Possible Solution
Replace with utility template that extracts attributes from carrier element.

```xsl
<xsl:template name="get-attributes" as="attribute()*">
  <xsl:param name="element" as="element()"/>
  <xsl:sequence select="$element/@*"/>
</xsl:template>
```

```xsl
<xsl:call-template name="get-attributes">
  <xsl:with-param name="element" as="element()">
    <dummy xsl:use-attribute-sets="topic.topic.topic.title"/>
  </xsl:with-param>
</xsl:template>
```

Alternative would be to used a template mode, but using it would require variables to store the carrier element

```xsl
<xsl:variable name="element" as="element()">
  <dummy xsl:use-attribute-sets="topic.topic.topic.title"/>
</xsl:variable>
<xsl:apply-templates select="$element" mode="get-attributes"/>
```

jelovirt:
## Description
Replace PDF2 attribute-set reflection with direct attribute-set use.

## Motivation and Context
Fixes #3827

## How Has This Been Tested?
Added integration test for PDF that compares `topic.fo` with expected output.

## Type of Changes

Backwards compatible as code for the old attribute-set reflection is retained but PDF2 now uses new attribute-set generation everywhere reflection used to be uses.

### DIFF
--- a/src/test/java/org/dita/dost/IntegrationTestPdf.java
+++ b/src/test/java/org/dita/dost/IntegrationTestPdf.java
@@ -1,0 +1,40 @@
+/*
+ * This file is part of the DITA Open Toolkit project.
+ *
+ * Copyright 2021 Jarno Elovirta
+ *
+ * See the accompanying LICENSE file for applicable license.
+ */
+
+package org.dita.dost;
+
+import org.junit.Test;
+
+import java.io.File;
+import java.nio.file.Paths;
+
+import static org.dita.dost.AbstractIntegrationTest.Transtype.HTML5;
+import static org.dita.dost.AbstractIntegrationTest.Transtype.PDF;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class IntegrationTestPdf extends AbstractIntegrationTest {
+
+    public AbstractIntegrationTest builder() {
+        return new IntegrationTestPdf();
+    }
+
+    @Override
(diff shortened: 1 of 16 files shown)


src/test/java/org/dita/dost/IntegrationTestPdf.java
```diff
@@ -0,0 +1,40 @@
...
+        return new IntegrationTestPdf();
+    }
+
+    @Override
+    Transtype getTranstype(Transtype transtype) {
+        return transtype;
+    }
+
+    @Test
+    public void pdf() throws Throwable {
```
jelovirt, line 33: This just tests that the generated FO file matches the expected file. So at least there is some test that fails when you change the FO output. The test case is suboptimal and should at some point be replaced with a test that has all DITA elements. We could then use that for HTML5 too.

robander (approved):
+100 on time to remove these

PR #3830, opened 2021-11-20
Refactor attribute generation in XSLT

jelovirt:
## Description
In XSLT stylesheets, refactor `<xsl:value-of select="…">` inside `<xsl:attribute>` into `<xsl:attribute select="…"/>`. Use `commonattributes` to generate `@class` attributes in HTML5 and XHTML output.

## Motivation and Context
The `<xsl:value-of>` inside `<xsl:attribute>` is XSLT 1.0 style. We now have XSLT 2.0 and moving to XSLT 3.0, the coding conventions should also evolve.

## How Has This Been Tested?
Existing tests pass.
## Type of Changes

None needed.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -264,7 +264,7 @@ See the accompanying LICENSE file for applicable license.
         </xsl:choose>
     </xsl:param>
     <xsl:element name="h{$headinglevel}">
-        <xsl:attribute name="class">topictitle<xsl:value-of select="$headinglevel"/></xsl:attribute>
+        <xsl:attribute name="class" select="concat('topictitle', $headinglevel)"/>
         <xsl:call-template name="commonattributes">
           <xsl:with-param name="default-output-class">topictitle<xsl:value-of select="$headinglevel"/></xsl:with-param>
         </xsl:call-template>
@@ -1446,14 +1446,14 @@ See the accompanying LICENSE file for applicable license.
       </xsl:call-template>
     </xsl:variable>
     <xsl:if test="not($height-in-pixel = '100%')">
-      <xsl:attribute name="height">
+      <xsl:attribute name="height" select="number($height-in-pixel)">
         <!--xsl:choose>
           <xsl:when test="../@scale and string(number(../@scale))!='NaN'">          
             <xsl:value-of select="number($height-in-pixel) * number(../@scale)"/>
           </xsl:when>
-          <xsl:otherwise-->
+          <xsl:otherwise>
             <xsl:value-of select="number($height-in-pixel)"/>
-          <!--/xsl:otherwise>
+          </xsl:otherwise>
         </xsl:choose-->
       </xsl:attribute>
     </xsl:if>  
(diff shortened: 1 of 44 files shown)


src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -2150,12 +2161,14 @@ See the accompanying LICENSE file for applicable license.
   
   <!-- Catch footnotes that should appear at the end of the topic, and output them. -->
   <xsl:template match="*[contains(@class, ' topic/fn ')]" mode="genEndnote">
-    <div class="p">
```
robander, line 2153: Don't know if removing this was intentional - I believe it will cause footnotes to be stacked slightly closer together by default, without the extra line break of paragraph-like content? CSS can still have control by matching `fn` just noted this seemed unrelated

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -1466,14 +1477,14 @@ See the accompanying LICENSE file for applicable license.
       </xsl:call-template>
     </xsl:variable>
     <xsl:if test="not($width-in-pixel = '100%')">
-      <xsl:attribute name="width">
+      <xsl:attribute name="width" select="number($width-in-pixel)">
         <!--xsl:choose>
```
robander, line 1481: Should this commented section be removed?

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -1247,8 +1257,9 @@ See the accompanying LICENSE file for applicable license.
     <xsl:apply-templates select="*[contains(@class, ' ditaot-d/ditaval-startprop ')]" mode="out-of-line"/>
     <xsl:call-template name="spec-title-nospace"/>
     <pre>
-      <xsl:attribute name="class" select="name()"/>
-      <xsl:call-template name="commonattributes"/>
+      <xsl:call-template name="commonattributes">
+        <xsl:with-param name="default-output-class" select="name()"/>
```
robander, line 1261: This seems odd ... normal behavior with the old one would be set the class, then it's basically ignored by the `commonattributes`, which includes the element name by default. Now we're calling common attributes, but passing in a default of the element name, which ... should match default processing?

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -383,8 +383,10 @@ See the accompanying LICENSE file for applicable license.
   
   <!-- section processor - div with no generated title -->
   <xsl:template match="*[contains(@class, ' topic/section ')]" name="topic.section">
-    <section class="section">
-      <xsl:call-template name="commonattributes"/>
+    <section>
+      <xsl:call-template name="commonattributes">
+        <xsl:with-param name="default-output-class" select="'section'"/>
```
robander, line 388: Several templates do this (I noted on a couple others) - where we previously set class on the element directly, then called commonattributes, we now just call common attributes but with the default. I think that the previous way of doing it almost made the original setting a no-op, because `commonattributes` resets class entirely, unless you override it to skip setting class. In that case - where you override `commonattributes` to avoid setting a class on `section` - the net result here is just to remove `@class` right? 

With a 4.0 release I'm not really worried about compatibility per-se, just want to make sure we do what is the best design. Might be better to delete the class as you've done, but without passing in `section` as a default, as that should already be set (by default) due to class ancestry

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -1446,14 +1451,14 @@ See the accompanying LICENSE file for applicable license.
       </xsl:call-template>
     </xsl:variable>
     <xsl:if test="not($height-in-pixel = '100%')">
-      <xsl:attribute name="height">
+      <xsl:attribute name="height" select="number($height-in-pixel)">
         <!--xsl:choose>
```
robander, line 1455: Could remove this commented code as well (missed commenting on it in last look)

robander (approved):
Looks good, still a couple redundant classes in `pr-d.xsl` and `sw-d.xsl` for XHTML, but it's XHTML so I'm not so concerned.

robander (approved):

PR #3831, opened 2021-11-27
Simplify move meta tests

jelovirt:
## Description
Move MetadataInheritance integration tests to MoveMetaModule unit test.

## Motivation and Context
Simplifying an integration test that runs the whole preprocess into a module unit test reduces test processing time, makes it easier to see what the module does, and allow easy test for in-memory store use.

## How Has This Been Tested?
New tests pass.
## Type of Changes

- New tests

None needed because this is not visible to end users.


robander (approved):
Files changed: 308 is amusing (but also makes sense for this test) -- good change

PR #3842, opened 2021-12-28
Fix link handling in conreffed content

jelovirt:
## Description
Link handling in conreffed content has bugs in how target elements are selected. This PR fixes the conref stylesheet query issues.

## Motivation and Context
The `conrefImpl.xsl` was originally written some 17 years ago with XSLT 1.0. When the conref code was moved to XSLT 2.0, the relavant code was not adjusted for the differences between XSLT 1.0 and 2.0. The code also contains a bug in selection of the first topic in a document.

## How Has This Been Tested?
Existing tests pass or have been fixed to expect correct output.

## Type of Changes

Mention in release notes. Fixes a bug that's been in the codebase for 17 year. Seventeen years.

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
@@ -547,7 +547,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="conref-topicid" as="xs:string">
       <xsl:choose>
         <xsl:when test="empty($topicid)">
-          <xsl:value-of select="//*[contains(@class, ' topic/topic ')][1]/@id"/>
+          <xsl:value-of select="/descendant-or-self::*[contains(@class, ' topic/topic ')][1]/@id"/>
         </xsl:when>
         <xsl:otherwise>
           <xsl:value-of select="$topicid"/>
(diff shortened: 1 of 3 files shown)


src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -547,7 +547,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="conref-topicid" as="xs:string">
       <xsl:choose>
         <xsl:when test="empty($topicid)">
-          <xsl:value-of select="//*[contains(@class, ' topic/topic ')][1]/@id"/>
+          <xsl:value-of select="/descendant-or-self::*[contains(@class, ' topic/topic ')][1]/@id"/>
```
jelovirt, line 550: A classic XPath bug where `//` desugars to `/descendant-or-self::node()/` and the `[1]` predicate no longer selects the first match of `//`, but rather anything that is the first child of parent node.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
```diff
@@ -559,10 +559,10 @@ See the accompanying LICENSE file for applicable license.
     <xsl:variable name="conref-gen-id" as="xs:string">
       <xsl:choose>
         <xsl:when test="empty($elemid) or $elemid = $href-elemid">
-          <xsl:value-of select="generate-id(key('id', $conref-topicid)[contains(@class, ' topic/topic ')]//*[@id = $href-elemid])"/>
+          <xsl:value-of select="generate-id((key('id', $conref-topicid)[contains(@class, ' topic/topic ')]//*[@id = $href-elemid])[1])"/>
```
jelovirt, line 562: When `$conref-topicid` starts to return correct values, `id` key will also start returning values. XSLT 1.0 will accept multiple argument values for `generate-id()`, but XSLT 2.0 will not. Thus we have to explicitly use only the first.

src/main/plugins/org.dita.base/xsl/preprocess/conrefImpl.xsl
robander: Ouch

robander (approved):

Issue #3431, opened 2019-12-12
Cannot properly expand abbreviated-form


raducoravu:
Based on this message:

https://groups.io/g/dita-users/message/44956

I'm attaching this sample project, when converted to PDF the warning appears and the abbreviated form reference is not properly expanded:

           [DOTX060W][WARN]: Key 'key_g_dccs' was used in an abbreviated-form element, but the key is not associated with a glossary entry. Abbreviated-form should ONLY be used to reference to a glossary entry.

[abbrevform-not-resolving.zip](https://github.com/dita-ot/dita-ot/files/3954354/abbrevform-not-resolving.zip)

I looked inside this XSLT stylesheet:

     plugins\org.dita.pdf2\xsl\fo\abbrev-domain.xsl

in this template:

    <xsl:template match="*[contains(@class,' abbrev-d/abbreviated-form ')]" name="topic.abbreviated-form">

this variable:

     <xsl:variable name="target" select="key('id', substring(@href, 2))[1]" as="element()?"/>

seems to match the topicref to the glossentry instead of the glossentry. Looking at the _MERGED.xml file, both the topicref to the glossentry and the glossentry element have the same @id attribute value which is wrong in my opinion. Also even if I manually set an @id attribute value on the glossref in the DITA Map:

           <glossref keys="key_g_dccs" href="glossary/g_dccs.dita#g_dccs" print="yes" id="test"/>

the @id attribute will not be preserved in the _MERGED.xml document.

A possible easy fix for this would be to rewrite the "target" variable as:

     <xsl:variable name="target" select="key('id', substring(@href, 2))[self::*[contains(@class,' glossentry/glossentry ')]]" as="element()?"/>

but the problem with the @id attribute set on two different elements with the same value in the _MERGED.xml document would remain.

shaneataylor:
I think this is a duplicate of #2773 .

raducoravu:
Duplicate of #2773

Issue #2773, opened 2017-09-08
DOTX060W is fired when using correct glossary entry reference


xephon2:
## Expected Behavior
No warning should be fired, because the DITA code should be correct.

**bookmap.ditamap**

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE bookmap PUBLIC "-//OASIS//DTD DITA BookMap//EN" "bookmap.dtd">
<bookmap>
  <booktitle>
    <mainbooktitle>test</mainbooktitle>
  </booktitle>
  <chapter href="chapter.dita"/>
  <backmatter>
    <mapref href="glossary.ditamap"/>
  </backmatter>
</bookmap>
```

**glossary.ditamap**

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "map.dtd">
<map>
  <title>Glossary</title>
  <glossref keys="GLS_Test" href="GLS_Test.dita"/>
</map>
```

**GLS_Test.dita**

```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE glossentry PUBLIC "-//OASIS//DTD DITA Glossary//EN" "glossary.dtd">
<glossentry id="GLS_Test">
  <glossterm>Test</glossterm>
  <glossdef>This is just a test.</glossdef>
  <glossBody>
    <glossSurfaceForm>Test (T)</glossSurfaceForm>
    <glossAlt>
      <glossAcronym>T</glossAcronym>
    </glossAlt>
  </glossBody>
</glossentry>
```

**chapter.dita**

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
<topic id="chapter">
  <title>test</title>
  <body>
    <p><abbreviated-form keyref="GLS_Test"/></p>
  </body>
</topic>
```

## Actual Behavior
Warning is fired

## Possible Solution
Debug/refactor glossary processing

## Steps to Reproduce

1. Download and extract ZIP archive.
2. Call `dita --input bookmap.ditamap --format pdf2 -v`.

[abbreviation-test.zip](https://github.com/dita-ot/dita-ot/files/1287595/abbreviation-test.zip)

## Copy of the error message, log file or stack trace

```
transform.topic2fo.main:
     ...
[...]

PR #3844, opened 2022-01-06
In PDF2: Select correct abbrev href target.

simtind:
## Description
Select glossref instead of topicref as target when evaluating abbreviated-form gloss entries in pdf transformation.

## Motivation and Context
Currently abbreviated-form gloss entriy references raises error DOTX060W as reported in #2773 and #3431. This issue seems to have been introduced in #3210. This PR does not revert the changes from #3210, but rather fixes the indexing so the correct href'd element is selected. 

A more thorough fix discussed in #2773 and #3431 would be to update the pipeline to decorate the elements with separate ids so they can be correctly selected in the xsl transform.

## How Has This Been Tested?
- Tested on internal code base that manifested the issue. 
- Also tested on reproducer examples provided in the issue tickets.
-- The initial example from #2773, abbreviation-test.zip did need to add print="yes" to the glossref in glossary.ditamap
-- Otherwise, all examples built successfully with the patch, and failed with DOTX060W without it.

## Type of Changes

None needed

### DIFF
--- a/src/main/plugins/org.dita.pdf2/xsl/fo/abbrev-domain.xsl
+++ b/src/main/plugins/org.dita.pdf2/xsl/fo/abbrev-domain.xsl
@@ -25,7 +25,7 @@ See the accompanying LICENSE file for applicable license.
   
   <xsl:template match="*[contains(@class,' abbrev-d/abbreviated-form ')]" name="topic.abbreviated-form">
     <xsl:variable name="keys" select="@keyref"/>
-    <xsl:variable name="target" select="key('id', substring(@href, 2))[1]" as="element()?"/>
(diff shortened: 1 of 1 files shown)


chrispy-snps:
@kirkilj - I believe I accidentally fixed this in #4002 without seeing this PR (although I think the fix my PR might be a bit more robust).

chrispy-snps:
@kirkilj, @simtind - I apologize, I did not see this PR when I submitted #4002 (which is a similar fix, although a bit more robust). The original issue should be fixed in DITA-OT 3.7.4 and this PR can be closed.

However, the testcase revealed a second issue, which I filed #4026 for.

PR #3851, opened 2022-01-20
New process skip properties

### LINKED ISSUE #3845: Need better design for "skip" properties
## Description

Many of our Ant targets in the preprocess code have "skip" properties that allow you to skip running part of our build pipeline. These are considered internal properties and subject to change, but available to advanced users who understand the code. There are several problems:

* They are not documented, even for advanced users (this has come up in Slack a few times).
* The naming is inconsistent, with most based on Ant target names rather than on function. You can disable the code that cascades map metadata, but you have to know the Ant target is `move-meta-entries` making the property name `preprocess.move-meta-entries.skip`. Some do not use the Ant target name, such as `preprocess.clean-map-check.skip`. Where preprocess and preprocess2 have overlapping targets, preprocess2 uses the name from preprocess; where they do not, it creates a property with the Ant target name. An odd one is `clean-temp.skip` which is outside of preprocess.
* Property names based on Ant target name makes it hard to split or combine steps
* You have to set the property to disable something, which is not intuitive; setting `preprocess.keyref.skip=true` disables keyref resolution.
* Because of how Ant works, setting the variable to *anything* disables the target, it just checks if the property is set. So, setting `preprocess.keyref.skip=false` also disables keyref. This is likely familiar to people with a lot of Ant experience, but many of those trying to tweak DITA-OT builds do not have that background.

List of current skip properties in base plugin:

- `clean-temp.skip`
- `preprocess.branch-filter.skip`
- `preprocess.chunk.skip`
- `preprocess.clean-map-check.skip`
- `preprocess.clean-preprocess.skip`
- `preprocess.coderef.skip`
- `preprocess.conref.skip`
- `preprocess.conrefpush.skip`
- `preprocess.copy-files.skip`
- `preprocess.copy-flag.skip`
- `preprocess.copy-html.skip`
- `preprocess.copy-image.skip`
- `preprocess.debug-filter.skip`
- `preprocess.gen-list.skip`
- `preprocess.keyref.skip`
- `preprocess.map-profile.skip`
- `preprocess.maplink.skip`
- `preprocess.mapref.skip`
- `preprocess.move-meta-entries.skip`
- `preprocess.normalize-codeblock.skip`
- `preprocess.profile.skip`
- `preprocess.topic-profile.skip`
- `preprocess.topicpull.skip`

Note that not all are used in `target/@unless`. Some are used e.g.

```xml
<filter class="org.dita.dost.writer.NormalizeCodeblock"
        unless:set="preprocess.normalize-codeblock.skip"/>
```

## Possible Solution

Ideally, in a 4.0 major release we could:
* Create a consistent pattern for all of the current "skip" properties.
* The property names should be based entirely on function, not connected to the Ant target, so that even if we split a single `chunk` target into 5 targets, the "process chunking" property does not change and works to disable all 5.
* The property be set to "false" to disable a step. Ant has `istrue` condition that maps case-insensitive "true", "yes", and "on" to `true`; otherwise `false`.
* The actual Ant property used to skip a target is internal-only, and our build should fail if someone tries to set it; going forward this ensures we do not break anyone if those properties have to change
* All of this makes it possible to document the properties in a straightforward manner - "Set X to false to disable conref resolution" - without going into detail about why it's "enable this property to disable this feature", or why "disable this property" also disables it, etc.

I'm not sure of the best pattern for these or I'd have started with a PR that updated one step. Adding some ideas, using conref as an example; I like having `run` or `buildstep` or `build-step` in there, but am not attached to anything. I do like having the feature name last so all of these have the same prefix:
* `run.conref.step=true`
* `run.step.conref=true`
* `run.buildstep.conref=true`
* `build-step.conref=true`

## Potential Alternatives

We could simply document the properties that exist today, but we've intentionally avoided this:
* It locks in the properties like an external API, making them hard to change
* The issues above all make it hard to document, requiring an explanation of Ant and of the "enable to disable" naming

## Additional Context

This has come up twice in Slack over the last few months, both times about the `preprocess.clean-map-check.skip` property as a way to skip that step and retain information added by preprocessing.

jelovirt:
## Description
Add new properties to enable or disable process steps.

## Motivation and Context
Fixes #3845
## How Has This Been Tested?
TBD
## Type of Changes

TBD

### DIFF
--- a/src/main/plugins/org.dita.base/build_init.xml
+++ b/src/main/plugins/org.dita.base/build_init.xml
@@ -92,7 +92,7 @@ See the accompanying LICENSE file for applicable license.
     <!-- end to check input parameters -->
 
     <!-- begin to init required parameters -->
-    
+
     <property name="filter-stage" value="early"/>
     <condition property="filter-on-parse" value="true">
       <and>
@@ -117,7 +117,7 @@ See the accompanying LICENSE file for applicable license.
     </condition>
 
     <property name="args.grammar.cache" value="yes"/>
-    <property name="args.xml.systemid.set" value="yes"/>    
+    <property name="args.xml.systemid.set" value="yes"/>
     <!-- end to init required parameters -->
 
     <!-- create required directories -->
@@ -131,7 +131,7 @@ See the accompanying LICENSE file for applicable license.
 
     <!-- Validate the xml file or not,default is validation(true)-->
     <property name="validate" value="true"/>
-    
+
     <!-- Related links to output: none, all, nofamily -->
     <condition property="include.rellinks" value="">
       <equals arg1="${args.rellinks}" arg2="none"/>
@@ -160,7 +160,7 @@ See the accompanying LICENSE file for applicable license.
     <property name="onlytopic.in.map" value="false"/>
 
     <property name="link-crawl" value="topic"/>
-    
+
     <!--
     default value : warn
   fail     :1a) Fail quickly if files are going to be generated/copied outside of that directory
@@ -170,24 +170,25 @@ See the accompanying LICENSE file for applicable license.
     <property name="outer.control" value="warn"/>
 
     <condition property="inner.transform">
-        <equals arg1="${generate.copy.outer}" arg2="1"/>
+      <equals arg1="${generate.copy.outer}" arg2="1"/>
     </condition>
 
     <condition property="old.transform">
       <equals arg1="${generate.copy.outer}" arg2="3"></equals>
     </condition>
-    
+
     <property name="conserve-memory" value="false"/>
 
     <!-- Process skip -->
     <macrodef name="preprocess-skip-init">
       <attribute name="name"/>
+      <attribute name="step"/>
       <sequential>
         <fail if:set="@{name}.skip">Internal property @{name}.skip may not be set directly.</fail>
         <condition property="@{name}.skip">
           <and>
-            <isset property="build-step.@{name}"/>
-            <isfalse value="${build-step.@{name}}"/>
+            <isset property="build-step.@{step}"/>
+            <isfalse value="${build-step.@{step}}"/>
           </and>
         </condition>
         <echo level="info" if:set="@{name}.skip">Skip @{name}</echo>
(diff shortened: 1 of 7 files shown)


jelovirt:
@robander Then there’s [`filter-stage`](https://www.dita-ot.org/dev/parameters/parameters-base.html#base__filter-stage) which I think no one ever has even noticed or used. It’s used to initialize `preprocess.profile.skip`.

Should we have one `build-step.profile` that defined whether filtering is used at all? And should that also apply to branch filtering?

robander:
Yeah I think I agree with your comment there -- prefer the first option, more verbose with 2 options (but true/false for running the step stays consistent).

Not sure if that one profile option should be used for both external DITAVAL filtering and for branch filtering. That's tough for me. They are clearly both filtering, but both end up with such different results; especially with early filtering (external DITAVAL) and late branch filtering.

I guess that's one argument for keeping them apart -- if there is only one option to control the build step(s), then there might be an expectation that `filter-stage=early|late` applies to both external and branch filtering?

src/main/plugins/org.dita.base/build_init.xml
```diff
@@ -45,6 +45,22 @@ See the accompanying LICENSE file for applicable license.
...
+    <sequential>
+      <fail if:set="@{name}.skip">Internal property @{name}.skip may not be set directly.</fail>
+      <condition property="@{name}.skip">
+        <and>
+          <isset property="build-step.@{step}"/>
+          <isfalse value="${build-step.@{step}}"/>
+        </and>
+      </condition>
+      <echo level="info" if:set="@{name}.skip">Skip @{name}</echo>
+      <echo level="info" unless:set="@{name}.skip">Run @{name}</echo>
```
jelovirt, line 60: These are for debug logging, to be removed before merging

src/main/plugins/org.dita.base/build_init.xml
```diff
@@ -45,6 +45,22 @@ See the accompanying LICENSE file for applicable license.
     </sequential>
   </macrodef>
 
+  <macrodef name="preprocess-skip-init">
+    <attribute name="name"/>
+    <attribute name="step"/>
+    <sequential>
+      <fail if:set="@{name}.skip">Internal property @{name}.skip may not be set directly.</fail>
```
jelovirt, line 52: We may want to change this into an error, not sure. And we could give this error an error code.

src/main/plugins/org.dita.base/plugin.xml
```diff
@@ -172,6 +172,78 @@ See the accompanying LICENSE file for applicable license.
       <val>true</val>
       <val default="true">false</val>
     </param>
+    <param name="build-step.clean-temp" desc="Run process clean-temp" type="enum">
```
jelovirt, line 175: This could be `build-step.cleanup`?

src/main/plugins/org.dita.base/plugin.xml
```diff
@@ -172,6 +172,78 @@ See the accompanying LICENSE file for applicable license.
...
+    </param>
+    <param name="build-step.branch-filter" desc="Run process branch-filter" type="enum">
+      <val default="true">true</val>
+      <val>false</val>
+    </param>
+    <param name="build-step.chunk" desc="Run process chunk" type="enum">
+      <val default="true">true</val>
+      <val>false</val>
+    </param>
+    <param name="build-step.clean-preprocess" desc="Run process clean-preprocess" type="enum">
```
jelovirt, line 187: This is totally internal step that has no direct meaning in DITA.

It's the step that will go through all files in temp and try to rewrite the filenames to something that remotely looks like what the source was. So if you have `topic.dita` as the source file and preprocess2 will mangle this to `<HASH>.dita` in temp dir, this step will rename `<HASH>.dita` back to `topic.dita`, taking uplevels into account.

So this step is not something I would expect end users to understand without docs.

src/main/plugins/org.dita.base/build_template.xml
```diff
@@ -25,10 +25,9 @@ See the accompanying LICENSE file for applicable license.
       </condition>
       <param name="1" value="${transtype}"/>
     </dita-ot-fail>
-    <property name="clean.temp" value="true"/>
-    <condition property="clean-temp.skip">
-      <isfalse value="${clean.temp}"/>
-    </condition>
+    <property name="clean.temp" value="yes"/>
```
jelovirt, line 28: The correct value space for `clean.temp` is `yes` and `no`.

src/main/plugins/org.dita.base/build_template.xml
```diff
@@ -25,10 +25,9 @@ See the accompanying LICENSE file for applicable license.
...
+    <property name="build-step.clean-temp" value="${clean.temp}"/>
+    <preprocess-skip-init name="clean-temp" step="clean-temp"/>
```
jelovirt, line 30: Instead of directly setting `*.skip` properties, set `build-step.*` and let the macro handle the rest.

jelovirt:
@raducoravu Does this raise any ideas?

src/main/plugins/org.dita.base/build_init.xml
```diff
@@ -45,6 +45,27 @@ See the accompanying LICENSE file for applicable license.
     </sequential>
   </macrodef>
 
+  <macrodef name="preprocess-skip-init">
```
jelovirt, line 48: This may not be the best name, because it can be used for transtype specific targets too. So maybe `init-skip-property`?

raducoravu:
@jelovirt I'm afraid I do not know much about this, so there are a bunch of parameters, like "build-step.normalize-codeblock" which are now advertised and can be toggled off by the end user to skip various stages, I have not needed this myself, is this useful for debugging or for making the transformation run faster by skipping stages?
One parameter I would find useful would be to switch a transtype between using "preprocess" and "preprocess2" by using a parameter instead of modifying its build file.

src/main/plugins/org.dita.base/plugin.xml
robander: Maybe even `build-step.cleanup-temp` ... not sure if "cleanup" would be more readable, but that might be because I've been setting `clean-temp` for 15 years. So maybe cleanup + temp (clear that it's cleaning some temp dir)?

robander:
@raducoravu it's for a couple things really. The current skip properties are not documented and the names are not good. If you set the "skip" property to true, it skips the step; if you set the "skip" property to false, it also skips the step. We've had people find and ask about these properties (and ask why they are not documented), but the skip properties are really internal properties. 

So this change does a few things
1. Gives the properties consistent, readable names that can be documented
2. Makes the values clearer, you can set the build step to "true" to run it and "false" to skip it

I've _very rarely_ used these properties to optimize processing - for example, skip keyref processing for a huge set of content that I knew did not have keys. But mostly I think it's to enable better documentation and to make these properties public for developers that want to toggle the steps.

robander (approved):

src/main/plugins/org.dita.base/plugin.xml
jelovirt: I think we can leave it to be `clean-temp` in this PR just because it's more readable for us for the mean time. All the descriptions etc. need to be sorted and we can rename it then.

jelovirt:
@raducoravu For example PDF2 disables chunk processing with the old skip property and not with the new `build-step.chunk`. This is because PDF2 will merge everything together in its own module and normal chunking would be just waste of time. So now we have a documented property for this.

src/main/plugins/org.dita.base/plugin.xml
jelovirt: @infotexture IIRC this step does more than just restore filenames. But I'll check.

PR #3855, opened 2022-01-24
Add HTML5 table XSpec

jelovirt:
## Description
Add HTML5 table XSpec.
## Motivation and Context
Improve test coverage.
## How Has This Been Tested?
Adds tests.
No documentation needed.


src/test/xsl/plugins/org.dita.html5/xsl/tables.xspec
```diff
@@ -0,0 +1,71 @@
...
+          <span>Cell span sample</span>
+        </caption>
+        <colgroup>
+          <col style="width:33.33333333333333%"/>
+          <col style="width:33.33333333333333%"/>
+          <col style="width:33.33333333333333%"/>
+        </colgroup>
+        <tbody>
+          <tr class="valign-middle">
+            <td class="align-center colsep-1 rowsep-1" colspan="2">horizontally spanned</td>
```
jelovirt, line 56: @robander the stylesheet wrapper with its mock templates will not output the element ancestor type names. For this test, IMO those can be disabled. We want to make sure these `colsep-1` class names are generated correctly.

jelovirt:
@robander For 4.0, we might want to refactor the stylesheets a bit so that these tests are easier to write. Namely, reduce the need to mock templates. Right now we have to mock templates, variables, keys, and functions because in a XSpec test we want to keep the number of XSLT stylesheet imports to a minimum. That way the code under test is limited to only that which is tested. So if we want to test `tables.xsl`, we should not have to import `topic.xsl` in order to get everything required to compile the stylesheet. That's why all common code should be in styleheets that serves those feature specific stylesheets like `topic.xsl` or `simpletable.xsl`.

robander (approved):

robander:
Makes sense, and 4.0 seems like a good time to refactor/bulk up the testing

Issue #3867, opened 2022-02-01
Child navigation links to nested topics are incorrect/missing if there is a <ph keyref/>


chrispitude:
## Actual Behavior
When all of the following are true:

- A topic file that has nested subtopics in the same file.
- This topic file is included in two different submaps (content reuse).
- Both submaps are included (in different keyscopes) in the same top-level map.
- The topic file contains a `<ph keyref="..."/>` reference.

then if I run the `html5` transformation, the child navigation links for the second submap's topic instance are missing, which causes all child links to point to the top-level topic instead of the nested topics:

- Feature Topic
  - Concepts
  - Configuration
    - Option 1
    - Option 2
  - Limitations
- Feature Topic
  - Feature Topic (incorrect)
  - Feature Topic (incorrect)
    - Feature Topic (incorrect)
    - Feature Topic (incorrect)
  - Feature Topic (incorrect)

If I run the `pdf2` transformation instead of the `html5` transformation (which uses `preprocess2` instead of `preprocess`), then I get an even weirder result of recursively nested links that are a mix of correct and incorrect links.

## Expected Behavior
Child links should resolve correctly, whether there is a `<ph keyref="..."/>` or not.

If the `<ph keyref="..."/>` element is deleted, then the child navigation links are gathered correctly:

- Feature Topic
  - Concepts
  - Configuration
    - Option 1
    - Option 2
  - Limitations
- Feature Topic
  - Concepts
  - Configuration
    - Option 1
    - Option 2
  - Limitations

## Possible Solution
I am not sure yet, but I will follow up if I figure something out.

## Steps to Reproduce
1. Extract the following testcase:
   [incorrect_nested_topic_links.zip](https://github.com/dita-ot/dita-ot/files/7980758/incorrect_nested_topic_links.zip)
2. Run the following command in the testcase directory:
   `dita -i map.ditamap -f html5 -o out`
3. View the index.html file in a browser and note the incorrect links.

## Environment

* DITA-OT version: 3.7.0
* Operating system and version:
  linux
* How did you run DITA-OT?
  `dita` command
* Transformation type:
  `html5`, `pdf2`

chrispitude:
The issue occurs in this part of the `keyref` preprocessing step:

```
    <pipeline message="Resolve keyref." taskname="keyref">
      <module class="org.dita.dost.module.KeyrefModule" parallel="${parallel}">
        <param name="transtype" value="${transtype}"/>
      </module>
    </pipeline>
```

[...]

PR #3874, opened 2022-02-07
fix for #3867: Child navigation links to nested topics are incorrect/missing if there is a <ph keyref/>

chrispy-snps:

## Description
This pull request captures @raducoravu 's suggested fix for #3867.

## Motivation and Context
It solves the issue in #3867, in which nested subtopic fragments are inadvertently lost when a topic is duplicated for `@keyref` uniquification.

## How Has This Been Tested?
I ran the simple testcase from #3867 and confirmed that it fixed the issues by inspecting the intermediate files during preprocessing and confirming the fragment preservation, and also by inspecting the results of an `html5` transformation to confirm that the gathered child navigation links were fixed.

I also ran a larger testcase (about 6500 topics). I transformed to `html5` "old" and "new" the fix, then diffed the results for differences. The only differences were in two content-reused topics. In both topics, the navigation links in the "new" results were correct, whereas the ones in the "old" results were incorrect. In addition, backmatter index links into the reused topics were similarly corrected by this fix.

There were no other differences between the "old" and "new" results.

I also ran the `test`, `integrationTest`, and `e2eTest` tests, and they passed (except for a `resource_map` test, which failed with and without the fix).

## Type of Changes

No documentation or script changes are required.

### DIFF
--- a/src/main/java/org/dita/dost/module/KeyrefModule.java
+++ b/src/main/java/org/dita/dost/module/KeyrefModule.java
@@ -329,11 +329,11 @@ final class KeyrefModule extends AbstractPipelineModuleImpl {
                     AttributeMap atts = ni.attributes();
                     final QName rewriteAttrName = getReferenceAttribute(node);
                     if (rewriteAttrName != null) {
-                        String referenceValue = node.getAttributeValue(rewriteAttrName);
+                        URI referenceValue = toURI(node.getAttributeValue(rewriteAttrName));
                         if (referenceValue != null) {
                             for (final KeyScope s : ss) {
-                                URI resolved = map.uri.resolve(referenceValue);
-                                String frag = resolved.getFragment();
+                                final URI resolved = map.uri.resolve(referenceValue);
+                                final String fragment = resolved.getFragment();
                                 final URI href = stripFragment(resolved);
                                 final FileInfo fi = job.getFileInfo(href);
                                 if (fi != null && fi.hasKeyref) {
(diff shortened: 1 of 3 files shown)


jelovirt:
The Java code needs some tweaking, I push those changes directly later on.

jelovirt:
There are no changes to tests in this PR, which means we don't have a test for the issue. That should also be added.

raducoravu:
Tests should probably be added to the "src/test/java/org/dita/dost/module/KeyrefModuleTest.java" test class.
But tests are quite low level, @jelovirt might the existing "testWalkMapAndRewriteKeydefHref()" be a good method to copy and adjust for this use case?

chrispy-snps:
@raducoravu - I don't know how to write/update tests. If you provide me with the updated files, I can update the pull request.

raducoravu:
@chrispy-snps a new test DITA Map should be added here "/dost/src/test/resources/KeyrefModuleTest/src/test3.ditamap" with the contents I'm attaching:
[test3.ditamap.zip](https://github.com/dita-ot/dita-ot/files/8030405/test3.ditamap.zip)

In the test class "KeyrefModuleTest" a new test method named "testWalkMapAndRewriteKeydefHrefKeepAnchor" should be added with the contents:

[KeyrefModuleTest.java.zip](https://github.com/dita-ot/dita-ot/files/8030415/KeyrefModuleTest.java.zip)

chrispy-snps:
Test added. (Thanks @raducoravu !)

raducoravu:
@jelovirt is there anything else that needs to be changed in this pull request?

jelovirt (approved):

PR #3881, opened 2022-02-18
#3846 Do not append two anchors to target chunked reference.

### LINKED ISSUE #3846: HTML5 Output fails when having links between inner topics and chunk="to-content select-topic"
## Expected Behavior
Uploaded a sample project, in the "untitled5.dita" topic there are two inner topics and one refers to the other:

    <topic id="topic2">
        <title>Topic 2</title>
        <body>
            <p>link to topic3: <xref href="#topic3"/></p>
        </body>
    </topic>
    <topic id="topic3">
        <title>Topic 3</title>
        <body>
            <p>Para 3</p>
        </body>
    </topic>

The DITA Map splits the main topic in multiple topics:

    <topicref href="untitled5.dita#topic2" chunk="to-content select-topic"/>
    <topicref href="untitled5.dita#topic3" chunk="to-content select-topic"/>

The HTML5 output should produce HTML files "topic2.html" and "topic3.html", the "topic2.html" should have a link to "topic3.html"

## Actual Behavior
The HTML5 output fails with a fatal error in the chunking stage:

    [Error: java.lang.IllegalArgumentException: Illegal character in fragment at index 21: untitled5.dita#topic2#topic3

## Possible Solution
In the "org.dita.dost.writer.AbstractChunkTopicParser.processAttributes(Attributes)" method in the if clause:

              if (href.startsWith(SHARP)) {

if the "relative" variable has an anchor inside it, avoid merging it with the href "relative + href". Something like:

            if (href.startsWith(SHARP)) {
                // if @href refers to a location inside current parsing file
                // update @href to point back to current file
                // if the location is moved to chunk, @href will
                // be update again to the new location.
            	if(relative.toString().contains("#")) {
            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative.resolve(href).toString());
            	} else {
            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
            	}
            }

## Steps to Reproduce
[sampleChunkSelfLinks.zip](https://github.com/dita-ot/dita-ot/files/7844006/sampleChunkSelfLinks.zip)

Publish the attached project to HTML5.

     [Error: java.lang.IllegalArgumentException: Illegal character in fragment at index 21: untitled5.dita#topic2#topic3

## Environment

* DITA-OT version: 3.6.1
* Operating system and version: any
* How did you run DITA-OT? dita command
* Transformation type:HTML5

raducoravu:

## Description
Avoid appending anchor twice to target reference when referencing a chunked topic.

## Motivation and Context
Fixes #3846

## How Has This Been Tested?
Using samples from #3846 and with an automated test in the ChunkMapReaderTest class.

## Type of Changes

Not needed.

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
+++ b/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
@@ -301,13 +301,9 @@ public abstract class AbstractChunkTopicParser extends AbstractXMLWriter {
                 // update @href to point back to current file
                 // if the location is moved to chunk, @href will
                 // be update again to the new location.
-            	if(relative.toString().contains(SHARP)) {
-            		//If the relative reference already has an anchor, we need to replace that with the current anchor.
-            		final URI res = URLUtils.setFragment(relative, href.substring(1)); 
-            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res.toString());
-            	} else {
-            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
-            	}
+                final URI hrefUri = URI.create(href);
+                final URI res = URLUtils.setFragment(relative.resolve(hrefUri), hrefUri.getFragment());
+                XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res.toString());
             } else if (relative.toString().contains(SLASH)) {
                 // if new file is not under the same directory with current file
                 // add path information to the @href value
--- a/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
+++ b/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
@@ -14,6 +14,7 @@ import org.dita.dost.module.reader.TempFileNameScheme;
 import org.dita.dost.util.Job;
 import org.dita.dost.util.Job.FileInfo;
 import org.dita.dost.util.TopicIdParser;
+import org.dita.dost.util.URLUtils;
 import org.dita.dost.util.XMLUtils;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
@@ -300,9 +301,10 @@ public abstract class AbstractChunkTopicParser extends AbstractXMLWriter {
                 // update @href to point back to current file
                 // if the location is moved to chunk, @href will
                 // be update again to the new location.
-            	if(relative.toString().contains("#")) {
+            	if(relative.toString().contains(SHARP)) {
             		//If the relative reference already has an anchor, we need to replace that with the current anchor.
-            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative.resolve(href).toString());
+            		final URI res = URLUtils.setFragment(relative, href.substring(1)); 
+            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res.toString());
             	} else {
             		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
             	}
--- a/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
+++ b/src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
@@ -300,7 +300,12 @@ public abstract class AbstractChunkTopicParser extends AbstractXMLWriter {
                 // update @href to point back to current file
                 // if the location is moved to chunk, @href will
                 // be update again to the new location.
-                XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
+            	if(relative.toString().contains("#")) {
+            		//If the relative reference already has an anchor, we need to replace that with the current anchor.
+            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative.resolve(href).toString());
+            	} else {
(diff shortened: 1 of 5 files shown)


src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
```diff
@@ -300,7 +300,12 @@ Attributes processAttributes(final Attributes atts) {
                 // update @href to point back to current file
                 // if the location is moved to chunk, @href will
                 // be update again to the new location.
-                XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
+            	if(relative.toString().contains("#")) {
+            		//If the relative reference already has an anchor, we need to replace that with the current anchor.
+            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative.resolve(href).toString());
+            	} else {
+            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
+            	}
```
jelovirt, line 308: This should probably use [`URLUtils.setFragment(URI, String)`](https://github.com/dita-ot/dita-ot/blob/develop/src/main/java/org/dita/dost/util/URLUtils.java#L551).

So something like

```java
final URI res = URLUtils.setFragment(relative, href.substring(1)); 
XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res);
```

`setFragment` should be improved so that it tests if the `fragment` argument starts with `#`. If so, it will strip it out.

jelovirt (changes requested):

src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
raducoravu: Thanks, I followed your advice and made the change. I did not make changes to URLUtils.setFragment yet, should I also do that and add an extra TC for the URLUtils.setFragment method?

src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
```diff
@@ -300,7 +301,13 @@ Attributes processAttributes(final Attributes atts) {
...
                 // if the location is moved to chunk, @href will
                 // be update again to the new location.
-                XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
+            	if(relative.toString().contains(SHARP)) {
+            		//If the relative reference already has an anchor, we need to replace that with the current anchor.
+            		final URI res = URLUtils.setFragment(relative, href.substring(1)); 
+            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res.toString());
+            	} else {
+            		XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative + href);
+            	}
```
jelovirt, line 310:
```suggestion
            	final URI res = URLUtils.setFragment(relative, href.substring(1)); 
            	XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res.toString());
```

Couldn't we just use `URLUtils.setFragment(relative, href.substring(1))` for both cases?

In this branch from row 299, `href` is always `#something`. The results of using `setFragment`

* `foo.dita` ->  `foo.dita#something`
* `foo.dita#old` -> `foo.dita#sommething`

jelovirt (changes requested):

src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
jelovirt: @raducoravu You don't have to make the change to automatically strip the fragment because this goes to bug fix release. We can do it in `develop` branch.

src/main/java/org/dita/dost/writer/AbstractChunkTopicParser.java
raducoravu: @jelovirt then how about further simplifying this entire thing:

            if (href.startsWith(SHARP)) {
            	// if @href refers to a location inside current parsing file
            	// update @href to point back to current file
            	// if the location is moved to chunk, @href will
            	// be update again to the new location.
            	final URI res = URLUtils.setFragment(relative, href.substring(1)); 
            	XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res.toString());
            } else if (relative.toString().contains(SLASH)) {
                // if new file is not under the same directory with current file
                // add path information to the @href value
                XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative.resolve(href).toString());
            }

to this:

            if (href.startsWith(SHARP) || relative.toString().contains(SLASH)) {
                XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, relative.resolve(href).toString());
            }

?

jelovirt: Then it would not have `setFragment` anymore. Maybe the whole code should be something like

```java
final URI hrefUri = URI.create(href);
final URI res = URLUtils.setFragment(relative.resolve(hrefURI), hrefUri.getFragment()); 
XMLUtils.addOrSetAttribute(resAtts, ATTRIBUTE_NAME_HREF, res.toString());
```

jelovirt: Pushed this change after adding a test case that covers both with and without fragment branches.

jelovirt (approved):

Issue #3635, opened 2020-11-19
NullPointerException in JobMapper when referencing image on another disk drive.


raducoravu:
On Windows I have a DITA topic on the D:\ drive like:
D:\projects\samples\dita\flowers\topics\introduction.dita
and it references using an absolute href an image on the c:\ drive:
file:/C:/Users/my_user/Desktop/Snowdrop.jpg
Using the latest DITA OT sources I publish to HTML5:

      dita -i D:\projects\samples\dita\flowers\flowers.ditamap -f html5 -v

and get a NullPointerException reported:

      DITA-OT3.x\plugins\org.dita.base\build_preprocess.xml:364: java.lang.NullPointerException
	at org.dita.dost.ant.types.JobMapper.mapFileName(JobMapper.java:79)
	at org.apache.tools.ant.util.CompositeMapper.lambda$mapFileName$0(CompositeMapper.java:32)
	at java.util.stream.ReferencePipeline$3$1.accept(Unknown Source)
	at java.util.stream.ReferencePipeline$2$1.accept(Unknown Source)

raducoravu:
The exception is very similar to the one here #2844 but as it stems from a different problem, I'm logging this issue separately.

raducoravu:
I can no longer reproduce this problem with DITA OT 3.6.

raducoravu:
It seems I can still reproduce the problem, the image which is referenced with an absolute reference needs to be placed on another drive than the original DITA Map.

raducoravu:
Adding a sample which can be used to reproduce the NPE (at least on Windows):
[jobMapperNPE.zip](https://github.com/dita-ot/dita-ot/files/6115503/jobMapperNPE.zip)

raducoravu:
Had two more people report this for distinct projects, the NullPointerException does not help them understand what the problem is.

jlacour31:
I created a pull request that resolve this issue, if the paths have nothing in common, the method returns the left URI. Which for example in the case of CleanPreprocessModule.getBaseDir() is the original calculated baseDir.

PR #3885, opened 2022-02-21
NullPointerException in JobMapper when referencing image on another disk drive.

jlacour31:
## Description
A NPE appears in a topic refers an image from another drive.

## Motivation and Context
Fix for https://github.com/dita-ot/dita-ot/issues/3635

## How Has This Been Tested?
I added a new assert in CleanPreprocessModuleTest.getCommonBase()

## Type of Changes

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
+++ b/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
@@ -263,11 +263,12 @@ public class CleanPreprocessModule extends AbstractPipelineModuleImpl {
                 //
             } else {
                 final int common = Math.max(0, i);
-                final String path = Arrays.asList(la)
-                        .subList(0, common)
-                        .stream()
-                        .collect(Collectors.joining("/")) + "/";
-                return URLUtils.setPath(left, path);
+                final List<String> commons = Arrays.asList(la).subList(0, common);
+                if (commons != null && commons.size() > 1) {
+                	final String path = commons.stream()
+                			.collect(Collectors.joining("/")) + "/";
+                	return URLUtils.setPath(left, path);
+                }
             }
         }
         return null;
--- a/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
+++ b/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
@@ -264,7 +264,7 @@ public class CleanPreprocessModule extends AbstractPipelineModuleImpl {
             } else {
                 final int common = Math.max(0, i);
                 final List<String> commons = Arrays.asList(la).subList(0, common);
-                if (commons != null && commons.size() > 1) {
+                if (commons.size() > 1) {
                 	final String path = commons.stream()
                 			.collect(Collectors.joining("/")) + "/";
                 	return URLUtils.setPath(left, path);
--- a/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
+++ b/src/main/java/org/dita/dost/module/CleanPreprocessModule.java
@@ -15,26 +15,20 @@ import org.apache.commons.io.FileUtils;
 import org.apache.xml.resolver.tools.CatalogResolver;
 import org.dita.dost.exception.DITAOTException;
 import org.dita.dost.pipeline.AbstractPipelineOutput;
-import org.dita.dost.util.CatalogUtils;
-import org.dita.dost.util.Job;
+import org.dita.dost.util.*;
 import org.dita.dost.util.Job.FileInfo;
-import org.dita.dost.util.URLUtils;
-import org.dita.dost.util.XMLUtils;
 import org.dita.dost.writer.AbstractXMLFilter;
 import org.dita.dost.writer.LinkFilter;
 import org.dita.dost.writer.MapCleanFilter;
 import org.w3c.dom.Document;
 import org.xml.sax.XMLFilter;
 
-import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.dom.DOMResult;
 import javax.xml.transform.dom.DOMSource;
-import javax.xml.transform.stream.StreamSource;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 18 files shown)


src/test/java/org/dita/dost/module/CleanPreprocessModuleTest.java
```diff
@@ -52,6 +52,7 @@ public void getCommonBase() {
         assertEquals(create("file:/foo/"), module.getCommonBase(create("file:/foo/a"), create("file:/foo/bar/b")));
         assertEquals(create("file:/foo/"), module.getCommonBase(create("file:/foo/bar/a"), create("file:/foo/b")));
         assertEquals(create("file:/foo/"), module.getCommonBase(create("file:/foo/bar/a"), create("file:/foo/baz/b")));
+        assertEquals(create("file:/C:/"), module.getCommonBase(create("file:/C:/a"), create("file:/D:/b")));
```
jelovirt, line 55: If you revert the code change, this test doesn't test for NPE.

jelovirt (changes requested):

src/main/java/org/dita/dost/module/CleanPreprocessModule.java
```diff
@@ -263,11 +263,12 @@ URI getCommonBase(final URI left, final URI right) {
                 //
             } else {
                 final int common = Math.max(0, i);
-                final String path = Arrays.asList(la)
-                        .subList(0, common)
-                        .stream()
-                        .collect(Collectors.joining("/")) + "/";
-                return URLUtils.setPath(left, path);
+                final List<String> commons = Arrays.asList(la).subList(0, common);
+                if (commons != null && commons.size() > 1) {
```
jelovirt, line 267: `common` can never be null.

jelovirt:
@jlacour31 could you attach a stacktrace of the the NPE that this fixes and a test case, because I can't reproduce a NPE with the tests that were added.

src/main/java/org/dita/dost/module/CleanPreprocessModule.java
jelovirt, line 267: With a test

```java
assertEquals(create("file:/foo/"), module.getCommonBase(create("file:/foo/a/b/c"), create("file:/bar/b/c/d")));
```

this code passes, but the result is incorrect. The correct results should be `file:/`. The code makes assumption that the first directory in the path token should be retained and the resulting path can never start from root. This only applies to Windows, not *NIX.

jelovirt (changes requested):

src/main/java/org/dita/dost/module/CleanPreprocessModule.java
jlacour31: I removed the null check.

src/test/java/org/dita/dost/module/CleanPreprocessModuleTest.java
jlacour31: You're right, this doesn't cover directly the NPE, I don't know where I can test it.

src/main/java/org/dita/dost/module/CleanPreprocessModule.java
jlacour31: The problem is in fact that if the common path is only `file:/` a NPE appears in the HTML5 transformation.
And you're right this problem is only Windows related as Linux doesn't have drives.

jlacour31:
@jelovirt I copied the stacktrace from the original issue posted by @raducoravu:
```
DITA-OT3.x\plugins\org.dita.base\build_preprocess.xml:364: java.lang.NullPointerException
at org.dita.dost.ant.types.JobMapper.mapFileName(JobMapper.java:79)
at org.apache.tools.ant.util.CompositeMapper.lambda$mapFileName$0(CompositeMapper.java:32)
at java.util.stream.ReferencePipeline$3$1.accept(Unknown Source)
at java.util.stream.ReferencePipeline$2$1.accept(Unknown Source)
```

raducoravu:
So I'm pretty sure while the fix in this PR will prevent a NPE, it will still result in incorrect output.

@jelovirt well it's not great at all for a process to fail with a NPE, as the end user does not know why it fails, ideally if it fails in such cases it should fail with a precise error message, or show a precise error message and continue to work (probably with a broken link or two somewhere). How about if your validation filter would report such problems as precise error messages? And we also guard the NPE.

jelovirt:
@raducoravu You have a Windows machine. Could you create a small test where you have e.g.

- `c:\foo\root.ditamap`
- `c:\foo\topic.dita`
- `d:\bar\other.dita`

And you run preprocess until cleanup method? So e.g. add `<fail>` as the first step of `clean-preprocess` target. Then, could you paste the `.job.xml` here? I'd like to see what the `FileInfo`s look like in this case.

raducoravu:
@jelovirt should I perform this fix with the patch made by @jlacour31 added or not, just with the default DITA OT 3.7 code?

raducoravu:
@jelovirt I did what you asked, my c:\foo\root.ditamap has the topicrefs like this:

        <map>
            <title></title>
            <topicref href="topic.dita"/>
            <topicref href="file:/D:/bar/other.dita"/>
        </map>

I published to HTML5 using DITA OT 3.6 (which does not have Julien's patch applied). Added a fail before the "clean-preprocess"
Attaching the entire contents of the temp folder:

[html5.zip](https://github.com/dita-ot/dita-ot/files/8213931/html5.zip)

I'm not sure if this test is similar to the current problem, when an image is referenced from another drive though.

jelovirt:
@raducoravu I pushed a fix for Windows to this PR branch. Could you run the unit tests in `CleanPreprocessModuleTest` and check what this fix does with the problematic image on a different drive?

jelovirt:
> following errors in the console at different stages

@raducoravu that's to be expected, and all those preprocess steps are ran _before_ clean-preprocess target. So this fix works as expected because it's a preprocess1 limitation that you cannot mix source from different drives on Windows.

What's the results with the original test set with the image? No NPE from there?

raducoravu:
@jelovirt if I remove the path to the bar/other.dita from the DITA Map and add inside "topic.dita" an absolute image reference like this:

      <image href="file:/D:/projects/eXml/samples/dita/flowers/images/Gardeniaflower.jpg"                 id="image_uv5_zxd_vsb"/>

then publish to HTML5 I get these errors in the console:

     copy-image:     
      [copy] Copying 1 file to C:\foo\out\html5 Attempt to copy C:\foo\D:\projects\eXml\samples\dita\flowers\images\Gardeniaflower.jpg to C:\foo\out\html5\D:\projects\eXml\samples\dita\flowers\images\Gardeniaflower.jpg using NIO Channels failed due to 'The filename, directory name, or volume label syntax is incorrect'.  Falling back to streams.      
       [copy]       Failed to copy C:\foo\D:\projects\eXml\samples\dita\flowers\images\Gardeniaflower.jpg to C:\foo\out\html5\D:\projects\eXml\samples\dita\flowers\images\Gardeniaflower.jpg due to The filename, directory name, or volume label syntax is incorrect

but the publishing continues, no NPE. The result topic.html contains an absolute path to the image file:

jelovirt:
@raducoravu That sounds about right. This fixed the NPE but the code.

jelovirt:
This PR contains a merge commit that should be rebased out, but to save time, I'm merging this as is.

PR #3891, opened 2022-02-24
Log invalid character in file include

jelovirt:
## Description
Log a `DOTJ084E` message when file include process fails due to charset mismatch.

## Motivation and Context
When file include is in charset that is neither the system default or matches the explicit charset definition in the include, reading the file may fail. A dedicated error message for this makes debugging source issues easier.

## How Has This Been Tested?
New unit tests added for success and failure cases.

## Type of Changes

Describe the `DOTJ084E` message in more detail. This error is thrown e.g. when

* Included file is UTF-8 encoded but default encoding is used to read it (no `@format` or `@encoding`)
* Included file is Windows-1252 encoded (Windows default) but `<include>` element has `@encoding` with value UTF-8.

### DIFF
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -423,6 +423,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ084E" type="ERROR">
+    <reason>Included file %1 contained invalid byte sequence for charset %2</reason>
+    <response></response>
+  </message>
+
   <!-- End of Java Messages -->
     
   <!-- Start of XSL Messages -->  
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -424,8 +424,8 @@ See the accompanying LICENSE file for applicable license.
   </message>
 
   <message id="DOTJ084E" type="ERROR">
-    <reason>Included file %1 contained invalid byte sequence for charset %2</reason>
-    <response></response>
+    <reason>Couldn’t read '%1' with the %2 character set.</reason>
+    <response>Save the file with %2 encoding.</response>
   </message>
 
   <!-- End of Java Messages -->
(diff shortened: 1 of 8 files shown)


src/main/config/messages_template.xml
```diff
@@ -423,6 +423,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ084E" type="ERROR">
+    <reason>Included file %1 contained invalid byte sequence for charset %2</reason>
+    <response></response>
```
jelovirt, line 428: @infotexture Could you help with the message text? Is this understandable to our users? They probably don't understand what "invalid byte sequence means".  Should we go for less programmery

```xml
<reason>Failed to read file %1 with charset %2.</reason>
<response>Check that file is saved as %2.</response>
```

Which would result in something like

> Failed to read file file://…/topic.dita with charset UTF-8. Check that file is saved as UTF-8.

infotexture, line 428: I'd suggest something like this:

```suggestion
    <reason>Couldn’t read '%1' with the %2 character set.</reason>
    <response>Save the file with %2 encoding.</response>
```

infotexture (changes requested):

infotexture (approved):

src/test/resources/messages_en_US.properties
```diff
@@ -183,3 +183,4 @@ DOTJ044W=There is a redundant conref action "pushbefore". Please make sure that
 DOTA014W=Attribute @{0} is deprecated. Use attribute @{1} instead.
 DOTJ080W=Integrator configuration ''{0}'' has been deprecated. Use plugin configuration ''{0}'' instead.
 DOTJ082E=Resource {0} is referenced with incorrect path capitalization
+DOTJ084E=Included file {0} contained invalid byte sequence for charset {1}
```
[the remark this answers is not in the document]
jelovirt: It doens't matter, the test message copy is never used anywhere.

src/main/config/messages_template.xml
```diff
@@ -423,6 +423,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ084E" type="ERROR">
+    <reason>Couldn’t read '%1' with the %2 character set.</reason>
```
robander, line 427: Curious, are there any illegal characters that will throw this message, which then break our log presentation because the logging can't properly handle it?

robander: ah nevermind for a moment I thought `%2` was the character, obviously it's the character set name

robander (approved):

PR #3895, opened 2022-03-03
Add profiles to publication project file

### LINKED ISSUE #3690: In DITA-OT project files, apply both <context> and <publication> DITAVAL filtering
## Expected Behavior

(This is a more educated follow-up to #3474.)

For DITA-OT project files, if a `<context>` applies a DITAVAL and a `<publication>` applies a DITAVAL, then a `<deliverable>` using both of them should apply both DITAVAL files:

## Actual Behavior
The `<context>` DITAVAL is applied but the `<deliverable>` DITAVAL is discarded.

If the `<context>` DITAVAL is removed, then the `<deliverable>` DITAVAL is applied as expected.

## Possible Solution
Both `<context>` and `<publication>` DITAVAL files should be applied. The need for this is demonstrated by replacing the `<context>` and `<publication>` definitions above with their respective DITAVAL files:

In this example, the `<context>` DITAVAL is filtering on `@product`, while the `<publication>` DITAVAL is filtering on `@audience` (and applying `@rev` highlighting). These filtering aspects are orthogonal.

## Steps to Reproduce
In the attached testcase:

[dita_project_file.zip](https://github.com/dita-ot/dita-ot/files/5920276/dita_project_file.zip)

run the following command:

`dita --project=books.xml --deliverable="del-pdf-review-prodA_ug" --verbose`

then note in the output log that the publication `flag_review.ditaval` DITAVAL is not applied.

## Environment

* DITA-OT version: 3.5.4
* Operating system and version: Ubuntu Linux 18.04 LTS
* Ran via `dita` command
* Transformation type: PDF

jelovirt:
## Description
Add profiles to publication in project files. When both context and publication contain profiles, they are applied in order: publication profiles first, context profiles second.

## Motivation and Context
Fixes #3690

## How Has This Been Tested?
New tests for project to properties conversion.

## Type of Changes

https://github.com/dita-ot/docs/blob/develop/topics/using-project-files.dita#L69 should say `publication` can link to publication related DITAVAL files. Adding profile to publication in examples would help in explaining where they should be used.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -36,7 +36,6 @@ import org.apache.tools.ant.property.ResolvePropertyMap;
 import org.apache.tools.ant.util.ClasspathUtils;
 import org.apache.tools.ant.util.FileUtils;
 import org.apache.tools.ant.util.ProxySetup;
-import org.dita.dost.log.MessageUtils;
 import org.dita.dost.platform.Plugins;
 import org.dita.dost.project.Project.Context;
 import org.dita.dost.project.Project.Publication;
@@ -357,7 +356,7 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
                     throw new BuildException("");
                 }
             } else {
-                projectProps = handleProject(conversionArgs.projectFile, definedProps);
+                projectProps = collectProperties(conversionArgs.projectFile, definedProps);
             }
             repeat = conversionArgs.repeat;
             // default values
@@ -400,9 +399,17 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
         readyToRun = true;
     }
 
-    private List<Map<String, Object>> handleProject(final File projectFile, final Map<String, Object> definedProps) {
+    private List<Map<String, Object>> collectProperties(final File projectFile, final Map<String, Object> definedProps) {
         final URI base = projectFile.toURI();
         final org.dita.dost.project.Project project = readProjectFile(projectFile);
+
+        return collectProperties(project, base, definedProps);
+    }
+
+    @VisibleForTesting
+    List<Map<String, Object>> collectProperties(final org.dita.dost.project.Project project,
+                                                final URI base,
+                                                final Map<String, Object> definedProps) {
         final String runDeliverable = (String) definedProps.get("project.deliverable");
 
         final List<Map<String, Object>> projectProps = project.deliverables.stream()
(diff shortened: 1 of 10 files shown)


chrispy-snps:
@jelovirt - this is great, thank you!!

jelovirt:
The code in `Main` should be refactored so that we could easily test `Project` to Ant property map.

jelovirt:
@raducoravu what's your take, is profiles in publication taking priority correct?

chrispy-snps:
@jelovirt - going by the guideline of more specific settings taking precedence over more general settings, then a `<context>` profile would take precedence over a `<publication>` profile. This would allow a publication to declare some default filtering, then allow map-specific overrides on a map-by-map basis.

And although you could define multiple `<publication>` definitions instead, maybe the per-map settings could be numerous or unique enough to make that impractical?

In our case, our filter/flagging settings between contexts/publications are completely orthogonal and I can't think of any specific use case to support a strong opinion. So I am just thinking out loud.

kirkilj:
We came up with a similar arrangement, where a project
wide.properties file could be defined and then if there was a properties file with the same basename as a DITA map, it was used to override the corresponding entries of the project.

Given the discussion in Thursday's Monthly call, I'll invite @thendarion to comment. We have other configurations to define deliverables, but it be preferable if the new OT project schema would enable us to simplify or do away with these home-grown mechanisms.

raducoravu:
@jelovirt I would also maybe see the context filters applied first with the publication filters applied last. Also the way in which in a project XML file a deliverable contains (context, publication) in this order would mean this is also the order in which the filters are applied.

But how does the order of the filters matter? 
If for example I declare a flagging color in the context filter, if the publication filter comes last it could overwrite that flagging to another color, right?
If I declare an exclusion for a certain profiling attribute value in the context filter, can the publication filter provide an explicit inclusion of that same attribute value to override the default exclusion and keep the element in the document?

jelovirt:
> But how does the order of the filters matter?

Multiple DITAVAL files are just like a single DITAVAL where they are merged into one, and order matters there. The key point here is _what_ you stick into various profiles. You should e.g.

* filter by audience in `context`
* flag to highlight or only show print content in `publication`.

If you start mixing those, then this feature is called a "foot gun".

The difficulty here is that both `publication` and `context` are reusable. You can use _site_ `publication` to publish product A and B `context`; you can use _novice_ `context` with PDF and HTML5 `publications`. So IMO there is no obvious correct priority order for these.

raducoravu:
@jelovirt right, in the end we need to settle for a certain order and document it as such. But as I said I would slightly incline towards having the context filters added before the publication filter.

robander:
I probably would have gone with “publication” first, but agree there is no obvious answer and do not have strong feelings.

My reasoning is that you just shouldn’t define things tightly with the publication if they might not apply. Just like with DITA topics — you should not hard-code metadata in the topic if that topic might be used where the metadata is not appropriate, and you should not hard-code related-links in a topic if that topic might be used where the links are not appropriate. In this case, you should not define DITAVAL rules together with the publication if they do not always apply to the publication.

All that said, I don't think it matters much - we should document that this is a "foot gun feature" as @jelovirt described it, and that you state 1) which order we evaluate and 2) that it is probably a Bad Idea to try and have overlapping / override rules.

PR #3898, opened 2022-03-05
Use separate temporary directory for each project deliverable

### LINKED ISSUE #3757: Separate temporary files folders for each deliverable
I can set the "clean.temp=no" parameter to specify that the transformation temporary files folder should not be deleted, but right now all deliverables are published using the same temporary files folder, making it hard to debug for each deliverable and see what temporary files have been created for them. 
So I think that for each deliverable we should have a separate subfolder in the transformation temporary files folder, to keep the temporary files separately. This would help also if at some point we want to make the processing parallel for all deliverables.

jelovirt:
## Description
Use separate temporary directory for each project deliverable.

Temporary directories with different argument combinations

| arguments | temporary directory |
| -----------| ---------------------|
| `dita -i root.ditamap` | `${java.io.tmpdir}/temp${timestamp}` |
| `dita -p project.xml` | `${java.io.tmpdir}/temp${timestamp}/${deliverable.id}` |
| `dita -i root.ditamap -t /tmp` | `/tmp` |
| `dita -p project.xml -t /tmp` | `/tmp/${deliverable.id}` |

Temporary directory for deliverables uses deliverable `@id` or a random UUID if not `@id` is set.

## Motivation and Context
Fixes #3757

## How Has This Been Tested?
Existing test pass.

## Type of Changes

This change is pretty deep in processing details and visible users only if they have strong expectations for temporary directory name. So in that sense this probably doesn't require long description in release notes. Docs currently don't have anything about the old temporary directory generation with project files.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Arguments.java
+++ b/src/main/java/org/dita/dost/invoker/Arguments.java
@@ -75,6 +75,7 @@ abstract class Arguments {
     boolean justPrintVersion;
     boolean justPrintDiagnostics;
     final Map<String, Object> definedProps = new HashMap<>();
+    int repeat = 1;
 
     Arguments() {
         useColor = getUseColor();
--- a/src/main/java/org/dita/dost/invoker/ConversionArguments.java
+++ b/src/main/java/org/dita/dost/invoker/ConversionArguments.java
@@ -79,7 +79,6 @@ public class ConversionArguments extends Arguments {
      * Project file
      */
     File projectFile;
-    int repeat = 1;
 
     public final List<String> inputs = new ArrayList<>();
     private final List<String> resources = new ArrayList<>();
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -51,6 +51,7 @@ import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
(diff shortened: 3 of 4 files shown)


jelovirt:
@raducoravu @robander do you think the temporary directory names are ok like this? Should we generate a subdirectory for each deliverable instead of parallel temporary directories with deliverable ID in the directory name? So instead of

`${java.io.tmpdir}/temp${deliverable.id}${timestamp}`

use subdirectory for deliverables

`${java.io.tmpdir}/temp${timestamp}/${deliverable.id}`

chrispy-snps:
@jelovirt - if an entire project file is published, would the timestamp string be identical for every deliverable in the set? I'm trying to understand the scenario in your nested-directory suggestion in which a timestamp directory could have multiple deliverable directories.)

raducoravu:
@jelovirt I think both approaches are fine by me, probably the "subdirectory for deliverables" would be a little bit more structured and it would be easier to delete everything manually by deleting the entire "${java.io.tmpdir}/temp${timestamp}" folder.

raducoravu (commented):
Code looks good to me, I tested running the publishing with the base temp dir specified or omitted, also works.
One thing, if the deliverable does not have an ID, that fallback UUID looks kind of long and scary, maybe we could have used instead something like "temp/deliverable-1" with some kind of counter. But if you do not want to do this I have no objections.

raducoravu:
@jelovirt one thing I forgot, let's say I'm publishing with clean.temp=yes, and a folder like this gets created "temp/deliverable-1", at the end of the publishing the "deliverable-1" subfolder is deleted by the DITA OT but the parent "temp" folder is probably not?

Issue #3889, opened 2022-02-21
Use official safety levels from ANSI Z535.4


xephon2:
I suggest to use the official safety levels as defined in the ANSI Z535.4, see [Neue ANSI Z535 - Vom vernünftigen Umgang mit Sicherheitshinweisen](https://docplayer.org/24219221-Neue-ansi-z535-vom-vernuenftigen-umgang-mit-sicherheitshinweisen.html), slide 21. If you agree, I'd send a PR.

Issue #3882, opened 2022-02-18
New translations for german and french


Mircea-Andrei16:
## Description
User came to Oxygen support asking for "Notice" translation in <hazardstatement> and proposes "Hinweis".

## Possible Solution
We have changed certain vars for german and french (variable id = value):
- Notices title = "Hinweises"
- Notice = "Hinweis"
- Caution = "VORSICHT"
- Remember = "Erinnern"
- Notices title = "Notices"
- Notice = "Notice"

## Additional Context
![hazard_notice](https://user-images.githubusercontent.com/72509527/154667095-04b481e7-1ff2-4f6f-8988-4f456d6e8393.png)

raducoravu:
@xephon2 any comments about making these changes?

infotexture:
@raducoravu Looks like there are a few duplicate entries in the list above.

The correct German plural for _Notices_ is _Hinweise_ — no trailing “s”.

jlacour31:
@infotexture the duplicates are the french versions, I already reviewed them.

xephon2:
@raducoravu, @infotexture 

Signal words are defined in the technical standards ANSI Z535.6 and IEC 82079. They are also named on the German Wikipedia site [Sicherheitshinweis](https://de.wikipedia.org/wiki/Sicherheitshinweis). The correct values are:

| **English** | **German** | Meaning                                                                                            |
| ----------- | ---------- | -------------------------------------------------------------------------------------------------- |
| Danger      | Gefahr     | Indicates a hazardous situation that, if not avoided, **WILL** result in death or serious injury.  |
| Warning     | Warnung    | Indicates a hazardous situation that, if not avoided, **COULD** result in death or serious injury. |
| Caution     | Vorsicht   | Indicates a hazardous situation that, if not avoided, could result in minor or moderate injury.    |
| Notice      | Hinweis    | Indicates a situation that, if not avoided, can result in property damage.                         |



> **Note**: The correct German ANSI term for `Notice` is `Hinweis`, but this is no good translation. A safety note of level `Notice` is used to refer to a material damage. `Hinweis` is a term which is usually used for useful information. But if want to explain to your user how he can avoid damage of his 1,000,000 USD machine, this is more than just a useful information. Therefore it is common practice to use the term `Achtung` instead of `Hinweis` for `Notice`. But the correct term is `Hinweis`. The DITA-OT should stick to the standard (even if the standard is not good).

[...]

PR #3899, opened 2022-03-07
Move Notice string from PDF2 to base plug-in

### LINKED ISSUE #3888: Moving the "Notice" string from PDF2 to base
As suggested in https://github.com/dita-ot/dita-ot/issues/3882, I suggest to move the string `Notice` from PDF2 to the base plugin in all languages. The reason is, that all safety levels should be kept in the base plugin. Please let me know if you'd accept this change, then I'd send a PR.

jelovirt:
## Description
Move `Notice` string from PDF2 to base strings file.

## Motivation and Context
Fixes #3888

## How Has This Been Tested?
Existing tests.

## Type of Changes

None needed


src/main/plugins/org.dita.base/xsl/common/strings-sl-si.xml
```diff
@@ -33,6 +33,7 @@ See the accompanying LICENSE file for applicable license.
    <variable id="Value">Vrednost</variable>
    <variable id="Note">Opomba</variable>
    <variable id="Notes">Opombe</variable>
+   <variable id="Notice">Notice</variable><!--TODO:Notice-->
```
robander, line 36: In the original `sl.xml` this one was
`<variable id="Notice">Objava</variable>`

infotexture (commented):
Shouldn't we target the `hotfix/3.7.1` branch for this and pull in @xephon2's latest **Notice** translations from #3882/#3883, which were already merged for 3.7.1?

That would allow these changes to merge cleanly, rather than modifying some of the same things on both `develop` and release branches.

If we're touching all of the **Notice** strings here, we might want to grab those ANSI Z535.4 localizations from #3889.

jelovirt:
Will rebase on `hotfix/3.7.1`

jelovirt:
The ANSI Z535.4 localizations from #3889 should be a separate feature PR. This refactor PR just moves variables to the correct files.

infotexture (approved):
👍 LGTM

PR #3907, opened 2022-03-26
Add param to publication ref element

### LINKED ISSUE #3682: Allow a PDF deliverable to specify its output file name
## Description
When a shared map is published to PDF in multiple profiling conditions, then each output PDF file must have a different name according to its condition.

`<deliverable>` elements allow the output _directory_ to be specified (great for HTML5!), but not the output _basename_ (needed for PDF).

As a result, the output PDF files derive their basename from the input map, resulting in duplicate PDF file names:

## Possible Solution
One approach would be to support `<param>` directly inside `<deliverable>`:

If this is too philosophically different from what `<deliverable>` was intended to contain, another approach is to allow `<param>` inside an `idref`'ed `<publication>` inside the `<deliverable>`

This second approach is very RelaxNG-like in its method of applying overrides to a refererence, which is quite nice!

Some workarounds (and their deficiencies) are:

* Uniquify the PDF `<publication>` definition.
  * In our case, we would need to uniquify our `<publication>` definitions by the hundreds, which is not ideal because then we must update _all_ of them whenever our PDF transformation changes.
* Rename the PDF files using a post-processing script.
  * Isn't the goal of DITA-OT project files to be able to replicate -- in a structured way -- what is normally performed on the `dita` command line (where the basename is normally specified), without the need for post-processing scripts?
  * If all PDFs go in the same directory, then they collide with themselves during full publication before such a script could run.
* Place the PDF files in different output directories.
  * This avoids file collisions, but it does not solve the duplicate file name issue (which would confuse our customers).

## Potential Alternatives
A new basename-specific element (`basename`?) could be added to `<deliverable>`, but a `<param>`-based approach would be more generally useful for deliverable-specific tweaks to publishing.

## Additional Context
To reproduce, you can try using the following project file (copied to `<DITA-OT>/docsrc/samples/project-files/pdf-test.xml`, perhaps):

```xml
<?xml-model href="https://www.dita-ot.org/rng/project.rnc" type="application/relax-ng-compact-syntax"?>
<project xmlns="https://www.dita-ot.org/project">

  <!-- single map to be published in two flavors -->
  <context id="context-ug1" name="User Guide 1">
    <input href="../../userguide-book.ditamap"/>
    <!-- pretend we have DITAVAL 1 applied here -->
  </context>

  <context id="context-ug2" name="User Guide 2">
    <input href="../../userguide-book.ditamap"/>
    <!-- pretend we have DITAVAL 2 applied here -->
  </context>


  <!-- single shared publication definition -->
  <publication id="pub-pdf2" transtype="pdf2">
    <param name="args.chapter.layout" value="BASIC"/>
    <param name="args.gen.task.lbl" value="YES"/>
    <param name="include.rellinks" value="friend"/>
    <param name="processing-mode" value="strict"/>
  </publication>


  <!-- publish book flavors with two different output base names -->
  <deliverable id="del-pdf-ug1">
    <context idref="context-ug1"/>
    <output href="."/>
    <publication idref="pub-pdf2">
      <param name="outputFile.base" value="userguide1"/>
    </publication>
  </deliverable>

  <deliverable id="del-pdf-ug2">
    <context idref="context-ug2"/>
    <output href="."/>
    <publication idref="pub-pdf2">
      <param name="outputFile.base" value="userguide2"/>
    </publication>
  </deliverable>

</project>
```

jelovirt:
## Description
Add support for overriding `param` in `publication` reference.

```xml
<project xmlns="https://www.dita-ot.org/project">
  <publication transtype="html5" id="common-html5">
    <param name="nav-toc" value="partial"/>
  </publication>
  <deliverable>
    <context>
      <input href="root.ditamap"/>
    </context>
    <output href="./out"/>
    <publication idref="common-html5">
      <param name="nav-toc" value="full"/> <!-- override common HTML publication -->
    </publication>
  </deliverable>
</project>
```

## Motivation and Context
Enables referencing a reusable publication and override parameters for it.

Fixes #3682

## How Has This Been Tested?
New tests.

## Type of Changes

The docs currently don't have a reference for project syntax, so this param override should go into the examples.

### DIFF
--- a/src/main/java/org/dita/dost/project/ProjectBuilder.java
+++ b/src/main/java/org/dita/dost/project/ProjectBuilder.java
@@ -14,6 +14,9 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import java.net.URI;
 import java.util.List;
 
+/**
+ * Project file AST builder.
+ */
 public class ProjectBuilder {
 
     @JsonProperty("deliverables")
--- a/src/main/java/org/dita/dost/project/ProjectFactory.java
+++ b/src/main/java/org/dita/dost/project/ProjectFactory.java
@@ -63,35 +63,61 @@ public class ProjectFactory {
         return new Project(
                 src.deliverables == null ? Collections.emptyList() :
                         src.deliverables.stream()
-                                .map(deliverable -> new Deliverable(
-                                        deliverable.name,
-                                        deliverable.id,
-                                        Optional.ofNullable(deliverable.context)
-                                                .flatMap(context -> Optional.ofNullable(context.idref))
-                                                .map(idref -> {
-                                                    final Context pub = src.contexts.stream()
-                                                            .filter(context -> Objects.equals(context.id, deliverable.context.idref))
-                                                            .findAny()
-                                                            .orElseThrow(() -> new RuntimeException(String.format("Context not found: %s", deliverable.context.idref)));
-                                                    return pub;
-                                                })
-                                                .orElse(deliverable.context),
-                                        deliverable.output,
-                                        Optional.ofNullable(deliverable.publication)
-                                                .flatMap(publication -> Optional.ofNullable(publication.idref))
-                                                .map(idref -> {
-                                                    final Publication pub = src.publications.stream()
-                                                            .filter(publication -> Objects.equals(publication.id, deliverable.publication.idref))
-                                                            .findAny()
-                                                            .orElseThrow(() -> new RuntimeException(String.format("Publication not found: %s", deliverable.publication.idref)));
-                                                    return pub;
-                                                })
-                                                .orElse(deliverable.publication)
-                                ))
+                                .map(deliverable -> {
+                                    final Context context = Optional.ofNullable(deliverable.context)
+                                            .flatMap(cntx -> Optional.ofNullable(cntx.idref))
+                                            .map(idref -> {
+                                                final Context pub = src.contexts.stream()
+                                                        .filter(cntx -> Objects.equals(cntx.id, deliverable.context.idref))
+                                                        .findAny()
+                                                        .orElseThrow(() -> new RuntimeException(String.format("Context not found: %s", deliverable.context.idref)));
+                                                return pub;
+                                            })
+                                            .orElse(deliverable.context);
+                                    final Publication publication = Optional.ofNullable(deliverable.publication)
+                                            .flatMap(publ -> Optional.ofNullable(publ.idref))
+                                            .map(idref -> {
+                                                final Publication pub = src.publications.stream()
+                                                        .filter(publ -> Objects.equals(publ.id, deliverable.publication.idref))
+                                                        .findAny()
+                                                        .orElseThrow(() -> new RuntimeException(String.format("Publication not found: %s", deliverable.publication.idref)));
+                                                return merge(pub, deliverable.publication);
+                                            })
+                                            .orElse(deliverable.publication);
+                                    return new Deliverable(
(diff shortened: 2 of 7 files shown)


chrispy-snps:
This is great, thank you @jelovirt!

infotexture (commented):
@jelovirt The description here shows only the XML example, but I see that the JSON and YAML test files are also modified, so I assume the same approach will work in all project file formats, right?

> The docs currently don't have a reference for project syntax

Should we change that? 🤔 

Not sure how hard it would be to generate or maintain that, or whether the existing examples in the [XML project files](https://www.dita-ot.org/dev/topics/project-files-xml.html) topic are enough for most people to work with.

@chrispy-snps Have you missed more detailed documentation here, or have the examples been enough to go on so far?

chrispy-snps:
@infotexture - Thank you for asking! I think more details would be useful. Some suggestions:

- On [this page](https://www.dita-ot.org/3.7/topics/using-project-files.html), perhaps start a new syntax section at "Though the exact syntax differs slightly".
- Mention that `<publication>` can also contain `<profile>` with a series of DITAVALs (#3895). (I'm sure this is already on your list.)
- For `<param>` definitions  in a `<publication>`, document the difference between `@href`, `@path`, and `@value`. (I think they are for file paths to be expanded, directory paths to be expanded, and simple string values to be passed as-is, but this is a guess.)
- Explicitly mention that inside a `<deliverable>`, a `<context>` or `<publication>` can either be entirely self-contained, or it can be referenced from elsewhere via an `@idref` reference. (It's not spelled out quite so directly now.)
  - Mention that inside a `<publication @idref>`, `<param> definitions can be added or overridden.

jelovirt:
@raducoravu what do you think about the backwards compatibility aspect. This new syntax will not work with DITA-OT 3.x, only 4.x. Will that be an issue for Oxygen because you support multiple DITA-OT versions at the same time.

raducoravu:
@jelovirt I do not see a problem with this. Does the current DITA OT 3.7.1 for example issue some kind of validation error when processing such a DITA OT project file?

jelovirt:
@raducoravu if you try to use 4.x project with `param` override with 3.x, you'll get a validation error from RNG.

raducoravu:
@jelovirt great, you get a "Failed to read project file" fatal error and the publishing breaks, works for me.

chrispy-snps:
@jelovirt - in addition, I have an XSLT stylesheet that can adapt the recent 4.0+ project file enhancements to run in 3.7:

[Oxygen XML Blog: Preprocessing DITA-OT Project Files](https://blog.oxygenxml.com/topics/preprocessing_ditaot_project_files.html)

infotexture:
@chrispy-snps Thanks for the docs suggestions. 🙏 

We'll consider those as we revise the documentation to reflect the changes in this PR.

> document the difference between `@href`, `@path`, and `@value`.

I'll ask @jelovirt to confirm your [guesses](https://github.com/dita-ot/dita-ot/pull/3907#issuecomment-1079917328) above, but agree it makes sense to clarify when each of these are appropriate and which types of values they expect.

jelovirt:
@infotexture those guesses about the attributes were not 100% correct. I'll need to remind myself by reading the code. Ping me in a docs PR to fill in the details.

robander (approved):

chrispy-snps:
We discovered another use for this enhancement. Some writers want indexes generated for their Oxygen WebHelp maps, but others do not. This allows each writer to enable/disable the index for their specific deliverable, without having to multiply the number of `<publication>` definitions we have. Thank you!

PR #3915, opened 2022-04-14
Feature/project reserved params 3913

### LINKED ISSUE #3913: In DITA-OT project files, do not quietly ignore an output.dir <param> definition
## Description
When the `output.dir` parameter is specified on a `dita` command line:

```
dita -i map.ditamap -f html5 -Doutput.dir=foo
Error: Property output.dir cannot be set with -D, use -o instead
```

then the DITA-OT warns that the parameter cannot be set directly and the `--output` option should be used instead:

But when a DITA-OT project file specifies the `output.dir` parameter:

```
  <param name="output.dir" path="foo"/>
</publication>
```

then the parameter is quietly ignored, potentially resulting in user confusion.

## Possible Solution

Issue a similar warning/error message that the `output.dir` parameter cannot be defined in that context.

## Potential Alternatives
Consider the `output.dir` parameter somehow, although precedence with (or concatenation with??) the `--output` command-line option must be considered.

## Testcase
[ditaot_project_output_dir_ignored.zip](https://github.com/dita-ot/dita-ot/files/8472392/ditaot_project_output_dir_ignored.zip)

jelovirt:
## Description
Log error message when `param` in project file defines a parameter that should be set with a dedicated key instead. 

## Motivation and Context
Fixes #3913.

## How Has This Been Tested?
Manual testing.

## Type of Changes

Mention in release notes. More detailed description of error `DOTJ085E`.

### DIFF
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -423,6 +423,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ085E" type="ERROR">
+    <reason>Parameter '%1' cannot be set with param in project file. The param '%1' will be ignored.</reason>
+    <response>Use '%2' key instead.</response>
+  </message>
+
   <!-- End of Java Messages -->
     
   <!-- Start of XSL Messages -->  
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -28,6 +28,7 @@ package org.dita.dost.invoker;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableMap;
 import org.apache.tools.ant.*;
 import org.apache.tools.ant.input.DefaultInputHandler;
 import org.apache.tools.ant.input.InputHandler;
@@ -36,7 +37,6 @@ import org.apache.tools.ant.property.ResolvePropertyMap;
 import org.apache.tools.ant.util.ClasspathUtils;
 import org.apache.tools.ant.util.FileUtils;
 import org.apache.tools.ant.util.ProxySetup;
-import org.dita.dost.log.MessageUtils;
 import org.dita.dost.platform.Plugins;
 import org.dita.dost.project.Project.Context;
 import org.dita.dost.project.Project.Publication;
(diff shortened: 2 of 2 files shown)


src/main/config/messages_template.xml
```diff
@@ -423,6 +423,11 @@ See the accompanying LICENSE file for applicable license.
     <response></response>
   </message>
 
+  <message id="DOTJ085E" type="ERROR">
+    <reason>Parameter '%1' cannot be set with param in project file. The param '%1' will be ignored.</reason>
+    <response>Use '%2' key instead.</response>
```
jelovirt, line 428: @infotexture Could you check the copy on this error message. Because we have multiple project formats and some parameters in XML are defined with elements and some are defined with attributes, the response can't be e.g. "Use <%2> element instead.". But _key_ is not a particularly good term for this either.

jelovirt: The error messages will not be

```
Error: [DOTJ085E][ERROR] Parameter 'output.dir' cannot be set with param in project file. The param 'output.dir' will be ignored. Use 'output' key instead.
Error: [DOTJ085E][ERROR] Parameter 'args.filter' cannot be set with param in project file. The param 'args.filter' will be ignored. Use 'profiles' key instead.
Error: [DOTJ085E][ERROR] Parameter 'args.input' cannot be set with param in project file. The param 'args.input' will be ignored. Use 'input' key instead.
Error: [DOTJ085E][ERROR] Parameter 'transtype' cannot be set with param in project file. The param 'transtype' will be ignored. Use 'transtype' key instead.
```

chrispy-snps: Could this be parameterized as

```
Use the %s instead.
```

where `%s` could be `the <output> element` or `the @transtype attribute`?

infotexture, line 428: I think we can simplify this to avoid repeating the parameter name, and the second sentence of the reason is sort of implied by the first. 

If we can't be more specific whether the suggested replacement is an attribute or element, I suggest we drop the noun entirely. How about this:

```suggestion
    <reason>The '%1' parameter can’t be set with &lt;param&gt; elements in project files.</reason>
    <response>Use '%2' instead.</response>
```

jelovirt: @infotexture If the project file is JSON or YAML format, then `set with &lt;param&gt; elements` doesn't apply because the project file doesn't have an XML element. If we need to know what format the source file is, we need to move the validation to different class and we'd also need a different error codes for JSON, YAML, and XML. 😬

jelovirt: @infotexture And because you can have a YAML project file that includes an XML project file that includes a JSON project file, we don't know what the format of the actual file that contains the param.

jelovirt: Because this fix is going into a patch release, I would prefer to not refactor the project parser code to make the error messages better. That should be done in the next major/minor version. So could the copy be changed to not assume XML format?

jelovirt: Currently we don't have support for something like that because the whole idea of messages is that the code doesn't contain any of the copy.

infotexture: Oh, right, forgot we don't know which of the 3 formats we're dealing with here. 
Then I guess this is the best we can do:

```suggestion
    <reason>The '%1' parameter can’t be set with 'param' in project files.</reason>
    <response>Use '%2' instead.</response>
```

chrispy-snps: @infotexture - I like your suggestion.

PR #3920, opened 2022-04-28
Remove Customization and custom.xsl overrides in PDF2

### LINKED ISSUE #3826: Remove support for Customization directory in PDF2
## Description
Remove support for [`Customization` directory and `cfg` URI scheme](https://www.dita-ot.org/dev/topics/pdf-plugin-structure.html) in PDF2.

## Possible Solution
The legacy extensibility API to PDF2 adds overhead and makes PDF2 more complex with limited benefits. By removing the legacy extensibility we can simplify PDF2 configuration code and offer users a one way to customization.

## Potential Alternatives
Retaining legacy customization and accepting complexity and limitations that come with it.

## Documentation
With legacy extensibility removed, we should offer a migration guide from `Customization` directory to plug-in based extensibility.

jelovirt:
## Description
Remove support for `Customization` directory and `custom.xsl` files referenced with `cfg` URI scheme in PDF2.

## Motivation and Context
Fixes #3826

## How Has This Been Tested?
Existing tests pass.

## Type of Changes

Release notes should describe this functionality has been removed.

Migration guide from previous version should describe a path to plug-in based override.


robander (approved):

raducoravu:
So a procedure like this will no longer work, right?
https://www.oxygenxml.com/doc/versions/24.1/ug-editor/topics/dita_pdf_output_customization.html
How about a PDF plugin which provides a customization dir, that will not work either, right?
How about this "dita.xsl.xslfo" plugin extension point? Will it still work?

jelovirt:
@raducoravu `dita.xsl.xslfo` extension point will still work. Instead of using the customization directory, you need to override the shell template and use it with `args.xsl.pdf` parameter.

raducoravu:
@jelovirt we might end up making some people upset about this (meaning they will no longer be able to use DITA OT 4.0), most of the PDF customizations I've seen use the customization folder and frequently the person who created them exited the company and the XSLTs are just slightly maintained.

jelovirt:
@raducoravu I still haven't merged this PR because I want to be sure we have a clear migration path. What I'm testing is to have a migration path plug-in `org.dita.pdf2.customization` that could be used to retain the old customization behaviour. It would make PDF2 code base cleaner by moving the customization code to plug-in that we could ship by default.

raducoravu:
@jelovirt yes it would help me to see a clear documentation about the migration path as well. Or maybe also a small tool/converter which takes a plugin in the old form and converts it to the new form.

Issue #3917, opened 2022-04-25
if parallel=true, some files are not rendered


thendarion:
## Expected Behavior
all files are rendered

## Actual Behavior
some files are skipped with error "folder could not be created". seemingly randomly in documents, but consistently between builds

## Steps to Reproduce
render a document with a lot of topics in the same subfolder (?)

## Copy of the error message, log file or stack trace
[xslt] Failed to transform document: Failed to transform document: Failed to create directory 

## Environment

* DITA-OT version: 3.7.1, 3.6.1
* Operating system and version:
  Linux, Windows
* How did you run DITA-OT?
  `dita` command
* Transformation type:
  HTML5

jelovirt:
@thendarion Could you share at which point in the log are `folder could not be created` output?

thendarion:
it happens in "html5.topics.common.inner"

thendarion:
I looked at the log more closely and I have a custom plugin that also suffers from  `folder could not be created` when `parallel="true"` so it isn't a HTML5-specific issue but more like a race condition between the XSLT threads it feels like

chrispy-snps:
@thendarion - I tried making a synthetic testcase with N topic files (I used N=1000), published with `--parallel=yes` repeatedly in a forever loop, then printing the number of HTML topic files in the output directory after each iteration. For me, I get no loss of topics:

```
$ ./do.sh
1000
1000
1000
1000
1000
...continues many times...
```

I tried two different machines - one 12-core, one 24-core.

Maybe you could try this testcase on your side, with your plugin installed, and see what happens?

[ditaot_parallel.zip](https://github.com/dita-ot/dita-ot/files/8564735/ditaot_parallel.zip)

thendarion:
@chrispy-snps thanks for the test case

there's a misunderstanding here though

you have a structure like
```ASCII
-- dita
  |-- map.ditamap
  |-- topic1.dita
  |-- topic2.dita
  ...
```
while I've been encountering the bug when the structure is like
```ASCII
-- dita
  |-- map.ditamap
  |-- subfolder1
  |  |-- topic1.dita
  |  `-- topic2.dita
  |-- subfolder2
  |  |-- topic1.dita
  |  `-- topic2.dita
  ...
```
because it's about XSLT trying to create the same folder twice in parallel and colliding with itself

chrispy-snps:
@thendarion, good news - adding directory structure reproduces the issue!

[...]

PR #3924, opened 2022-05-11
Fix mkdir issue when parallel=true

chrispy-snps:

## Description
This fixes the following error that could occur when `--parallel=yes` was used:

```
     [xslt] Failed to transform document: Failed to transform document: Failed to
            create directory /chrispy/dita/out/826_to_858
```

@raducoravu describes the root cause [here](https://github.com/dita-ot/dita-ot/issues/3917#issuecomment-1118201374).

## Motivation and Context
This is a fix for #3917 (if parallel=true, some files are not rendered).

## How Has This Been Tested?
I created a testcase to specifically reproduce the issue, then tested on two machines (12-core and 24-core) to reproduce the original problem and validate the fix.

## Type of Changes

Only release notes needed.

## Checklist

I did not update any unit tests, but I don't think we have unit tests for parallel-processing transformations.

### DIFF
--- a/src/main/java/org/dita/dost/store/StreamStore.java
+++ b/src/main/java/org/dita/dost/store/StreamStore.java
@@ -235,7 +235,7 @@ public class StreamStore extends AbstractStore implements Store {
     }
 
     @VisibleForTesting
-    Serializer getSerializer(final URI dst) throws IOException {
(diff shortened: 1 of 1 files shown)


jelovirt:
I prefer a different approach to handle this issue. Opened #3925

PR #3925, opened 2022-05-11
Fix concurrent directory creation

### LINKED ISSUE #3917: if parallel=true, some files are not rendered

jelovirt:
## Description
Concurrent directory creation fails because we treat all failure cases as an exception. If the directory already exists, we can ignore that exception.

## Motivation and Context
Fixes #3917

Replaces #3924

## How Has This Been Tested?
Existing tests pass.
## Type of Changes


robander (approved):

chrispy-snps:
@jelovirt - I confirmed your revised fix on both my 12-core and 24-core machines.

PR #3935, opened 2022-06-07
Add dir=auto on Bidi code elements

jason-fox:
## Description

For *RTL renderings only*, add `dir="auto"` to monospace elements, effectively encoding a `<bdi>` within the `<span>` to ensure that the browser will look at the first strongly typed character in the element and work out from that what the [base direction](https://www.w3.org/International/questions/qa-html-dir#dirauto)  of the element should be.

## Motivation and Context

When reading an HTML document in an RTL language such as Arabic or Hebrew, code blocks (typically written in Latin script) are read LTR. This makes the rendered HTML document inherently bidirectional. HTML browsers usually make a good attempt at rendering bidirectional text correctly by splitting characters into three types, LTR, RTL and neutral. The problem is that when a monospace code block begins with a neutral character (like punctuation or a number) the directionality of the character is inherited from the parent element.

Take the following `<filepath>` for example:

```xml
<section>
      <title>Styling DITA elements</title>
      <p xml:lang="he" dir="rtl">
           החל מחלקות שירות עקביות של Bootstrap על פני DITA על ידי תיקון 
          <filepath>/Customization/xsl/utility-classes.xsl</filepath>
      </p>
</section>
```

It currently renders in HTML (without a `dir` attribute) as:

```xml
<p class="p" lang="he" dir="rtl">
  החל מחלקות שירות עקביות של Bootstrap על פני DITA על ידי תיקון 
  <span class="ph filepath">/Customization/xsl/utility-classes.xsl</span>
</p>
```

Which causes the first neutral character  in the `<filepath>` -  the leading `/` to be **incorrectly** rendered RTL at the end of the line:

Screenshot 2022-06-07 at 09 33 12

With ` dir="auto"`, it should read the code block as all LTR latin text and render as

Screenshot 2022-06-07 at 09 32 47

```xml
<p class="p" lang="he" dir="rtl">
  החל מחלקות שירות עקביות של Bootstrap על פני DITA על ידי תיקון 
  <span class="ph filepath" dir="auto">/Customization/xsl/utility-classes.xsl</span>
</p>
```

Which causes the browser to consider the first strong character it finds and then render the whole `<span>` correctly as left-to- right with the leading slash at the beginning. Note that had the `<span>` started with RTL text it would have rendered correctly as well, since ` dir="auto"`  is only a processing instruction not a forced direction like  `dir="ltr"`

## How Has This Been Tested?

Local testing, creating documents with a `default.language=ar`

## Type of Changes

None


jason-fox:
Of course, this is only really an issue when the `default.language` is set to an RTL language like `he`, `ar` etc. If only a small snippet is effected, the author can already add `dir="rtl"` and `dir="auto"`  on each individual element throughout:

```xml
<section>
      <title>Styling DITA elements</title>
      <p xml:lang="he" dir="rtl">
           החל מחלקות שירות עקביות של Bootstrap על פני DITA על ידי תיקון 
          <filepath dir="auto">/Customization/xsl/utility-classes.xsl</filepath>
      </p>
</section>
```

But this removes the chore of adding `dir="auto"` on elements which are defined in the DITA spec as part of the programming domain and as such are semantically define as code would rendered LTR in Arabic, Hebrew, Farsi etc. anyway.

PR #3942, opened 2022-06-15
Add support for DITA 2.0 split chunking

jelovirt:

## Description
Add support for DITA 2.0 split chunking

## Motivation and Context
Implement DITA 2.0 chunking.

## How Has This Been Tested?
New tests for split and existing tests for regression bugs.

## Type of Changes

TBS

### DIFF
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -437,6 +437,16 @@ See the accompanying LICENSE file for applicable license.
     <response>Use '%2' instead.</response>
   </message>
 
+  <message id="DOTJ086W" type="WARN">
+    <reason>Split chunk attribute found on '&lt;%1>' element that does not reference a topic.</reason>
+    <response>Ignoring chunk operation.</response>
+  </message>
+
+  <message id="DOTJ087W" type="WARN">
+    <reason>Found chunk attribute with value '%1' inside combine chunk.</reason>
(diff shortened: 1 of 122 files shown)


src/main/config/messages_template.xml
```diff
@@ -437,6 +437,16 @@ See the accompanying LICENSE file for applicable license.
...
   </message>
 
+  <message id="DOTJ086W" type="WARN">
+    <reason>Split chunk attribute found on '%1' element that does not reference a topic.</reason>
+    <response>Ignoring chunk operation.</response>
+  </message>
+
+  <message id="DOTJ087W" type="WARN">
+    <reason>Found chunk attribute with value '%1' inside combine chunk.</reason>
+    <response>Ignoring chunk operation.</response>
```
jelovirt, line 447: @infotexture Do you want to come up with better copy?

* `DOTJ086W` is a case where `chunk="split"` is set on `<topicref>` that doesn't have `@href` attribute, or on `<topicgroup>` or `<topichead>`. Per spec, [figure 2](https://nervous-franklin-bfb551.netlify.app/archSpec/base/example-chunk-ignored.html)
* `DOTJ087W` is case where `@chunk` is set on a descendant element of `<topicref>` that already has `chunk="combine"`. [figure 1](https://nervous-franklin-bfb551.netlify.app/archSpec/base/example-chunk-ignored.html)

PR #3951, opened 2022-06-24
do not call dita-ot:matches-linktext-class() for <prop> or <revprop>

### LINKED ISSUE #3814: Transformation fails if <topicref> has @rev and references filtered-out topic
## Expected Behavior
If I have a `<topicref>` with `@rev` defined in a map:

```xml
<map>
  <title>Map</title>
  <topicref href="topic2.dita" rev="1.0"/>
</map>
```

and that topic's content is hidden using `@props` (or other exclusion filtering):

```xml
<topic id="topic2_id" rev="1.0" props="hide">
  <title>Topic 2</title>
  <body>
    <!-- ... -->
  </body>
</topic>
```

and the content is published with a filter that highlights `@rev` content _and_ excludes `@props` content:

```shell
dita -i map.ditamap -f html5 -filter filter_props_rev.ditaval
```

then map publishing should complete, although messages should be issued for topics with filtered content as expected.

## Actual Behavior
In this case, the XSLT processing goes down an unhandled path, with transformation failing as follows:

```
[move-meta] Error evaluating (fn:not(...)) in xsl:when/@test on line 810 column 68 of mappullImpl.xsl:
[move-meta]   XPTY0004: An empty sequence is not allowed as the first argument of
[move-meta]   dita-ot:matches-linktext-class(). Found while atomizing the first operand of '='
[move-meta] at template getmetadata on line 1037 of mappullImpl.xsl:
[move-meta]      invoked by xsl:call-template at file:/home/chrispy/dita-ot/plugins/org.dita.base/xsl/preprocess/mappullImpl.xsl#790
[move-meta] at template get-stuff on line 692 of mappullImpl.xsl:
[move-meta]      invoked by xsl:call-template at file:/home/chrispy/dita-ot/plugins/org.dita.base/xsl/preprocess/mappullImpl.xsl#179
[move-meta]   In template rule with match="*[fn:contains(...)]" on line 80 of mappullImpl.xsl
[move-meta]      invoked by xsl:apply-templates at file:/home/chrispy/dita-ot/plugins/org.dita.base/xsl/preprocess/mappullImpl.xsl#1148
[move-meta]   In template rule with match="*[fn:contains(...)]" on line 1146 of mappullImpl.xsl
[move-meta]      invoked by built-in template rule (text-only)
Error: Failed to run pipeline: Failed to transform document: An empty sequence is not allowed as the first argument of dita-ot:matches-linktext-class(). Found while atomizing the first operand of '='
```

This failure only occurs when the DITAVAL applies `@revprop` highlighting to `@rev` attributes, _and_ a map `<topicref>` attempts to apply `@rev` highlighting to a filtered-out topic.

## Possible Solution
I will add something here when I've looked into the XSLT code more closely.

## Steps to Reproduce
Download the attached testcase:

[testcase.zip](https://github.com/dita-ot/dita-ot/files/7412543/testcase.zip)

then run:

```shell
# this fails - excludes @props content, applies @rev highlighting
dita -i map.ditamap -f html5 -filter filter_props_rev.ditaval

# this completes - excludes @props content, no @rev highlighting
dita -i map.ditamap -f html5 -filter filter_props_norev.ditaval
```

## Environment

* DITA-OT version: 3.6.1
* Operating system and version: Ubuntu 20.04

chrispy-snps:

## Description
This change ensures that `dita-ot:matches-linktext-class()` is not called with an empty sequence, which causes the function to fail with an exception because it requires a `@class` attribute as input.

`<prop>` and `<revprop>` elements do not have a `@class` attribute. The current code can inadvertently try to evaluate their `@class` attribute with this function, causing the issue.

## Motivation and Context
This fixes #3814.

## How Has This Been Tested?
This fix resolves the following four code paths that previously failed:

```
<topicref href="NOTFOUND.dita" rev="1.0"/>
<topicref href="topic.pdf" rev="1.0" scope="external" format="pdf"/>
<topicref href="topic.dita" rev="1.0" scope="peer"/>
<topicref href="topic.dita" rev="1.0" scope="external"/>
```

It was also tested by running `./gradlew test`.

## Type of Changes

The code paths related to the four cases above were updated to avoid calling the function with an empty sequence. Other approaches are described in #3814, but this change seemed the most precise.

This change introduces no compatibility issues. For documentation, a release notes mention is sufficient.

## Checklist

I do not know Java; can someone update unit tests to include this condition? Any map including the four constructs above is sufficient, along with a DITAVAL file that highlights `@rev`:

```
<val>
   <!-- highlight @rev revisions -->
   <revprop action="flag" backcolor="#FEFEE1"/>
</val>
```

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/mappullImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/mappullImpl.xsl
@@ -587,8 +587,8 @@ Other modes can be found within the code, and may or may not prove useful for ov
           <xsl:when test="*[contains(@class, ' map/topicmeta ')]/*[contains(@class, ' topic/navtitle ')]">
             <xsl:copy-of select="*[contains(@class, ' map/topicmeta ')]/*[contains(@class, ' topic/navtitle ')]/node()"/>
           </xsl:when>
-          <xsl:when test="*/*[dita-ot:matches-linktext-class(@class)]">
-            <xsl:value-of select="*/*[dita-ot:matches-linktext-class(@class)]"/>
+          <xsl:when test="*/*[@class][dita-ot:matches-linktext-class(@class)]">
+            <xsl:value-of select="*/*[@class][dita-ot:matches-linktext-class(@class)]"/>
           </xsl:when>
           <xsl:otherwise>
             <xsl:value-of select="@href"/>
@@ -604,8 +604,8 @@ Other modes can be found within the code, and may or may not prove useful for ov
           <xsl:when test="@navtitle">
             <xsl:value-of select="@navtitle"/>
           </xsl:when>
-          <xsl:when test="*/*[dita-ot:matches-linktext-class(@class)]">
-            <xsl:value-of select="*/*[dita-ot:matches-linktext-class(@class)]"/>
+          <xsl:when test="*/*[@class][dita-ot:matches-linktext-class(@class)]">
+            <xsl:value-of select="*/*[@class][dita-ot:matches-linktext-class(@class)]"/>
           </xsl:when>
           <xsl:otherwise>#none#</xsl:otherwise>
         </xsl:choose>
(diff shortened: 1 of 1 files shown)


robander:
Wonder if we should update these 3 functions to make the passed in `@class` value optional, and return false if the value is empty? https://github.com/dita-ot/dita-ot/blob/9d0ad7495db0240ef7acee99b3593368acd90a89/src/main/plugins/org.dita.base/xsl/common/functions.xsl#L19-L32

That would fix all similar calls to this function, not just the ones hit in the original commit

chrispy-snps:
@robander - is there any change you'd like to see here to consider this for 3.7.4?

jelovirt:
IMO the correct fix is to change those functions to allow empty sequence as parameter. Will make this change in another PR.

chrispy-snps:
@jelovirt - or, I can cancel this PR and you or I can add the `?` to the function parameter in a separate PR. (Or I can redo and force-push this one to do that.)

Any approach is fine with me. :)

jelovirt:
Replaced by #4028

PR #3955, opened 2022-07-02
wrap <note> body content in its own <span> (to match the note title)

chrispy-snps:

## Description
This change wraps the body content of a `<note>` element in its own `<div class="note__body">`:

```
<div class="note note note_note">
  <span class="note__title">Note:</span> <span class="note__body">This is a note.</span></div>
                                    <!-- ^^^^^^^^^^^^^^^^^^^^^^^^^               ^^^^^^^ -->
</div>
```

This is consistent with the note title, which is already wrapped in its own `<div class="note__title">`.

The space between the title and body is placed between the `<span>` elements. This ensures correct behavior whether the title and body are styled as inline or block objects.

## Motivation and Context
This change allows note body content to be styled independently from both the title and the parent note container.

For example, a note title and body could be styled as `display: block`, with the body indented from the title.

Or, the note container could be styled as a CSS flexbox container, with the title and body styled responsively as flex items for better rendering across desktop and mobile devices.

## How Has This Been Tested?
I ran `gradlew test` and `gradlew e2etest`. I also tested DITAVAL `<startflag>`/`<endflag>` for correct behavior.

## Type of Changes

A single `<span>` is added to the `<note>` template in `org.dita.html5`. (`org.dita.xhtml` uses a different processing structure for notes, and this change does not apply.)

A release notes mention is sufficient.

Compatibility impact should be minimal. Styling-wise, the `<span>` is transparent unless specifically styled. Functionally, existing DITA-OT plugins targeting note elements should consider the new `<span>` as regular content, such as a `<span>` from a `<ph>`-based element.

But because a compatibility impact cannot be ruled out (plugin precedence, strict expectations on note body plaintext in user plugins), this pull request is being submitted for DITA-OT 4.0 instead of 3.7.3.

## Checklist


chrispy-snps:
For reference, here is the plugin-based approach we currently use today. This could be entirely removed with the implementation of this pull request.

```
  <!-- provide separate spans for note title and note body, so we can style them independently -->

  <!-- get existing transformation results, then postprocess with a moded template -->
  <xsl:template match="*[contains-token(@class, 'topic/note')]">
    <xsl:variable name="results">
      <xsl:next-match/>
    </xsl:variable>
    <xsl:apply-templates select="$results" mode="snps-add-note-body"/>
  </xsl:template>

  <!-- define our mode -->
  <xsl:mode name="snps-add-note-body" on-no-match="shallow-copy"/>

  <!-- leave everything up to the note__title as-is, then put everything after that in a node__body -->
  <xsl:template match="*[contains-token(@class, 'topic/note')]
                        [*[contains-token(@class, 'note__title')]]
                        [not(*[contains-token(@class, 'note__body')])]" mode="snps-add-note-body">
    <xsl:copy>
      <xsl:apply-templates select="@*" mode="#current"/>
      <xsl:variable name="note-title" select="node()[contains-token(@class, 'note__title') or following-sibling::*[contains-token(@class, 'note__title')]]"/>
      <xsl:sequence select="$note-title"/>
      <xsl:text> </xsl:text>
      <span class="note__body">
        <xsl:apply-templates select="node() except $note-title" mode="#current"/>
      </span>
    </xsl:copy>
  </xsl:template>

  <!-- remove existing leading space from newly created note body (we add it above ourselves) -->
  <xsl:template match="text()[preceding-sibling::node()[1][contains-token(@class, 'note__title')]]" mode="snps-add-note-body">
    <xsl:value-of select="replace(., '^\s+', '')"/>
  </xsl:template>
```

src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -2814,10 +2814,13 @@ See the accompanying LICENSE file for applicable license.
         </xsl:call-template>
       </span>
       <xsl:text> </xsl:text>
-      <xsl:apply-templates select="*[contains(@class, ' ditaot-d/ditaval-startprop ')]/revprop" mode="ditaval-outputflag"/>
-      <xsl:apply-templates/>
-      <!-- Normal end flags and revision end flags both go out after the content. -->
-      <xsl:apply-templates select="*[contains(@class, ' ditaot-d/ditaval-endprop ')]" mode="out-of-line"/>
+      <span class="note__body">
```
robander, line 2817: A note can have all sorts of block elements - won't this end up with paragraphs/divs/lists inside of `<span>`?

chrispy-snps: @robander - I believe I confirmed that a `<span>` containing block elements becomes inconsequential and is handled by CSS layout engines as expected, but let me confirm and reply back here.

chrispy-snps:
I am withdrawing this pull request.

When I originally created this pull request, I confirmed that browsers treat `<span>` elements as elastic containers that conform to whatever inline or block element content they contain. Accordingly, they seemed like a suitable CSS-styleable container for arbitrary content.

However, @robander's question above prompted me to dig a bit deeper into HTML compliance. Unfortunately, block elements are not valid in [the `<span>` content model](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element). The proposed structure in this pull request fails the [W3C validation service](https://validator.w3.org/nu/#textarea). So while this structure behaves as desired, it is not valid HTML and the DITA-OT should not produce it.

It would be possible to be "clever" and use a `<div>` self-styled to be an elastic inline:

```
    <div class="note note note_note">
      <span class="note__title">Note:</span> <div class="note__body" style="display:inline">This is a note.</div>
    </div>
```

This is compliant per W3C validation and it behaves as desired for both block and inline content. But, it's not elegant or suitable for default HTML generation.

My company will continue using our plugin-based `<note>` override, and I hereby (and a bit sadly) withdraw this pull request.

src/main/plugins/org.dita.html5/xsl/topic.xsl
robander: I expect it would display correctly, as browsers are very forgiving. My concern is more that it's invalid HTML5 (spans allow only [phrase-level content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span)), and I don't think we should knowingly generate HTML5 that will break the spec. 

_Note ... I say knowingly because I would not be surprised to find out we have invalid HTML somewhere else. But I don't think we should set out to do that._

I'd be happy with this if it used a `<div>` rather than span, just because that would give us the container for styling but leave us with valid HTML. That said, we would need to put in some default CSS for that, unless we are setting out to change the current style where text-only notes appear directly after the **Note:** generated text.

robander:
I only saw your follow-up withdrawal after adding my comment above. I don't see any problem with using a div in this case, but if you find that less desirable then of course we can close the pull request. My only issue with div is that we need to either set up our default CSS so that it defaults to inline presentation, or we need to be explicit that we are changing the overall style. (At this point I'd prefer CSS to retain the style, given the long history here.)

chrispy-snps:
Thanks @robander! I forgot that we have default CSS files too. This makes the `<div>` approach much cleaner.

In the following testcase:

[3955.zip](https://github.com/dita-ot/dita-ot/files/11343383/3955.zip)

a diff of the output is minimal:

```
$ diff -r out-ORIG/ out-NEW/
diff -r out-ORIG/commonltr.css out-NEW/commonltr.css
476a477,480
> .note .note__body {
>   display: inline;
> }
>
diff -r out-ORIG/commonrtl.css out-NEW/commonrtl.css
483a484,487
> .note .note__body {
>   display: inline;
> }
>
diff -r out-ORIG/topic.html out-NEW/topic.html
7c7
<         <div class="note note note_note"><span class="note__title">Note:</span> Here is some inline text.</div>
---
>         <div class="note note note_note"><span class="note__title">Note:</span> <div class="note__body">Here is some inline text.</div></div>
9c9
<         <div class="note note note_note"><span class="note__title">Note:</span>
---
>         <div class="note note note_note"><span class="note__title">Note:</span> <div class="note__body">
11c11
<         </div>
---
>         </div></div>
13c13
<         <div class="note note note_note"><span class="note__title">Note:</span>
---
>         <div class="note note note_note"><span class="note__title">Note:</span> <div class="note__body">
19c19
<         </div>
---
>         </div></div>
```

The resulting HTML passes W3C validation and the browser rendering is identical to the original HTML. The testcase archive includes these output directories so you can examine them without building clients.

chrispy-snps:
Should I update the `xhtml` transtype to match this behavior?

chrispy-snps:
I was going to update `xhtml` to match, but it uses type-dependent title classes (`notetitle`, `tiptitle`, etc.). Replicating note body container types (and their CSS entries) for consistency is not worthwhile for a legacy output type.

chrispy-snps:
[This comment](https://github.com/dita-ot/dita-ot/issues/4254#issuecomment-1679748417) confirms that intrinsic styles belong in the CSS and not in hardcoded `@style` properties.

jelovirt (approved):

PR #3964, opened 2022-07-05
Fix for #3724, properly compute absolute references to files.

### LINKED ISSUE #3724: Error in mapref stage when passing dita map as resource
I'm attaching a sample DITA project
[chemistry_crash_single_topic.zip](https://github.com/dita-ot/dita-ot/files/6242590/chemistry_crash_single_topic.zip)

on Windows I'm running this command line:
```batch
dita -i C:\Users\radu_coravu\Downloads\chemistry_crash_single_topic\iugux\installation_preparation\overview_of_installation_process.dita ^
  --resource=C:\Users\radu_coravu\Downloads\chemistry_crash_single_topic\iugux.ditamap ^
  -f html5 ^
  --temp=tmp ^
  --clean.temp=no ^
  -v
```

then I get an error reported in the "mapref" stage:

     [mapref]   I/O error reported by XML parser processing
     [mapref]   file:/C:/Users/radu_coravu/Desktop/dita-ot-3.6/bin/tmp/C:/Users/radu_coravu/Downloads/chemistry_crash_single_topic/iugux.ditamap: C:\Users\radu_coravu\Desktop\dita-ot-3.6\bin\tmp\C:\Users\radu_coravu\Downloads\chemistry_crash_single_topic\iugux.ditamap (The filename, directory name, or volume label syntax is incorrect)

As you can see the absolute path to the dita map gets concatenated to the absolute path to the temporary files folder, producing an invalid file path on Windows.
Looking in the .job.xml at the entry for the DITA Map:

```xml
        <file src="file:/C:/Users/radu_coravu/Downloads/chemistry_crash_single_topic/iugux.ditamap"
            uri="file:/C:/Users/radu_coravu/Downloads/chemistry_crash_single_topic/iugux.ditamap"
            path="C:\Users\radu_coravu\Downloads\chemistry_crash_single_topic\iugux.ditamap"
            result="file:/C:/Users/radu_coravu/Downloads/chemistry_crash_single_topic/iugux.ditamap"
            format="ditamap" resource-only="true"/>
```
it is the only entry with an absolute `@path` attribute.

When the method "org.dita.dost.module.MaprefModule.processMap(FileInfo)" gets called, the input file gets computed by appending two absolute file paths one to another:

```java
    File inputFile = new File(job.tempDir, input.file.getPath());
```
which is invalid on Windows.

The same thing works on Mac because the file path does not contain a drive name and "new File(job.tempDir, input.file.getPath())" manages to produce a valid file path.

raducoravu:

## Description
Properly compute file path from parent and child path when the child path may be absolute.

## Motivation and Context
Fixes #3724

## How Has This Been Tested?
Automatic test.

## Type of Changes

No doc

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/util/FileUtils.java
+++ b/src/main/java/org/dita/dost/util/FileUtils.java
@@ -8,12 +8,12 @@
  */
 package org.dita.dost.util;
 
-import static org.apache.commons.io.FilenameUtils.*;
-import static org.dita.dost.util.Constants.*;
-
 import java.io.File;
 import java.net.URI;
 import java.util.*;
+
+import static org.apache.commons.io.FilenameUtils.normalize;
+import static org.dita.dost.util.Constants.*;
 
 /**
  * Static file utilities.
@@ -623,7 +623,7 @@ public final class FileUtils {
      * @return The file path, never <code>null</code>
      */
     public static File getFilePath(File parent, URI path) {
-        return getFilePath(parent, path.getPath());
+        return parent.toPath().resolve(path.getPath()).toFile();
     }
     
     /**
@@ -633,7 +633,7 @@ public final class FileUtils {
      * @return The file path, never <code>null</code>
      */
     public static File getFilePath(File parent, File path) {
-        return getFilePath(parent, path.getPath());
+        return parent.toPath().resolve(path.toPath()).toFile();
     }
 
     /**
@@ -643,10 +643,6 @@ public final class FileUtils {
      * @return The file path, never <code>null</code>
      */
     public static File getFilePath(File parent, String path) {
-        File filePath = new File(path);
-        if(!filePath.isAbsolute()) {
-            filePath = new File(parent, path);
-        }
-        return filePath;
+        return parent.toPath().resolve(path).toFile();
     }
 }
--- a/src/main/java/org/dita/dost/util/FileUtils.java
+++ b/src/main/java/org/dita/dost/util/FileUtils.java
@@ -615,34 +615,5 @@ public final class FileUtils {
             return c.getPath().startsWith(d.getPath());
         }
     }
-    
-    /**
-     * Get a file path.
-     * @param parent The parent file.
-     * @param path An URI containing a path which can be either absolute or relative. 
-     * @return The file path, never <code>null</code>
-     */
-    public static File getFilePath(File parent, URI path) {
-        return parent.toPath().resolve(path.getPath()).toFile();
-    }
-    
-    /**
-     * Get a file path.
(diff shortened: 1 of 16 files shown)


jelovirt (commented):
The new `FileUtils.getFilePath` method is basically same as `Path.resolve` or `URI.resolve`. I would prefer to change the code from

```java
File out = FileUtils.getFilePath(destDir, path);
```

to just use

```java
File out = destDir.toPath().resolve(path).toFile();
```

In the long run we should remove basically all use of `File` and replace it with `URI`, or in cases where we must use system paths, replace with `Path`.

src/main/java/org/dita/dost/util/FileUtils.java
```diff
@@ -615,5 +615,38 @@ public static boolean directoryContains(final File directory, final File child)
...
+     * @param parent The parent file.
+     * @param path A file path which can be either absolute or relative. 
+     * @return The file path, never <code>null</code>
+     */
+    public static File getFilePath(File parent, String path) {
+        File filePath = new File(path);
+        if(!filePath.isAbsolute()) {
+            filePath = new File(parent, path);
+        }
+        return filePath;
```
jelovirt, line 650: If we change this to

```suggestion
        return parent.toPath().resolve(Paths.get(path)).toFile();
```

It would be an easy way to check if that tests still pass. Then the other `getFilePath` methods that call this could be inlined and made more simple. Like

```java
public static File getFilePath(File parent, File path) {
    return getFilePath(parent, path.getPath());
}
```

would be rewritten to

```java
public static File getFilePath(File parent, File path) {
    return parent.toPath().resolve(path.toPath()).toFile();
}
```

jelovirt:
@raducoravu I pushed my suggested changes. Are you able to run the tests on Windows to see if they still work?

raducoravu:
@jelovirt I'm confused, I do not understand where you pushed your changes.

jelovirt:
@raducoravu I pushed my changes to your branch.

jelovirt:
@raducoravu See https://github.com/dita-ot/dita-ot/pull/3964/commits

jelovirt:
@raducoravu I rebased the branch on `hotfix/3.7.3` so pull with rebase when you test this on Windows.

raducoravu:
@jelovirt I confirm that when running my original example in #3724 on Windows with your changes the mapref stage properly finds the DITA Map:

       mapref:
         [mapref] Processing file:/C:/Users/radu_coravu/Downloads/chemistry_crash_single_topic/iugux.ditamap

jelovirt (approved):

PR #4003, opened 2022-09-08
#3940 Propagate scope and processing-role in ForceUniqueFilter

### LINKED ISSUE #3940: `processing-role="resource-only"` does not cascade downward in map
## Expected Behavior
In a map, I have two versions of the same topic file (topic2.dita):

* The unpublished version is resource-only without a key definition.
* The published version is local with a key definition.

wrong_ref

When I have an `<xref>` to that key, the reference should resolve to the published version.

## Actual Behavior
The transformation creates a duplicate HTML5 file for the unpublished version, and the key-derived reference incorrectly points to that instead:

wrong_ref_html

## Possible Solution
A workaround (but not a fix) is to add `@processing-role="resource-only"` to the unpublished `<topicref>` directly:

wa

In this case, the HTML5 output is as expected:

wa_html

## Steps to Reproduce
Testcase: [ditaot_wrong_ref.zip](https://github.com/dita-ot/dita-ot/files/8884518/ditaot_wrong_ref.zip)

To run: `dita -i map -f html5 -Dforce-unique=true`

## Environment

* DITA-OT version: 3.7.2
* Operating system and version: Linux

raducoravu:

## Description
Propagate scope and processing-role attributes from ancestors in the ForceUniqueFilter stage.

## Motivation and Context
Fixes #3940

## How Has This Been Tested?
Auto test.

## Type of Changes

No doc

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/writer/ForceUniqueFilter.java
+++ b/src/main/java/org/dita/dost/writer/ForceUniqueFilter.java
@@ -61,10 +61,11 @@ public final class ForceUniqueFilter extends AbstractXMLFilter {
             final URI href = toURI(res.getValue(ATTRIBUTE_NAME_HREF));
             final URI copyTo = toURI(res.getValue(ATTRIBUTE_NAME_COPY_TO));
             final URI source = copyTo != null ? copyTo : href;
-            final String scope = res.getValue(ATTRIBUTE_NAME_SCOPE);
+            String currentScope = res.getValue(ATTRIBUTE_NAME_SCOPE);
+            String scope = getCascadingScope(currentScope);
             final String format = res.getValue(ATTRIBUTE_NAME_FORMAT);
-            // FIXME: handle cascading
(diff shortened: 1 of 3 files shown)


chrispy-snps:
@raducoravu - this fix worked perfectly for me. On our help collection map of about 7k topics, it eliminated the expected 48 duplicate topics correctly. I also compared your fix against my hacky `mapref` workaround:

```
  <!-- work around DITA-OT #3940 (processing-role="resource-only" does not cascade downward in map) -->
  <xsl:template match="*[@processing-role = 'resource-only']//*[contains-token(@class, 'map/topicref')]">
    <xsl:copy>
      <xsl:attribute name="processing-role" select="'resource-only'"/>
      <xsl:apply-templates select="@*|node()"/>
    </xsl:copy>
  </xsl:template>
```

and confirmed that both approaches produced identical results.

Thank you for submitting the fix!

jelovirt (approved):

PR #4005, opened 2022-09-10
Fix #4001 - Add lazy loading for external images

### LINKED ISSUE #4001: Should HTML Images be lazily loaded by default?
## Description

Whilst fixing a chunking issue in a custom plugin, I discovered a more generic issue related to the rendering of `` elements. The issue can be seen here: https://jason-fox.github.io/dita-bootstrap/Chunk1343199979.html

I have a page which contains a side-toc and lots of content. In this particular case, one image is failing to load, but that doesn't really matter, what does matter is that the DOM is only fully rendered once the whole page is loaded. In this case it takes 45 seconds for the bottom half of the DOM after the slow image to complete rendering. In the meantime the side-toc remains unusable.

When loading on mobile, for a massive full of images, all images would need to be downloaded (wasting bandwidth) before the page is rendered. This would also improve rendering speed of longer documents throughout.

## Possible Solution

[Lazy loading](https://blog.hubspot.com/website/lazy-loading-eager-loading) has been implemented across many browsers.  The suggestion is to add [`loading="lazy`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading) to rendered images:

```html
<img  src="/media/cc0-images/grapefruit-slice-332-332.jpg"
     alt="Grapefruit slice atop a pile of other slices"
     loading="lazy"
/>
```

The `loading` attribute is ignored by older browsers.

The questions are as follows: 

#### When should `loading="lazy"` be added?

Looking at the DITA spec, I can see that [`<image>`](https://www.oxygenxml.com/dita/1.3/specs/langRef/base/image.html) has an [`@scope`](https://www.oxygenxml.com/dita/1.3/specs/langRef/attributes/thescopeattribute.html) attribute, the suggestion would be to add `loading="lazy"`  when `scope="external"` or defaults to external because the `@href` begins with `http` and `@scope` is not defined.

#### When should `loading="eager"` be added?

If the `@scope` attribute of  an `<image>` is deliberately set to `scope="local"`, it should probably be set to  `loading="eager"`

#### What about `scope="peer"`? 

Given the definition _might not accessible at build time_ that would seem to indicate `loading="lazy"`.

## Potential Alternatives
This behaviour could added as a custom plugin, but seems too trivial to warrant one. I think it would make a better PR request, but I'd like feedback before I spend more time looking at it and work out if the proposed categories of lazy/eager are correct. Furthermore, a properly implemented XSL template could still be applied to override the defaults - e.g. set lazy for all images or none.

jason-fox:
Signed-off-by: Jason Fox <jason-fox@users.noreply.github.com>

## Description

[Lazy loading](https://blog.hubspot.com/website/lazy-loading-eager-loading) has been implemented across many browsers.  The suggestion is to add [`loading="lazy`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading) to rendered images:

```html
<img  src="/media/cc0-images/grapefruit-slice-332-332.jpg"
     alt="Grapefruit slice atop a pile of other slices"
     loading="lazy"
/>
```

The `loading` attribute is ignored by older browsers.

## Motivation and Context

See attached Issue #4001 

## How Has This Been Tested?

Local builds. Manual Testing

## Type of Changes


chrispy-snps:
@jason-fox - I suggest removing the `priority="-10"` from the last template, as it's not needed.

jason-fox:
Setting a DITA `<image>` like this:

```xml
<image scope="external" href="https://...etc"/>
```

Causes a match against both templates, which results in the following if `priority` is missing.

```text
     [xslt] Ambiguous rule match for /dita/topic[1]/topic[2]/body[1]/bodydiv[1]/image[1]
     [xslt] Matches both "*[fn:matches(...)]" on line 1980 of file:/Users/jasonfox/Workspace/dita/dita-ot-3.7/plugins/org.dita.html5/xsl/topic.xsl
     [xslt] and "*[xs:string(@scope) eq "external"]" on line 1977 of file:/Users/jasonfox/Workspace/dita/dita-ot-3.7/plugins/org.dita.html5/xsl/topic.xsl
```

Hence the addition of `priority`. Now it may be the case that the `<xsl:template match="*[matches(@href, '^https?:')]">` is  likely to be unnecessary because pre-process already complains and sets the sets the `<image>`  to `scope="external"` anyway, but I think having both better explains the **intent** of the change since it doesn't rely on preprocessing always doing the right thing.

chrispy-snps:
But I think the default template for `"*"` is taking precedence over the `priority="-10"` template, or at least that's the behavior I see when I test these templates in isolation in an [XSLT fiddle](https://xsltfiddle.liberty-development.net/). Perhaps instead a `priority="10"` can be added to the `@scope="external"` template to break the ambiguity?
```

raducoravu:
One remark, the stylesheet "org.dita.xhtml/xsl/dita2xhtml-util.xsl" should probably have an extra template to remove this attribute as the XHTML generation should be valid according to the XHTML standard which does not allow it.

jason-fox:
> I think the https-match template can be omitted completely.

@ chrispy-snps - removed  6af6e29

> the stylesheet "org.dita.xhtml/xsl/dita2xhtml-util.xsl" should probably have an extra template to remove this attribute 

@ raducoravu - Is that really necessary? There is no call to the `<xsl:apply-templates mode="set-image-loading"/>` in the XHTML transforms compare https://github.com/jason-fox/dita-ot/blob/patch-1/src/main/plugins/org.dita.html5/xsl/topic.xsl#L1396 to the equivalent  https://github.com/jason-fox/dita-ot/blob/patch-1/src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl#L1556

raducoravu:
@jason-fox please disregard my previous comment then.

robander (approved):

chrispy-snps:
Thanks @robander!

PR #4010, opened 2022-09-21
#4007 Avoid NPE

### LINKED ISSUE #4007: NullPointerException crash occurs when a peer mapref and local mapref use the same keyscope
## Expected Behavior
When I run a transformation on the following files, where (1) there is a peer and local mapref of the same submap, and (2) there is a keyref inside a submap topic:

bug

I expect it to complete successfully.

## Actual Behavior
The transformation fails as follows:

```
keyref:
   [keyref] Reading file:/tmp/temp20220915211429383/map.ditamap
Error: java.lang.NullPointerException: Cannot invoke "org.dita.dost.util.KeyScope.equals(Object)" because "rt.scope" is null
```

In `submap.ditamap`, if I comment out one of the two `<topicref>` elements, then it fails as follows:

```
keyref:
   [keyref] Reading file:/tmp/temp20220915212742017/map.ditamap
Error: java.lang.NullPointerException: element cannot be mapped to a null key
```

The issue seems to somehow be caused by the following:

* The peer and local maprefs sharing the same keyscope value in `map.ditamap`
* The key references to be resolved in the topic files

If I remove any of these, the crash does not occur.

## Possible Solution
When I swap the order of the peer mapref and the local mapref, the crash no longer occurs. Maybe the key resolution attempt is not handling the peer mapref properly.

## Steps to Reproduce
1. Download and extract the following testcase:
   [4007.zip](https://github.com/dita-ot/dita-ot/files/9583577/4007.zip)
2. Run the following command:
   `dita -i map.ditamap -f html5`

## Environment
* DITA-OT version 3.7.3
* Operating system and version: Linux (Ubuntu 20.04)

raducoravu:
Avoid NPE with peer map refs

## Description
Avoid NPE when using peer map refs

## Motivation and Context
Fixes #4007

## How Has This Been Tested?
Auto test

## Type of Changes

No doc

## Checklist


src/test/java/org/dita/dost/module/KeyrefModuleTest.java
```diff
@@ -363,4 +364,38 @@ public void testWalkMapAndRewriteKeydefHrefKeepAnchor() throws URISyntaxExceptio
...
+                .result(new File(baseDir, "src" + File.separator + "test4.ditamap").toURI())
+                .format("ditamap")
+                .isInput(true)
+                .build();
+        job.add(inputMapFileInfo);
+        final XdmNode act = parse(inputMapFileInfo.src);
+        Map<String, KeyDef> defsMap = new HashMap<>();
+        defsMap.put("Tool", new KeyDef("Tool", null, "local", "dita", inputMapFileInfo.src, null));
+        
+        KeyScope submap1 = new KeyScope("submap", " mapref[6]map[6].submap", defsMap, Arrays.asList(new KeyScope[] {}));
```
chrispy-snps, line 382: Should the second argument have a leading space?

src/test/resources/KeyrefModuleTest/src/test4.ditamap
```diff
@@ -0,0 +1,15 @@
...
+        <topicref class="- map/topicref " href="topic1.dita" xtrc="topicref:1;6:35" xtrf="file:/Users/raducoravu/Downloads/4007/submap.ditamap"/>
```
chrispy-snps, line 12: Are the absolute `file:/` paths okay here?

src/test/java/org/dita/dost/module/KeyrefModuleTest.java
raducoravu: What second argument?

chrispy-snps: On line 382, the second argument to KeyScope() is `" mapref[6]map[6].submap"` which is different than line 383, so I was curious about the difference.

src/test/resources/KeyrefModuleTest/src/test4.ditamap
raducoravu: They do not influence the test, but I just removed the path which was specific to my operating system nonetheless.

src/test/java/org/dita/dost/module/KeyrefModuleTest.java
raducoravu: I removed that extra space as well.

chrispy-snps:
@raducoravu - thank you for the fix, and for creating this pull request!

PR #4018, opened 2022-10-10
Remove obsolete FINALOUTPUTTYPE param

robander:
## Description

Removes the obsolete `FINALOUTPUTTYPE` parameter. This is never passed in from our scripts, and refers to legacy formats and attributes.

## Motivation and Context

Motivation is code cleanup and removing technical debt.

The check for `IDD` as an output format is a legacy of DITA-OT 1.0 and the beta code that came from IBM, this referred to the SGML format "IBMIDDoc" that was used for a PDF path at the time. This is no longer used in IBM and should not be in this code even if it was.

Because the parameter is never passed in by any of our Ant scripts, the tests always evaluate to false. In many cases the tests were for `not(FINALOUTPUTTYPE=X)` which is equivalent to `not(false())`, so these tests always passed; direct tests of the value always fail.

## How Has This Been Tested?

Unit tests continue to pass

## Type of Changes

Technically a breaking change as anyone setting this parameter and passing it to `maplink` or `mappull` with a plugin will no longer see it passed in.

n/a

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
@@ -21,7 +21,6 @@ See the accompanying LICENSE file for applicable license.
   
   <!-- =========== DEFAULT VALUES FOR EXTERNALLY MODIFIABLE PARAMETERS =========== -->
   <!-- output type -->
-  <xsl:param name="FINALOUTPUTTYPE" select="''"/>
   <xsl:param name="INPUTMAP" select="''"/>
   <xsl:param name="WORKDIR">
     <xsl:apply-templates select="/processing-instruction('workdir-uri')[1]" mode="get-work-dir"/>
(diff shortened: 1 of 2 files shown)


src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
```diff
@@ -119,17 +118,14 @@ See the accompanying LICENSE file for applicable license.
         <xsl:with-param name="pathFromMaplist" select="$pathFromMaplist"/>
       </xsl:call-template>
     </xsl:variable>
-    <!-- If going to print, and @print=no, do not create links for this topicref -->
-    <xsl:if test="not(($FINALOUTPUTTYPE = 'PDF' or $FINALOUTPUTTYPE = 'IDD') and @print = 'no')">
```
robander, line 123: The param is never set so
`not(($FINALOUTPUTTYPE = 'PDF' or $FINALOUTPUTTYPE = 'IDD') and @print = 'no')`
==>
`not((false() or false()) and @print = 'no')`
==>
`not((false()) and @print='no')`
==>
`not(false())`
==>
`<xsl if test="true()">`

src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
```diff
@@ -535,8 +531,7 @@ See the accompanying LICENSE file for applicable license.
     <!-- child found tag -->
     <xsl:param name="found" select="true()" as="xs:boolean"/>
     <!-- If going to print, and @print=no, do not create links for this topicref -->
-    <xsl:if test="not(($FINALOUTPUTTYPE = 'PDF' or $FINALOUTPUTTYPE = 'IDD') and @print = 'no') and
```
robander, line 538: Same test as above, this was equivalent to
`<xsl:if test="true() and not(@processing-role = 'resource-only') and $found">`

src/main/plugins/org.dita.base/xsl/preprocess/maplinkImpl.xsl
```diff
@@ -591,15 +586,10 @@ See the accompanying LICENSE file for applicable license.
         <!--figure out the linktext and desc-->
         <xsl:apply-templates select="*[contains(@class,' ditaot-d/ditaval-startprop ')]" mode="add-props-to-link"/>
         <xsl:if test="*[contains(@class, ' map/topicmeta ')]/*[dita-ot:matches-linktext-class(@class)]">
-          <!--Do not output linktext when The final output type is PDF or IDD
-            The target of the HREF is a local DITA file
-            The user has not specified locktitle to override the title -->
-          <xsl:if test="not(($FINALOUTPUTTYPE = 'PDF' or $FINALOUTPUTTYPE = 'IDD') and (not(@scope) or @scope = 'local') and (not(@format) or @format = 'dita') and (not(@locktitle) or @locktitle = 'no'))">
```
robander, line 597: `($FINALOUTPUTTYPE = 'PDF' or $FINALOUTPUTTYPE = 'IDD')` == `false()`

So this was `<xsl:if test="not(false() and X and Y and Z)">` ==> which gets `<xsl:if test="not(false())">` so again, always true

jelovirt (approved):

PR #4044, opened 2022-11-11
Add PDF theme plug-in as bundled plug-in

jelovirt:
## Description
Add PDF theme plug-in as bundled plug-in.

## Motivation and Context
Add PDF theme plug-in as bundled plug-in to make it easier to customize PDF output.

## Type of Changes


infotexture (changes requested):
See https://github.com/jelovirt/pdf-generator/issues/87

infotexture (approved):

PR #4050, opened 2022-11-14
Incremental publishing

jelovirt:
## Description
Incremental publishing.

## Type of Changes

  _(Provide links to existing documentation topics that will require updates)_

## Checklist

### DIFF
--- a/src/main/java/org/dita/dost/util/Graph.java
+++ b/src/main/java/org/dita/dost/util/Graph.java
@@ -1,0 +1,69 @@
+/*
+ * This file is part of the DITA Open Toolkit project.
+ *
+ * Copyright 2022 Jarno Elovirta
+ *
+ * See the accompanying LICENSE file for applicable license.
+ */
+
+package org.dita.dost.util;
+
+/**
+ * Implementation of directed Graph using Adjacent Matrix.
+ *
+ * @see <a href="https://en.wikipedia.org/wiki/Adjacency_matrix#Directed_graphs">Adjacent Matrix</a>
+ */
+public class Graph {
+    private int size;
+    private boolean adjacentMatrix[][];
+
+    public Graph(int size) {
+        if (size < 0) {
+            throw new IllegalArgumentException();
+        }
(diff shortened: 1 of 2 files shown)


chrispy-snps:
@jelovirt - will this use some kind of SHA of the input files to determine only what needs to be reprocessed? (I have been considering doing something like this in our builds to detect what deliverables must be rebuilt when writers modify the DITA content.)

kirkilj:
I was having SHA inklings about this as well.

jelovirt:
Brain dump

1. pass `--cache=path/dir`
2. read `path/dir/.job.xml` to get file infos and get dependency graph
3. start processing from input document:
    1. check if file has changed by timestamp and hash
    2. if file has not changed, copy file from cache and mark file as fully processed
    3. else,  continue processing normally
4. run rest of preprocessing
5. convert every changed file to HTML

raducoravu:
So step (6) would not benefit of incremental publishing. But all preprocessing steps would. Also if at least one map is found to be modified, all incremental caches for all resources are off.

raducoravu:
For step (6) are you considering some kind of ditafileset which only enumerates not changed files from the temp folder?
For our WebHelp output, as we need to generate breadcrumbs inside each topic we will probably not benefit of incremental publishing caching in this step, but benefit in the previous steps.

chrispy-snps:
I wonder if this approach could be scaled up to implement deliverable-level caching, as described in the last approach here:

[dita-users.groups.io > some thoughts on how a DITA/Git CI/CD pipeline could work](https://dita-users.groups.io/g/main/topic/some_thoughts_on_how_a/95871034)

For example, maybe the DITA could store deliverable checksums (perhaps by deliverable ID), then it would simply be a matter of doing

`dita --project project.xml --cache /path/to/cache`

which would be extremely cool.

Regarding the timestamp check - note that it might be useful to disable this in Git-based flows. Git does not store or restore timestamps. If I temporarily change branches or make then discard changes, timestamps can differ even when contents do not.

PR #4052, opened 2022-11-15
Add directed graph for URIs

jelovirt:
Add directed graph for URIs to track dependencies between resources identified with a URI.


kirkilj:
What took you so long? 😉

PR #4054, opened 2022-11-16
PDF2: Remove duplicate `space-before` attribute

infotexture:
The `topic.topic.title` attribute set currently contains 2 lines that both set the `space-before` attribute to different values.

Assuming the second instance here overrides the first, this change removes the redundant `15pt` setting, leaving only the effective `12pt` value.


infotexture:
The duplicate was apparently introduced back in 2011 with [69cdbb9](https://github.com/dita-ot/dita-ot/commit/69cdbb924112f3c4da6d0956bdfc101798e56002#diff-11b722afc2158538797ea44c16f9d3409931c188d53626c40f24a3f790e3b79eL128), when margin top/bottom was replaced with with space before/after.

robander:
Tested and confirmed that the 15pt one was ignored, so removing it has no impact on the PDF

Issue #3435, opened 2019-12-28
Comment or deprecate "place-tbl-lbl" named template


chrispy-snps:
## Expected Behavior
The "place-tbl-lbl" named template (in org.dita.html5/xsl/tables.xsl) should be either commented or deprecated if it is not currently used.

When I tried to override figure and table titles, I saw that "place-fig-lbl" affected figure titles but "place-tbl-lbl" did not affect table titles. A closer inspection revealed that the figure label template is defined and used:

```
$ fgrep 'place-fig-lbl' -r .
./xsl/topic.xsl:      <xsl:call-template name="place-fig-lbl"/>
./xsl/topic.xsl:  <xsl:template name="place-fig-lbl">
```

but the table label template is defined but never used:

```
$ fgrep 'place-tbl-lbl' -r .
./xsl/tables.xsl:  <xsl:template name="place-tbl-lbl">
```

I spent a couple hours trying to figure out why the override was failing before realizing the template was never used.

## Actual Behavior
If you search for "table--title-label" in tables.xsl, you'll see some piecemeal templates further down the file that implement table output. This is fine; the user just needs some guidance (in comments, etc.) to know where they should be working.

## Environment
* Ubuntu 19.10
* DITA-OT 3.4
* `html5` transformation

robander:
Linked issue above marks this template deprecated. Waiting to remove from `develop` branch until after the hotfix change is merged into `develop`

PR #4056, opened 2022-11-17
Mark obsolete template as deprecated #3435

robander:
Signed-off-by: Robert D Anderson <gorodki@gmail.com>

## Description

Adds a comment to mark an unused template as deprecated, so we can remove it in 4.1

## Motivation and Context

Requested in #3435

## How Has This Been Tested?

n/a, only adds an XML comment

## Type of Changes

no functional change

For release notes, can just note that the `place-tbl-lbl` template in HTML5 processing has been marked deprecated. This template was carried over from XHTML code (which still has a copy that is used), but the copy in HTML5 is not called.


infotexture (approved):
I'll make a note to add this to the 4.0 migration topic.

Issue #4048, opened 2022-11-13
Warn about profiled DITA content that is not affected by any DITAVAL filtering rule


chrispy-snps:
## Description
For us, content with profiling conditions should reach users only when explicitly allowed by a DITAVAL file. There is an existing DOTJ031I information message:

```
[exec]    [filter] [DOTJ031I][INFO] No specified rule for 'product=B' was found in the ditaval file. This value will use the default action, or a parent prop action if specified. To remove this message, you can specify a rule for 'product=B' in the ditaval file.
```

but it has some limitations.

### Limitation 1 - DOTJ031I check is never performed if filtering is not used

If neither global filtering nor branch filtering is used, then the DOTJ031I check for missing DITAVAL is never performed.

In the following run, no DITAVAL is applied and all profiled content is published, but no DOTJ031I message is issued:

```
$ ./run-no-filtering.sh
...

DOTJ031I messages indicating a missing DITAVAL rule for `@product`:

Profiled content that was published without DITAVAL consideration:
out/no-filtering-mapref/topicA.html:        <p class="p">This is content for product A.</p>
out/no-filtering-mapref/topicB.html:        <p class="p">This is content for product B.</p>
-
out/no-filtering-topic/topic.html:        <p class="p">This is content for product A.</p>
out/no-filtering-topic/topic.html:        <p class="p">This is content for product B.</p>
-
out/no-filtering-topicref/topicA.html:        <p class="p">This is content for product A.</p>
out/no-filtering-topicref/topicB.html:        <p class="p">This is content for product B.</p>
```

### Limitation 2 - DOTJ031I does not differentiate between default-exclusion filtering and missing filtering

The DOTJ031I check does not differentiate between a completely missing DITAVAL filter and a default DITAVAL exclusion (`@action="exclude"` with no `@val`):

```
<val>
    <prop att="product" action="exclude"/>
    <prop att="product" val="A" action="include"/>
</val>
```

In the following run, DITAVAL global filtering is applied to intentionally exclude `@product="B"` content via a default exclusion action, but DOTJ031I is issued for the `@product="B"` content:

```
$ ./run-global-filtering.sh
...

DOTJ031I messages indicating a "missing" filtering rule for product="B":
log/global-filtering-mapref.log: [gen-list] [DOTJ031I][INFO] No specified rule for 'product=B' was found in the ditaval file. This value will use the default action, or a parent prop action if specified. To remove this message, you can specify a rule for 'product=B' in the ditaval file.
[...]

PR #4073, opened 2022-12-08
Do not issue information messages if the filter has default exclude rules

raducoravu:
Do not issue information messages that certain values were not specified in the ditaval filter if the filter has default exclude rules

## Description
Sometimes the DITA OT issues an information message like this:

     [filter] [DOTJ031I][INFO] No specified rule for 'product=B' was found in the ditaval file.

but if the DITAVAL has default exclude rules like this:

     <val>
	<prop action="exclude"/>
	<prop action="include" att="product" val="product_A"/>
    </val>

or like this:

     <val>
	<prop action="exclude" att="product"/>
	<prop action="include" att="product" val="product_A"/>
     </val>

there is no need to issue such information messages because the default exclude rules take care of all other attribute values.

## Motivation and Context
Fixes most of the problems reported in #4048

## How Has This Been Tested?
Automatic test

## Type of Changes

No doc.

## Checklist

### DIFF
--- a/src/test/java/org/dita/dost/util/FilterUtilsTest.java
+++ b/src/test/java/org/dita/dost/util/FilterUtilsTest.java
@@ -263,6 +263,41 @@ public class FilterUtilsTest {
     }
     
     @Test
+    public void testNoInfoMessageFilter() {
+        //Do not issue information messages that certain values
+        //were not specified in the ditaval filter if the filter has default exclude rules.
+        final Map<FilterKey, Action> fm = new HashMap<>();
+        fm.put(new FilterKey(OS, null), Action.EXCLUDE);
+        fm.put(new FilterKey(OS, "amiga"), Action.INCLUDE);
+        FilterUtils f = new FilterUtils(false, fm, null, null);
+        StringBuilder infoMsg = new StringBuilder();
+        f.setLogger(new TestUtils.TestLogger() {
+            @Override
+            public void info(String msg) {
+                infoMsg.append(msg);
+            }
+        });
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/dita/dost/util/FilterUtilsTest.java
```diff
@@ -262,6 +262,41 @@ public void testExtCheckExclude() {
         assertFalse(f.extCheckExclude(new QName[] {PLATFORM, OS}, Arrays.asList( "unix", "amiga", "windows")));
     }
     
+    @Test
+    public void testNoInfoMessageFilter() {
+        //Do not issue information messages that certain values
```
jelovirt, line 267: Could you refactor this test to something like to remove duplication and also make it easier to turn this into a parameterized test when we move to JUnit 5

```java
for (FilterKey filterKey: new FilterKey[]{ new FilterKey(OS, null), FilterUtils.DEFAULT }) {
    final Map<FilterKey, Action> fm = new HashMap<>();
    fm.put(filter, Action.EXCLUDE);
    fm.put(new FilterKey(OS, "amiga"), Action.INCLUDE);
    FilterUtils f = new FilterUtils(false, fm, null, null);
    StringBuilder infoMsg = new StringBuilder();
    f.setLogger(new TestUtils.TestLogger() {
        @Override
        public void info(String msg) {
           infoMsg.append(msg);
       }
    });

    assertTrue(f.extCheckExclude(new QName[] {OS}, Arrays.asList("osx")));
    assertEquals("Should not output info message " + infoMsg, "", infoMsg.toString());
}
```

jelovirt (changes requested):

jelovirt (approved):

PR #4078, opened 2022-12-09
Fix homebrew release workflow

jelovirt:
## Description
Fix homebrew workflow and add homebrew release to release workflow.

When homebrew release has been proven to work inside release workflow, separate homebrew workflow can be removed.


infotexture (approved):
@jelovirt Was this version of the `bump-homebrew-formula.yml` action used to generate https://github.com/Homebrew/homebrew-core/pull/117752 ?

If so, this should be good to go.

PR #4089, opened 2022-12-28
remove "mailto:" prefix from default "mailto:" link target text

### LINKED ISSUE #4020: Improve the way emails are presented
I have the following DITA topic: 
image

I publish it to HTML5 and I obtain this: 
image

I expected that the "mailto:" prefix not be presented.  (the "mailto:" should be removed from the text inside the "a" element).

chrispy-snps:

## Description
For `mailto:` links,

* In `pdf2`, `xhtml`, and `html5`, removes the `mailto:` prefix from the default target text.
* In `xhtml` and `html5`, suppresses `target="_blank"` to avoid a blank browser tab from briefly appearing before the mail client opens.

Both `<xref>` and `<link>` elements are updated. If explicit target text is provided, it is used normally, but `target="_blank"` is still suppressed.

## Motivation and Context
Fixes #4020.

## How Has This Been Tested?
I ran this testcase: [4020.zip](https://github.com/dita-ot/dita-ot/files/10316747/4020.zip) (`dita --project project.xml`).

I also ran `gradlew test` and `gradlew e2etest`.

## Type of Changes

I initially tried to override the `name="href"` template, but I found that it is used to compute `@href` attribute values as well as to compute target text.

A release notes mention should be sufficient. I would not expect any plugins to break. If plugins override the default processing, they will still override this updated default processing.

## Checklist

I did not update unit tests because no tests broke with the change.

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/rel-links.xsl
@@ -77,6 +77,9 @@ See the accompanying LICENSE file for applicable license.
                   <xsl:apply-templates select="*[not(contains(@class, ' topic/desc '))] | text()"/>
                   <!--use xref content-->
                 </xsl:when>
+                <xsl:when test="@scope = 'external' and starts-with(@href, 'mailto:')">
+                  <xsl:value-of select="replace(@href, '^mailto:', '')"/><!--remove mailto: prefix from href text-->
+                </xsl:when>
                 <xsl:otherwise>
                   <xsl:call-template name="href"/><!--use href text-->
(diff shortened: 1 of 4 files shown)


lief-erickson:
@chrispy-snps, not everyone has an external email client installed. What happens if the email client is a web browser?

chrispy-snps:
@lief-erickson - thanks for asking this question! When I tried `mailto:` links with a browser-based email client, I found that Firefox ignored the `target="blank"` specification but Chrome considered it.

To avoid a regression in behavior, I will remove this aspect of the PR.

jelovirt (approved):

PR #4093, opened 2023-01-01
remove unreferenced XSLT variables

chrispy-snps:

## Description
Removes `<xsl:variable>` definitions that are not referenced in subsequent code.

## Motivation and Context
I noticed an unreferenced variable while exploring the code. I used the following refactoring operation to identify and remove them in all XSLT files:

[remove-unreferenced-variables.zip](https://github.com/dita-ot/dita-ot/files/10328895/remove-unreferenced-variables.zip)

## How Has This Been Tested?
All tests pass.

During code review, please look at the following:

* plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
  * line 518 - can this remaining comment can be modified/removed?
  * line 518 - can we pass @type instead of $type below and remove this remaining variable?
* plugins/org.dita.pdf2/xsl/fo/topic.xsl
  * line 56 - this looks like a bad variable reference, but after "fixing" it, I was still not able to sensitize this code branch with testcases, so I removed it

## Type of Changes
- Unused code removal - no behavior change

No documentation or release notes needed.

## Checklist


src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
```diff
@@ -53,17 +53,7 @@ See the accompanying LICENSE file for applicable license.
     </xsl:template>
 
     <xsl:template match="*[contains(@class,' topic/topic ')]/*[contains(@class,' topic/title ')]">
-        <xsl:variable name="topicType" as="xs:string">
```
jelovirt, line 56: This should now be removed, with the change it will run normal processing in every case.

jelovirt (changes requested):

src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
chrispy-snps: @jelovirt - I updated the PR to keep the code but fix the variable reference. Please let me know if this is correct, as I was unable to sensitize this code path to validate the behavior. (We do not use the `pdf2` transformation and I am not familiar with it.)

src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
```diff
@@ -58,7 +58,7 @@ See the accompanying LICENSE file for applicable license.
         </xsl:variable>
         <xsl:choose>
             <!--  Disable chapter title processing when mini TOC is created -->
-            <xsl:when test="(topicType = 'topicChapter') or (topicType = 'topicAppendix')" />
+            <xsl:when test="($topicType = 'topicChapter') or ($topicType = 'topicAppendix')" />
```
robander, line 61: This looks like it fixes a very old bug, and will change the layout of the mini-TOC in PDF; I generally turned off the mini-TOC but I wonder if this will give unexpected results. (That is, maybe if this had worked in the first place, the behavior would be wrong, and we haven't noticed because the code bug masked that issue.)

chrispy-snps: @robander - I also wondered about behavior change when filing the PR, but I could not sensitize this statement. For example, if I add the following debug statement:

```
<xsl:message expand-text="yes">$topicType = '{$topicType}'</xsl:message>
```

then transform the following `<bookmap>` testcase with `<part>`/`<chapter>`/`<appendices>`/`<appendix>` structure:

[4093-bookmap.zip](https://github.com/dita-ot/dita-ot/files/11342341/4093-bookmap.zip)

I only get `topicSimple` values:

```
transform.topic2fo.main:
...
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
 [pipeline] $topicType = 'topicSimple'
```

Maybe there is another bug with `topicChapter` and `topicAppendix` values that has been hiding this bug? We don't use the `pdf2` transform so this code is unfamiliar to me.

chrispy-snps: @robander - I think the code to check `$topicType` here might be orphaned code. In `plugins/org.dita.pdf2/xsl/fo/commons.xsl`, there are separate templates for the bookmap structures, and these process the `<title>` elements themselves:

```
$ rg 'xsl:template name=.*processTopic\w' build/tmp/dist/plugins/org.dita.pdf2/xsl/fo
build/tmp/dist/plugins/org.dita.pdf2/xsl/fo/preface.xsl
41:     <xsl:template name="processTopicPreface">

build/tmp/dist/plugins/org.dita.pdf2/xsl/fo/commons.xsl
231:    <xsl:template name="processTopicChapter">
306:    <xsl:template name="processTopicAppendix">
381:  <xsl:template name="processTopicAppendices">
463:    <xsl:template name="processTopicPart">
543:    <xsl:template name="processTopicNotices">
```

Perhaps in earlier days, the titles of all topic types went through this `topic/title` template, but now that's no longer true? If so, we could remove this check completely and redirect the `topic/title` template processing right into `mode="processTopicTitle"`.

src/main/plugins/org.dita.pdf2/xsl/fo/topic.xsl
chrispy-snps: @jelovirt - with the code change to keep the code, do you consider this aspect resolved?

jelovirt (approved):

infotexture:
> No documentation or release notes needed.

We should probably add a migration topic with a list all of the variables that have been removed, in case anyone was using them in custom code.

(For example, like we did the the [3.5 migration topic](https://www.dita-ot.org/dev/topics/migrating-to-3.5#migrating-to-3.5__3.5-legacy-target-removal).)

chrispy-snps:
@infotexture - in this case, I think we are okay. I specifically looked only at variables inside `<xsl:template>` blocks, which can't be referenced outside that scope. Finding unused global variables would indeed be a harder problem, especially because they *could* be used by plugin code even if we stopped using them ourselves.

Issue #3948, opened 2022-06-22
Report deliverable id for each encountered problem when publishing a DITA OT project file


raducoravu:
## Description
I publish a DITA OT project file, various problems are reported in the console, how do I know for which deliverable I obtain which problem? 

## Possible Solution
Maybe besides the [DOT...] problem ID we should also add a [deliverable.id] in the reported message string.

## Potential Alternatives
Not sure if there are any.

xephon2:
One example could be when a keyref points to a missing key because of a profiling filter or a missing key definition map.

raducoravu:
As an example, let's say you publish a DITA OT project file with lots of deliverables, each with a context and profiling setting. And you receive this problem in the console:

     [DOTJ048I] Unable to find key definition for key reference X

You have no idea for what context (map + filters) this problem was encountered.

jelovirt:
It would be difficult to add the deliverable as a context to each message. Would it be enough if we have a info log message that "Processing delivarable {}" and you would need to search the log to find which deliverable it belongs to? At least for 3.7.x.

raducoravu:
@jelovirt yes this would help and would probably be better than adding extra information to each message.

chrispy-snps:
@jelovirt - that would work perfectly fine for us too.

chrispy-snps:
An additional thought as I watch the messages for 60+ deliverables scroll by... it would be useful if the message could indicate the current and total deliverable numbers, something like:

```
Publishing deliverable del-pdf-final-dcug (3 of 60)
```

PR #4097, opened 2023-01-05
#3948 Log the current deliverable when publishing DITA-OT project

DunaMariusCosmin:
## Description
I added an info log with the current deliverable description. 
The log message has the following format: 'Publishing deliverable: "_deliverable.id/deliverable.name_" (x of y)'

## Motivation and Context
This is a fix for the following issue:  #3948

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -84,6 +84,7 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
             "args.input", "input",
             "args.filter", "profiles"
     );
+    private static final String DELIVERABLE_DESC = "deliverable.desc";
 
     /**
      * File that we are using for configuration.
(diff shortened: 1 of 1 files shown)


raducoravu:
Our original intention was to prefer to display the deliverable ID in the console output but I think it makes sense to display also the deliverable human readable name if present.
How about something like?

    Publishing deliverable 1 of 3: Product A Online Help [del-olh-A-map]

so present the deliverable name if present and present also the ID in brackets.
For the set of errors which occur when a certain deliverable is published, we plan to present in Oxygen in the Problems view the error prefixed with the deliverable ID, something like:

     [del-olh-A-map][DOTZZZ] Class value for conref is ....

DunaMariusCosmin:
Thank you for your feedback. 
The initial implementation prefers to present the deliverable ID in that message. The deliverable name was presented only when the ID is not present. This I wanted to represent with the `/` character in my first message.
 
I agree that is better to present both of them in the message. 
I close this pull request and I will create another one that will show the message like this one: 
`Publishing deliverable 1 of 3: Product A Online Help [del-olh-A-map]`

DunaMariusCosmin:
Thanks. It was easier for me to create a new pull request: https://github.com/dita-ot/dita-ot/pull/4100

PR #4111, opened 2023-01-26
Working around a bug in CMD.exe on Windows: lower-case drive letter stored as CWD by cd command

mkarg:
This PR is working around a bug in CMD.exe on Windows: The `cd` command does not only change into a directory, but also stores the drive letter in the case *given by the command issuer*, resulting in CWD sometimes starting with `c:` instead of `C:`. As this happens on all versions on Windows, and as it happens to everybody running DITA-OT on Windows, we should working around this using a simple trick: Not just "absoluting" the CWD using `getAbsolutePath`, but actually finding out the *correct* case by reading the *real* name from the file system using `toRealPath`.

### DIFF
--- a/src/main/java/org/dita/dost/writer/ValidationFilter.java
+++ b/src/main/java/org/dita/dost/writer/ValidationFilter.java
@@ -198,7 +198,7 @@ public final class ValidationFilter extends AbstractXMLFilter {
                     final File p = new File(URLUtils.setQuery(URLUtils.stripFragment(abs), null));
                     try {
                         final File canFile = p.getCanonicalFile();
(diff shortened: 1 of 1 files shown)


chrispy-snps:
@mkarg - commits in a pull request must be committed with the `--signoff` option. Try running the following commands to update the commit:

```
git --amend --signoff
git push --force
```

chrispy-snps:
@mkarg - are you able to rebase this fix and add the signoff flag to your commit?

mkarg:
Sorry for my late response, I missed your original comment from 2023. 😅

PR #4114, opened 2023-01-30
Disable debug logging for non-DITA-OT loggers

### LINKED ISSUE #4106: DITA OT enables full debug logging for all Java packages
The logback.xml file "config/logback.xml" enables by default full debug logging for all Java packages. This means it interferes with Java packages from other plugins as well. I think it should enable full debug logging only on the "org.dita.dost" and "com.idiominc" packages and leave the default debug level as "INFO" instead of DEBUG.
For example we have some indexing Java code in our Oxygen WebHelp plugins, once the plugins are added to a stock DITA OT distribution, the end user will see all our debug messages in the console.

raducoravu:
Leave full debug logging by default only on DITA OT provided Java packages.

## Description
Avoid debugging Java code from third party plugins on full debug.

## Motivation and Context
Fixes #4106

## How Has This Been Tested?
No automatic tests.

## Type of Changes

No doc

## Checklist

### DIFF
--- a/src/main/config/logback.xml
+++ b/src/main/config/logback.xml
@@ -8,12 +8,12 @@
       </pattern-->
     </encoder>
   </appender>
-  <logger name="FOP" level="INFO"/>
-  <logger name="org.apache.fop" level="INFO"/>
   <logger name="org.apache.fop.apps.FopConfParser" level="WARN"/>
-  <logger name="org.apache.xmlgraphics" level="INFO"/>
-  <logger name="org.apache.http" level="INFO"/>
-  <root level="DEBUG">
+  <logger name="org.dita" level="DEBUG"/>
+  <logger name="org.ditang" level="DEBUG"/>
+  <logger name="com.idiominc" level="DEBUG"/>
+  <logger name="com.suite" level="DEBUG"/>
+  <root level="INFO">
     <appender-ref ref="STDOUT" />
   </root>
 </configuration>


jelovirt:
@raducoravu could you add a short example of log messages before and after this change? So what will still be shown and what is not?

raducoravu:
> @raducoravu could you add a short example of log messages before and after this change? So what will still be shown and what is not?

Let's say I run the DITA OT with the "-v" flag.
A DITA OT containing the base plugins will output the same logging information as it did before this change. 
If the DITA OT contains a third party plugin which has Java extensions containing logger.debug messages, the DITA OT used to output by default all these debug messages of the third party plugin to the console. After this change, the logger.debug messages from the third party plugin will no longer be shown in the console output by default. They can be shown if someone explicitly modifies their logback.xml.
The logger.error/warn/info messages from the third party plugin's Java code will still be shown by default in the console output.

raducoravu:
Seeing that the DITA OT code actually uses its own slf4j implementation "DITAOTAntLogger" I assume the way in which it interprets debug levels is not at all related or controlled by the "logback.xml" file to I could probably also remove these lines from the logback.xml file as well:

    <logger name="org.dita" level="DEBUG"/>
    <logger name="org.ditang" level="DEBUG"/>
    <logger name="com.idiominc" level="DEBUG"/>
    <logger name="com.suite" level="DEBUG"/>

jelovirt (approved):

Issue #4105, opened 2023-01-12
Cascading of metadata elements from submap


infotexture:
### Discussed in https://github.com/orgs/dita-ot/discussions/3927

<div type='discussions-op-text'>

<sup>Originally posted by **lief-erickson** May 17, 2022</sup>
What is the expected behavior for element metadata cascading from submaps?

**root-map.ditamap**
```xml
<map id="root-map">
  <topicmeta>
    <category>one</category>
  </topicmeta>
  <mapref href="submap.ditamap"/>
</map>
```

**submap.ditamap**
```xml
<map id="submap">
  <topicmeta>
    <category>two</category>
  </topicmeta>
  <topicref href="topic.dita"/>
</map>
```

**topic.dita**
```xml
<concept>
  <prolog>
    <metadata>
      <category>three</category>
    </metadata>
  </prolog>
  ...
</concept>
```

According to the [specification](http://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part3-all-inclusive/archSpec/base/cascading-of-metadata-elements-from-map-to-map.html#cascading-of-metadata-elements-from-map-to-map), `<category>` should be additive. And it kinda is. Using the `html5` and normalized `dita` transtypes, the output for **topic.dita** in **temp.dir** is:
```xml
<concept>
  <prolog>
    <metadata>
      <category>one</category>
      <category>three</category>
    </metadata>
  </prolog>
  ...
</concept>
```
The `<audience>two</audience>` element is discarded and nowhere to be found. Is this expected behavior?

This alternative construct using the root map produces the same results.

**root-map-alternative.ditamap**
```xml
<map id="root-map-alt">
  <topicmeta>
    <category>one</category>
  </topicmeta>
  <mapref href="submap.ditamap">
    <topicmeta>
      <category>two</category>
    </topicmeta>
  </mapref>
</map>
```

DITA-OT version 3.7.0
</div>

kirkilj:
Same behavior in 4.0.1

chrispy-snps:
@robander - saving submap metadata in a `<foreign>`-specialized `<submap-topicmeta-container>` element was a clever way to stub out for future functionality!

Currently, metadata inside a `<mapref>` is ignored:

```
<mapref href="submap.ditamap">
    <topicmeta>...</topicmeta>
</mapref>
```

[...]

PR #4115, opened 2023-01-31
Save submap metadata inside `<mapref>` elements

chrispy-snps:

## Description
This pull request updates `mapref` processing to also save metadata inside `<mapref>` elements.

## Motivation and Context
`mapref` preprocessing copies submap metadata into a temporary `<submap-topicmeta-container>` element:

```
<map>
   <title>My Map</title>
   <submap>
      <submap-topicmeta>
         <submap-title>My Submap</submap-title>
      </submap-topicmeta>
      <submap-topicmeta-container>
         ....
      </submap-topicmeta-container>
```

Currently metadata in the submap is saved ([Example 4 here](https://github.com/dita-ot/dita-ot/issues/4105#issuecomment-1399501514)):

4105_ex4

but metadata in the `<mapref>` element is not ([Example 3 here](https://github.com/dita-ot/dita-ot/issues/4105#issuecomment-1399501514)):

4105_ex3

This pull request updates `mapref.xsl` to also save the `<mapref>` metadata in Example 3 into a new `<mapref-topicmeta-container>` container (named similarly to `<submap-topicmeta-container>`).

This change to `mapref.xsl` applies to both `preprocess` and `preprocess2`.

This change does not affect any DITA-OT transformation results because both container elements are specialized from `<foreign>` and thus ignored by DITA-OT preprocessing, then deleted during cleanup. This is an incremental step toward a solution for #4105.

## How Has This Been Tested?
All tests passed. In addition, I also tested examples 3 and 4 from [here](https://github.com/dita-ot/dita-ot/issues/4105#issuecomment-1399501514), plus a combined 3+4 testcase.

## Type of Changes

A release notes mention is sufficient.

There should be no compatibility issues; the existing `<submap-topicmeta-container>` behavior is unchanged, and the new additional data is saved in a new `<mapref-topicmeta-container>` container.

## Checklist

No unit test changes were needed.

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
@@ -202,7 +202,9 @@ See the accompanying LICENSE file for applicable license.
               </xsl:choose>
             </xsl:variable>
             <xsl:variable name="targetTitleAndTopicmeta" as="element()*"
-              select="$file/*/*[contains(@class,' topic/title ') or contains(@class,' map/topicmeta ')]"/>
+              select="$file/*/*[contains(@class,' topic/title ') or contains(@class,' map/topicmeta ')]"/>  <!-- submap title and topicmeta -->
+            <xsl:variable name="maprefTopicmeta" as="element()?"
+              select="*[contains(@class,' map/topicmeta ')]"/>  <!-- mapref topicmeta -->
             <xsl:variable name="contents" as="node()*">
               <xsl:choose>
                 <xsl:when test="not(contains($href,'://') or empty($element-id) or $file/*[contains(@class,' map/map ')][@id = $element-id])">
@@ -275,6 +277,7 @@ See the accompanying LICENSE file for applicable license.
                   </xsl:choose>
                 </xsl:with-param>
               </xsl:apply-templates>
+              <xsl:apply-templates select="$maprefTopicmeta" mode="preserve-mapref-topicmeta"/>
               <xsl:apply-templates select="*[contains(@class, ' ditavalref-d/ditavalref ')]"/>
               <xsl:apply-templates select="$contents">
                 <xsl:with-param name="refclass" select="$refclass"/>
@@ -536,6 +539,11 @@ See the accompanying LICENSE file for applicable license.
       <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
       <xsl:apply-templates select="*|processing-instruction()|text()"/>
     </submap-topicmeta-container>
+  </xsl:template>
+  <xsl:template match="*[contains(@class,' map/topicmeta ')]" mode="preserve-mapref-topicmeta">
+    <mapref-topicmeta-container class="+ topic/foreign ditaot-d/mapref-topicmeta-container ">
+      <xsl:apply-templates select="*|processing-instruction()|text()"/>
+    </mapref-topicmeta-container>
   </xsl:template>
 
   <xsl:template match="*" mode="reltable-copy" priority="10">
--- a/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
@@ -530,19 +530,19 @@ See the accompanying LICENSE file for applicable license.
     <submap-topicmeta class="+ map/topicmeta ditaot-d/submap-topicmeta ">
       <submap-title class="+ topic/navtitle ditaot-d/submap-title ">
         <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
-        <xsl:apply-templates select="*|processing-instruction()|text()"/>
+        <xsl:apply-templates/>
       </submap-title>
     </submap-topicmeta>
   </xsl:template>
   <xsl:template match="*[contains(@class,' map/topicmeta ')]" mode="preserve-submap-title-and-topicmeta">
     <submap-topicmeta-container class="+ topic/foreign ditaot-d/submap-topicmeta-container ">
       <xsl:apply-templates select="@*" mode="preserve-submap-attributes"/>
-      <xsl:apply-templates select="*|processing-instruction()|text()"/>
+      <xsl:apply-templates/>
     </submap-topicmeta-container>
   </xsl:template>
   <xsl:template match="*[contains(@class,' map/topicmeta ')]" mode="preserve-mapref-topicmeta">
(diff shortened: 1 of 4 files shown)


chrispy-snps:
I want to take a closer look at how `mapref` and `keyref` interact here.

chrispy-snps:
I rebased the PR from `hotfix/4.0.2` into `develop`.

As mentioned above, this saves the additional data in a new `<submap-topicmeta-container>` element that follows the existing `<submap-topicmeta>` element. The change is entirely internal to preprocessing; both elements are specialized from `<topicmeta>` and thus are deleted during preprocessing cleanup.

src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
```diff
@@ -156,8 +156,14 @@ See the accompanying LICENSE file for applicable license.
                 </xsl:otherwise>
               </xsl:choose>
             </xsl:variable>
-            <xsl:variable name="targetTitleAndTopicmeta" as="element()*"
-              select="$file/*/*[contains(@class,' topic/title ') or contains(@class,' map/topicmeta ')]"/>
+            <xsl:variable name="targetTitleAndTopicmeta" as="element()*">
+              <!-- submap title -->
+              <xsl:sequence select="$file/*/*[contains(@class,' topic/title ')]"/>
+              <!-- metadata in mapref -->
+              <xsl:sequence select="*[contains(@class,' map/topicmeta ')]"/>
```
robander, line 163: This looks pretty simple / straightforward. Couple of questions:
1. Do we have any way to distinguish between the mapref `topicmeta` and the submap `topicmeta`? There could be a case where I want to do something with submap metadata (like I did at one point when doing the original bit here, but *not* want to do that with mapref metadata. I think with this, if we had only mapref metadata, I would think it came from the map. Kind of hard to come up with a reason where that might matter but it's possible - maybe alt-titles on a mapref would throw things off if present in a submap?
2. Looking at where this is processed, we pass in `relative-path` based on the path of the referenced map. I think this means that any relative links in the topicmeta for the mapref, will be updates as if they are located in the submap. If the mapref and submap are in the same directory, no issue, but if the submap is in a subdirectory, metadata like `<source href="../somefile.dita"/>` would fail. Definitely uncommon but something to consider

chrispy-snps: @robander - you raise good points.

For #1, indeed there is currently no way to differentiate between submap (`3:`) and mapref (`4:`) metadata:

```
<submap>
   <submap-topicmeta>
      <submap-title>My Submap</submap-title>
   </submap-topicmeta>
   <submap-topicmeta-container>
      <shortdesc>3:topicmeta/shortdesc</shortdesc>
   </submap-topicmeta-container>
   <submap-topicmeta-container>
      <shortdesc>4:topicmeta/shortdesc</shortdesc>
   </submap-topicmeta-container>
   <topicref href="topic.dita">
      <topicref href="subtopic.dita"/>
   </topicref>
</submap>
```

How about if I put mapref metadata in its own `<submap-mapref-topicmeta-container>` element?

```
<submap>
   <submap-topicmeta>
      <submap-title>My Submap</submap-title>
   </submap-topicmeta>
   <submap-topicmeta-container>
      <shortdesc>3:topicmeta/shortdesc</shortdesc>
   </submap-topicmeta-container>
   <submap-mapref-topicmeta-container>  <!-- note the "-mapref" -->
      <shortdesc>4:topicmeta/shortdesc</shortdesc>
   </submap-mapref-topicmeta-container>
   <topicref href="topic.dita">
      <topicref href="subtopic.dita"/>
   </topicref>
</submap>
```

This allows for filtering, precedence, and aggregation via XPath. I think I like this approach. If you are also agreeable, I'll update the pull request.

For #2, I will need to look at the code some more and try to sensitize it with a testcase. Would you be agreeable to handling that through a follow-up issue and pull request?

robander: For item 1 -- that sounds like a good solution, yes. (Note that the container today is a pseudo-specialization of `foreign`, which should also be the case for `submap-mapref-topicmeta-contaner`). Typing that out, it could also be shortened to `mapref-topicmeta-container`...

For item 2 - yes, it could be a follow-up. This is a pretty extreme edge case, and the same issue with paths probably still exists in some dark corners of the code.

chrispy-snps: @robander - I force-pushed the updated code. I think this is much better, thank you! In this implementation, issue 2 shouldn't be an issue because the current map's file context remains in place when resolving the mapref topicmeta.

I know you had concerns about end-of-line comments [here](https://github.com/dita-ot/dita-ot/pull/4176#discussion_r1177968948), but I see many such existing comments so I decided to sneak some more in...

chrispy-snps:
Switching back to draft status, there are failing tests to be resolved.

robander:
@chrispy-snps looking at this now, we've removed it from the next release because of the failing tests -- if you are able to get the tests worked out in the next week we should be able to get it back in

jelovirt:
The new element would need to added to https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.base/xsl/preprocess/clean-map.xsl

chrispy-snps:
To confirm - these are the tests that we run these days?

```
./gradlew test
./gradlew check
./gradlew e2etest
```

When I suggested [this change](https://github.com/dita-ot/dita-ot/commit/9b15b3bc36e5e25f73de946fa2ab65295d7e1ab1) to the README, I thought that `check` was all-inclusive, but now I suspect that it isn't.

src/main/plugins/org.dita.base/xsl/preprocess/maprefImpl.xsl
```diff
@@ -537,6 +540,11 @@ See the accompanying LICENSE file for applicable license.
       <xsl:apply-templates select="*|processing-instruction()|text()"/>
     </submap-topicmeta-container>
   </xsl:template>
+  <xsl:template match="*[contains(@class,' map/topicmeta ')]" mode="preserve-mapref-topicmeta">
+    <mapref-topicmeta-container class="+ topic/foreign ditaot-d/mapref-topicmeta-container ">
+      <xsl:apply-templates select="*|processing-instruction()|text()"/>
```
jelovirt, line 545: We can just use `node()` here so use default select.

jelovirt (dismissed):

jelovirt:
Added a test and cleaned the code.

jelovirt (approved):

PR #4126, opened 2023-02-19
Dispatch docs release #4125

### LINKED ISSUE #4125: Run additional GitHub Actions on release
## Description

We currently have several GitHub Actions that need to be manually run for each release.

These were originally set up this way because the release artifacts need to be available for certain actions to run.

## Possible Solution

Now that job steps can be configured to wait for another step, we should consider integrating these actions into the main [release workflow](https://github.com/dita-ot/dita-ot/blob/develop/.github/workflows/release.yml):

- https://github.com/dita-ot/docs/blob/develop/.github/workflows/release.yml
- https://github.com/dita-ot/dita-ot/blob/develop/.github/workflows/bump-homebrew-formula.yml
- https://github.com/dita-ot/dita-ot-action/blob/master/.github/workflows/release.yml

  See also:
  - https://github.com/dita-ot/dita-ot-action/issues/5
  - https://github.com/dita-ot/dita-ot-action/blob/master/.github/workflows/chron.yml

jelovirt:
## Description
Dispatch release workflow for docs from core repo release workflow.

## Motivation and Context
Fixes #4125


infotexture (commented):
Looks like this uses the changes from https://github.com/dita-ot/docs/pull/462, but I don't see anything here that bumps the Homebrew formula yet, so wondering if that workflow should be deleted here. 🤔

jelovirt:
@infotexture If you read the `release.yml`, you will find that the Homebrew step has already been added there in earlier commit but the separate Homebrew workflow was not removed in that PR.

infotexture (approved):

PR #4130, opened 2023-02-21
Use Saxon-HE 12

jelovirt:
## Description
Change Saxon dependency to Saxon-HE 12.2.

## Motivation and Context
Update dependency to latest stable Saxon.

[Saxon release notes](https://www.saxonica.com/documentation12/#!changes/all/10-12)


chrispy-snps:
This PR works in linux, but there is a strange `mapref` issue in Windows. Given a map referencing a submap:

```
<!-- map.ditamap -->
<?xml-model href="urn:oasis:names:tc:dita:rng:map.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
<map>
  <title>My Map</title>
  <mapref href="submap.ditamap"/>
</map>

<!-- submap.ditamap -->
<?xml-model href="urn:oasis:names:tc:dita:rng:map.rng" schematypens="http://relaxng.org/ns/structure/1.0"?>
<map/>
```

`mapref` fails in Windows as follows:

```
mapref:
   [mapref] Processing file:/C:/test_saxon/temp-SAXON/submap.ditamap
   [mapref] Processing file:/C:/test_saxon/temp-SAXON/map.ditamap
Error: Failed to run pipeline: Failed to replace temporary file C:\test_saxon\temp-SAXON\submap.ditamap.temp: Unable to delete file: C:\test_saxon\temp-SAXON\submap.ditamap
```

Maybe the newer Saxon is somehow leaving a lingering lock on the file?

This occurs only with the updated-Saxon distribution running in Windows. A development distribution running in Windows completes as expected.

chrispy-snps:
The Windows `mapref` file lock issue still occurs with the updated branch.

raducoravu:
I get the same problems when attempting to delete files on Windows. For example:

      [image-metadata] Failed to read image metadata: Failed to replace file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508122140090/concepts/glossarySepal.dita: Unable to delete file: C:\Users\RADU_C~1\AppData\Local\Temp\temp20230508122140090\concepts\glossarySepal.dita

If I set "store-type=memory" param the publishing works.

jelovirt:
@chrispy-snps @raducoravu I pushed commit that may or may not have effect on this. Could you retry on Windows.

raducoravu:
@jelovirt this still fails for me, as @chrispy-snps  says it starts failing in the maplink stage. As seen below the topic-fragment stage works and it probably calls the same API. 


        topic-fragment:
        [topic-fragment] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/tasks/pruning.dita
        [topic-fragment] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/topics/care.dita
        [topic-fragment] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/tasks/gardenPreparation.dita
        [topic-fragment] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/topics/introduction.dita
        
        chunk-check:
        
        chunk:
            [chunk] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/flowers.ditamap
           [chunk2] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/flowers.ditamap
        
        move-meta-entries:
        [move-meta] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/flowers.ditamap
        [move-meta] Loading stylesheet C:\Users\radu_coravu\Downloads\dita-ot-feature-saxon-11\src\main\plugins\org.dita.base\xsl\preprocess\mappull.xsl
        [move-meta] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/flowers.ditamap
        
        maplink-check:
        
        maplink:
          [maplink] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/topics/care.dita
          [maplink] Failed to insert links: Failed to replace file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/topics/care.dita: Unable to delete file: C:\Users\RADU_C~1\AppData\Local\Temp\temp20230508192807272\topics\care.dita
          [maplink] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/tasks/pruning.dita
          [maplink] Failed to insert links: Failed to replace file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/tasks/pruning.dita: Unable to delete file: C:\Users\RADU_C~1\AppData\Local\Temp\temp20230508192807272\tasks\pruning.dita
          [maplink] Processing file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/tasks/gardenPreparation.dita
          [maplink] Failed to insert links: Failed to replace file:/C:/Users/RADU_C~1/AppData/Local/Temp/temp20230508192807272/tasks/gardenPreparation.dita: Unable to delete file: C:\Users\RADU_C~1\AppData\Local\Temp\temp20230508192807272\tasks\gardenPreparation.dita

raducoravu:
If for example I comment out the "move-meta-entries" target, the processing no longer reports the error in the  "maplink" stage but instead reports it in the [topicpull] stage which comes after it.
It looks like some kind of cache Saxon (maybe that cache of document() functions) which somehow also keeps an input stream opened to each cached resource.

jelovirt:
@raducoravu @chrispy-snps I made some additional changes to close resources. Could you pull the changes and test again?

chrispy-snps:
With the latest changes (0beae8a501c76da9640e31c499e35acff412d615), I get the failure in `mapref`:

`Error: Failed to run pipeline: Failed to replace temporary file C:\test_saxon\temp-SAXON\using_help\keys.ditamap.temp: Unable to delete file: C:\test_saxon\temp-SAXON\using_help\keys.ditamap`

jelovirt:
@chrispy-snps That seems to be a different test source than what you had earlier. I assume the same problem persists with `map.ditamap` and `submap.ditamap` example files?

chrispy-snps:
@jelovirt - sorry for the confusion. Yes, it still happens in the tiny testcase too.

I see you force-pushed the branch. Let me know if you want me to rebuild and retest.

raducoravu:
I can still reproduce the problem with the latest changes, this time during the topicpull stage:
[flowers.zip](https://github.com/dita-ot/dita-ot/files/11429896/flowers.zip)

Sometimes if I remove for example two more topics from the DITA Map I can no longer reproduce the problem. Probably depends on how the garbage collector finalizes and closes dangling file input streams which are no longer referenced.

chrispy-snps:
@raducoravu - on your side, does it occur with the two-file testcase from [this post](https://github.com/dita-ot/dita-ot/pull/4130#issuecomment-1536185064)?

raducoravu:
I overwrote the URL file url handler protocol in the JVM and found where file input streams are created which are not closed afterwards:

	at org.xmlresolver.Resolver.openConnection(Resolver.java:327) 
	at org.xmlresolver.Resolver.resolve(Resolver.java:128)
 	at org.dita.dost.util.DelegatingURIResolver.resolve(DelegatingURIResolver.java:32) 
	at net.sf.saxon.lib.ResourceResolverWrappingURIResolver.resolve(ResourceResolverWrappingURIResolver.java:58)
 	at net.sf.saxon.lib.ResourceRequest.resolve(ResourceRequest.java:131) 
	at net.sf.saxon.functions.DocumentFn.resolveURI(DocumentFn.java:414) 
	at net.sf.saxon.functions.DocumentFn.makeDoc(DocumentFn.java:249) 
	at net.sf.saxon.functions.DocumentFn$DocumentMappingFunction.mapItem(DocumentFn.java:159)

raducoravu:
@jelovirt in org.dita.dost.util.DelegatingURIResolver.resolve(String, String) you need to break the for loop after the first resolution, otherwise you get a resolved stream source with an opened input stream and you ignore it, returning the second resolved stream source.

raducoravu:
@jelovirt in org.dita.dost.util.DelegatingURIResolver.resolve(String, String) you need to break the for loop after the first resolution, otherwise you get a resolved stream source with an opened input stream and you ignore it, returning the second resolved stream source.

raducoravu:
@jelovirt in org.dita.dost.util.DelegatingURIResolver.resolve(String, String) you need to break the for loop after the first resolution, otherwise you get a resolved stream source with an opened input stream and you ignore it, returning the second resolved stream source.

jelovirt:
@raducoravu the for-loop in `DelegatingURIResolver` should explicitly _not_ be broken out of, because it's not what it's doing. But, you're right, it needs to check that the resource is closed.

`DelegatingURIResolver`  is a chain of resolvers, like a reducer function. Maybe it should be renamed to `ReducingURIResovelver`. The problem is that the actual resolvers return streams, when this class would only like to get the system URIs of the resolved resources. Maybe it needs to be redesigned.

jelovirt:
Maybe `DelegatingURIResolver` could be replaced by passing the catalog resolver to store, and if the catalog resolver needs to be overridden with a resolver from Ant, some sort of wrapper to Store could handle that 🤔

raducoravu:
@jelovirt looking at what the DelegatingURIResolver does, maybe you just need to iterate the other way through the list of resolvers as you seem to prefer the last resolver which returns any significant value anyway.
Should I also try to add an issue on the Saxonica issues list about adding an extra resolver feature to disable the creation of input streams in the returned source?

jelovirt:
@raducoravu That will not work. The idea is that I can first use catalog resolver to resolve magic URIs etc. that are mapped to real results, and then use the Store to get e.g. in-memory resources. The Store will then default to normal resolvers. I'm refactoring the whole code so that we no longer need the `DelegatingURIResolver`.

jelovirt:
@raducoravu Looked at the code and it actually can be simplified to reversing the order and short circuiting the resolver result. Are you able to test now again? Ping @chrispy-snps

chrispy-snps:
@jelovirt - this updated PR looks great. On multiple testcases (small and large), the log files and HTML5 output directories match identically between `develop` and this PR.

raducoravu:
@jelovirt the changes make sense for me. They also work well for me.

PR #4142, opened 2023-03-06
Support note type `other` in HTML output

### LINKED ISSUE #4136: Other type notes cannot be identified by their class in HTML5 outputs
## Expected Behavior
I should be able to select a default `<note>` and a `<note type="other" othertype="Custom Safety">` separately in a CSS or XSLT stylesheet.

## Actual Behavior
Both default and other type notes are rendered as `<div class="note note note_note">` in the HTML output.

## Possible Solution
Modify `<xsl:template match="*" mode="process.note.other">` in `org.dita.html5\xsl\topic.xsl`:
```
we can set:
    <xsl:with-param name="type" select="replace(normalize-space(@othertype), ' ', '_')"/>
instead of:
    <xsl:with-param name="type" select="'note'"/>
```
The class will then be: `class="- topic/note note Custom_Safety note_Custom_Safety"`

## Steps to Reproduce
1. Create a ditamap referring a topic with the following content:
```
<topic>
    <title>Notes</title>
    <body>
        <note type="other" othertype="Custom Safety">An othertype note called "Custom Safety".</note>
    </body>
</topic>
```
2. Transform it in HTML5

## Environment
* DITA-OT version: 4.0.2
* Operating system and version: Windows
* How did you run DITA-OT? `dita` command and oXygen
* Transformation type: HTML5

jlacour31:
## Description
Added the othertype attribute value in the <note> class.

## Motivation and Context
Fixes #4136.

## Type of Changes
- Bug fix

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -548,7 +548,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:choose>
       <xsl:when test="@othertype">
         <xsl:apply-templates select="." mode="process.note.common-processing">
-          <xsl:with-param name="type" select="'note'"/>
+          <xsl:with-param name="type" select="replace(normalize-space(@othertype), ' ', '_')"/>
           <xsl:with-param name="title" select="string(@othertype)"/>
         </xsl:apply-templates>
       </xsl:when>
--- a/src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl
+++ b/src/main/plugins/org.dita.xhtml/xsl/xslhtml/dita2htmlImpl.xsl
@@ -645,7 +645,7 @@ See the accompanying LICENSE file for applicable license.
   <xsl:choose>
     <xsl:when test="@othertype">
       <xsl:apply-templates select="." mode="process.note.common-processing">
-        <xsl:with-param name="type" select="'note'"/>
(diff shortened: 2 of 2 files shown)


src/main/plugins/org.dita.html5/xsl/topic.xsl
```diff
@@ -548,7 +548,7 @@ See the accompanying LICENSE file for applicable license.
     <xsl:choose>
       <xsl:when test="@othertype">
         <xsl:apply-templates select="." mode="process.note.common-processing">
-          <xsl:with-param name="type" select="'note'"/>
+          <xsl:with-param name="type" select="replace(normalize-space(@othertype), ' ', '_')"/>
```
jelovirt, line 551: This will not work because the value of `$type` is used to look up a string variable and if the source has `othertype="cat"`, there will likely not be a string variable with ID `Cat`. A plug-in can add one, but if one is not added, the note comes out with out a label or defaults to `note` in every language.

robander, line 551: This assumes that the othertype is a token, rather than a string; many of the cases I've seen where `@othertype` is used are full natural-language titles. So we would end up with something like a type parameter of `My_type_of_note`. That othertype value is already passed in the next `title` parameter.

In the common processing, this will assume that the variable `My_type_of_note` is defined, and throw an error if it is not. Today we print out the string (and I know some tools translate it appropriately), while this change would result in broken output until the variable is defined.

PR #4148, opened 2023-03-12
Add processing mode to default error handler

jelovirt:
## Description
Add processing mode to default SAX error handler. Instead of always treating errors as fatal errors by throwing an exception, the handler can throw the error only in strict processing mode.

## Motivation and Context
Per SAX interface documentation, default behaviour for error should be to ignore them. We don't want to ignore them, but we don't want to treat all errors as fatal errors either. By using processing mode to decide how to handle errors, we allow parsing to continue in lax mode and only log the error.

## Type of Changes

### DIFF
--- a/src/main/java/org/dita/dost/exception/DITAOTXMLErrorHandler.java
+++ b/src/main/java/org/dita/dost/exception/DITAOTXMLErrorHandler.java
@@ -9,7 +9,7 @@
 package org.dita.dost.exception;
 
 import org.dita.dost.log.DITAOTLogger;
-
+import org.dita.dost.util.Configuration;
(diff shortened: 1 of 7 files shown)


raducoravu:
Just to see if I understand what this means for DITA OT users, there are two types of SAX exceptions:
- Exceptions when the processed XML is not wellformed. And such exceptions break the parsing anyway.
- Exceptions when the XML does not adhere to the DTD/XML Schema/RNG. For example the XML has an invalid element somewhere. And in these cases the processing of a topic would stop on the first error and the topic would be ignored but after your fix the processing of the topic might proceed until the end if there is processing-mode=lax?

PR #4157, opened 2023-04-01
Prettier Java formatter

jelovirt:
## Description
Add Java code formatter using [Prettier for Java](https://github.com/jhipster/prettier-java).

## Motivation and Context
Using Prettier to format Java code allows consistent Java format to be used. It can be extended to format other formats too that are supported by Prettier.

Coding conventions and PR contribution guidelines should be changed to mention formatting code before committing.


infotexture:
@jelovirt If we choose this route, we could also use Prettier to also auto-format the DITA, XML, CSS, Sass, and Markdown files in the repo like we do for the docs & website repos.

For sample settings, see

- https://github.com/dita-ot/docs/blob/develop/.prettierrc.json
- https://github.com/dita-ot/docs/blob/develop/package.json

robander (approved):

PR #4165, opened 2023-04-14
Support empty @specializations attribute value

jelovirt:
## Description
Support empty `@specializations` attribute value.

## Motivation and Context
Fix `StringIndexOutOfBoundsException` that is thrown when `@specializations` attribute has empty or all whitespace value.

## How Has This Been Tested?
New unit test.

## Type of Changes

This is a bug that only manifested with DITA 2.0 content so it's unlikely that anyone outside the DITA-OT developers has encountered this. So a mention in release notes is enough.


jelovirt:
In theory this could go to 4.0.x patch release, but this can only be encountered with latest version on LwDITA plug-in that only supports 4.1 or never, so it's more simple to add this to 4.1.

Issue #4196, opened 2023-05-17
gradlew fails with Java versions newer than 17


chrispy-snps:
## Expected Behavior
According to the [DITA-OT prerequisites](https://github.com/dita-ot/dita-ot/#prerequisites-java-17), Java versions newer than version 17 should be usable to build the DITA-OT:

> To build and run DITA-OT, you’ll need Java Development Kit (JDK), version 17 or newer.

## Actual Behavior
When OpenJDK 18 is used, the initial `gradlew` command fails with the following error:

```
* What went wrong:
Could not compile settings file '/home/chrispy/git/dita-ot/settings.gradle'.
> startup failed:
  General error during conversion: Unsupported class file major version 62
```

When OpenJDK 19 is used, a similar error occurs:

```
* What went wrong:
Could not compile settings file '/home/chrispy/git/dita-ot/settings.gradle'.
> startup failed:
  General error during conversion: Unsupported class file major version 63
```

## Possible Solution
If this is a known limitation, the prerequisites statement should be updated.

## Steps to Reproduce
1. Install OpenJDK 19:
   `sudo apt install openjdk-19-jdk`
2. Select version 19:
   `sudo update-alternatives --config java`
3. In a clone of the DITA-OT development repository, build the DITA-OT:
   `rm -rf ~/.gradle && git clean -fdx && ./gradlew`

## Copy of the error message, log file or stack trace
```
$ ./gradlew
Downloading https://services.gradle.org/distributions/gradle-7.2-bin.zip
..........10%...........20%...........30%...........40%...........50%...........60%...........70%...........80%...........90%...........100%

Welcome to Gradle 7.2!

Here are the highlights of this release:
 - Toolchain support for Scala
 - More cache hits when Java source files have platform-specific line endings
 - More resilient remote HTTP build cache behavior

For more details see https://docs.gradle.org/7.2/release-notes.html

Starting a Gradle Daemon, 3 stopped Daemons could not be reused, use --status for details

FAILURE: Build failed with an exception.

* Where:
Settings file '/home/chrispy/git/dita-ot/settings.gradle'

* What went wrong:
Could not compile settings file '/home/chrispy/git/dita-ot/settings.gradle'.
> startup failed:
  General error during conversion: Unsupported class file major version 63

  java.lang.IllegalArgumentException: Unsupported class file major version 63
        at groovyjarjarasm.asm.ClassReader.<init>(ClassReader.java:196)
[...]

PR #4197, opened 2023-05-18
Update Gradle wrapper to 7.6.1

infotexture:
Update Gradle distribution version & regenerate wrapper to pick up recent changes per https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper

    ./gradlew wrapper --gradle-version 7.6.1
    ./gradlew wrapper --gradle-version 7.6.1

https://docs.gradle.org/7.6/release-notes.html

Fixes #4196.

PR #4198, opened 2023-05-18
Update Gradle wrapper to 7.6.1

jelovirt:
Update Gradle wrapper to 7.6.1

https://docs.gradle.org/7.6/release-notes.html

See https://github.com/dita-ot/dita-ot/issues/4196.


infotexture:
Duplicates #4197, which [also updates](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper) `gradle-wrapper.jar`.

> Note that running the wrapper task once will update gradle-wrapper.properties only, but leave the wrapper itself in gradle-wrapper.jar untouched. This is usually fine as new versions of Gradle can be run even with ancient wrapper files. If you nevertheless want all the wrapper files to be completely up-to-date, you’ll need to run the wrapper task a second time.

PR #4211, opened 2023-06-09
Update `outer` parameter descriptions per #4201

infotexture:
This revises the parameter descriptions for `generate.copy.outer` & `outer.control` to reflect the fact that recent toolkit versions also support topics as input resources (not only DITA maps).

Per @jelovirt's suggestion in https://github.com/orgs/dita-ot/discussions/4201#discussioncomment-6062543.


chrispy-snps:
This change looks good to me.

jelovirt (approved):

PR #4217, opened 2023-06-29
Add support for .ditaotrc

jelovirt:
## Description
Add support for local configuration file `.ditaotrc`. Look for the configuration files in 

1. `./.ditaotrc`
2. `$HOME/.ditaotrc`
3. `$DITA_HOME/.ditaotrc`
4. `$DITA_HOME/local.properties`

Files are read in order and combined, with first occurrence of a property taking precedence.

### Example

User has `.ditaotrc` in current directory:

```properties
pdf.formatter=fop
```

The user home directory has `.ditaotrc`:

```properties
pdf.formatter=xep
args.grammar.cache=no
```

DITA-OT is ran as if the user had set

```bash
--pdf.formatter=fop --args.grammar.cache=no
```
on the command line with other arguments.
## Motivation and Context
Extend existing local configuration cover more use cases.

## How Has This Been Tested?
Manual testing.

## Type of Changes

https://www.dita-ot.org/dev/parameters/local-properties-file.html should be rewritten to describe `.ditaotrc`. Support for `local.properties` is retained for backwards compatibility, but users should migrate to `.ditaotrc`.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -89,6 +89,11 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
     "args.filter",
     "profiles"
   );
+  private static final String CONFIGURATION_FILE = ".ditaotrc";
+
+  @Deprecated
+  /** @deprecated since 4.2 */
+  private static final String CONFIGURATION_FILE_OLD = "local.properties";
 
   /**
    * File that we are using for configuration.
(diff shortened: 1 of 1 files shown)


src/main/java/org/dita/dost/invoker/Main.java
```diff
@@ -438,23 +440,35 @@ private void processArgs(final String[] arguments) {
...
-          .entrySet()
-          .stream()
-          .collect(Collectors.toMap(e -> e.getKey().toString(), Map.Entry::getValue));
-      } catch (IOException e) {
-        System.err.println("Failed to read " + localPropertiesFile);
-      }
+    final Map<String, Object> res = new HashMap<>();
+    Stream
+      .of(
+        new File(ditaDir, CONFIGURATION_FILE_OLD),
```
jelovirt, line 446: The priority order here is reversed, because we use `putAll` to overwrite existing values.

robander (approved):

chrispy-snps:
Cool feature, @jelovirt!

raducoravu:
Would this feature cover also this usecase?
https://stackoverflow.com/questions/76573984/override-plugindirs-configuration-property-before-running-dita-install
Overriding the "plugindirs" property from the configuration.properties?

jelovirt:
@raducoravu No, plug-in directory cannot be overridden because the installation process changes the classpath.

raducoravu:
@jelovirt I may not understand this but when running "dita --install" the installation process could read content from a configuration.properties located in the user home folder + from the configuration.properties placed in the DITA OT folder.

PR #3926, opened 2022-05-13
Grammar caching and validation for RNG-based topics

raducoravu:
FIxes for #3661 and #3253
Signed-off-by: Radu Coravu <radu_coravu@sync.ro>

## Description
Added grammar caching for Relax NG based DITA topics and maps. Makes publishing faster for such RNG-based DITA projects.
Added validation for Relax NG based DITA topics and maps. Reports in the console validation errors if the topics are invalid according to the RNG schema.

## Motivation and Context
FIxes for #3661 and #3253

## How Has This Been Tested?
Tested manually with a small RNG-based DITA project on Mac OS.

## Type of Changes
Speed performance improvement and removed limitation which did not present validation errors for RNG-based topics.
- New feature _(non-breaking change which adds functionality)_

## Documentation and Compatibility
Maybe mention in the release notes:
Added grammar caching for Relax NG based DITA topics and maps. Makes publishing faster for such RNG-based DITA projects.
Added validation for Relax NG based DITA topics and maps. Reports in the console validation errors if the topics are invalid according to the RNG schema.

- What documentation changes are needed for this feature?
Not sure if the limitation that RNG-based publishing did not validate was mentioned anywhere in the user's guide.
- Will this change affect backwards compatibility or other users' overrides?
No

## Checklist
- My code follows the code style of this project.
    -  <https://github.com/dita-ot/dita-ot/wiki/Java-Coding-Conventions>
    -  <https://github.com/dita-ot/dita-ot/wiki/XSLT-Coding-Conventions>
    -  <https://github.com/dita-ot/docs/wiki/coding-guidelines>
- I have updated the unit tests to reflect the changes in my code.

PR #4238, opened 2023-07-16
Fix for #4234 XSD-based parsing works + small TC.

raducoravu:
Signed-off-by: Radu Coravu <radu_coravu@sync.ro>

## Description
Fixed default attributes expansion when parsing XML Schema based topics and maps.

## Motivation and Context
Fixes #4234

## How Has This Been Tested?
Automatic test.

## Type of Changes
- Bug fix _(non-breaking change which fixes an issue)_

## Documentation and Compatibility
No
## Checklist
- My code follows the code style of this project.
    -  <https://github.com/dita-ot/dita-ot/wiki/Java-Coding-Conventions>
    -  <https://github.com/dita-ot/dita-ot/wiki/XSLT-Coding-Conventions>
    -  <https://github.com/dita-ot/docs/wiki/coding-guidelines>
- I have updated the unit tests to reflect the changes in my code.

PR #4242, opened 2023-07-17
Fix XSD validation configuration

jelovirt:
## Description
Fix configuration bug to enable XSD validation of input documents.

## Motivation and Context
RELAX NG grammar caching PR #3926 introduced a configuration bug in DITA-OT 4.0 that disabled XSD validation.

## How Has This Been Tested?
Manual testing with source files that use XSD.

## Type of Changes

### DIFF
--- a/src/main/java/org/ditang/relaxng/defaults/RelaxDefaultsParserConfiguration.java
+++ b/src/main/java/org/ditang/relaxng/defaults/RelaxDefaultsParserConfiguration.java
@@ -163,9 +163,8 @@ public class RelaxDefaultsParserConfiguration extends XIncludeAwareParserConfigu
   public void setFeature(String featureId, boolean state) throws XMLConfigurationException {
     if (XERCES_SCHEMA_VALIDATION.equals(featureId)) {
       relaxNGValidation = state;
-    } else {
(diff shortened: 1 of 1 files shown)


src/main/java/org/ditang/relaxng/defaults/RelaxDefaultsParserConfiguration.java
```diff
@@ -163,9 +163,8 @@ protected void insertRelaxDefaultsComponent() {
   public void setFeature(String featureId, boolean state) throws XMLConfigurationException {
     if (XERCES_SCHEMA_VALIDATION.equals(featureId)) {
```
jelovirt, line 164: The RNG support reuses the `http://apache.org/xml/features/validation/schema` XSD schema validation feature and fails to pass the feature value to parent configuration. In the future, we might want to consider using some RNG specific feature URI to make it more explicit.

raducoravu: Yes, I used that as a convenience to avoid defining yet another feature.
My PR here also has an automated test: https://github.com/dita-ot/dita-ot/pull/4238

PR #4150, opened 2023-03-19
Describe how to test potential contributions to the DITA-OT

chrispy-snps:
## Description
Updates the Developers section of the main `README.markdown` file to describe how to test potential contributions,

## Motivation and Context
As a novice contributor, I was unsure of how to test potential contributions to the DITA-OT. I wanted to capture this information.

Questions for those more knowledgeable:

* Are my comment descriptions correct?
* Is the test ordering correct, or should they be run in a different order?
* Are any tests missing?

## How Has This Been Tested?
No tests needed.

## Type of Changes
Small README change only; no code changes.

## Checklist
I tried to follow the existing style of the `README.markdown` file.

PR #4243, opened 2023-07-17
Show development content in README

jelovirt:
## Description
Show development content in README

## Motivation and Context
Hiding development content in README makes is more difficult to find development related info. Our end user documentation is as dita-ot.org, whereas developer documentation is at GitHub. 

Fixes #4150


chrispy-snps:
Please also consider #4150.

Issue #2420, opened 2016-06-17
Can no longer use conref in keydef metadata [DOT 2.3]


raducoravu:
If in my DITA Map I define a key which has a keyword with a conref to another one defined in a topic:

```
     <keydef keys="company_name_full">
      <topicmeta>
        <keywords>
          <keyword
            conref="abc.dita#abc/company_full"/>  
        </keywords>
      </topicmeta>
    </keydef>
```

and in some topic I use the key:

```
<ph keyref="company_name_full"/>
```

it no longer gets expanded. Using DITA OT 1.8 the key got expanded.

raducoravu:
Any hint in the specs about what stage should be applied first? Conref or keyref?
In what issue was the order of these stages changed?

infotexture:
@raducoravu: This was changed for 2.0 with #1631. From the [Release Notes](http://www.dita-ot.org/2.0/readme/changes/rel2.0.html):

> The order of preprocessing stages has been optimized to move keyref processing before conref processing. This provides significant performance gains for source that uses multiple key references within reusable content.

robander:
> Any hint in the specs about what stage should be applied first? Conref or keyref?

Because no order was specified way back in 1.0 the spec has had great difficulty giving a mandated processing order for any of the features. Even newer features have trouble, because giving an order like "Between existing features A and B" establishes an order for A and B. An additional complication is that while I think the spec implies you should evaluate keyref before conref, that doesn't rule out "evaluating" as a way of simply moving the conref from the map context into the topic.

jelovirt:
I think the cause may be that the key resolution module fails to verify a resolved topic has conrefs.

drmacro:
My analysis is that the branch filtering and key scope features effectively require a specific order of processing and that implementations should do preprocessing in a "filtering aware" way. I captured this analysis in a dita-ot wiki page here: https://github.com/dita-ot/dita-ot/wiki/DITA-1.3-Preprocessing-Architecture-proposal

jelovirt:
This is something that would require processing keyref and conref in a loop when you keep on resolving them until no keyref or conref remains.

drmacro:
Resolution of direct conrefs should be done before any key-based processing is performed, otherwise there is no way to ensure you get the right answer.

Once key spaces are constructed it should then be possible to resolve key references in any context, including in the content of key defining-topicrefs.

raducoravu:
👍 We got one more report here:

[...]

PR #4258, opened 2023-08-04
Fix conref in keydef value in preprocess2

jelovirt:
## Description
Fix the case where a map uses a topic as a conref target but the topic is not referenced with a topicref. 

## Motivation and Context
Partial fix for #2420.

## How Has This Been Tested?
Manual testing.

## Type of Changes


raducoravu:
I understand, so with preprocess2 if there is a conref to a topic from a DITA Map, the topic will be taken into account without being the need to also have a topicref to it in the DITA Map.

Issue #4254, opened 2023-07-31
HTML5: Teletype output contains inline style attributes


Jon-Hill-i2:
## Expected Behavior

When MDITA source contains text wrapped in backticks, the generated HTML output should contain either `<span>` elements with appropriate class attributes, or `<tt>` elements as described in the release notes.

## Actual Behavior

In the situation described above, the generated HTML contains inline style attributes, so that (for example)

```
`backticks`
```

in the MDITA source becomes

```
<span class="ph tt" style="font-family: monospace">backticks</span>
```

in the HTML output. This appears to contradict the explanation in the release notes, which goes like this:

> Standard Markdown syntax can now be used to indicate a span of code by wrapping it with backtick quotes (`). In Markdown DITA input, these constructs are converted to DITA \<codeph\> elements on import, and rendered as \<code\> elements in HTML output. In MDITA input, this syntax is converted to teletype \<tt\> elements to align with LwDITA, which supports teletype highlighting in XDITA.

## Possible Solution

Unknown

## Steps to Reproduce

The attached [`demo.zip`](https://github.com/dita-ot/dita-ot/files/12215053/demo.zip) file contains source and output that demonstrates the problem.

## Copy of the error message, log file or stack trace

N/A

## Environment

* DITA-OT version: 4.1.1
* Operating system and version: Windows 10 version 22H2
* How did you run DITA-OT? `dita` command
* Transformation type: HTML5

jelovirt:
> In MDITA input, this syntax is converted to teletype <tt> elements to align with LwDITA

This means when MDITA is parsed, backtick code spans are parsed as `<tt>` elements. This means the transtype specific code will get a `<tt>` for transformation. In the case of HTML5 transtype, `<tt>` is converted into a HTML5 `<span>` element with classes. So there's no bug here, it's how HTML5 plug-in works. This is not related to MDITA, the same will happen with DITA input documents.

jelovirt:
HTML5 plug-in doesn't generate HTML5 `<tt>` element for DITA `<tt>` element because the [HTML5 `<tt>` element is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt).

Jon-Hill-i2:
> > In MDITA input, this syntax is converted to teletype `<tt>` elements to align with LwDITA
> 
[...]

PR #4267, opened 2023-08-15
Move teletype styling to Sass partial

infotexture:
Per https://github.com/dita-ot/dita-ot/issues/4254#issuecomment-1679748417

- [x] Add teletype styling to Sass partial for highlighting domain
- [x] Remove hard-coded style attribute from `<tt>` element output

@jelovirt Can we safely remove [Lines 39-40 from `hi-d.xsl`](https://github.com/dita-ot/dita-ot/blob/feature/teletype-style/src/main/plugins/org.dita.html5/xsl/hi-d.xsl#L39-L40), or do we need to modify that to preserve DITAVAL styles?

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/domains/_hi-d.scss
+++ b/src/main/plugins/org.dita.html5/sass/domains/_hi-d.scss
@@ -26,6 +26,7 @@
 }
 
 .tt {
+  font-family: monospace;
 }
 
 .u {
--- a/src/main/plugins/org.dita.html5/xsl/hi-d.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/hi-d.xsl
@@ -36,8 +36,6 @@ See the accompanying LICENSE file for applicable license.
     <xsl:template match="*[contains(@class,' hi-d/tt ')]" name="topic.hi-d.tt">
       <span>
         <xsl:call-template name="commonattributes"/>
-          <!-- Combine TT style with style from ditaval, if present -->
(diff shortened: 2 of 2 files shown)


chrispy-snps:
> Can we safely remove [Lines 39-40 from `hi-d.xsl`](https://github.com/dita-ot/dita-ot/blob/feature/teletype-style/src/main/plugins/org.dita.html5/xsl/hi-d.xsl#L39-L40), or do we need to modify that to preserve DITAVAL styles?

@infotexture - I think those lines can be removed because the `commonattributes` template ([link](https://github.com/dita-ot/dita-ot/blob/cf5995088dc113fdf644af7607d6eec9e1558c89/src/main/plugins/org.dita.html5/xsl/topic.xsl#L1789)) indeed applies the DITAVAL styling:

```
  <xsl:template match="@* | node()" mode="commonattributes">
...omittted...
    <xsl:apply-templates select="*[contains(@class, ' ditaot-d/ditaval-startprop ')]/@style" mode="add-ditaval-style"/>
```

but @jelovirt can give official confirmation.

infotexture:
@chrispy-snps Thanks for confirming. I added that change in 898e84b & force-pushed.

This should be ready for @jelovirt’s review.

infotexture:
@jelovirt Targeting `develop` here for now, but we could add it to the next maintenance release if you think that's better.

jelovirt (approved):

Issue #4269, opened 2023-08-20
Footnote in <title>, footnote text appears in <xref>


markgif:
## Expected Behavior

An xref to a title element should show the content of <title> without the <fn> text. When published, the <fn> text should only appear in a footnote.

## Actual Behavior

The <fn> text appears in the cross reference.

## Possible Solution

I fixed this in version 2.5.4, so I will refrain from doing a PR. But I tested it with DITA OT 4.0.2 and the bug is still there. Here is my fix in OT 2.5.4 in links.xsl (see "markg" changes):

    <xsl:template match="*[contains(@class,' topic/xref ')]" name="topic.xref">
    <xsl:variable name="destination" select="opentopic-func:getDestinationId(@href)"/>
    <xsl:variable name="element" select="key('key_anchor',$destination, $root)[1]"/>
      
    
    <xsl:variable name="element2">
      <xsl:apply-templates select="$element" mode="removefn"/>
    </xsl:variable>

    <xsl:variable name="referenceTitle" as="node()*">
      <xsl:apply-templates select="." mode="insertReferenceTitle">
        <xsl:with-param name="href" select="@href"/>
        <xsl:with-param name="titlePrefix" select="''"/>
        <xsl:with-param name="destination" select="$destination"/>

        
        <xsl:with-param name="element" select="$element2"/>
      </xsl:apply-templates>
    </xsl:variable>
    [...]
    </xsl:template

    
    <xsl:template match="fn" mode="removefn" />
    <xsl:template match='@* | node()' mode="removefn">
      <xsl:copy>
        <xsl:apply-templates select='@* | node()' mode="removefn"/>
      </xsl:copy>
    </xsl:template>

## Steps to Reproduce

1. Add a footnote to a <title> element like this:
    &lt;fig id="figureID">
         &lt;title>Washing the car &lt;ph>&lt;fn>HEY( fig footnote )&lt;/fn>&lt;/ph>&lt;/title>
         &lt;image href="../image/carwash.jpg" alt="washing the car" height="171" width="249"/>
    &lt;/fig>
2. Do a cross ref to the <title> element:
&lt;p> For more information, see &lt;xref href="washingthecar.dita#washcar/figureID" />&lt;/p>

3. Run a PDF (this was only tested on pdf output), and the footnote text appears in the cross reference like this:
For more information, see Washing the car HEY( fig footnote )

[garage.zip](https://github.com/dita-ot/dita-ot/files/12387358/garage.zip)

## Environment

[...]

PR #4276, opened 2023-08-28
ignore `<fn>` elements when computing link text for a target element

chrispy-snps:
## Description
For a link to a target element that has no explicit link text, the DITA-OT computes default link text from the target element. This link text computation should ignore any `<fn>` elements it encounters, as the footnote numbering in the target document has no meaning when the source document is different.

Note the following:

* For links to self-titled elements, the `<title>` element does not directly allow `<fn>` elements. But, they can "sneak into" the title when contained in other elements (such as `<ph>`).
  * This also applies to links to `<dt>` elements (and possibly others).
* For links to other elements, the `<fn>` element might exist in whatever target element content is used for link text computation.

## Motivation and Context
Fixes the link text aspect of #4269.

## How Has This Been Tested?
I ran the automated tests, plus the two testcases in #4269.

## Type of Changes

In `topicpull`, `<fn>` elements are ignored during link text computation.

In PDF transformations, `<fn>` elements are ignored during link text recomputation.

No documentation update is needed. The fix should not affect any existing flows (unless they rely on this questionable behavior).

A release notes entry might be as follows:

> For a link to a target element that has no explicit link text, the DITA-OT computes default link text from the target element. This link text computation now ignores any `<fn>` elements it encounters, as the footnote numbering convention in the target document has no meaning when the source document is different.

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
+++ b/src/main/plugins/org.dita.base/xsl/preprocess/topicpullImpl.xsl
@@ -1225,8 +1225,8 @@ mode="topicpull:figure-linktext" and mode="topicpull:table-linktext"
     </xsl:choose>
   </xsl:template>
 
+  <!-- use common "dita-ot:text-only" templates for link text computation -->
   <xsl:template match="*|text()|processing-instruction()" mode="text-only">
-    <!-- Redirect to common dita-ot module -->
     <xsl:apply-templates select="." mode="dita-ot:text-only"/>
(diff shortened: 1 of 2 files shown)


jelovirt:
@chrispy-snps Could you rebase/cherry-pick this PR on `hotfix/4.1.2`

chrispy-snps:
@jelovirt - PR rebased/cherry-picked. Thanks!

jelovirt (approved):

Issue #4094, opened 2023-01-03
Cascaded DITAVAL filtering is not properly applied to multiple-condition profiled content


chrispy-snps:
## Expected Behavior
This is a follow-up issue from discussion in #4085.

In [3.5.4.1 `<ditavalref>`](https://docs.oasis-open.org/dita/dita/v1.3/os/part2-tech-content/langRef/base/ditavalref.html), I interpret the paragraph about cascaded `<ditavalref>` elements as follows:

* If a "condition" (attribute name and value) is excluded at a higher level, that condition cannot be reenabled for inclusion at a lower level.
* Processing-wise, `<ditavalref>` is applied from top to bottom, with higher levels removing conditions from consideration by lower levels.
* It is the exclusions, not the inclusions, that combine in an OR-like way from top to bottom.

As an industrial DITA user, I would expect this behavior so that sensitive content excluded from a higher level cannot unexpectedly be re-included at a lower level, such as when reusing common submaps. (However, we have yet to hit this scenario in practice, so there is no urgency from us.)

## Actual Behavior
Content separately profiled for A or for B is filtered as expected. But content profiled for "A or B" is published, which is unexpected.

This occurs for cascaded branch filtering:

FIGURE

It also occurs for global filtering followed by branch filtering:

FIGURE

## Possible Solution
In addition to removing content elements that are excluded, perhaps the global and branch filtering code can also remove the excluded keywords themselves from attribute values when only some keywords in the value are excluded. This would more accurately reflect the propagation of exclusions described in section 3.5.4.1.

## Steps to Reproduce
1. Download the following testcase:
   TESTCASE
2. Run the following commands:
   `dita --project cascaded-branch/project.xml`
   `dita --project global-then-branch/project.xml`

## Environment
* DITA-OT version: 4.0
* Operating system and version: Linux
* How did you run DITA-OT? `dita` command
* Transformation type: HTML5

kirkilj:
This would be a huge problem for us, since several of our projects use composable ditavals to produce a composite condition set. 

Btw, We've seen some topicmeta elements, specifically <category>, not cascading as specified in the DITA spec, and we're in the process of validating our findings and coming up with test content we can use to write a bug report if needed. I'm mentioning this in case there's a wider issue with cascading, whether it be ditavals or metadata elements.

chrispy-snps:
@kirkilj - is your content profiled for multiple (space-separated) condition values?

lief-erickson:
[...]

PR #4277, opened 2023-09-01
Treat ungrouped profiling values as group with attribute name

### LINKED ISSUE #4256: DITAVAL with conditions for group default values
## Expected Behavior
If we have a conditional attribute group such as `groupA(a b)` and we set the group to **exclude** in a ditaval file using `<prop action="exclude" att="audience" val="groupA"/>` it should set the default values for `a` and `b` to **exclude** and the entire group should evaluate to **exclude**, according to [DITA Specification](https://docs.oasis-open.org/dita/dita/v1.3/os/part1-base/langRef/ditaval/ditaval-val.html).

## Actual Behavior
The default values of `a` and `b` are not changed and the group evaluates to **include**.

## Steps to Reproduce

1. Download the following testcase: [test-condition.zip](https://github.com/dita-ot/dita-ot/files/12248441/test-condition.zip)
2. Publish main_map.ditamap to PDF

## Environment

* DITA-OT version: DITA-OT version:4.1
  _(macOS Big Sur Version 11.6)_
* How did you run DITA-OT?
  _(oXygen)_
* Transformation type:
  _(PDF)_

jelovirt:
## Description
Improve profiling code to match DITA spec.

## Motivation and Context
Fixes  #4256
## How Has This Been Tested?
Improved unit tests

## Type of Changes

TBD


chrispy-snps:
@jelovirt - would your code changes in this area also be able to resolve #4094?

PR #4284, opened 2023-09-22
Fix metadata cascade

### LINKED ISSUE #4275: Different topicmeta/navtitle should keep their value even if the target file is the same
Hi, maybe this goes against the spec but I didn't find anything about it.

## Expected Behavior

In a ditamap, different `topicmeta/navtitle` should keep their value even if the target file (`@href`) is the same.

expected-behavior

## Actual Behavior

Transformation (here HTML5, but I think it's related to preprocess) keeps the string of the first `navtitle` with the same `@href`

actual-behavior

## Steps to Reproduce

Create a ditamap with several `topicmeta/navtitle` referencing the same file. Then, run a transformation (HTML5, PDF).

```xml
<!DOCTYPE map PUBLIC "-//OASIS//DTD DITA Map//EN" "map.dtd">
<map>
    <title>Test map</title>
    <topicref href="test_navtitle.dita" locktitle="yes">
        <topicmeta>
            <navtitle>Navtitle 1</navtitle>
        </topicmeta>
    </topicref>
    <topicref href="test_navtitle.dita" locktitle="yes">
        <topicmeta>
            <navtitle>Navtitle 2 (using same href as previous)</navtitle>
        </topicmeta>
    </topicref>
    <topicref href="test_navtitle2.dita" locktitle="yes">
        <topicmeta>
            <navtitle>Navtitle 3 (using a different href)</navtitle>
        </topicmeta>
    </topicref>
</map>
```

## Environment

* DITA-OT version: 4.0.2
* Operating system and version: MacOS 13.2.1
* How did you run DITA-OT? oXygen 25.1
* Transformation type: HTML5, PDF

jelovirt:
## Description
Fix metadata cascading within map and collection of metadata per topic when the same resource is referenced multiple times. On duplicate topicrefs metadata is no longer cascaded from first to second occurrence in the map

## Motivation and Context
Fixes #4275

This is a very old bug, dating back to DITA-OT 1.x or older. The cause is premature optimization where the original code was written to cascade metadata within map and also collect metadata to be pushed to topics. This code worked when each topic was only referenced once. When there were duplicates, only the first topicref got the correct cascaded metadata; the second topicref got the correct cascaded metadata and incorrectly the metadata from the first topicref.

The code was fixed by separating metadata cascading within map from the collection of metadata to push to topics.

## How Has This Been Tested?

Improved unit tests.

##Type of Changes

-    Bug fix (non-breaking change which fixes an issue)


robander (approved):
Makes sense, good fix for longstanding issue

PR #4288, opened 2023-10-01
Update FOP to 2.9

### LINKED ISSUE #4272: Update Apache FOP to version 2.9
A new FOP version is available: https://xmlgraphics.apache.org/fop/2.9/releaseNotes_2.9.html
With all its satellite libraries:
 - Apache XML Graphics Commons
 - Apache FOP-PDF-Images
 - Apache Batik (1.17)

Also in `org.dita.pdf2.fop/lib` there's an extra fop.jar which seems useless.

jelovirt:
## Description
Update FOP and related libraries to version 2.9.

## Motivation and Context
Fixes #4272

## How Has This Been Tested?
Existing tests pass and manual testing.

## Type of Changes

### DIFF
--- a/src/main/config/logback.xml
+++ b/src/main/config/logback.xml
@@ -9,6 +9,7 @@
     </encoder>
   </appender>
   <logger name="org.apache.fop.apps.FopConfParser" level="WARN"/>
+  <logger name="org.apache.fop.fonts.truetype.TTFFile" level="INFO"/>
   <logger name="org.dita" level="DEBUG"/>
   <logger name="org.ditang" level="DEBUG"/>
   <logger name="com.idiominc" level="DEBUG"/>
--- a/src/main/plugins/org.dita.pdf2.fop/build.gradle
+++ b/src/main/plugins/org.dita.pdf2.fop/build.gradle
@@ -12,19 +12,13 @@ repositories {
     }
 }
 dependencies {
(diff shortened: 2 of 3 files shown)


src/main/config/logback.xml
raducoravu, line 1: I do not understand what this line is for.
There are still errors like:

     [ERROR] Unsupported TrueType font: No Unicode or Symbol cmap table not present. Aborting

reported in the console when running the DITA OT on my mac. If you want such lines to disappear which makes sense you could maybe turn it off:

     <logger name="org.apache.fop.fonts.truetype.TTFFile" level="OFF"/>

raducoravu (changes requested):

src/main/config/logback.xml
jelovirt: @raducoravu if you remove the configuration and run `dita` CLI, you notice what messages the configuration removes. That error should not be ignored.

raducoravu: I tried this but I do not notice what messages the configuration removes. There are usually lots of messages when fonts are loaded and cached by Apache FOP on the first run. Once they are cached on subsequent runs I do not have any extra message that the line you added would hide.

raducoravu (approved):

src/main/config/logback.xml
jelovirt: I can't reproduce the log messages now because I don't remember how I got them.

PR #4297, opened 2023-10-08
Refactor subject scheme reader

jelovirt:
## Description
Refactor `SubjectSchemeReader` for code readability and more modern Java.

Support multiple keys in a single `@keys` attribute. For example

```xml
<subjectdef keys="linux windows"/>
```

## Motivation and Context
Refactor the code to make it easier to fix bugs and maintain the code in the future.

## How Has This Been Tested?
Existing test and new tests.
## Type of Changes

Document that we now support multiple keys in a single `@keys` attribute on `<subjectdef>` element. There are no examples of this use case in the DITA spec nor is it explicitly stated that this is supported. But because subject definitions use `@keys` attribute, multiple values per single attribute are supported.

### DIFF
--- a/src/main/java/org/dita/dost/reader/SubjectSchemeReader.java
+++ b/src/main/java/org/dita/dost/reader/SubjectSchemeReader.java
@@ -180,20 +180,15 @@ public class SubjectSchemeReader {
   }
 
   public void loadSubjectScheme(final Element schemeRoot) {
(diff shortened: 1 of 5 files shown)


raducoravu:
@jelovirt thanks for the work to bringing the subject scheme processing code up to date. We started using them more and more, in Oxygen 26 they are used in conjunction with classification maps to define category filters in our WebHelp output 
Screenshot 2023-10-09 at 09 24 53

PR #4302, opened 2023-10-13
Move map specific reader code to map module

jelovirt:
## Description
Refactor subject scheme code for preprocess2 to support all features of subject scheme that preprocess supports. This changes how preprocess2 does filtering. Before preprocess2 map filtering was lazy, meaning it happened at the end of preprocess2 map step. This changes the map filtering to be early, after initial map parsing and mapref merge. It's not as early as in preprocess2 topic processing, because filtering requires reading subject schemes and those are part of the map.

## Motivation and Context
Support all subject scheme features in preprocess2 that are supported by preprocess.

## How Has This Been Tested?
Existing and new tests.

## Type of Changes

https://www.dita-ot.org/dev/reference/map-first-preprocessing could mention that since DITA-OT 4.2 preprocess2 supports more subject scheme features.

### DIFF
--- a/src/main/java/org/dita/dost/module/FilterModule.java
+++ b/src/main/java/org/dita/dost/module/FilterModule.java
@@ -9,7 +9,7 @@ package org.dita.dost.module;
 
 import static org.dita.dost.util.Configuration.printTranstype;
 import static org.dita.dost.util.Constants.*;
-import static org.dita.dost.util.FilterUtils.*;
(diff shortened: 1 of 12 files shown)


infotexture:
> https://www.dita-ot.org/dev/reference/map-first-preprocessing could mention that since DITA-OT 4.2 preprocess2 supports more subject scheme features.

@jelovirt While it's fresh in your mind: could you provide details on _which_ features?

PR #4312, opened 2023-10-23
Use preprocess2 in XHTML transtype

jelovirt:
## Description
Use preprocess2 in XHTML transtype instead of old preprocess.

## Motivation and Context
Preprocess2 is the way of the future and has been proven to work in PDF2.

## How Has This Been Tested?
Existing test.

## Type of Changes

TBD


jelovirt:
IMO it may be better to leave XHTML transtype with the all at once preprocessing.

PR #4357, opened 2024-01-12
Simplify `docsrc/` exclude patterns in dist builds

infotexture:
Hidden files & folders were previously excluded from the distribution package by name, which allowed new dotfiles to creep in if not explicitly excluded in the Gradle build.

(This happened with `.git-blame-ignore-revs`, which shipped with 4.1.1 & 4.1.2 after it was moved to the repo root in dita-ot/docs@dd111ba.)

This replaces the existing path references with the same `.*` pattern that is used 5 lines above to exclude all dotfiles and hidden folders from the core repo.

Tested with local distribution build:

- [x] All previously excluded `docsrc/` resources are still excluded ✅ 
- [x] The `.git-blame-ignore-revs` file is also excluded ✅


infotexture:
@jelovirt The `platform.ditaval` is still included in the distribution package despite [this line](https://github.com/dita-ot/dita-ot/pull/4357/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R270), which tries to exclude it:

```
exclude "docsrc/platform.ditaval"
```

I'm guessing this is because the file is generated by the docs Gradle build [L64-L83](https://github.com/dita-ot/docs/blob/develop/build.gradle#L64-L83) — **after** — the `copyDistTemp` task, where we're excluding things.

In that case, we could probably remove this exclusion (and the next line `exclude "docsrc/temp"` — which also doesn't exist at this point) in favor of the `cleanGenerateDocs` task that runs afterward.  

Does that sound right?

build.gradle
```diff
@@ -336,6 +333,9 @@ generateDocs.onlyIf { skipDocs() }
 task cleanGenerateDocs(type: Delete) {
     delete "${distTempDir}/temp"
     delete "${distTempDir}/docsrc/.gradle"
+    delete "${distTempDir}/docsrc/build"
+    delete "${distTempDir}/docsrc/node_modules"
+    delete "${distTempDir}/docsrc/out"
```
infotexture, line 338: @jelovirt Do we need these at all? 🤔 

These shouldn't exist in the `docsrc` submodule when the build runs, since they're generated at build time, and not versioned in the docs repo.

You added the `node_modules` exclusion in 5ad7508 so I'm guessing there was a reason, but I'm not sure it's necessary.

But I suppose it also does no harm to explicitly ensure these are never included.

Thoughts?

jelovirt: They may exist when dist is run from a Git clone that has been used for development, i.e. developer workstation.

jelovirt (approved):

PR #4303, opened 2023-10-14
use `@class="compact"` instead of `@compact="yes"` for HTML outputs

chrispy-snps:
## Description
Use `@class="... compact"`  instead of `@compact="yes"` for the XHTML and HTML5 outputs. The relevant CSS rules are updated accordingly.

Although the deprecation applies only to HTML5, the XHTML transformation was also updated for consistency.

## Motivation and Context
Fixes #4298.

## How Has This Been Tested?
I ran the testcase in #4298 for both XHTML and HTML5 and confirmed that the browser results were visually identical. I also ran `./gradlew check`.

## Type of Changes
- Bug fix _(non-breaking change which fixes an issue)_

## Documentation and Compatibility
No documentation change is needed for this fix.

A release notes entry could be something like:

> The DITA standard defines a `@compact` attribute for list elements. Previously, this attribute was published to XHTML and HTML5 outputs as an HTML `@compact` attribute. However, the `@compact` attribute was deprecated in HTML4 (over 20 years ago). Now, DITA `@compact` attributes are published as `@class="compact"` keywords in both the XHTML and HTML5 outputs. The default CSS rules have been updated accordingly. Custom CSS rules referencing the `@compact` attribute should be updated.

## Checklist
- My code follows the code style of this project.
    -  <https://github.com/dita-ot/dita-ot/wiki/XSLT-Coding-Conventions>

PR #4358, opened 2024-01-12
Deprecate compact attributes with Sass @warn rules

infotexture:
## Description

@chrispy-snps submitted #4303 assuming it would be merged into a pending 4.1.3 hotfix, but since that's no longer planned, this updates the code comments to reflect the version in which the changes will be included per https://github.com/dita-ot/dita-ot/pull/4303#issuecomment-1872150926.

While we're at it, we may as well implement the deprecation notices as [Sass warnings](https://sass-lang.com/documentation/at-rules/warn/) as we did for 2.3, so they're not forgotten and to remind us to remove the deprecated rules in a future version.

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/_lists.scss
+++ b/src/main/plugins/org.dita.html5/sass/_lists.scss
@@ -16,7 +16,7 @@ ul.simple {
   font-weight: bold;
 }
 
-/* Use CSS to expand lists with @compact="no" */
+/* Use CSS to expand compact lists */
 .dltermexpand {
   font-weight: bold;
   margin-top: 1em;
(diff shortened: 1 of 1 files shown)


infotexture (commented):
@jelovirt @robander Speaking of deprecated CSS rules: 

We have these Sass warnings in place for various table-related rules that have been deprecated since 2.3:

https://github.com/dita-ot/dita-ot/blob/develop/src/main/plugins/org.dita.html5/sass/_tables.scss#L21

Should we remove those in the upcoming minor release, or wait for the next major version?

jelovirt (approved):

Issue #4086, opened 2022-12-22
Go through all DITA OT information messages, avoid reporting some and increase severity of others


raducoravu:
In general the DITA OT should in my opinion not show information messages and if there are information messages which are very useful, like a missing keyref, it could be presented as a warning maybe.
Because I'm looking in "/DITA-OT/config/messages.xml" at "INFO" messages, some of them are like this:

     <message id="DOTX029I" type="INFO">
       <reason>The type attribute on a %1 element was set to %3, but the reference is to a more specific %4 %2. This may cause your links to sort incorrectly in the output.</reason>
       <response/>
      </message>

which in my opinion should be a warning.

Other messages make up part of the usual functionality:

     <message id="DOTX077I" type="INFO">
        <reason>Resolving content references results in duplicate ID '%1'.</reason>
        <response>Rewriting resolved version to '%2'.</response>
     </message>

and in my opinion should not be output at all because end user start asking if this is a problem and if not how they can avoid such messages from being presented.

Or an info message like this:

     <message id="DOTJ047I" type="INFO">
       <reason>Unable to find key definition for key reference "%1" in root scope.</reason>
       <response>The href attribute may be used as fallback if it exists</response>
     </message>

which should be a warning.

raducoravu:
<table>
         <thead>
            <tr>
               <th>Description</th>
               <th>XPath Location</th>
               <th>File</th>
               <th>Location</th>
            </tr>
         </thead>
         <tbody>
            <tr>
               <td>Duplicate condition in filter file for rule '%1'.</td>
               <td>/messages[1]/message[11]/reason[1]</td>
               <td>DITA-OT3.x/config/messages.xml</td>
               <td>
                  <p>Start line 64:5</p>
                  <p>End line 64:5</p>
               </td>
            </tr>
            <tr>
               <td>Log file '%1' was generated successfully in directory '%2'.</td>
               <td>/messages[1]/message[18]/reason[1]</td>
               <td>DITA-OT3.x/config/messages.xml</td>
               <td>
                  <p>Start line 100:5</p>
                  <p>End line 100:5</p>
               </td>
            </tr>
            <tr>
               <td>No 'domains' attribute was found for element '&lt;%1&gt;'.</td>
               <td>/messages[1]/message[27]/reason[1]</td>
               <td>DITA-OT3.x/config/messages.xml</td>
               <td>
                  <p>Start line 147:5</p>
                  <p>End line 147:5</p>
               </td>
            </tr>
[...]

Issue #4068, opened 2022-11-30
DOTJ047I (unresolved key reference) should not be issued for references into declared peer maps


chrispy-snps:
## Expected Behavior
If there are scoped keyrefs into a declared, keyscoped peer map, then a new **information** message (not DOTJ047I) should be issued because the peer map could possibly define those keys:

![DOTJ047I_expected](https://user-images.githubusercontent.com/50950969/204831617-afb4c1d3-353e-48f7-9bff-371c2d6c48fb.svg)

## Actual Behavior
An unresolved key message (DOTJ047I) is displayed for correctly defined keyrefs into peer maps, making it different to separate correct references from bad references (undefined keyscopes, missing map files, etc.):

![DOTJ047I_actual](https://user-images.githubusercontent.com/50950969/204831662-6ef8a26a-ce41-45b1-bc6d-c81097fd95d8.svg)

## Possible Solution
Implement a new check that detects whether the keyref's scope reaches into any declared keyscoped peer map. Subsequent nested scope levels "inside" the peer map can be ignored, as we don't know what scope structures exist inside that map.

Also implement #3950 to increase the severity of bad keyrefs with bad scopes.

## Steps to Reproduce
1. Download the following testcase:
   [4068.zip](https://github.com/dita-ot/dita-ot/files/11968824/4068.zip)
2. Run the following command:
   `dita --project project.xml -verbose`

## Environment
* DITA-OT version: 4.0
* Operating system and version: Ubuntu 20.04 LTS
* How did you run DITA-OT? `dita` command
* Transformation type: `html5`, `pdf2`

chrispy-snps:
I am not sure why these scoped keyrefs do not issue DOTJ048I instead of DOTJ047I, which is specifically for scoped key references:

```
  <message id="DOTJ048I" type="INFO">
    <reason>Unable to find key definition for key reference "%1" in scope "%2".</reason>
    <response>The href attribute may be used as fallback if it exists</response>
  </message>
```

chrispy-snps:
As part of this fix, `KeyrefModule` should add an attribute or PI to references into peer maps, as requested in [this comment](https://github.com/dita-ot/dita-ot/issues/4069#issuecomment-1360689049):

![keyref_add_peer_scope](https://github.com/dita-ot/dita-ot/assets/50950969/23a862c9-74dc-45e2-b97a-429c3ca42bb9)

[...]

PR #4362, opened 2024-01-25
Rewrite messages.xml in integration to support overriding and deduplication

jelovirt:
## Description
Add integration process to rewrite `messages.xml` to

* allow overriding messages in part or completely
* remove duplicates.

## Motivation and Context
Allows plug-ins to override messages completely or only override part of the message configuration.

Relates to #4086, #4068

## How Has This Been Tested?
Manual testing

## Type of Changes

For example we have a built-in message

```xml
<message id="DOTX029I" type="INFO">
    <reason>The @type attribute on a &lt;%1&gt; element was set to '%3', but the reference is to a more specific '%4' '%2'.</reason>
    <response>This may cause links to sort incorrectly in the output.</response>
</message> 
```

A custom plug-in can use the old extension point to add a new messages file that would change the severity level

```xml
<message id="DOTX029I" type="WARN">
</message> 
```

or replace the response text

```xml
<message id="DOTX029I" >
    <response>Complain to person X who handles all reltables.</response>
</message> response
```

The only required info is the `@id` attribute. You can omit `@type`, `<reason>`, or `<response>` in which case the previous declaration of the message will provide that value.

### DIFF
--- a/src/main/config/configuration.properties
+++ b/src/main/config/configuration.properties
@@ -15,7 +15,7 @@ default.coderef-charset=UTF-8
 # Integration
 plugindirs = plugins;demo
 plugin.ignores =
-plugin.order = org.dita.base org.oasis-open.dita.v1_3 org.oasis-open.dita.v1_2
+plugin.order = org.dita.base org.dita.xhtml org.dita.htmlhelp org.dita.pdf2 org.oasis-open.dita.v1_3 org.oasis-open.dita.v1_2
 registry = https://plugins.dita-ot.org/
 
 # PDF2 defaults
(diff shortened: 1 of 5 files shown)


chrispy-snps:
@jelovirt - in [this discussion post](https://github.com/orgs/dita-ot/discussions/3922#discussioncomment-5780836), I explored a promising way of overriding a message's severity via the `dita.xsl.messages` extension point. In a later point, I found that this plugin-based severity override could work but it would need an update in `MessageUtils.java`.

Does this update possibly have anything to do with that?

robander (dismissed):

robander:
@chrispy-snps it's closely related - you could bump an info message up to an error, for example. If you're running in strict mode, and the build is correctly stopping the build for an error, that would stop publishing. (The only reason I don't directly say "this solves the issue" is that I haven't tested lately to verify Error messages from all of Java + XSLT + Ant stop the build in strict mode.)

stefan-jung:
Hi @jelovirt, has this already been shipped? Increasing the severity like so does not work on `4.3.0`.

```xml
<message id="DOTJ047I" type="FATAL">
</message> 
```

PR #4369, opened 2024-01-28
Use generated copy-to from branch filtering as key value

jelovirt:
## Description
Use generated copy-to from branch filtering attribute `@filter-copy-to` as key value instead of `@href`.

## Motivation and Context
Fix bug in map-first processing where key definitions are branch filtered with `<dvrResourcePrefix>`. The key resolution code needs to use `@filter-copy-to` instead of `@href`.

## How Has This Been Tested?
Existing tests.

## Type of Changes

### DIFF
--- a/src/main/java/org/dita/dost/module/filter/MapBranchFilterModule.java
+++ b/src/main/java/org/dita/dost/module/filter/MapBranchFilterModule.java
@@ -48,7 +48,7 @@ import org.w3c.dom.*;
  */
 public class MapBranchFilterModule extends AbstractBranchFilterModule {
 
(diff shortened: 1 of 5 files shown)


infotexture (approved):
@jelovirt Tested with our docs, seems to work as intended: https://github.com/dita-ot/website/commit/1b1421a33e0df110c8fbaf1019408ee830b974f6

The only diffs there are the restored `<dvrResourcePrefix>` values mentioned earlier: 
https://github.com/dita-ot/org.dita-ot.html/pull/13#issuecomment-1913332262.  

:shipit:

PR #4386, opened 2024-02-03
Add deliverable option to CLI help

jelovirt:
## Description
Add `--deliverable` option to CLI help.

## How Has This Been Tested?
Manual testing.

## Type of Changes


infotexture (approved):
👍 Thanks for the quick fix.

infotexture:
If I understand correctly, this option only applies if the `--project` argument is also set, but I'm not sure how to indicate that in a compact way, or if any other such sub-options exist, so this is fine for now.

PR #4408, opened 2024-02-16
Add message for generic parsing exceptions

jelovirt:
## Description
When the input XML is in some way invalid but it can still be parsed, for example

```xml
<topic id="%50">
```

the XML parser will throw an exception with message

> Attribute value "%50" of type ID must be an NCName when namespaces are enabled.

We currently capture this and log out as an error message.

In order to give as many messages an ID as possible, we add a message ID for generic parsing errors where the `<reason>` of the message is provided by the XML parser and the `<response>` is left empty. The logged message will be

> Error: file:/Volumes/tmp/src/topic.dita:3:17: [DOTJ088E] Attribute value "%50" of type ID must be an NCName when namespaces are enabled.

or for example with incorrect attribute name that goes against the DTD

> Error: file:/Volumes/tmp/src/topic.dita:3:36: [DOTJ088E] Attribute "xml:langx" must be declared for element type "topic".

## Motivation and Context
Add error ID for parsing exceptions to improve debugging.

## How Has This Been Tested?
Manual testing.

## Type of Changes

This is a generic error message, so we don't know in advance what the error message will be. Users may override the message configuration to be e.g.

```
<reason>Error in parsing file: %1</reason>
```

if they want to add more context to the message.

### DIFF
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -451,7 +451,7 @@ See the accompanying LICENSE file for applicable license.
   </message>
 
   <message id="DOTJ088E" type="ERROR">
-    <reason>%1</reason>
+    <reason>XML parsing error: %1</reason>
     <response></response>
   </message>
 
--- a/src/main/config/messages_template.xml
+++ b/src/main/config/messages_template.xml
@@ -450,6 +450,11 @@ See the accompanying LICENSE file for applicable license.
     <response>Ignoring chunk operation.</response>
   </message>
 
+  <message id="DOTJ088E" type="ERROR">
+    <reason>%1</reason>
+    <response></response>
+  </message>
+
   <!-- End of Java Messages -->
 
   <!-- Start of XSL Messages -->
--- a/src/main/java/org/dita/dost/exception/DITAOTXMLErrorHandler.java
+++ b/src/main/java/org/dita/dost/exception/DITAOTXMLErrorHandler.java
@@ -9,6 +9,7 @@
 package org.dita.dost.exception;
 
 import org.dita.dost.log.DITAOTLogger;
+import org.dita.dost.log.MessageUtils;
 import org.dita.dost.util.Configuration;
 import org.xml.sax.ErrorHandler;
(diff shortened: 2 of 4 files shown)


chrispy-snps:
@jelovirt - this is a nice enhancement, especially now being able to override this class of message as desired.

src/main/config/messages_template.xml
```diff
@@ -450,6 +450,11 @@ See the accompanying LICENSE file for applicable license.
     <response>Ignoring chunk operation.</response>
   </message>
 
+  <message id="DOTJ088E" type="ERROR">
+    <reason>%1</reason>
+    <response></response>
```
infotexture, line 455: If we know this is an XML parsing error, should we say so?

```suggestion
    <reason>XML parsing error: </reason>
    <response>%1</response>
```

jelovirt: The `%1` in this case is the error message thrown by the XML parser. It describes what the error condition is, like "Attribute value "%50" of type ID must be an NCName when namespaces are enabled.".

This means it's not a response that describes what the user should do next. Because this is a generic parsing, we don't know what the next step should be. It could be "fix your specialization" or "fix your content".

infotexture: OK, so we leave the placeholder in the `<reason>` element, but shall we add the intro for context:

```suggestion
    <reason>XML parsing error: %1</reason>
    <response></response>
```

jelovirt: I would prefer to not use a prefix in the reason, but it may be useful for users 🤷🏽. I'll need to get more examples of those messages to help in deciding whether it's helpful of not.

jelovirt: Xerces produces error messages listed at https://github.com/apache/xerces2-j/blob/trunk/src/org/apache/xerces/impl/msg/XMLMessages.properties#L230.

jelovirt: Added `XML parsing error: ` prefix to the message.

PR #4409, opened 2024-02-16
Fix missing log level prefix when not using colours

### LINKED ISSUE #4406: Better mark validation error in topic
If I have an invalid ID in a topic:

```xml
<topic id="%50">
```

I will get in the console output something like this:

```
file:../flowers/topics/introduction.dita:3:17: Attribute value "%50" of type ID must be an NCName when namespaces are enabled.
```
so the error line is not marked with an `[ERROR]` tag.

jelovirt:
## Description
Fix missing log level prefix `Error: ` or `Warning: ` when not using colours in CLI.

## Motivation and Context
Fixes #4406
## How Has This Been Tested?
Manual testing.
## Type of Changes

### DIFF
--- a/src/main/java/org/dita/dost/log/AbstractLogger.java
+++ b/src/main/java/org/dita/dost/log/AbstractLogger.java
@@ -233,18 +233,24 @@ public abstract class AbstractLogger extends MarkerIgnoringBase implements DITAO
     }
     StringBuilder buf = null;
     if (!legacyFormat) {
-      if (useColor && level == Project.MSG_ERR) {
-        buf =
-          new StringBuilder()
-            .append(ANSI_RED)
-            .append(Main.locale.getString("error_msg").formatted(""))
-            .append(ANSI_RESET);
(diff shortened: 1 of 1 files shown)


raducoravu (approved):
I also built a DITA OT based on your changes and now the "Error:" marker appears.

raducoravu:
In general it's crazy hard to programmatically parse console lines in order to gather errors, some error lines look like this "[ERROR]", some look like this "Error!", and now some look like "Error:". Same for all severity types from info/warning/fatal error.

jelovirt:
Since the CLI default logger is intended for human consumption, it may be easier to use some other logger that will output log messages in easier to parse format.

jelovirt:
I will need to write initial docs for using customized or alternative logger for different log use cases.

raducoravu:
@jelovirt yes this might help, see also this discussion here:

https://github.com/orgs/dita-ot/discussions/4340

There are various places from where errors may arise: SAX Parser, XSLT stylesheets, DITA OT code, Apache FOP.

PR #4418, opened 2024-02-24
Collect keywords only from top level topicmeta in map

### LINKED ISSUE #4405: Variable definition values should not be included in `index.html` file
## Expected Behavior
If I publish the following:

FIGURE

the `<meta name="keywords">` element in `index.html` file should contain only keywords intended to be metadata.

## Actual Behavior
In #4352, the HTML5 transformation was updated to include map-level metadata in the `index.html` file. In `get-meta.xsl`, `<keyword>` values are harvested from the following paths ([link to code here](https://github.com/dita-ot/dita-ot/pull/4352/files#diff-44db18c4206d6b9300448634510cdbe6599828f7db24ce6f7ca2a330ee55d3e1R250)):

```
topic/prolog/metadata/keywords/descendant-or-self::*
map//topicmeta/keywords/descendant-or-self::*
map//topicmeta/metadata/keywords/descendant-or-self::*
```

However, this also matches `<keyword>` elements used in variable definitions:

FIGURE

which is unintentionally and undesirably includes all variable definitions in the metadata:

```
<html lang="en">
  <head>
    ...
    <meta name="keywords" content="keyword1, variable1, variable2">
                                             ^^^^^^^^^  ^^^^^^^^^
    <title>My Map</title>
```

## Possible Solution
The metadata template should be updated to exclude variable definitions in a robust and reliable way.

## Steps to Reproduce
1. Download this testcase:
2. Run the following command:
   `dita -i map.ditamap -f html5`

## Environment
* DITA-OT version: 4.2
* Operating system and version: Linux
* Transformation type: HTML5

jelovirt:
## Description
Collect `<keyword>` element only from the top level `<topicmeta>` for `index.html`

## Motivation and Context
Fixes #4405. Collecting keywords from all levels is unintuitive behaviour.

## How Has This Been Tested?
Improved HTML5 unit test to cover keyword metadata generation.

## Type of Changes

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/get-meta.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/get-meta.xsl
@@ -258,7 +258,7 @@ See the accompanying LICENSE file for applicable license.
 
   <xsl:template match="*[contains(@class, ' map/map ')]" mode="gen-keywords-metadata">
     <xsl:variable name="topicmeta" as="element()*"
-                  select="descendant::*[contains(@class,' map/topicmeta ')]"/>
+                  select="*[contains(@class,' map/topicmeta ')]"/>
     <xsl:variable name="keywords" as="element()*"
                   select="($topicmeta | $topicmeta/*[contains(@class,' topic/metadata ')])/
                             *[contains(@class,' topic/keywords ')]/
--- a/src/main/plugins/org.dita.html5/xsl/get-meta.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/get-meta.xsl
@@ -13,6 +13,7 @@ See the accompanying LICENSE file for applicable license.
 
   <!-- Get each value in each <keywords>. Nested indexterms should have unique entries. Other
        elements (based on keyword) cannot nest. -->
+  <!-- Deprecated since 4.2.1 -->
   <xsl:key name="meta-keywords" match="*[ancestor::*[contains(@class,' topic/keywords ')]]" use="text()[1]"/>
 
   <!-- Deprecated since 3.6 -->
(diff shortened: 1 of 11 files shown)


src/main/plugins/org.dita.html5/xsl/get-meta.xsl
```diff
@@ -244,24 +245,26 @@ See the accompanying LICENSE file for applicable license.
...
-          <xsl:if test="position() > 2">
-            <xsl:text>, </xsl:text>
-          </xsl:if>
-          <xsl:value-of select="normalize-space(text()[1])"/>
-        </xsl:if>
-      </xsl:for-each>
-    </xsl:variable>
+  <xsl:template match="*[contains(@class, ' topic/topic ')]" mode="gen-keywords-metadata">
+    <xsl:variable name="keywords" as="element()*"
+                  select="*[contains(@class,' topic/prolog ')]/
```
robander, line 250: I don't think we should turn off the descendant matching for prolog here. Some of the background covered in this comment: https://github.com/dita-ot/dita-ot/pull/4410#issuecomment-1962357794

But basically - this code has been around for a long time, and -- the memories are hazy -- but I'm pretty confident this was an explicit reaction to user complaints when we did *not* get descendants. Topics are designed for reuse, and when you pull 3 topics into one file for HTML output, you only have one spot to put the keywords. You've created an HTML document where your content explicitly defines keywords for all three, say `keyword-topic1`, `keyword-topic2`, and `keyword-topic3`. Those should all apply because the meta tag applies to the whole document. If we only grab from the root topic, then we're explicitly dropping the keywords from those nested topics; search engines that still pick up this meta tag (if any still do that) will be losing important information about what's in the document.

Unlike in maps ... this isn't something that can really be "fixed" in the source. Most likely, the keywords from nested topics do *not* apply to the body of the root topic, so it's improper to move them up to that level. If the topics were combined through chunking, there is even less of an option.

jelovirt: Good catch, deleted that by accident.

jelovirt:
@robander Fixed the issue with nested topics and improved the test case to cover those too.

@chrispy-snps If you need to add those `<topicgroup>` keywords into `index.html`, I think the code is now easier to override. I separated the topic and map templates because it made more sense to keep them separate. Maybe it would make sense to move the map template into the stylesheet that generates `index.html` and not keep it in the generic stylesheet for topics 🤔.

The test cases are not better. They don't have `<keyref>` elements, but those shouldn't be processed now.

jelovirt:
(Off-topic, I added @chrispy-snps as the `Co-authored-by` to this PR. https://github.com/dita-ot/dita-ot/pull/4418/commits shows that nicely, "jelovirt and chrispy-snps committed 7 minutes ago")

robander (approved):

chrispy-snps:
@jelovirt - thanks, all your changes look great! I appreciate the unit tests too; this is good coverage to have because processing changes can have tricky unseen side effects.

I also had the same thought about breaking map-level metadata generation into its own stylesheet. Plus, I noticed opportunities for code improvement. Perhaps once we start using XSLT3 features, this is an area we could revisit.

And thanks for the co-authored reference. Remind me not to reply to PR feedback before I've had my first cup of coffee...

Issue #4432, opened 2024-03-12
Set CSS class="inline" for inline image elements


stefan-jung:
## Description
It would be nice, if an additional CSS class would be added for inline **`<img>`** elements. Currently, this does not happen, as you can see in the [**`<xsl:template match="*[contains(@class, ' topic/image ')]" name="topic.image">`**](https://github.com/dita-ot/dita-ot/blob/e94fff117b4329aff20d2ec6e8d33ea1a4ff5b70/src/main/plugins/org.dita.html5/xsl/topic.xsl#L1366C3-L1366C81) template and the [**`<xsl:template name="topic-image">`**](https://github.com/dita-ot/dita-ot/blob/e94fff117b4329aff20d2ec6e8d33ea1a4ff5b70/src/main/plugins/org.dita.html5/xsl/topic.xsl#L1410) template. This would simplify the styling of inline images, e.g. to change formatting, positioning, and scaling.

This would be a non-breaking change, quite easy to implement. If you would accept the change, I could send a pull request.

## Possible Solution
Set an additional class, for instance `inline`.

chrispy-snps:
This is a great idea! I encourage you to make the pull request; I will support it.

stefan-jung:
@chrispy-snps, because you have said that you want to support it: Maybe you can take the reviewer role and review https://github.com/dita-ot/dita-ot/pull/4434.

PR #4434, opened 2024-03-12
Add CSS class "inline" to inline images.

stefan-jung:
## Description
Add CSS class `inline` to inline images in HTML5 output, as suggested in https://github.com/dita-ot/dita-ot/issues/4432.

## Motivation and Context
Simplifies formatting, positioning, and scaling of inline images in HTML5.

## How Has This Been Tested?
[html5-inline-image.zip](https://github.com/dita-ot/dita-ot/files/14573648/html5-inline-image.zip)

## Type of Changes

    _none_
    _no_

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -1412,13 +1412,14 @@ See the accompanying LICENSE file for applicable license.
     <img>
       <xsl:call-template name="commonattributes">
         <xsl:with-param name="default-output-class">
-          <xsl:if test="@placement = 'break'"><!--Align only works for break-->
-            <xsl:choose>
-              <xsl:when test="@align = 'left'">imageleft</xsl:when>
-              <xsl:when test="@align = 'right'">imageright</xsl:when>
-              <xsl:when test="@align = 'center'">imagecenter</xsl:when>
-            </xsl:choose>
-          </xsl:if>
+          <xsl:choose>
+            <!-- Align only works for break -->
+            <xsl:when test="@placement = 'break' and @align = 'left'">imageleft</xsl:when>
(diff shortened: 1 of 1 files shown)


chrispy-snps:
@stefan-jung - Thanks for filing this pull request!

Would you be willing to switch the polarity of this change? It might make more sense to add a `break` class value instead of an `inline` class value for the following reasons:

- In DITA, the default value for the `@placement` attribute is `inline`.
- In HTML5, the default behavior for `` elements is `display: inline`.
- A `class ~= "break"` class property would indicate the non-default behavior; the lack of the property would indicate the default behavior.
- Because inline images tend to occur within running-text content, it might be nicer to have that be the default (shorter) case.

stefan-jung:
Hi @chrispy-snps, you have requested:

```
Would you be willing to switch the polarity of this change? It might make more sense to add a break class value instead of an inline class value for the following reasons:
```

I don't really understand what you ask me for. I've updated the PR, so that `break` will also be written to the `@class` attribute. But I don't understand how the "polarity" should be changed.

https://github.com/dita-ot/dita-ot/pull/4434/files#diff-bc5b82365276214ccdf2367749393dd4df08388067aaef15535a8f86e6826945R1421

chrispy-snps:
@stefan-jung, I apologize if I was not clear. What I meant was to include `break` for `@placement = "break"` images, but *not* to include `inline` for `@placement = "inline"` images (explicit or default). In other words, switch the condition where you add a keyword (`break` instead of `inline`).

chrispy-snps:
@stefan-jung - maybe we can restart this discussion?

chrispy-snps:
@stefan-jung - I submitted #4537 with the alternate approach. Are you agreeable to closing this request?

PR #4436, opened 2024-03-15
Pass exceptions to logging and output stacktrace on debug logging

### LINKED ISSUE #4430: Cannot see exception stack trace in DITA OT console
I have a custom Java ANT task which extends org.apache.tools.ant.Task and is used from a DITA OT plugin.
Using DITA OT 4.2.1 at some point I log an exception:

     getProject().log(ZendeskHelpCenterUpdateTask.this, ex.getMessage(), ex, Project.MSG_ERR);

but I do not see its stack trace in the console even if I run with the debug flag.
Possibly the problem is in the "DefaultLogger" class.
Older DITA OT versions used to show the stack trace of the exception as it is useful for debugging problems.

jelovirt:
## Description
Fix passing exception with error log and log stacktrace in debug logging mode.

## Motivation and Context
Fixes #4430

## How Has This Been Tested?
Manual testing.
## Type of Changes

### DIFF
--- a/src/main/java/org/dita/dost/invoker/DefaultLogger.java
+++ b/src/main/java/org/dita/dost/invoker/DefaultLogger.java
@@ -386,8 +386,8 @@ class DefaultLogger extends AbstractLogger implements BuildLogger {
         message.append(event.getMessage());
       }
       final Throwable ex = event.getException();
-      if (Project.MSG_DEBUG <= msgOutputLevel && ex != null) {
-        message.append(StringUtils.getStackTrace(ex));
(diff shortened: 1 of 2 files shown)


jelovirt:
@raducoravu Are you able to test if this logging fix works for you?

raducoravu:
@jelovirt thanks for this, I confirm that when running with "-d", errors logged like this from my ANT task now have the stack trace visible:

     getProject().log(mytask, ex.getMessage(), ex, Project.MSG_ERR);

Issue #4480, opened 2024-06-03
foreign xlink:href attributes are processed, resulting into NullPointer exception


DrRataplan:
Hey! I am setting up a Dita OT plugin that (among other things) includes some elements in a foreign element that I specialized and I place in a `map`.
This `<foreign />` contains some element that has an xlink:href attribute on it.

## Expected Behavior
I would expect Dita OT to not process the contents of `<foreign />` elements. Just copy over the contents.

## Actual Behavior
I am getting a NullPointer exception if I have an `xlink:href` attribute on an element in a `foreign` element

## Possible Solution
Ignore the contents of `foreign` elements in the preprocess step. Or ignore anything that is not on a topicref/@href. Or ignore `@href` in other namespaces than the null one.

## Steps to Reproduce

1. Take the map at https://gist.github.com/DrRataplan/ded9a31559cceb5188f74c53e980c59a
2. Run it through Dita OT. Try to output XHTML for instance, but this seems to break in an early step.
3. Observe some errors because of no specialization. 
4. Observe error.

## Copy of the error message
See https://gist.github.com/DrRataplan/b768cde94dfc5fc22831d9474c918ca6
The important part is this:

```
Caused by: java.lang.NullPointerException: Cannot invoke "java.net.URI.toString()" because the return value of "org.dita.dost.writer.DitaWriterFilter.replaceHREF(String, org.xml.sax.Attributes)" is null
	at org.dita.dost.writer.DitaWriterFilter.processAttributes(DitaWriterFilter.java:163)
	at org.dita.dost.writer.DitaWriterFilter.startElement(DitaWriterFilter.java:133)
	at java.xml/org.xml.sax.helpers.XMLFilterImpl.startElement(Unknown Source)
```

## Environment

* DITA-OT version: 4.3.2
* Operating system and version: macOS
* How did you run DITA-OT? Docker
* Transformation type: xhtml

DrRataplan:
I am up for making a PR to fix it, but I am a bit stuck on how to write a test. I cannot find any direct tests for the `DitaWriterFilter` class or a good place to start from. I do have a fix ready. I think if we chane to this in the DitaWriterFilter.java file, we're golden:

```diff
modified   src/main/java/org/dita/dost/writer/DitaWriterFilter.java
@@ -147,8 +147,14 @@ public final class DitaWriterFilter extends AbstractXMLFilter {
     for (int i = 0; i < attsLen; i++) {
[...]

PR #4483, opened 2024-06-11
Fix nullpointer caused by namespaced attributes

DrRataplan:
If a foreign element contains an element with an attribute called 'href', it would be processed as a dita href. Even if the href was actually in a namespace, so not really a dita href, but an xlink:href.

Do not process any attributes in namespaces to fix this.

## Description
Do not process attributes in namespaces in DitaWriterFilter.java. They are not Dita-related.

## Motivation and Context
I am making a dita specialization where I place some non-dita XML in a specialized `<foreign />` element. This contains an `xlink:href` attribute. Having this `xlink:href` somewhere in my map causes a NullPointerException, as outlined in #4480.

## How Has This Been Tested?
I tried changing an unit test to include my case. I am having difficulty running the unit tests, so I might need some help fixing that test. Can someone help me?

## Type of Changes

- Bug fix _Fix NullPointer exception caused by non-dita href attributes_

This should not need any updates. The new code should just prevent crashes.

  _None_
 _It should not_
## Checklist


DrRataplan:
Hey @jelovirt, what are your thoughts on this PR? Is this how you would've approached this crash?

src/main/java/org/dita/dost/writer/DitaWriterFilter.java
```diff
@@ -147,8 +147,11 @@ private Attributes processAttributes(final String qName, final Attributes atts)
     for (int i = 0; i < attsLen; i++) {
       final String attName = atts.getLocalName(i);
       final String origValue = atts.getValue(i);
+      final String uri = atts.getURI(i);
       String attValue = null;
-      if (origValue.equals(ATTR_VALUE_DITA_USE_CONREF_TARGET)) {
+      if (!uri.isEmpty()) {
```
robander, line 152: @jelovirt will know better than I do, but is this a reliable indicator of an _unexpected_  namespace? Wondering if there is any impact on the handful of attributes with `ditaarch:` or `xml:` but I don't think this if/else/else currently has any processing for those.

DrRataplan:
I'm more than happy to make it more specific to my issue: `xlink:href` being interpreted, but I'm a bit afraid this will open the door to other non-dita href attributes (or non-dita attributes in general) a bit too much.

Just give me the call! I'll also need a bit more help with the unit tests, if anyone can step in?

jlacour31:
The processAttributes method only processes conref / href / copy-to / format, we can safely skip attributes with a namespace:
```
final String attNamespace = atts.getURI(i);
if (attNamespace != null && !attNamespace.isEmpty()) {
  continue;
}
```

PR #4486, opened 2024-06-22
Multiple format in CLI

jelovirt:
## Description
Add support to define multiple format arguments in DITA-OT CLI.

## Example
```shell
$ dita -f html5 -f pdf -i topic.dita
```
Is equivalent to running multiple commands
```shell
$ dita -f html5 -i topic.dita
$ dita -f pdf   -i topic.dita
```

## Motivation and Context
Add functionality to combine multiple CLI calls to reuse the JVM and allowing to run more complex publishing combinations with a single command. This can be seen as a CLI level short hand for a DITA-OT project file that defines multiple deliverables.

## How Has This Been Tested?
Unit tests and manual testing.

## Type of Changes

If you use `-l` to write log to a file, the logs for each run will be output to the same log file, in order.


raducoravu:
Interesting, I assume the temp and output folders will be reused by all transformations and that they will take place sequentially. Or by default an unique temp folder is created for each transtype? Maybe skipping the pre-processing stages for the second transform and reusing the content from the temp folder would be interesting.

jelovirt:
@raducoravu Yes, the output directory and all other properties will the be the same for all transformation and they are run sequentally. If you explicitly set the temporary directory, then the same is used for all; by default it will use a different temp dirertory based on timestamp.

The preprocessing steps cannot be skipped right now, because filtering takes place early and you can filters based on `@print` attribute.

jelovirt:
@infotexture I update the CLI help copy to add "This option can be passed multiple times." to input and format. Should there be a paragraph at the end of the Arguments section to say that "Passing multiple input or format arguments will create a Cartesian product of argument combinations and will perform a publishing run for each combination" in user speak?

jelovirt:
Removed support for multiple inputs and only retained multiple formats. This allows tackling multiple inputs separately because it's more complex feature.

PR #4509, opened 2024-09-08
Add init subcommand

jelovirt:
## Description
Add `init` subcommand to CLI to create project files.

## Motivation and Context
#4523

## Type of Changes

New topics are required.


src/main/plugins/org.dita.base/init/startup/topics/welcome.dita
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE topic PUBLIC "-//OASIS//DTD DITA Topic//EN" "topic.dtd">
+<topic id="welcome">
+  <title>My first topic</title>
+  <body>
+    <p>Welcome to your first <ph keyref="product"/>.</p>
+    <p>You can add images:</p>
+    <image href="../images/welcome.png" id="image_j3n_cc1_bpb" placement="break"/>
```
infotexture, line 8: Missing image reference

src/main/plugins/org.dita.base/init/startup/topics/welcome.dita
```diff
@@ -0,0 +1,12 @@
...
+    <p>You can also use reusable content:</p>
+    <note conref="../reusables/reusable_components.dita#reusable_notes/reusable_note" id="note_rtg_hc1_bpb"/>
```
infotexture, line 10: Unreferenced generated IDs here, which we typically [avoid](https://github.com/dita-ot/docs/wiki/coding-guidelines#avoid-unnecessary-markup).

src/main/resources/cli_en_US.properties
```diff
@@ -40,10 +40,23 @@ uninstall.error.identifier_not_defined=You must specify plug-in identifier when
...
+init.usage=dita init <template> [options]
+init.argument.template=Template name
+init.option.output=Output directory
+init.option.list=List installed templates
+init.error.template_not_defined=Template must be defined
+init.error.file_already_exists=Failed to create project, file already exists: %s
+init.error.create_failed=Failed to create project: %s
+init.error.template_not_found=Template %s not found
+init.error.list_failed=Failed to list templates
+init.info.create=Init project using %s template to %s
```
infotexture, line 53: Let's spell this out and quote the placeholders as we do in other messages since https://github.com/dita-ot/dita-ot/commit/9736df0:

```suggestion
init.info.create=Initialized new project in '%s' using the '%s' template
```

2 questions:
1. Do the quotes need to be escaped in `.properties` files?
2. Is it OK to flip the order of arguments here, or would that require other changes?

src/main/plugins/org.dita.base/init/startup/topics/welcome.dita
jelovirt: This is not meant to actually be something we use as is, because we'd need to create some image or use an existing image that has a suitable license. This is just a in progress example of what we could have. The question what should init project contain.

src/main/plugins/org.dita.base/init/startup/topics/welcome.dita
jelovirt: Again same thing, this is just an in progress draft that needs to be edited to something we want to ship.

src/main/resources/cli_en_US.properties
jelovirt: Quotes are ok. If we want to switch the order, that requires code changes. What should the new copy be?

PR #4537, opened 2024-11-09
Use CSS to style `@placement="break"` images in HTML5 and XHTML transformations

### LINKED ISSUE #4506: Avoid relying on default image placement attribute value in stylesheets
## Expected Behavior
Based on this discussion:
https://github.com/orgs/dita-ot/discussions/4501#discussioncomment-10291439

If a Markdown file contains an image reference, it should be interpreted as having placement inline by default.

## Actual Behavior
If a Markdown file contains an image reference, it is interpreted by default as having placement break.

## Possible Solution
In the stylesheet "plugins/org.dita.html5/xsl/topic.xsl" there is a condition like this:

    <xsl:if test="not(@placement = 'inline')"><br/></xsl:if>

maybe it can be changed to remove the negation like:

    <xsl:if test="@placement = 'break'"><br/></xsl:if>

## Environment

* DITA-OT version: 4.2
macOS
* How did you run DITA-OT?
oXygen
* Transformation type:
  HTML5

chrispy-snps:
## Description
For HTML5 transformations, the `<image>` templates are updated to implement `@placement="break"` using a CSS `display: block` property instead of adjacent `<br>` elements.

## Motivation and Context
Fixes #4506 and #4432, and also allows `margin-top` and `margin-bottom` properties of image and non-image elements to collapse properly.

## How Has This Been Tested?
I used the following testcase, which tests `@placement=inline/break` images in various contexts, and with various `@align` values, using both DITA and Markdown as input formats:

[4537.zip](https://github.com/user-attachments/files/17688366/4537.zip)

## Type of Changes

Existing CSS rules for `<br>` elements (such as for custom margins) would need to be updated.

No documentation change is needed.

A release notes entry is needed. A draft write could be as follows:

> In previous releases, the HTML5 transformation used `<br>` elements to separate `<image placement="break">` images from surrounding content. Starting with this release, these images now use a CSS property (applied to `img.break`) to implement this separation. Existing CSS rules for `<br>` elements (such as margin rules) should be reviewed and updated. The XHTML transformation behavior of using `<br>` is unchanged.

### DIFF
--- a/src/main/plugins/org.dita.html5/sass/_figures.scss
+++ b/src/main/plugins/org.dita.html5/sass/_figures.scss
@@ -56,6 +56,12 @@
 
 /* Align images based on @align on topic/image */
 
+img.break {
+  display: block;
+  margin-top: 1em;
+  margin-bottom: 1em;
+}
+
 div.imageleft {
   text-align: left;
 }
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -1364,15 +1364,7 @@ See the accompanying LICENSE file for applicable license.
   <!-- =========== IMAGE/OBJECT =========== -->
   
   <xsl:template match="*[contains(@class, ' topic/image ')]" name="topic.image">
-    <!-- build any pre break indicated by style -->
-    <xsl:choose>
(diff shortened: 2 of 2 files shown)


chrispy-snps:
This did not make the 4.4 release. What is the next release we can merge it into?

chrispy-snps:
Rebased to `develop`.

PR #4544, opened 2024-11-17
Keyref with only whitespace is processed

### LINKED ISSUE #4543: keyref element's with whitespace replaces the text content
## Expected Behavior
According to the specification: http://docs.oasis-open.org/dita/dita/v1.3/errata02/os/complete/part3-all-inclusive/archSpec/base/processing-keyref-for-text.html#processing_key_references

Empty elements that include a key reference with a defined key might get their effective content from the key definition. Empty elements are defined as elements that meet the following criteria:

* Have no text content, including white space 

So I would expect XML such as `<ph keyref='key'> </ph>` to display empty text.

## Actual Behavior
The XML `<ph keyref="key"> </ph>` results in the following HTML: `<span class="ph"> some key</span>`, which is in contradiction with the specification.  There are no errors reported.

## Steps to Reproduce

[dita-keyref-whitespace-issue.zip](https://github.com/user-attachments/files/17757904/dita-keyref-whitespace-issue.zip)

There are no errors in the console.

## Environment
Windows 11

* DITA-OT version: 4.2.3
* Operating system and version: Windows
* How did you run DITA-OT? `dita`

jelovirt:
## Description
Keyref with only whitespace is processed when it should be skipped.

## Motivation and Context
Fixes #4543 

## How Has This Been Tested?
New unit tests.

## Type of Changes

This configuration needs to be documented in release notes and in [The `configuration.properties` file](https://github.com/dita-ot/docs/blob/hotfix/4.2.4/parameters/configuration-properties-file.dita) topic.

The key thing in release notes is that this bug has been fixed in 4.2.4, but because it may break existing content, the fix has not been enabled. This is a really old bug and a lot of content has been written with this is place. So common exception _may_ be that whitespace only content is not treated as something that a key can replace.

This toggle will be enabled by default in the next minor or major release, but even then the toggle will be retained for those who need to disable the fix.

### DIFF
--- a/src/main/config/configuration.properties
+++ b/src/main/config/configuration.properties
@@ -22,5 +22,5 @@ registry = https://plugins.dita-ot.org/
 org.dita.pdf2.i18n.enabled = true
 
 # Compatibility modes
-# Threat keyref elements that contain only whitespace as empty.
-compatibility.keyref.treat-whitespace-only-as-empty=true
+# Treat keyref elements that contain only whitespace as empty.
+compatibility.keyref.treat-blank-as-empty=true
--- a/src/main/config/configuration.properties
+++ b/src/main/config/configuration.properties
@@ -20,3 +20,7 @@ registry = https://plugins.dita-ot.org/
 
 # PDF2 defaults
 org.dita.pdf2.i18n.enabled = true
+
+# Compatibility modes
+# Threat keyref elements that contain only whitespace as empty.
+compatibility.keyref.treat-whitespace-only-as-empty=true
--- a/src/main/java/org/dita/dost/writer/KeyrefPaser.java
+++ b/src/main/java/org/dita/dost/writer/KeyrefPaser.java
@@ -177,6 +177,10 @@ public final class KeyrefPaser extends AbstractXMLFilter {
   private Set<URI> normalProcessingRoleTargets;
   private MergeUtils mergeUtils;
 
+  private boolean compatibilityMode = Boolean.parseBoolean(
+    Configuration.configuration.get("compatibility.keyref.treat-whitespace-only-as-empty")
+  );
+
   /**
    * Constructor.
    */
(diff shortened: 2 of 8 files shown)


src/main/config/configuration.properties
```diff
@@ -20,3 +20,7 @@ registry = https://plugins.dita-ot.org/
 
 # PDF2 defaults
 org.dita.pdf2.i18n.enabled = true
+
+# Compatibility modes
+# Threat keyref elements that contain only whitespace as empty.
```
infotexture, line 25: Minor nits:

```suggestion
# Treat keyref elements that contain only whitespace as empty
```

src/main/config/configuration.properties
```diff
@@ -20,3 +20,7 @@ registry = https://plugins.dita-ot.org/
...
+compatibility.keyref.treat-blank-as-empty=true
```
jelovirt, line 26: @robander @infotexture do you think the name is descriptive enough? It requires docs too, but as a key name

src/main/config/configuration.properties
```diff
@@ -20,3 +20,7 @@ registry = https://plugins.dita-ot.org/
 
 # PDF2 defaults
 org.dita.pdf2.i18n.enabled = true
+
+# Compatibility modes
+# Treat keyref elements that contain only whitespace as empty.
```
jelovirt, line 25: @infotexture This explanation needs some expansion. Are you able to improve this description:

Value `false` is correct behaviour per DITA spec. Value `true` is not valid behaviour per DITA spec, but it currently used for backwards compatibility to not break existing content.

jelovirt: Or should we just say here that we use `true` for backwards compatibility and will change in future. Kinda like `XXX: ` prefixed comment. Then in docs describe this property and how it works?

PR #4558, opened 2025-01-09
allow @ character in plugin directory

### LINKED ISSUE #4354: Installing plugin from directory with '@' character results in error
## Expected Behavior
Successfully install a plugin zip file which is stored in a directory containing a '@' character.

## Actual Behavior
exits with InvalidArgumentException

## How to reproduce  
Plugin path: /path/containing@sign/plugin  
Command line call: dita install plugin

## Explanation
[This line](https://github.com/dita-ot/dita-ot/blob/6e5d9ba7d7ce842852d2d6100372deec62360874/src/main/java/org/dita/dost/invoker/Main.java#L477) sets the plugin version.  
The error is thrown [here](https://github.com/dita-ot/dita-ot/blob/6e5d9ba7d7ce842852d2d6100372deec62360874/src/main/java/org/dita/dost/platform/SemVerMatch.java#L120)  
But [pluginInstall](https://github.com/dita-ot/dita-ot/blob/6e5d9ba7d7ce842852d2d6100372deec62360874/src/main/java/org/dita/dost/platform/PluginInstall.java#L91) does not use the _pluginVersion_ member for installing a local plugin which makes file names like _plugin@1.3.2_ superfluous.  

## Possible Solution
Option 1: Completely ignore pluginVersion in this case since the specified version in _plugin.xml_ is used anyway.  
Option 2: Only use the filename for this algorithm like in #4332 instead of the whole path.

pflotscher:
installing plugins that are located in directories that contain the '@' character does not lead to an InvalidArgumentException anymore

## Description
A zipped plugin file may or may not have a version number given in the form of filename@1.2.3. The current install mechanism splits up the whole path into /path/to/filename and 1.2.3.
If the path contains the '@' character e.g. /path@to/filename@1.2.3 it is split up into /path and to/filename@1.2.3 which leads to an InvalidArgumentException as described in #4354

As I already state in the issue: as far as I understand the version information is not further processed but if it will be in the future my approach should work as expected.

## Motivation and Context
Both Windows and Linux allow the '@' character in directory names.
This PR fixes #4354 

## How Has This Been Tested?
locate a file `plugin.zip` in a directory /test@test
install with `dita install /test@test/plugin.zip`  (dita version 4.2.4) --> fails
install with `dita install /test@test/plugin.zip`  (dita version local build with fix) --> successful

## Type of Changes

No documentation changes required

### DIFF
--- a/src/main/java/org/dita/dost/invoker/Main.java
+++ b/src/main/java/org/dita/dost/invoker/Main.java
@@ -612,29 +612,7 @@ public class Main extends org.apache.tools.ant.Main implements AntMain {
       pluginInstall.setForce(Boolean.parseBoolean(installArgs.definedProps.get("force").toString()));
     }
     if (installArgs.installFile != null) {
-      final File f = new File(installArgs.installFile.replace('/', File.separatorChar)).getAbsoluteFile();
-      final String pluginFile = f.exists() ? f.getAbsolutePath() : installArgs.installFile;
-      try {
-        pluginInstall.setPluginFile(Paths.get(pluginFile));
-      } catch (InvalidPathException e) {
-        // Ignore
-      }
-      try {
-        final URI uri = new URI(pluginFile);
-        if (uri.isAbsolute()) {
-          pluginInstall.setPluginUri(uri);
-        }
-      } catch (URISyntaxException e) {
-        // Ignore
-      }
-      if (pluginFile.contains("@")) {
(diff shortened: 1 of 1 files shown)


jelovirt:
Changed the implementation to work more correctly, i.e. not try to parse the value as a plug-in ID with a version when it's a path or a URI.

jelovirt (approved):

Issue #3731, opened 2021-04-21
Running the dita command sometimes does not show the entire error stack trace


raducoravu:
I have a small DITA Map with a single topic containing a link like the one in #3730

    <xref href="http://www,oxygenxml.com" format="html" scope="external"/>

I publish to HTML5 from the command line using the "-v" flag:

     dita -i D:\projects\flowers\flowers.ditamap -f html5 -v -d

the console does not have the entire stack trace of the error inside it:

       [antcall] Exiting d:\projects\eXml\frameworks\dita\DITA-OT3.x\build.xml.
        Error: java.lang.RuntimeException: Expected scheme-specific part at index 5: http:

If I run the publishing without using the "dita" command, just by using the ANT jar libraries, the stack trace is fully included in the console output:

    D:\projects\DITA-OT3.x\plugins\org.dita.base\build_preprocess.xml:96: java.lang.RuntimeException: Expected scheme-specific part at index 5: http:
	at org.dita.dost.util.URLUtils.setFragment(URLUtils.java:559)
	at org.dita.dost.util.URLUtils.stripFragment(URLUtils.java:541)
	at org.dita.dost.writer.ValidationFilter.validateReference(ValidationFilter.java:199)
	at org.dita.dost.writer.ValidationFilter.startElement(ValidationFilter.java:96)
	at org.dita.dost.writer.ProfilingFilter.startElement(ProfilingFilter.java:121)
	at org.xml.sax.helpers.XMLFilterImpl.startElement(Unknown Source)
	at org.dita.dost.writer.DebugFilter.startElement(DebugFilter.java:78)
	at org.apache.xerces.parsers.AbstractSAXParser.startElement(Unknown Source)
	at org.apache.xerces.parsers.AbstractXMLDocumentParser.emptyElement(Unknown Source)
	at org.ditang.relaxng.defaults.RelaxNGDefaultsComponent.emptyElement(RelaxNGDefaultsComponent.java:666)
	at org.apache.xerces.impl.dtd.XMLDTDValidatorXerces.emptyElement(XMLDTDValidatorXerces.java:852)
	at org.apache.xerces.impl.XMLNSDocumentScannerImpl.scanStartElement(XMLNSDocumentScannerImpl.java:219)
	at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl$FragmentContentDispatcher.dispatch(XMLDocumentFragmentScannerImpl.java:1655)
[...]

PR #4579, opened 2025-02-17
Add stacktrace CLI option

jelovirt:
## Description
Add `--stacktrace` CLI option to print out Java stack trace on error.

## Motivation and Context
Most end-users don't need the stack traces by default in verbose logging,. Disable stack trace print and add a way to enable it with an option.

## Type of Changes


raducoravu:
@jelovirt  will this stack trace be printed when running with --stacktrace?

https://github.com/dita-ot/dita-ot/issues/3731

jelovirt:
@raducoravu not sure, the test case in that issue no longer fails.

PR #4581, opened 2025-02-23
JSON logging

jelovirt:
## Description
Add JSON logging format with `--logger=json`. When log is output to standard out, the log format is row based where each row is a JSON object. When log is output to a file, the log format is a JSON array with JSON objects.

## Open issues

- [ ] What should the format and accuracy of `duration` be?

## Motivation and Context
A structured log format allows for easier log processing and introspection.

## How Has This Been Tested?
New unit tests.
## Type of Changes

New section to https://www.dita-ot.org/dev/topics/logging about structured logging. This only applies to CLI.


jelovirt:
Add configuration to enable field `id` and `timestamp`.

PR #4592, opened 2025-03-07
Support format option for validate

### LINKED ISSUE #4590: The "validate" transtype should work to be invoked also with the "-f" flag
With DITA OT 4.3.0 running the DITA OT with the "-f validate" flags throws an "Unsupported option -f" error. I think it should work. As Oxygen is an application which offers all available transtypes to the end user by introspecting the DITA OT, they should all be able to start up using the "-f" format.

jelovirt:
## Description
Support `--format=validate` as a fallback.

## Motivation and Context
The correct way to run validation is using the `validate` subcommand, but support running validation via `--format=validate` when users attempt to use validation with `--format` option.

Fixes #4590

## Type of Changes

This is a fallback behaviour which users should not use. Thus this should not be documented as a feature.

### DIFF
--- a/src/main/java/org/dita/dost/invoker/ValidateArguments.java
+++ b/src/main/java/org/dita/dost/invoker/ValidateArguments.java
@@ -56,14 +56,14 @@ class ValidateArguments extends Arguments {
       final String arg = args.pop();
       if (arg.equals("validate")) {
         definedProps.put("transtype", "validate");
-        break;
-      } else if (arg.startsWith("-")) {
-        parseCommonOptions(arg, args);
-      }
-    }
-    while (!args.isEmpty()) {
-      final String arg = args.pop();
-      if (isLongForm(arg, "-project") || arg.equals("-p")) {
+      } else if (isLongForm(arg, "-format") || arg.equals("-f")) {
+        final Map.Entry<String, String> entry = parse(arg, args);
+        if (entry.getValue().equals("validate")) {
+          definedProps.put("transtype", "validate");
+        } else {
+          // Ignore, this only happens when both validate subcommand and --format option are used and the --format option is not validate.
+        }
+      } else if (isLongForm(arg, "-project") || arg.equals("-p")) {
         handleArgProject(arg, args);
       } else if (isLongForm(arg, "-input") || arg.equals("-i")) {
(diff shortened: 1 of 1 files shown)


raducoravu (approved):

jelovirt:
@raducoravu are you able to verity this works for you?

raducoravu:
@jelovirt  unfortunately we also pass automatically the -o option from Oxygen:

````
Executing:
sh "/Users/dita-ot-feature-validate-transtype-option/src/main/bin/dita" -v -no-color -f validate -o /Users/samples/dita/flowers/out/validate -t /Users/samples/dita/flowers/temp/validate/oxygen_dita_temp -i /Users/samples/dita/flowers/flowers.ditamap

Unsupported option -o

````

jelovirt:
Yes, because `-o` is an invalid option for validate. What we could do instead of failing, just ignore the option with a warning. A less stricts option. I have to check what other tools do, whether they accept invalid options.

raducoravu:
@jelovirt ok if you decide not to support also ignoring -o, probably not very many people would want to use DITA OT "Validate" from Oxygen anyway. I'm afraid our transformation scenarios always pass "-o some/path", so it's compulsory for the end user to pass an output folder from the transformation scenario, in a way the fact that "Validate" exists is a bit unusual for a publishing engine. 
Maybe to turn lemons to profit, if an output folder is given, the errors could be saved to a "report.json" placed there..

jelovirt:
I tested a bunch of command and they all fail and print out an error message with you give them an option they don't support. Generally speaking I think it's the right thing to do.

Oxygen could support this, by checking which plug-in defines `validate` transtype and then ignoring/disabling `-o` for that transtype. But in the end, I think Oxygen has a better map validation feature already, so it's unlikely many users would even want to use `validate` "transtype" from Oxygen.

raducoravu:
@jelovirt  ok if you want you can decide not to integrate the PR, if we start adding hardcoded transtype checks in Oxygen in the future we could just call "-validate" instead of "-f validate"

jelovirt:
It would need to be `validate`, not `-validate`. I have to think about this a bit more, because there's still that option that a custom plug-in that extends validate would want to write to an output directory, like that `report.json"` example you gave. 🤔

jelovirt:
I'll merge this PR as is, and we can think about how and if to continue in the next one.

PR #4600, opened 2025-03-13
Move simple DITA 1.3 chunking to DITA 2.0 chunking module

jelovirt:
## Description
Simple chunking cases in DITA 1.x maps can be processed with DITA 2.x chunk module in compatibility mode.

## Motivation and Context
The old chunking code for DITA 1.x is complex and has bugs. For cases where we can rewrite DITA 1.x chunk actions to DITA 2.x, it makes sense to process chunks with the next chunk code.

## How Has This Been Tested?
Unit tests.

## Type of Changes

No user facing changes.


robander:
Just noting that I really like this idea -- the original chunk module is very tangled, as it tried to support a lot of complex edge cases and conditions. If we can transition the simple chunk usage to use the new 2.0 based processing, we will likely get much more stable results. Also nice that this has a configuration switch to go back to the older module if something changes in a way that is not desired.

jelovirt:
Changed to only remove test cases that no longer work and keep 1.x chunk tests.

PR #4617, opened 2025-04-18
Added check.key.duplicates.in.keyref parameter (default = no); if set…

rajko-furlan:
… to yes then key duplication is checked for in keyref processing instead of at the beginning

## Description
A parameter was added that allows key duplication to be checked for after filtering (in keyref target [preprocess] and map-keyref [preprocess2]  to be exact). The parameter is _check.key.duplicates.in.keyref_ and it defaults to "no" which means that by default the current behavior doesn't change. To change it the parameter must be explicitly set to "yes".

## Motivation and Context
The key duplication currently reports keys as duplicated even when the duplicates are filtered out in filtering stage. One of our clients is having problems because of that as for them a key duplication warning is fatal and they quit the processing if that warning is logged.

## How Has This Been Tested?
The test suite was executed with gradlew check. One of the test cases failed, but after trying the original forked code it failed in the same test case; therefore, the failure is not due to the new code. The following test case fails in 4.3.1 HEAD (as forked a week ago) on my machine (Windows 11):

IntegrationTestPreprocess2 > testconref() FAILED
    org.apache.tools.ant.BuildException at ExtensibleAntInvoker.java:214
        Caused by: org.dita.dost.exception.DITAOTException at CopyToModule.java:72
            Caused by: org.dita.dost.exception.DITAOTException at AbstractStore.java:113

The change should be backward compatible and it has no impact on the output. It just potentially changes the log.

## Type of Changes

The change requires addition of a new parameter here: <https://www.dita-ot.org/dev/parameters/parameters-base>. The parameter was already added to org.dita.base/plugin.xml.

The change shouldn't affect backwards compatibility or other users' overrides. 

## Checklist

- Please note I _did not_ update the unit tests as the changes only effect the log and I'm not sure how to test that in unit tests.

### DIFF
--- a/src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
+++ b/src/main/java/org/dita/dost/module/GenMapAndTopicListModule.java
@@ -123,6 +123,8 @@ public final class GenMapAndTopicListModule extends SourceReaderModule {
 
   /** Profiling is enabled. */
   private boolean profilingEnabled;
+  /** Show key duplicates */
+  private boolean showKeyDuplicates;
   /** Absolute path for filter file. */
   private File ditavalFile;
   /** Number of directory levels base directory is adjusted. */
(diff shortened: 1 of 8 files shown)


stefan-jung:
@jelovirt and @robander, it would be nice if you'd merge this one.

jomjohn:
Hi @stefan-jung and @jelovirt I am wondering if we can pick this PR up again since it seemed at the time to be a good idea. It looks like we could catch 2 birds (issues) with one stone (PR) here. 

@rajko-furlan would it be easy enough for you to rebase against the latest develop branch?

PR #4630, opened 2025-05-30
Support HTML5 attribute group passthrough (#4488)

### LINKED ISSUE #4488: DITAVAL "passthough" action does not work for profiling applied to an attribute group
## Expected Behavior
If I publish the following topic: to HTML:

```
<topic id="topic">
  <title>Topic</title>
  <body>
    <p audience="keep">KEEP (no attribute group)</p>
    <p audience="llm(keep)">KEEP (attribute group)</p>
  </body>
</topic>
```

with the following DITAVAL that applies `action="passthrough"` to the "keep" values:

```
<val>
   <prop att="audience" val="keep" action="passthrough"/>
   <prop att="llm" val="keep" action="passthrough"/>
</val>
```

then I expect both `<p>` elements to include passthrough annotations in the HTML output.

## Actual Behavior
Only the profiling applied directly to the attribute contains the passthrough annotations; the profiling applied to the attribute group does not:

```
<!DOCTYPE html
  SYSTEM "about:legacy-compat">
<html lang="en"><head><meta charset="UTF-8"><meta name="copyright" content="(C) Copyright 2024"><meta name="generator" content="DITA-OT"><title>Topic</title><link rel="stylesheet" type="text/css" href="commonltr.css"></head><body id="topic"><main role="main"><article role="article" aria-labelledby="ariaid-title1">
  <h1 class="title topictitle1" id="ariaid-title1">Topic</h1>
  <div class="body">
    <p class="p" data-audience="keep">KEEP (no attribute group)</p>
    <p class="p">KEEP (attribute group)</p>
  </div>
</article></main></body></html>
```

## Possible Solution
Not sure.

## Steps to Reproduce
1. Download the following testcase:
   TESTCASE
3. Run the following command:
   `dita -i topic.dita -f html5 --filter filter.ditaval`

## Environment
* DITA-OT version: 4.2.3
* Operating system and version: _Linux Ubuntu 22.04 LTS
* Transformation type: HTML5

chrispy-snps:
## Description
Extends DITAVAL `passthrough` actions in HTML5 output to support profiling attribute groups. (Previously, it supported only simple profiling attribute values.)

Per DITA 2.0 specification sections 7.4.5 and 9.8.2.3, the implementation is as follows:

* The `att` value of a `passthrough` action can match either a profiling attribute name or a group name.
* Ungrouped values belong to an implicit group named after the attribute.
* HTML5 `data-*` attributes are always named after their group (explicit or implicit).

## Motivation and Context
Fixes #4488.

## How Has This Been Tested?
I used an extended version of the testcase from #4488 that included more attributes and more values. I also compared the new `passthrough` behavior against the existing `flag` behavior (which does support attribute groups).

A unit test failed because it expected a `@props` value with groups to be passed through as a literal string. However, per this statement in the draft DITA 2.0 specification, grouping syntax also applies within `@props`:

> The grouping syntax exactly matches the syntax used for generalized attributes, making it valid inside `@props` and any attribute specialized from `@props`, including those integrated by default in the OASIS-provided document-type shells: `@audience`, `@deliveryTarget`, `@platform`, `@product`, `@otherprops`.

## Type of Changes

The change is backwards compatible (no change in passthrough behavior for ungrouped values).

I did not find any DITA-OT documentation on passthrough behaviors to update.

A release notes description could be as follows:

> In previous releases, DITAVAL `passthrough` actions in HTML5 transformations supported only simple (ungrouped) profiling attribute values. In this release, passthrough support is extended to profiling attribute groups. When an HTML5 `data-*` passthrough attribute is created for a value in a group, it is named after the group name.

## Checklist

### DIFF
--- a/src/main/plugins/org.dita.html5/xsl/topic.xsl
+++ b/src/main/plugins/org.dita.html5/xsl/topic.xsl
@@ -107,7 +107,7 @@ See the accompanying LICENSE file for applicable license.
       </xsl:choose>
     </xsl:variable>
     
-  <xsl:variable name="FILTERDOC"
+  <xsl:variable name="FILTERDOC" as="document-node()?"
                 select="if (string-length($FILTERFILEURL) > 0)
                         then document($FILTERFILEURL, /)
                         else ()"/>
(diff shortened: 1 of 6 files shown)


chrispy-snps:
@jelovirt - I added an HTML5 unit test specific to attribute group passthrough:

./src/test/resources/html5_ditaval_groups/

Let me know if you want me to consolidate it into the existing HTML5 unit test for regular attribute passthrough (the more I think about it, the more that might make sense):

./src/test/resources/html5_ditaval/

jelovirt:
Cleaned commit to remove redundant whitespace changes

jelovirt (approved):

chrispy-snps:
Thank you @jelovirt!

PR #4635, opened 2025-06-04
Rename DITAVAL @outputclass to @add-outputclass

jelovirt:
## Description
Rename DITAVAL `@outputclass` to `@add-outputclass`. Keep backwards compatibility support for `@outputclass` with `DOTA014W` warning message.

## Motivation and Context
Attribute has been renamed in the DITA 2.0 draft.

## How Has This Been Tested?
New unit tests.
## Type of Changes

Implementation has been update to match DITA 2.0 draft. Support for old attribute name `@outputclass` is retained for backwards compatibility.

### DIFF
--- a/src/main/java/org/dita/dost/reader/DitaValReader.java
+++ b/src/main/java/org/dita/dost/reader/DitaValReader.java
@@ -210,6 +210,18 @@ public final class DitaValReader implements AbstractReader {
 
   private Flag readFlag(Element elem) {
     final String style = getValue(elem, ATTRIBUTE_NAME_STYLE);
+    var outputClass = getValue(elem, ATTRIBUTE_NAME_ADD_OUTPUTCLASS);
+    if (outputClass == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/dita/dost/reader/DitaValReader.java
```diff
@@ -210,6 +210,18 @@ public void readProp(final Element elem) {
 
   private Flag readFlag(Element elem) {
     final String style = getValue(elem, ATTRIBUTE_NAME_STYLE);
+    var outputClass = getValue(elem, ATTRIBUTE_NAME_ADD_OUTPUTCLASS);
+    if (outputClass == null) {
+      outputClass = getValue(elem, ATTRIBUTE_NAME_OUTPUTCLASS);
+      if (outputClass != null) {
+        logger.warn(
```
jelovirt, line 217: This warning doesn't contain file location because we don't generate `@xtrf` and `@xtrc` attributes to DITAVAL elements. Still, I think it makes sense to warn about this.

PR #4651, opened 2025-06-24
Spotless is prettier

thendarion:
## Description
Introduce the Gradle plugin Spotless to handle Prettier.
Spotless brings the following tasks:
* spotlessCheck
* spotlessApply

## Motivation and Context
To make a simple contributor's life easier, eliminate the need to install a separate toolchain for formatting. You only need Java and Gradle.

## How Has This Been Tested?
Ran the build locally as well as on GitHub with and without a formatting error in a .java file. I also tried the fix and it produces the same output as the old npm call.

## Notes
* I used the old prettier and prettier-plugin-java versions from the package-lock.json
* the line endings are only interesting on Windows, but prettier likes LF, hence the change
* .prettierignore is obsolete as Gradle handles the inclusions and exclusions


jelovirt:
Looks good, I have to test this out.

infotexture:
If we're only formatting Java here, it might make sense to switch if that's the only thing we're using Node for here — but then why use Spotless as a wrapper to run Prettier when Spotless can format Java code itself?

If the rules Spotless uses to format Java are more configurable (less opinionated) than Prettier, it might be easier to replicate the preferred coding style with Spotless.

Might also be worth testing how useful Spotless would be for formatting the XML and other file types in the repo while we're at it, but IIRC, XSLT is usually the tricky part.

thendarion:
I just ran into something weird on my computer, I'd like to investigate before moving forward 😅

jelovirt (approved):

PR #4654, opened 2025-06-26
use gradle daemon in test GHA

thendarion:
## Description
daemon make build go zoom

## Motivation and Context
It's good practice to not leave rogue daemons running about, but IMHO it's better to kill them when they're no longer useful than to create new ones all the time. In the test build where there's multiple subsequent gradlew calls, not having to initialise everything every time can reduce build times by 20-25%.

## How Has This Been Tested?
ran the test build with the new changes to see the daemon in action
also ran the test build on a failing branch to make sure that the daemon is always stopped 

## Type of Changes
- New feature

n/a

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,13 +23,18 @@ jobs:
             build/spotless-*
             build/tmp/spotless-register-dependencies
           key: ${{ runner.os }}-${{ hashFiles('.prettierrc') }}-${{ hashFiles('.nvmrc') }}
+      - name: Start Gradle Daemon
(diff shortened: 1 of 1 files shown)


jelovirt:
Start Gradle Daemon takes 7 seconds. Maybe that could be removed and just let the Spotless check start the daemon.

thendarion:
Yes, I thought about that too. Whether to make the initialisation/validation process part of the first gradle step with a task. 
 
I decided to add a separate Start step for symmetry and transparency. 

Now each step has a single responsibility.

PR #4656, opened 2025-06-30
add the missing unit tests to the build

thendarion:
## Description
Add the missing unit tests to the build.

## Motivation and Context
When browsing the folders, I noticed that most test files ended with "Test", but some started with "Test", the latter of which were not picked up.

## How Has This Been Tested?
Ran the tests locally and noted the extra tests that were picked up after the change.
In the github build it's hard to tell the difference because it doesn't tell you the number of tests run at the end.

## Type of Changes
- Bug fix

n/a

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -121,6 +121,8 @@ artifacts {
 test {
     useJUnitPlatform {
         filter {
+            includeTestsMatching "Test*"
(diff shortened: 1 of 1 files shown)


thendarion:
#4657 made this change obsolete

PR #4665, opened 2025-07-10
Upgrade XSpec 

thendarion:
## Description
Upgrade XSpec to version 3.2.2.

## Motivation and Context
Some features are not available in the old XSpec version.
Most of the tests were OK after the upgrade too, only one set needed a little tweaking.
Now we can write tests more easily.
While I was at it, I realised that before if there were multiple failing tests in an XSpec file, you only got the first one reported. Now all of them are reported and you can easily see them.

## How Has This Been Tested?
The existing XSpec tests run better than ever. I used the new framework to update some tests and also write some new ones on a separate experimental branch. It's been working fine so far.

## Type of Changes
- New feature (internal)

n/a

### DIFF
--- a/src/test/java/org/dita/dost/util/XSpecTest.java
+++ b/src/test/java/org/dita/dost/util/XSpecTest.java
@@ -1,7 +1,7 @@
 /*
  * This file is part of the DITA Open Toolkit project.
  *
- * Copyright 2016 Jarno Elovirta, 2025 David Bertalan
+ * Copyright 2016 Jarno Elovirta
  *
  * See the accompanying LICENSE file for applicable license.
  */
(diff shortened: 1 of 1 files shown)


src/test/java/org/dita/dost/util/XSpecTest.java
```diff
@@ -55,54 +47,168 @@ private static void findXSpec(final File f, final List<File> res) {
...
+
     CatalogUtils.setDitaDir(ditaDir);
     final Resolver catalogResolver = CatalogUtils.getCatalogResolver();
-    resolver = new ClassPathResolver(catalogResolver);
+    URIResolver resolver = new ClassPathResolver(catalogResolver);
     transformerFactory.setURIResolver(resolver);
-    final Source stylesheet = resolver.resolve("classpath:///XSpec/generate-xspec-tests.xsl", "");
-    compiler = transformerFactory.newTransformer(stylesheet);
-    compiler.setURIResolver(resolver);
+    System.err.println(System.getProperty("java.class.path"));
```
jelovirt, line 62: Remove debug output

jelovirt (changes requested):
This looks good. Few comments.

jelovirt (approved):

Issue #4478, opened 2024-05-29
Handling content outside the map directory: incorrect links in nav toc (HTML5)


PieterjanVandenweghe:
## Expected Behavior
I have a folder structure where my ditamap file is not at the root level, for example:
* folder 1
    - file1.dita
    - test.ditamap
 * folder 2
    - file2.dita

When I use the nav-toc="full" parameter and set the parameter _generate.copy.outer_ to 3 in the HTML5 output format, I expect to be able to navigate to the different HTML files in the side-toc.

## Actual Behavior
Links in the side-toc are not correct: linking from file2.html to ../1/file1.html is not correct

## Steps to Reproduce
1. Create a ditamap according to this structure [sample_map.zip](https://github.com/dita-ot/dita-ot/files/15481770/sample_map.zip)
2. Set the parameter _generate.copy.outer_ to 3
3. Set the parameter _nav-toc_ to full
4.  Generate HTML5

## Copy of the error message, log file or stack trace
Not applicable

## Environment
* DITA-OT version: 4.2.3
* Operating system and version:
Windows
* How did you run DITA-OT?
OxygenXML and command line
* Transformation type:
HTML5

Issue #4527, opened 2024-10-30
generate.copy.outer option always generating content outside of the map's repository


gdelory:
I'm opening this bug following [this discussion](https://github.com/orgs/dita-ot/discussions/4522). I do think it's a bug, but I'm not entirely sure, hence the discussion. I'd like confirmation and I'm happy to try to help to fix that if this is a bug, but I find it surprising that no one never complained about that, so maybe I just misunderstood how it is supposed to work?

## Expected Behavior
`generate.copy.outer=1` should not generate content outside of the map's folder, and should not push down the map's directory in the output folder.

## Actual Behavior
`generate.copy.outer=1` acts like `generate.copy.outer=3`, plus a few warnings, see  [the discussion](https://github.com/orgs/dita-ot/discussions/4522) for the full explanation.

## Possible Solution
None I could find

## Steps to Reproduce

1. Download the [minimum sample to reproduce](https://github.com/user-attachments/files/17530384/minimum_sample.zip)
2. Run `dita -t temp -f html5 -o output -v -i map.ditamap -Dgenerate.copy.outer=1`
3. The output  folder contains subfolders instead of simply `task.html`

## Environment

* DITA-OT version: 3.3.4 and 4.2.3
* Operating system and version:
  Linux
* How did you run DITA-OT?
  `dita` command
* Transformation type:
  HTML5

thendarion:
I did some more digging, and it looks to me that `generate.copy.outer=1` was kind of wonky in 2.3.3 too, it was more like a mixture of `generate.copy.outer=1` and `generate.copy.outer=2` in terms of output

currently `generate.copy.outer=1` seems to be completely broken


what are the real life use cases for `generate.copy.outer=1` anyway?
can we justify its existence?
or do we need a different approach?

---
related links: 
* #4522 
* https://www.dita-ot.org/1.8/readme/generate-copy-outer

gdelory:
We use almost exclusively `generate.copy.outer=1` today (or more exactly a version of it I created a plugin for, to make it work since it doesn't work 😉 ). The use case is that we upload to our web server where the structure does matter and needs to be predictable. We always want only what's at the map level or below. If our users are using conref or even xref out of scope, we don't need them uploaded, users need to build them and upload them as part of the other doc set.
We don't want  `generate.copy.outer=3` because it will end up pushing down the content to sub-folder to include out-of-scope content, and we would upload sub-folder we don't except as part of your bigger documentation set which used navref to reference it.
[...]

PR #4668, opened 2025-07-15
Fix uplevels processing in Preprocess1 and Preprocess2

thendarion:
## Description
[fix] #4527 generate.copy.outer option always generating content outside of the map's repository

## Motivation and Context
This bug has been bugging me for years. The root cause runs deep and there's more to clean up, I'm sure. But I managed to put a working version together that at least passes the tests if nothing else. 

## How Has This Been Tested?
Through the automated test suite and some manual runs.

## Type of Changes
- Bug fix (more like rewrite 😞)

??


jelovirt:
@thendarion This is a quite large a PR, it will take me a while to review 😅

thendarion:
> @thendarion This targets `hotfix/4.3.4` branch but because this is not just a bug fix, it should be moved to `develop` branch. We will release 4.3.4 soon, so I think if you wait until the release, it will be easy to rebase on `develop`.

it's targeting hotfix/4.3.4 because the previous uplevels changes were merged there, 
but I already checked and it's relatively easy to merge into develop too, with only like two minor conflicts.
I can make this work with develop if we don't want it in 4.3.4

thendarion:
on the other hand, when we fixed #4478 we focused on the nav-toc links, and we overlooked the path2project links in `index.html`, so the CSS path is broken. this PR fixes that too
maybe we should add a note

note: this is the bad CSS link in 4.3.4
https://github.com/dita-ot/dita-ot/blob/49f905e9063a5797420d805957e2c49455c90c24/src/test/resources/html5_uplevels3_nav_toc_full/exp/html5/maps/index.html#L8

we could also add some kind of hotfix. let me check

PR #4677, opened 2025-07-29
Intorduce more uplevels tests

thendarion:
## Description
- builds on #4676
- add uplevels tests to html5 and xhtml_with_preprocess2 formats
- add uplevels tests for resource only situations
- make sure that uplevels1 doesn't write outside the temp or out folders (this was important during development, hence the addition here)
- reflect the desired output in the content files
- reflect the desired output (the lack thereof) in the tests

## Motivation and Context
Set up a solid base for fixing the bugs around uplevels.

## How Has This Been Tested?
The test suite still succeed.

## Type of Changes
- New feature (internal)

n/a


src/test/java/org/dita/dost/ITContentUplevels.java
```diff
@@ -0,0 +1,114 @@
...
+
+public interface ITContentUplevels {
+  AbstractIntegrationTest builder();
+
+  @Test
+  default void uplevels1() {
+    var outerFiles = getListOfPotentialOuterFilesInTmpAndOut();
+    cleanUpAnyLeftovers(outerFiles);
+
+    assertThatThrownBy(() ->
```
jelovirt, line 26: My preference would be to not add additional dependencies if possible, even if they are for testing. Could this be replaced by

```java
var e = assertThrows(AssertionError.class, () -> ...);
assertTrue(e.getMessage().contains("Missing file"));
```

I admin it's not as fluent as the AssertJ, but

src/test/java/org/dita/dost/ITPreprocess2.java
```diff
@@ -14,7 +14,7 @@
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
-public class ITPreprocess2 extends AbstractIntegrationTest implements ITPreprocessBase {
+public class ITPreprocess2 extends AbstractIntegrationTest implements ITPreprocessBase, ITContentUplevels {
```
jelovirt, line 17: Because `ITFormatHtml5` already uses preprocess2, also running the uplevels tests here is duplicated tests. Thus IMO this should be removed to make the integration tests faster.

jelovirt (changes requested):

jelovirt (approved):

PR #4678, opened 2025-07-29
check output image paths in integration tests

thendarion:
## Description
Now we check whether they exist in the right place or not. There's no content comparison still.

## Motivation and Context
Images are part of the output, and in uplevels cases they tend to end up in unexpected places. This check makes sure that images are copied alongside other content as expected.

Note: since there's no content comparison, in the same way we could check CSS file paths in the output too if we ever so desire.

## How Has This Been Tested?
Existing integration tests.

## Type of Changes

- New feature (internal)

n/a

### DIFF
--- a/src/test/java/org/dita/dost/ITFormatXhtml.java
+++ b/src/test/java/org/dita/dost/ITFormatXhtml.java
@@ -356,7 +356,7 @@ public class ITFormatXhtml extends AbstractIntegrationTest implements ITContentU
       .isInstanceOf(AssertionError.class)
       .hasMessageContaining("Multiple Failures (2 failures)")
       .hasMessageContaining("Missing file")
-      .hasMessageContaining("filterlist/exp/xhtml/subdir/delta2.gif")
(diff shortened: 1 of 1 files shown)


thendarion:
the tests fail because apparently `delta2.gif` is not copied to the expected place in the test `filterlist`

```
org.gradle.internal.exceptions.DefaultMultiCauseException: Multiple Failures (2 failures)
	org.opentest4j.AssertionFailedError: Missing file: /home/runner/work/dita-ot/dita-ot/src/test/resources/filterlist/exp/xhtml/subdir/delta2.gif

	org.opentest4j.AssertionFailedError: Missing file: /home/runner/work/dita-ot/dita-ot/build/tmp/integrationTest/filterlist/out/xhtml/delta2.gif
```

@jelovirt  which one is the desired output in `filterlist`?
`./delta2.gif` or `./subdir/delta2.gif`?

thendarion:
added a FIXME note

jelovirt (approved):

PR #4680, opened 2025-07-29
Clean CleanPreprocessModule 

thendarion:
## Description
- organize code for better readability
- move methods to more appropriate locations

## Motivation and Context
- CleanPreprocessModule took me a long time to understand when I first encountered it. I wanted to make it easier to navigate, especially as a new contributor.
- I found that getBasedir only operated on objects that belonged to Job, so I moved it. This'll make development smoother later.

## How Has This Been Tested?
All the tests pass still (the moved ones too).

## Type of Changes
- New feature (internal)

## Documentation and Compatibility
n/a

PR #4682, opened 2025-07-30
Make generate.copy.outer work as documented in uplevels scenarios

thendarion:
**note**: builds on #4680, #4678, and #4677

## Description
This is the final piece of #4668, refactored to work on `develop` instead of `4.3.4`. After this, most long-standing uplevels bugs should be fixed, including #4527.
### Bug fix
- fix uplevels with resource-only outerfiles only
- fix unnecessary copying of some resource-only files (e.g. images)
- fix unexpected uplevels directory shift
- fix path2project in uplevels situations
### New feature
- the temp folder contains all files at all times, no filtering happens here
- JobMapper ensures that the files end up at the appropriate level (ignores resource-only files)
- in `generate.copy.outer=1` situations, outer files are handled as resource-only so that they're ignored in the output

## Motivation and Context
I ran into uplevels processing bugs a couple years ago. I've always wanted to dive in and fix them, and finally I had the time. The issues ran deeper than I thought. Here I'd like to thank the other contributors and users who have raised questions and took notes; those were invaluable during the investigation.

## How Has This Been Tested?
Enabled the content tests from #4677, which alongside the tests from #4678 ensure that in the covered situations files are created where expected with the correct relative paths to, for example, CSS files.

## Type of Changes
- Bug fix (externally)
- New feature (internally)

n/a

Thanks for reviewing!


thendarion:
thanks for merging the hotfix branch. this change was building on those fixes 
I rebased it and cleaned it up a little

thendarion:
@jelovirt what do you think of this fix? do we need to change something still, or can it get into the 4.4 release?

PR #4734, opened 2025-12-28
Add support for OASIS linktitle and navtitle elements

robander:
## Description

Adds support for DITA 2.0 element `<linktitle>`, and support for both the DITA 1.3 and DITA 2.0 class attributes for `<navtitle>`

## Motivation and Context

Adding support for new element in DITA 2.0 that is not yet supported

## How Has This Been Tested?

Updating unit tests to match it. Also tested with DITA 2.0 draft spec, replacing `<linktext>` with `<linktitle>`, and got equivalent output.

## Type of Changes

Plugins that support `<navtitle>` or `<linktext>` may wish to check that they handle the new elements properly. In the preprocessed map, if using a DITA 2.0 root map, both the `<linktext>` (for 1.3) and `<linktitle>` (for 2.0) elements will be generated.


src/main/plugins/org.dita.base/xsl/common/functions.xsl
```diff
@@ -21,6 +21,18 @@ See the accompanying LICENSE file for applicable license.
     <xsl:sequence select="contains($class, ' topic/linktext ') or contains($class, ' map/linktext ')"/>
   </xsl:function>
   
+  <xsl:function name="dita-ot:matches-linktitle-titlealt" as="xs:boolean">
+    <xsl:param name="class" as="attribute(class)?"/>
+    <xsl:param name="titlerole" as="attribute(title-role)?"/>
+    <xsl:sequence select="contains($class, ' topic/titlealt ') and contains(concat(' ',$titlerole,' '), ' linking ')"/>
```
jelovirt: Could that titlerole be `tokenize($titlerole) = 'linking'` with test for empty value.

jelovirt (approved):

Issue #1774, opened 2014-11-04
1.3: Support MathML and Equation domains


drmacro:
Proposal 13111

I have basic support for both domains that I can contribute. For MathML I generate inline markup for HTML and FO. For HTML can also generate references to the MathJax Javascript library.

Note that FOP (with jEuclid) and Antenna House support MathML rendering natively. XEP does not.

This means that a full FO implementation would require the ability to separately generate SVG or raster images from MathML source. This could be done using jEuclid. However, I'm not particularly motivated to do it in the OT since FOP supports MathML directly. RenderX can provide the necessary MathML rendering support themselves as they see fit.

However, having said that, in the context of EPUB generation I may need to implement the option to generate rasters and SVGs from the MathML in any case (to support EPUB readers that don't support MathML sufficiently [which currently is all of them]), so I may end up doing this anyway.

raducoravu:
+1, right now I have to use a combination of DITA OT 2.x + Eliot's plugins to have this support available.

drmacro:
My expectation had kind of been that someone would migrate my OT 1.x plugins to the 2.x OT as the basis for formal 1.3 support. Most of them should just work as is or with minor updates. I didn't do it myself because I wasn't ever sure what the implementation plan details were.

raducoravu:
Added a like to this, somebody on the DITA Users List requested it:
https://groups.yahoo.com/neo/groups/dita-users/conversations/messages/40080

raducoravu:
:+1: from one of our users:

        The documentation indicates that with DITA 1.3 I should get automatic equation numbering if I 
        leave the <equation-number> tag empty, but I'm not getting that with either the DITA PDF or DITA XHTML transformation scenario.

raducoravu:
👍 One of our customers uses "equation-figure" and the generated HTML creates the static text "Figure X" instead of "Equation X".

raducoravu:
Another report about separating the numbering for equation-figure from the plain figure numbering:
https://www.oxygenxml.com/forum/post42584.html

raducoravu:
+1
https://groups.yahoo.com/neo/groups/dita-users/conversations/messages/43027

raducoravu:
Related issue which also needs to be fixed to get MathML support in PDF to work: https://github.com/dita-ot/dita-ot/issues/1623

raducoravu:
👍 Another request for MathML in DITA:

[...]

PR #4735, opened 2026-01-03
#1774 Initial support for Eq and MathML domains

super-structure:
## Description
- updates common strings to include "Equation"
- preprocessing for links to equation-blocks to use the equation number
- processing for equation domain and mathml domains in HTML5 and PDF2
- automatic numbering for equation number (when not defined)
- processing to copy any .mml files for HTML5 output

## Motivation and Context
The DITA-OT does not have any default processing for inline MathML markup nor referenced MathML files.

This PR adds "pass through" type processing for DITA MathML content for HTML5 and PDF2 to address issue #1774 .
Modern browsers can support MathML quite well directly and the DITA OT currently does not have any templates to even pass on this content. This also adds a "pass through" type handling of MathML for PDF2 output, though FOP and XEP cannot natively handle that content. It does not cause the FOP build to fail, it just doesn't result in the MathML displaying without some further work. But this is a necessary first step.

## How Has This Been Tested?
Tested with DITA OT 4.3.5 in MacOS and Windows 11.

## Type of Changes
- New feature 

    none
    Doubtful, as these templates are incorporated in the usual method and any user overrides should work as normal. 
    Given that there were no previous templates for MathML markup in DITA, there are no backward compatability concerns I can think of.

## Checklist

- [x] My code follows the code style of this project.
- [ ] I have updated the unit tests to reflect the changes in my code.
    There are no existing unit tests for MathML markup. The changes are XSLT only.


super-structure:
Closing as the signoff issue is proving to be more of an issue than I can correct, as it's older commits now. I'd rather just get my git config corrected and create a new PR.

